PR #14, opened 2026-07-08
feat(core): agent loop — tool registry + turn engine

naingthet:
The centerpiece that ties the pieces we've built (provider · tools · store · event stream) into a working **turn**.

## What it does
- **`ToolRegistry`** — name-keyed `Box<dyn Tool>`: `register` / `get` / `specs` (what the model is offered) / `is_empty`.
- **`Agent` + `AgentConfig`** — `run_turn(session, user_input, events)` drives one turn:
  1. assemble `ChatRequest` (system prompt + stored transcript + tool specs),
  2. stream the model call, mapping `ProviderEvent` → `AgentEvent` (text/reasoning/tool-call deltas),
  3. if the model called tools, run them and feed `ToolResult` blocks back,
  4. loop until the model stops — bounded by a **`max_steps`** guard.
  It persists user/assistant/tool messages to the `Store` and emits the full `AgentEvent` lifecycle (`TurnStarted` → … → `TurnCompleted`/`TurnFailed`) over a channel.
- **Error discipline:** provider/store failures emit `TurnFailed` and return `Err`; **tool** failures come back as `ToolOutput::error` so the model adapts — never `Err`.
- **Runtime-agnostic:** the engine uses only `futures` channels/streams — no `tokio` needed in `openwave-core`'s engine itself.

## v1 scope (deliberate — each is a follow-up)
- Tool calls run **sequentially** (concurrency for independent calls next).
- Approval: **auto** for `ReadOnly`/`Workspace`; **`Sensitive` refused** (emits `ApprovalRequired` + a "not yet supported" result) — the park-and-resume approval flow is the next slice.
- Cross-turn context is the stored **text** messages; structured tool-call persistence (the `tool_call` table) + context summarization come later.

## Test
A scripted fake provider drives `read_file` against a real workspace file through a real SQLite store, asserting the event lifecycle **and** that `user`/`tool`/`assistant` messages persist in order.

Verified: `cargo fmt`, `cargo clippy --workspace --all-targets -D warnings`, `cargo test -p openwave-core` (**26 tests**), `--no-default-features` build, `cargo doc -D warnings`.

---
**Design-heavy slice — flagged for your review** before merge. Key calls to sanity-check: the in-memory transcript + text persistence (vs a structured `tool_call` table now), the sequential-tools / auto-approval v1 policy, and the `futures`-channel event sink (the axum surface will forward it to WS next). Not merging until you've looked.

### DIFF
--- a/crates/openwave-core/src/agent.rs
+++ b/crates/openwave-core/src/agent.rs
@@ -1,0 +1,649 @@
+//! The agent loop: the turn engine that drives a conversation.
+//!
+//! One [`Agent`] ties together a [`ModelProvider`], a [`ToolRegistry`], and a
+//! [`Store`], and runs a *turn* — one user input through to a final answer —
+//! emitting [`AgentEvent`]s as it goes.
+//!
+//! Per turn the loop: assembles the request → streams the model call →, if the
+//! model called tools, runs them and feeds the results back → repeats until the
+//! model stops, bounded by a max-steps guard.
+//!
+//! v1 scope (deliberately small; each is a tracked follow-up):
+//! - tool calls run **sequentially** (concurrency for independent calls later);
+//! - approval is **auto** for `ReadOnly`/`Workspace`; `Sensitive` is refused
+//!   (the park-and-resume approval flow lands next);
+//! - cross-turn context is the stored text messages (structured tool-call
+//!   persistence + context summarization come later).
+
+use std::collections::HashMap;
+use std::sync::Arc;
+
+use chrono::Utc;
+use futures::channel::mpsc::UnboundedSender;
+use futures::StreamExt;
+use serde_json::Value;
+
+use crate::error::{AgentError, Result};
+use crate::event::AgentEvent;
+use crate::id::{CallId, MessageId, TurnId};
+use crate::model::{Message, Role, Session};
+use crate::provider::{
+    ChatMessage, ChatRequest, ContentBlock, ModelProvider, ProviderEvent, StopReason, Usage,
+};
+use crate::storage::Store;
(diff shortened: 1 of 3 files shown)


naingthet:
Both review findings fixed:

- **#1 (tools run on final step):** the step guard now bails *before* executing tools when no step remains to consume their results — `if step + 1 >= max_steps { return Err(..) }` after detecting tool calls. New test `max_steps_guard_fails_before_running_tools` (max_steps=1) asserts the turn errors, no `ToolCallCompleted` is emitted, and nothing tool-related is persisted (only the `User` message).
- **#2 (usage under-reporting):** `usage` now accumulates across steps via `AddAssign for Usage` (`total_usage += reported`), so `TurnCompleted` reflects the whole turn. The end-to-end test now asserts the summed counts (5+3 in, 2+4 out → 8/6).

27 tests green; clippy/rustdoc/`--no-default-features` clean.

naingthet:
Pulled the tool-result size cap forward (was going to be a follow-up):

- Configurable via **`AgentConfig.max_tool_result_bytes`** (default `DEFAULT_MAX_TOOL_RESULT_BYTES` = 256 KiB).
- Applied at the **dispatch layer** (`run_tool`) so it covers every tool uniformly, truncating on a UTF-8 char boundary with a `[truncated: N of M bytes shown]` notice — so a large `read_file` (or any future tool) can't blow the model's context.
- Test `large_tool_results_are_truncated` verifies a 10 KB read under a 100-byte cap comes back truncated (not an error).

Offset-based paging (`read_tool_result`-style) is the natural later enhancement; truncate-with-notice is the v1.

PR #36, opened 2026-07-09
feat: mid-turn steer — inject a message into a running turn

naingthet:
## Summary
- Mid-turn **steer**: `SteerInbox` drains at step boundaries; `interrupt: true` preempts the provider stream (same race pattern as cancel) then continues the turn.
- New `UserSteered` event + `POST /chats/{id}/steer` (`202` / `409` / `404` / `400`). Cancel still wins when both are ready.
- Desktop UI recognizes `user_steered` / `turn_cancelled`; API client gains `steer` / `cancel`.

## Test plan
- [ ] `cargo test -p openwave-core interrupt_steer` / `cancel_wins_over_steer`
- [ ] `cargo test -p openwave-server interrupt_steer` / `steer_without_a_running_turn`
- [ ] `cargo test --workspace` / clippy / fmt
- [ ] Manual: start a turn, `POST /chats/{id}/steer` with `interrupt: true`, confirm `UserSteered` then `TurnCompleted` (not cancelled)

### DIFF
--- a/crates/openwave-core/src/agent.rs
+++ b/crates/openwave-core/src/agent.rs
@@ -33,6 +33,7 @@ use crate::model::{Chat, Message, Role};
 use crate::provider::{
     ChatMessage, ChatRequest, ContentBlock, ModelProvider, ProviderEvent, StopReason, Usage,
 };
+use crate::steer::SteerInbox;
 use crate::storage::Store;
 use crate::tool::{ApprovalClass, Tool, ToolCtx, ToolOutput, ToolSpec};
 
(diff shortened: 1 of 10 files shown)


naingthet:
Review follow-up: release the turn slot when `run_turn` returns (before journal drain), so a late `POST /steer` gets 409 instead of a silent 202. Interrupt-during-approval remains documented v1 scope (queued until next boundary; cancel still unblocks approval).

naingthet:
Follow-up: close cancel/steer ingress when run_turn returns (slot still held for journal drain / seq). Late POST /cancel or /steer during drain now gets 409, not a silent 202. Covered by slot_stays_held_until_journal_drains.

PR #471, opened 2026-07-24
refactor(desktop): give a conversation its own request state

### LINKED ISSUE #450: Desktop: move per-conversation request state out of the root component
## Problem

`App.tsx` is over 1200 lines and owns boot, the chat list, transcript hydration, the event socket lifecycle, agent-run polling, folder-access polling, approvals, steering, cancellation, chat CRUD, and model selection. It then passes roughly thirty props into `ChatView`, most of which are one conversation's request state threaded through by hand alongside its own in-flight flags, error maps, and fences.

That shape is why every new chat-scoped surface has to be wired through the top of the tree, and it makes the side-by-side layout harder than it should be: two surfaces reading the same conversation state both have to receive it as props from the root.

## Scope

- Move the per-conversation request concerns — approvals, folder access, agent runs, steering, cancellation — out of the root component and next to the state they own.
- Preserve the existing fences and generation counters exactly; they are what keeps a stale response from landing on a newly selected chat.
- Reduce the chat surface's props to what a caller genuinely supplies.
- Keep the root responsible for boot, the client, and chat selection.

## Acceptance criteria

- Selecting or deleting a chat mid-request still discards the stale response.
- Reconnect, replay, and terminal hydration behave as before.
- No change to rendered output; the existing suites cover the move.
- The chat surface no longer receives per-request state it only forwards.

Depends on #447. Should land before #449.

naingthet:
Selecting Sources or Outputs, cancelling a turn, and answering an approval all
read the same conversation state, and all of it lived in the root component.
`App.tsx` owned the polling for agent runs, folder access, and user questions,
the fences that keep a stale response off a newly selected chat, and the steer
and cancel controls — then handed roughly thirty props down to the chat surface,
most of which it only forwarded.

That is why every chat-scoped surface has had to be wired through the top of the
tree, and why two surfaces reading the same conversation both had to receive it
from the root.

## What changed

A `ConversationRequestsProvider` now owns what one conversation has in flight,
and the surfaces that render it read the context directly. It is mounted with
`key={chat.id}`: the state belongs to the conversation, so selecting another one
replaces it rather than resetting it field by field — which is also what makes a
response that arrives after the switch land on nothing.

The root keeps what it genuinely owns — boot, the client, the event socket, and
chat selection — and drives the rest through a small handle: the three refreshes
a stream event can ask for, what a turn beginning or resolving means for the
cancel and steer controls, and the one case where a chat is deleted while its own
conversation is still on screen and the fences have to be invalidated in place.

The fences move with the state they guard rather than changing. `ChatView` drops
from forty props to fourteen; `App.tsx` from 1376 lines to 751. `MessageList`
stays prop-driven — it is a leaf renderer, and keeping it that way is what its
suites test.

## Worth knowing

- The steer control now returns a promise that settles when the guidance is
  accepted or rejected, not when it is submitted. The composer awaits it to
  decide where focus belongs afterwards, so a fire-and-forget wrapper would have
  restored focus while the request was still open. There is a test for it.
- The per-chat key sits above the workspace rather than on the transcript, so the
  scoped surfaces remount on a chat switch too. They already refetched on a
  `chatId` change; the one thing that behaves differently is the Sources search
  box, which used to carry a query over from the previous conversation.
- The provider wraps both the workspace and Settings, so polling and the
  new-question attention hint still run while Settings is open, as before.

## Verification

`tsc`, 345 vitest tests, and a production build. The four new fence tests were
checked by mutation — removing each guard (the selection check on a failed
answer, the sandbox stop fence, the per-chat key) fails the corresponding test.

Closes #450

### DIFF
--- a/crates/openwave-desktop/ui/src/App.tsx
+++ b/crates/openwave-desktop/ui/src/App.tsx
@@ -1,27 +1,17 @@
 import { useEffect, useRef, useState } from "react";
 import {
   ApiClient,
-  type AgentRun,
   type Chat,
   type ModelInfo,
   type ModelSelectionKey,
-  type PendingFolderAccessRequest,
-  type PendingUserQuestions,
   type ProviderInfo,
   type ReasoningEffort,
   type SequencedEvent,
   type ServerInfo,
-  type UserQuestionAnswer,
 } from "./api";
 import { modelForSelection } from "./ModelSelection";
 import { resolveServerInfo } from "./boot";
-import {
-  hasMacOverlayTitlebar,
-  hasNativeHost,
-  requestUserAttention,
-  resolveFolderAccessRequest,
-  type FolderAccessDecision,
-} from "./host";
+import { hasMacOverlayTitlebar, hasNativeHost } from "./host";
 import { Logomark } from "./Logomark";
 import { useTheme } from "./theme";
 import { SettingsView } from "./SettingsView";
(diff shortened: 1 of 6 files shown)


naingthet:
Closing in favour of #472, which landed first and solved the same issue with per-concern hooks rather than a context. Reverting three merged hooks to swap in a different shape isn't worth the churn, and #472's argument for keeping `MessageList` prop-tested is a fair one.

Two things from this branch worth carrying forward rather than losing with it:

- **#450 isn't finished.** Its scope named approvals, folder access, *agent runs, steering, and cancellation*. `agentRuns`, `steerFenceRef`, `sandboxStopFenceRef`, and `cancelRequestTurnRef` are all still in the root. Tracked in #484.
- **A hazard for whoever moves steering next.** Wiring the steer control as `async () => requests.steerActiveTurn()` resolves the promise on submit rather than on the response. `Composer` awaits it to time focus restoration, so focus returns to the textarea while the request is still open and the stale-submission guard can never fire. The branch fixed it by typing the control as a promise that settles with the request. Not a bug on `main` today, since steering never moved.

The audit that came out of this branch also found three regressions in #472; #482 fixed the folder-picker latch, and the other two are tracked in #485 and #486.

PR #598, opened 2026-07-27
ci: combine clippy and headless test lanes

### LINKED ISSUE #589: CI: stop compiling the workspace twice for clippy
Slice of #586.

`clippy` takes 461s and `test` takes 591s, on separate runners. They compile substantially the same workspace; clippy just runs a different lint pass over it. Roughly 460s of the critical path is duplicated work.

## Scope

- Run `cargo clippy --workspace --all-targets --locked -- -D warnings` and `cargo test --workspace` in **one** job so a single compilation serves both. Note `--all-targets` already builds the test targets, so the test run after it should mostly hit warm artifacts.
- Keep the two results **separately legible**. A merged job must not make a clippy failure look like a test failure — whoever reads a red X should still know which gate broke, and the roll-up job (`fmt · clippy · build · test`) has to keep reporting truthfully.
- Check the same overlap between `desktop test` (378s) and `test`. If they build overlapping trees on separate runners, the same argument applies.
- Preserve the change-scope gate's behavior. It currently skips lanes for docs-only and dependency-only PRs, and merging jobs must not accidentally make those PRs start compiling Rust.

## The tradeoff to state explicitly

Merging serializes what is currently parallel. If clippy and test each take ~8 minutes of *compile* plus a little run time, merging is a clear win. If they are mostly independent work, it is a wash or worse. Measure the merged job's wall clock against `max(clippy, test)` today, and only keep it if it wins.

## Acceptance criteria

- Reported before/after CI wall clock.
- A clippy violation and a test failure are still distinguishable at a glance from the PR checks list.

naingthet:
## Summary

- Run Clippy and the headless workspace suite serially in one job, so the test command can reuse the first command's Cargo artifacts.
- Publish separate `clippy` and `test` statuses on same-repository pull requests while the aggregate gate follows the combined job.
- Keep desktop tests independent: the headless command explicitly excludes `openwave-desktop`.

## Validation

- `actionlint .github/workflows/ci.yml`
- `git diff --check`

## Measurement

The recent full `main` run [30290540014](https://github.com/brightwave-inc/openwave/actions/runs/30290540014) took 8m08s end to end; its parallel Rust lanes took 6m40s for Clippy and 7m35s for headless tests. This PR's combined job will be retained only if its wall clock beats that 7m35s critical-path baseline.

Closes #589

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -185,11 +185,14 @@ jobs:
         run: cargo fmt --all --check
 
   lint:
-    name: clippy
+    name: clippy · test
     needs: changes
     if: ${{ needs.changes.outputs.rust == 'true' }}
     runs-on: ubuntu-latest
     timeout-minutes: 20
+    permissions:
+      contents: read
+      statuses: write
     env:
       CARGO_INCREMENTAL: 0
       CARGO_PROFILE_DEV_DEBUG: 0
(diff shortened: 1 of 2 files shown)


naingthet:
Closing without merge: the combined `clippy · test` job passed but took 9m07 (18:13:24Z–18:22:31Z), slower than the 7m35 headless-test critical-path baseline in main run 30290540014. #589 explicitly says to retain the merge only if it wins, so this experiment does not meet acceptance.

Issue #543, opened 2026-07-27
Desktop: open a source document in a panel


naingthet:
Part of #533 (phase C).

Clicking a source does nothing today. There is no viewer of any kind.

- A document detail panel addressed as `sources.{documentId}`, with a breadcrumb back to the list.
- Two views per document: the extracted text we already store, and the original. Documents with no viewer for their format show only the extracted text — a better floor than refusing to open them.
- A viewer dispatcher keyed on media type, plus the two cheap viewers: text/markdown and images.
- A download action.

PR #656, opened 2026-07-28
feat(desktop): rebuild the sources panel around per-document state

### LINKED ISSUE #541: Desktop: rebuild the sources panel around per-document state and actions
Part of #533 (phase B).

The sources list is a flat set of unclickable rows with a status pill and no actions. Delete and retry exist as server routes with nothing calling them.

- A table with title, type, size, and date, sorted and searchable.
- Status that renders nothing when a document is ready, a pulsing indicator while it is being processed, and an explanation plus a retry affordance when it fails — with retry offered only when the failure reason is actually retriable.
- Delete, retry, and open, wired to the routes that already exist.
- An empty state that says what can be dropped and what the limits are.

naingthet:
The sources list was a flat set of unclickable rows carrying a status pill and nothing else. Delete and retry had existed as server routes since the catalog landed, with no caller anywhere.

## The panel

A table of name, type, size, and date, sortable on any column and filtered by name, with open, retry, and remove per row. Opening a source navigates to `sources.{documentId}` — the panel address the URL grammar already understood but nothing constructed. The detail panel that address resolves to is #543; until that lands the navigation is a no-op rather than a break.

Status only speaks when there is something to say. A ready source is undecorated, because a list where every row is decorated makes the rows needing attention harder to find. A source still being prepared gets a pulsing indicator. A failed one carries a sentence explaining what stopped it, and a retry button when retrying could help.

The empty state says what can be dropped and what will happen to it: which formats become searchable, that the rest are stored and citable anyway, and that folders and empty files are refused.

Remove goes through the shared confirmation dialog and drops the row optimistically — the delete is durable when it returns, and waiting on the refresh round trip leaves a just-removed source sitting in the list looking like the action failed.

## Making "retriable" true rather than plausible

Offering retry only when the failure is actually retriable needed the failure to reach the client, and it did not. The document row records only the word `failed`. The category lives on the semantic job as `last_error_code`, and the transient/permanent distinction existed solely as a `retryable` boolean passed into `record_failure` and discarded once the job reached its terminal state — so nothing read it back, and the retry endpoint re-queued a doomed `parse_failed` job exactly like a `vector_store_failed` one.

Rather than reconstruct that judgement from an error string in the renderer, `record_failure` now derives it from the code it is about to record. The classification becomes one table instead of a bool threaded through thirteen call sites, drift between the two is not expressible, and the catalog reports the code alongside the same verdict the worker used. Retry is offered only when repeating the job could reach a different answer: never for content this host has no reader for, bytes that no longer match what was stored, or a pipeline that has moved on. An unrecognised code counts as retriable — a new transient failure should cost a wasted attempt rather than silently strand a source.

`DocumentSummary` also gained `size_bytes` and the `created_at` the table's columns need. Both were already on the document row and simply not projected.

The native bridge passes the failure through as a bounded lowercase identifier and never the diagnostic detail recorded beside it, which is free text from the pipeline and could carry a path. The renderer turns the identifier into a sentence.

## Tests

- `catalog_names_a_failure_and_says_whether_retrying_could_help` drives one source through a transient failure and then a permanent one, asserting the catalog reports each category and verdict. This is the contract the panel's retry decision rests on and the one most likely to be reverted by accident.
- Three renderer tests: retry offered for a retriable failure and absent for a terminal one, remove confirming before it deletes, and the empty state naming the limits.
- `documents.test.ts` gains a case that a failure reason reaching the renderer must be a category rather than free text.
- Replaced `DocumentsView.test.tsx`, a static-markup smoke test that asserted the old headings existed and could not have caught anything this change is about.

The orphaned `.document-*` and `.documents-*` CSS the old markup used came out with it; `.document-error` stays for `FoldersView`, which still uses it.

## Verification

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, `cargo test --workspace` (30 binaries, no failures), and `cargo check --workspace --locked` with no lockfile movement. Under `ui`: `tsc`, 513 vitest tests, and a production build.

Closes #541


naingthet:
Closing unmerged. #664 landed the same slice first and has the floor; rebasing this over it would revert reviewed work.

Salvaged the differences rather than forcing them through:

- #669 — the merged panel offers Retry for `pipeline_changed`, which the server can only ever answer with 409, because the retry path matches on the *current* pipeline fingerprint and that failure means the job's fingerprint is stale. Filed with the underlying cause: the transient/permanent judgement is written down twice (the worker's `retryable` bool and the bridge's code table) with nothing keeping them in agreement.

Two smaller differences not worth their own issues: this branch exposed `created_at` on the catalog so the date column could mean "added" rather than "last changed", and it split the row into `document-status-pill` / `document-icon` / `search-input` components using the shadcn primitives instead of adding panel CSS. Both are preference, not defect, against work that already landed.

Issue #774, opened 2026-07-28
Desktop: managed-mode sign-in gate


abaas-bw:
Part of #742. Sign-in gate when managed policy is active.

- On boot, when `GET /policy` reports managed and no gateway session is
  stored, the renderer shows a full-screen gateway sign-in gate (reusing the
  existing browser OAuth flow) instead of the composer.
- Signed in, the app behaves normally; sign-out returns to the gate, not to
  the open product.
- Unmanaged profiles see no change.
Scope note: this is the renderer gate only — it reads `getPolicy()` and
`getGatewayStatus()`; no server changes expected beyond what exists.

PR #795, opened 2026-07-28
feat(desktop): pair with a gateway via openwave:// deep link

### LINKED ISSUE #775: Desktop: deep-link provisioning (openwave:// pairing)
Part of #742. Deep-link pairing: `openwave://provision?gateway=<url>`.

- Register the `openwave://` URL scheme (tauri deep-link plugin) and stop
  discarding single-instance args (desktop lib.rs currently drops them) —
  that is the delivery path on Windows/Linux.
- A provision link validates the gateway (e.g. GET /api/v1/meta) and calls
  `managed_policy::provision`, then routes the renderer to gateway sign-in.
- Carried from the #753 review: `provision()` is a non-atomic
  check-then-write — if two pairing links can be in flight, move the
  conflict check into a single store transaction. A conflicting re-provision
  is refused at the write path; an explicit user-confirmed re-pair flow, if
  wanted, lives here.

abaas-bw:
Registers the `openwave://` URL scheme and implements deep-link pairing: `openwave://provision?gateway=<url>` validates the gateway and provisions the profile's managed-mode policy.

## What happens on a provision link

1. The desktop shell parses the link against a strict contract — scheme `openwave`, action `provision`, exactly one `gateway` query parameter — and refuses malformed links whole.
2. The main window is surfaced, and the handler waits for the embedded server's store (a provision link commonly *launches* the app, so it must not race boot).
3. The new server-side seam `openwave_server::pair_with_gateway` takes over: it holds the gateway URL to the connectors URL contract, probes the deployment with unauthenticated `GET /api/v1/meta` **before anything is written**, then calls `managed_policy::provision` and points the ModelGateway provider config at the gateway (`base_url` + `enabled`, dropping any model snapshot synced from a previously configured base URL — sign-in resyncs the entitled set).

A failed pairing — malformed link, unreachable gateway, conflicting re-provision — changes nothing and leaves one bounded, secret-free line on stderr and in `pairing.log` under app-data (the raw link is never echoed; gateway errors already strip URLs and token material).

## Delivery path per platform

- **macOS**: opened URLs arrive as the runtime's opened-URL events, which `tauri-plugin-deep-link` re-emits to the one open-URL listener (warm and cold start). The bundler generates the `Info.plist` URL-type registration from the deep-link config in `tauri.conf.json`.
- **Windows / Linux**: the OS launches a second instance with the link as its only argument. `tauri-plugin-single-instance` is now built with its `deep-link` feature, so those arguments are forwarded to the deep-link plugin in the primary process (raising the same open-URL event) instead of being discarded. Cold starts are picked up from the plugin's recorded launch URL right after the listener is registered. Runtime scheme registration (`register_all`) covers dev builds and portable installs.

## Concurrency (carried from the #753 review)

`managed_policy::provision` is a non-atomic check-then-write and the `Store` API offers no cross-call transaction, so `pair_with_gateway` serializes end to end behind a process-local mutex. That is sufficient because exactly one desktop process owns the store — the server's instance lock guarantees it — so there is no cross-process writer for the mutex to miss. The `#[cfg_attr(not(test), allow(dead_code))]` on `provision()` is removed now that it has its production caller.

## Renderer scope: none

The sign-in gate is a parallel slice (#774); once policy flips to managed, the gate presents itself on its next poll/launch. This PR only focuses the main window on receipt. The deep-link plugin's commands are deliberately **not** added to the webview capability, and provisioning is a direct in-process call on the embedded server — no HTTP route, nothing renderer-writable can reach the policy write path.

## Dependencies

`tauri-plugin-deep-link` pinned `=2.4.9` (matches lockfile resolution); `tauri-plugin-single-instance` stays `=2.4.3` with the `deep-link` feature added. Lockfile diff is additive only (the plugin and its transitive deps); no existing pins moved.

## Tests

- `pairing::tests::pairing_provisions_policy_and_points_the_provider` — probe against a fixture gateway, policy provisioned, provider pointed/enabled, stale model snapshot dropped, idempotent re-pair.
- `pairing::tests::a_failed_pairing_changes_nothing` — unreachable gateway writes nothing; a reachable but conflicting gateway is refused after the probe and the original pairing survives in both policy and provider config.
- `deep_link::tests::provision_links_are_held_to_the_contract` — one table over the link contract (well-formed, encoded, wrong action, missing/empty/duplicate/unknown parameters, trailing path, foreign scheme). No test needs a running Tauri app.

## Verification

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, `cargo test --workspace` (1426 passed, 0 failed), `cargo check --workspace --locked` — all green locally on top of current `main`. UI untouched.

Closes #775

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2965,6 +2965,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "dlv-list"
+version = "0.5.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "442039f5147480ba31067cb00ada1adae6892028e40e45fc5de7b7df6dcc1b5f"
+dependencies = [
+ "const-random",
+]
+
+[[package]]
 name = "dom_query"
 version = "0.27.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -6714,6 +6723,7 @@ dependencies = [
  "sha2 0.11.0",
  "tauri",
  "tauri-build",
+ "tauri-plugin-deep-link",
  "tauri-plugin-dialog",
  "tauri-plugin-shell",
  "tauri-plugin-single-instance",
@@ -6877,6 +6887,16 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "b7d950ca161dc355eaf28f82b11345ed76c6e1f6eb1f4f4479e0323b9e2fbd0e"
 dependencies = [
  "num-traits",
+]
+
+[[package]]
+name = "ordered-multimap"
+version = "0.7.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "49203cdcae0030493bad186b28da2fa25645fa276a51b6fec8010d281e02ef79"
+dependencies = [
+ "dlv-list",
+ "hashbrown 0.14.5",
 ]
 
 [[package]]
@@ -8238,6 +8258,16 @@ name = "roxmltree"
 version = "0.20.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6c20b6793b5c2fa6553b250154b78d6d0db37e72700ae35fad9387a46f487c97"
+
+[[package]]
+name = "rust-ini"
+version = "0.21.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "796e8d2b6696392a43bea58116b667fb4c29727dc5abd27d6acf338bb4f688c7"
+dependencies = [
+ "cfg-if",
+ "ordered-multimap",
+]
 
 [[package]]
 name = "rust-stemmers"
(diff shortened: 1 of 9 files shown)


abaas-bw:
Pushed 623365c addressing the review:

- **Renderer event forgery**: the main window capability now carries `core:event:deny-emit` and `core:event:deny-emit-to`, so the webview can no longer emit the deep-link plugin's event to spoof an incoming `openwave://provision` link. Host-to-renderer events (updater, document import) are Rust-side `Emitter::emit` calls and are unaffected; the renderer only ever calls `listen`.
- **User consent**: pairing now shows a native confirmation dialog naming the gateway origin before the identity probe or any write. Declining is a no-op.
- **Partial-write window**: the conflict check runs before either write, then the provider config is written before the sticky managed policy. A failure between the two now leaves a still-unmanaged profile recoverable from settings, rather than a managed profile with no configured provider.
- **Parser strictness**: provision links with userinfo, a port, a query, or a fragment are rejected, as are non-http(s) gateway values; logs carry only the gateway origin, never the full link.
- **Concurrency**: provision links arriving while a pairing is in flight are dropped with a log line; the server-side mutex remains the authority.
- **Test determinism**: the unreachable-gateway test reserves and releases an ephemeral port instead of assuming port 9 is closed.

Verified: `cargo fmt --check`, `cargo clippy --workspace --all-targets --locked -D warnings`, `cargo test --workspace`, `cargo check --workspace --locked` all green; UI `tsc`, vitest (556 tests), and production build green. Capability JSON is validated by tauri-build at compile time; the ACL change compiles clean.

PR #814, opened 2026-07-28
refactor(desktop): download a source's bytes one way

### LINKED ISSUE #678: Desktop: two document-download hooks landed in parallel
The viewing slices landed out of order and each brought its own way to pull a
document's bytes, so `main` now carries two hooks that do the same job:

- `crates/openwave-desktop/ui/src/document/useFileDownload.ts` — used by
  `UniverSpreadsheetViewer` and `UniverDocumentViewer`
- `crates/openwave-desktop/ui/src/components/document/useFileDownload.ts` —
  used by `image-viewer` and `markdown-viewer`

Same name, same purpose, ~76 and ~80 lines, different shapes: one takes an
`ApiClient`, the other reaches for `useApp()`. Nothing is broken, but the next
viewer has to guess which one to import, and a fix to caching or error handling
will land in one and not the other.

Collapse them into one. The `ApiClient`-taking version is the easier one to test
and matches how the dispatcher already passes its client down.

While in there: `components/document/` and `document/` split the same concern —
the first holds the extracted-text side and the panel's view switcher, the
second holds the original-bytes viewers and the media-type dispatcher. That
split is defensible but undocumented, and the directory names do not say it.
Either write it down at the top of each, or merge them.

naingthet:
Three download paths had grown for the same job — a hook under `document/`
taking an `ApiClient`, a hook under `components/document/` reaching for
`useApp()`, and a hand-rolled `useEffect` in `PdfViewer`. All seven viewers now
go through one hook, `document/useFileDownload`.

## The two endpoints were not interchangeable, and one of them was wrong

Worth calling out, because it means this is not purely a refactor.
`/documents/{id}/file-content` serves a document only when it belongs to no
chat and no project — `serve_document_file_content` filters on
`chat_id == None && project_id == None`, and `documents.rs` in the server tests
asserts a chat-owned document 404s there. Every document the desktop can open
is chat-owned: the Tauri side only ever ingests via
`/chats/{chat_id}/documents/raw`, and the two documents in my own app database
(one of them an xlsx) both carry a `chat_id`.

So the PDF, spreadsheet and word-document viewers were asking for chat-owned
bytes at an address that cannot answer for them. The unified hook is
chat-scoped, and `ApiClient.getDocumentFileContent` is deleted rather than left
lying around, so the type checker now stops anything drifting back to the
unscoped address. The server route stays — unscoped documents are a real server
concept — it just has no caller in the renderer.

## Shape

`useFileDownload(client, chatId, documentId, { parseAs })`, the
`ApiClient`-taking shape the issue preferred, extended with the conversation
the address needs. Every viewer now takes `client` and `chatId` and asks the
hook for the shape it wants, so there is one convention rather than two.

`getChatDocumentFile` returns `{ bytes, contentType }` instead of a `Blob`.
That is what lets the hook cache bytes once and derive per-viewer: an
`ArrayBuffer` copy for the parsers that take ownership, UTF-8 text, or a `Blob`
carrying the stored media type for an object URL. Progress reporting is
untouched — the `onProgress` callback and `streamBytes` contract are the same,
and all seven viewers keep their `FileDownloadProgressIndicator`.

Two shape differences resolved: `isError: boolean` gives way to
`error: Error | null` (the richer of the two), and the `enabled` option is
dropped — nothing passed it, and deferring the download is already handled by
the panel not rendering a viewer it has nothing for.

## Keeping the byte cache

The cache the workbook viewers had now covers all seven, and it earns that. The
panel unmounts the viewer whenever the reader switches between the original and
the extracted text, so without a cache every flip back re-downloads the whole
file and redraws from a spinner. A hit cannot be stale: replacing a file makes
a new document with a new id.

It is still unbounded, as it was — a long session that opens many large sources
holds them all. That risk already existed for the formats with the biggest
files, so unifying does not change its order of magnitude, but it is worth
bounding at some point and I have not done it here.

## Directories

Documented rather than merged, per the cheaper option: a note at the top of
each side's entry file (`document/DocumentViewer.tsx`,
`components/document/document-details.tsx`) saying what the split actually is —
engine-backed, code-split viewers on one side; the extracted text and the
viewers that need nothing beyond the file's characters on the other — and that
both sides download the same way. Moving eight files for no behaviour change
would have made this diff much harder to read.

## Tests

Added one: flipping to the extracted text and back draws the original without a
second download. That is the one deliberate behaviour change here and the
cache's whole justification; I checked it fails (2 calls, not 1) with the
`byteCache.set` commented out.

Deleted none — neither duplicate had a test of its own. The three transfer
tests in `api.test.ts` moved to the surviving client method, since they cover
`streamBytes` semantics rather than the endpoint.

`DocumentDetailRoot.dom.test.tsx` drives four of the seven viewers end to end
and carries the refactor. It now clears the cache between cases, since each
case opens the same document id with different content.

## Verification

`npx tsc --noEmit`, `npx vitest run` (584 passed), and `npx vite build` all
pass; the build confirms `PdfViewer` and `UniverSpreadsheetViewer` are still
code-split. I have not driven the running app, so the 404 above is diagnosed
from the server's own filter and tests plus the app database rather than
observed in the UI, and I have not watched a real progress bar or a rendered
PDF after the change.

Closes #678


naingthet:
Rebased onto #811, which landed while this was in CI and touched three of the
same files. Both intents kept:

- `DocumentDetailRoot.dom.test.tsx` — its `afterEach` reset now also clears the
  byte cache, and its `openCitation` helper mocks `getChatDocumentFile` in the
  new `{ bytes, contentType }` shape. Its stub for `getDocumentFileContent` is
  dropped along with the method.
- `DocumentViewer.tsx` — `isPaginatedOriginalViewer` and `targetPage` untouched;
  only `chatId` and the client's method name are mine.
- `document-details.tsx` — citation placement untouched.

Its citation tests and mine both run: 587 pass locally.

PR #838, opened 2026-07-28
feat(desktop): show what a connected-folder call found

### LINKED ISSUE #809: Desktop: connected-folder tool results reach the transcript
Follows #783. `ToolResultPreview::Entries` (#800), its failures channel (#805),
and retention across a reload (#807) have all landed, but no client-executed
tool can use any of it.

## Why they can't

`ClientExecutionResolution` carries only a model-facing string:

```rust
pub enum ClientExecutionResolution {
    Completed { result: String },
    Failed { result: String, error_code: String, error_detail: Option<String> },
    Cancelled { result: String },
}
```

The desktop already builds the structure — `serialize_result` in
`client_execution/folder_operations.rs` produces
`{"status":"ok","folders":[{root_id, display_name}]}` — and then stringifies it
away for the model. Nothing structured survives the trip, so
`ToolResultPreview::build` has no `data` to read and every connected-folder call
settles into a bare rail line.

## What alpha does

The desktop authors the *client-facing result* directly and the server persists
it. `ClientLocalFileRootsResult`, `ClientLocalFilesImportResult`, and
`ClientLocalFilesWriteResult` are typed payloads posted by the client, each with
a `failures` list beside its successes. This is the model to follow — the
desktop knows what it did, and reverse-engineering a projection out of the
model-facing string on the server would be fragile and backwards.

## Work

- [ ] `StoredResolution::Completed` carries `entries` and `failures`. It is a
      persisted receipt read after a crash, so both need `#[serde(default)]`;
      receipts written by an older build must still load.
- [ ] `ClientExecutionResolution::Completed` carries them too, and the route
      validates them into a `ToolResultPreview::Entries` before they reach the
      store — the desktop is a client of this API like any other, so this is a
      trust boundary, not a passthrough.
- [ ] A client-side `resolve_*_with_artifacts` counterpart so the projection
      lands in the `result_preview` column #807 added.
- [ ] `serialize_result` builds the rows: folders from `ListRoots`, files and
      directories from `ListDirectory`, the file from `ReadFile`, and the
      imported source from `import_connected_file`. `write_output_to_connected_folder`
      too. Partial import failure is the case the failures channel was built for.

## Open sub-problem: the live event

A recovered call rebuilds its `ToolOutput` with `data: None`
(`accept_server_call` in `agent.rs`), so `ToolResultPreview::build` returns
`None` on that path regardless of what was stored. Worth confirming whether a
client call's completion reaches the renderer live at all, or only through the
terminal transcript — if it is the latter, a folder card would appear only after
reopening the chat, which is a strange half-state and should be fixed in the
same slice.

naingthet:
The last piece of #809, on top of #800, #805, #807 and #813.

Connected-folder calls announce that they finish now (#813) but still showed
nothing about what they did. Their results come back from the desktop over a
wire carrying only a model-facing string, so `ToolResultPreview::build` had no
structured data to read.

The structure already existed and was being thrown away: `serialize_result`
built `{"status":"ok","folders":[…]}` and then stringified it for the model. It
now reports rows alongside, built from the same values the model-facing payload
is, so the card and the model cannot disagree about what the folder held.

## The executor posts rows, not a projection

This is the part worth reviewing. The desktop is a client of this API like any
other, so it sends its `{entries, failures}` **unvalidated** and the store builds
the projection from them against the name on the call's own row:

```rust
let projected = rows.and_then(|rows| {
    crate::ToolResultPreview::build(&resolved_name, /* … */)
});
```

So the tool-name allowlist and every clamp are applied where the name is
authoritative rather than trusted from a client that could name any tool. An
executor cannot award itself a card for a tool that projects nothing, nor an
unbounded row. Tested from identical rows: `list_folder` projects,
`request_folder_access` does not.

Alpha posts typed client results and stores them as JSONB — same division of
labour (the executor knows what it did), but our boundary is closed, so the
validation happens server-side rather than being implied by the type.

## Receipt compatibility

`rows` is also carried on `StoredResolution`, which is a receipt read after a
crash, so the field defaults. A recovery format that refused older receipts
would strand the very work it exists to finish; an older receipt reports no
rows, which is what it knew.

## What each tool reports

- `list_connected_folders` — a folder row per authorized root
- `list_folder` — files and directories, kinded from the broker's own entry kind
- `read_connected_file` — the file and its size, named from the *request*
  (a read result is only bytes), marked when truncated. It reports the read
  rather than replaying the text: that is what the model gets, and it is far too
  much for a card
- `import_connected_file` — the source it added, when it added one
- `write_output_to_connected_folder` — the output and its size

`request_folder_access` and `connect_folder` deliberately stay off the
allowlist: their outcome is already the whole of what the rail line says, and a
card would restate it.

Closes #809

### DIFF
--- a/crates/openwave-core/src/db/mod.rs
+++ b/crates/openwave-core/src/db/mod.rs
@@ -2884,6 +2884,31 @@ impl Store for DbStore {
             now,
             resolution,
             resolved_at,
+            None,
+        )
+        .await
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    async fn resolve_client_tool_call_and_append_event_with_rows(
+        &self,
+        id: CallId,
+        chat_id: ChatId,
+        lease_token: uuid::Uuid,
+        now: chrono::DateTime<Utc>,
+        resolution: &ToolCallResolution,
+        resolved_at: chrono::DateTime<Utc>,
+        rows: Option<&serde_json::Value>,
+    ) -> Result<JournaledClientToolCallOutcome> {
+        ops::client_execution::resolve_client_tool_call_and_append_event(
+            self,
+            id,
+            chat_id,
+            lease_token,
+            now,
+            resolution,
+            resolved_at,
+            rows,
         )
         .await
     }
@@ -2905,6 +2930,31 @@ impl Store for DbStore {
             now,
             resolution,
             resolved_at,
+            None,
+        )
+        .await
+    }
+
+    #[allow(clippy::too_many_arguments)]
+    async fn resolve_expired_client_tool_call_and_append_event_with_rows(
+        &self,
+        id: CallId,
+        chat_id: ChatId,
+        lease_token: uuid::Uuid,
+        now: chrono::DateTime<Utc>,
+        resolution: &ToolCallResolution,
+        resolved_at: chrono::DateTime<Utc>,
+        rows: Option<&serde_json::Value>,
+    ) -> Result<JournaledClientToolCallOutcome> {
+        ops::client_execution::resolve_expired_client_tool_call_and_append_event(
+            self,
+            id,
+            chat_id,
+            lease_token,
+            now,
+            resolution,
+            resolved_at,
+            rows,
         )
         .await
     }
--- a/crates/openwave-core/src/db/ops/client_execution.rs
+++ b/crates/openwave-core/src/db/ops/client_execution.rs
@@ -297,6 +297,7 @@ pub(in crate::db) async fn resolve_server_tool_call(
         resolution,
         None,
         None,
+        None,
     )
     .await?
     .outcome)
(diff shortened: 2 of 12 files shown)


naingthet:
Added a second commit: a doc note on `ResultFailure::error` about who is meant to author the reason.

I had been flagging that field as a boundary concern. Checking it: no tool populates `failures` anywhere (the only `ResultFailure` in the tree is in a test), and the field is already clamped to `MAX_RESULT_ENTRY_CHARS` with control characters stripped. So there was nothing urgent about it.

What survives is narrower and not enforceable by the type: the justification for the field crossing at all is that *our own tool wrote it for a person to act on*. Forwarding a `std::io::Error` or a broker failure compiles just as well and interpolates whatever it was handed, which is how a host path would reach a card unintentionally. That is now written next to the field for whichever tool reports failures first.

PR #842, opened 2026-07-28
fix(desktop): stop the phase line rendering as an empty pulsing row

naingthet:
Found by actually running the app.

The tool rail's phase label typed itself out from an empty string at a fixed
30ms per character:

```ts
const TYPE_INTERVAL_MS = 30;
// ...
setDisplayed("");
timerRef.current = window.setInterval(() => {
  setDisplayed(target.slice(0, i));
}, TYPE_INTERVAL_MS);
```

So a phase labelled "Requesting folder access" spent ~720ms showing an icon, a
chevron, and nothing between them. It also carries `animate-pulse`, because the
phase is live — and an empty *pulsing* row does not read as text arriving, it
reads as something that failed to load.

Length made it worse rather than better, in exactly the wrong place. Phase
labels grow as their calls settle: "Checking connected folders and 1 other
task" is 43 characters, so the blank-to-legible stretch ran past a second right
when the most was happening.

## Two changes

**The first character lands synchronously**, so there is never a blank frame.

**Paced by the whole label, not per character.** A longer label advances further
each tick rather than ticking slower — slowing the tick would only trade a crawl
for a stutter, and going below one frame buys nothing. Every phase line now
appears in about the same moment (~360ms) whatever its length.

## Why no test caught it

Every assertion we had checks the label's *text*, which was always correct — it
was correct while invisible, and correct a second later. Nothing asserted what
was on screen in between. The three tests added here pin that: no blank first
frame, a long label finishing in the same budget as a short one, and a settled
label still appearing at once with no animation.

The hook is shared with the chat-title rename animation, which gets the same
treatment and wants it for the same reason.

### DIFF
--- a/crates/openwave-desktop/ui/src/useTypewriterOnce.test.ts
+++ b/crates/openwave-desktop/ui/src/useTypewriterOnce.test.ts
@@ -1,0 +1,36 @@
+// @vitest-environment jsdom
+import { renderHook, act } from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { useTypewriterOnce } from "./useTypewriterOnce";
+
+beforeEach(() => vi.useFakeTimers());
+afterEach(() => vi.useRealTimers());
+
+describe("phase label typing", () => {
+  it("never renders a blank frame", () => {
+    // The row it labels also pulses while its phase is live, so an empty frame
+    // does not read as text arriving — it reads as something that failed to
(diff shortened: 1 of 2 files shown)


naingthet:
Rebased onto `main`. This branch was cut while standing on `thet/connected-folder-results`, so it was silently carrying #838's three commits — its diff showed changes it had no business owning, and it inherited a stale-bindings failure from them. It is now the two files it should be: the hook and its test.

Issue #843, opened 2026-07-28
Desktop: a citation's page is never recorded, so a paginated source cannot open on it


naingthet:
#674 taught the panel to open a paginated source on the page a citation was recorded on, and #812's audit of what a citation carries turned up that the page list feeding that is always empty in practice.

A citation's pages are derived from the `SourceRegion`s of the evidence behind it, which are clipped from the document's own regions. No parser produces any:

- `ParsedDocument` carries `text` and `source_regions`, and every `DocumentParser` returns `ParsedDocument::from_text(...)` — the PDF, office and image parsers included (`crates/openwave-retrieval/src/liteparse_parser.rs`, `liteparse_office_parser.rs`, `liteparse_image_parser.rs`, and `parse.rs` for plain text and the fallback).
- `document_worker.rs` stores `parsed.source_regions`, which is therefore always empty, so `Chunk.source_regions` is empty, so the evidence's regions are empty, so `AssistantCitationSnapshot.pages` is `[]`.
- `SourceLocation::Page { number }` is constructed only in tests.

So the whole page pathway is plumbed and validated end to end and never fed. The page hand-off in the desktop is correct and dead.

`liteparse` already returns per-page data the PDF parser drops: `ParseResult.pages` carries page numbers and dimensions alongside the text. Emitting a `Page` region per page, with the byte span of that page's text in the emitted Markdown, is what makes the citation's page real. Existing documents would need reparsing — `LITEPARSE_FINGERPRINT` in `liteparse_parser.rs` is the lever, and the source blob is retained so a reparse is possible.

Worth having a test that a PDF ingest produces a citation whose page is the page the passage is on, since that is the assertion the current plumbing cannot make.

Issue #812, opened 2026-07-28
Desktop: draw a citation's highlight in the original-document views


naingthet:
#674 landed the citation's highlight in the extracted text and its page in the PDF viewer. The original-document views still open a citation without showing where it points.

The seams exist and have no caller:

- `MarkdownViewer` takes `highlightRange` and already uses it to pick which chunk to render first and to scroll to an anchor, but does not draw the highlight.
- `JsonViewer` and `XmlViewer` implement `highlightPath` fully — highlight, ancestor expansion, scroll.
- `PdfViewer` opens on the recorded page, and nothing on the page is marked.

Neither prop can be fed from a citation as things stand, which is why #674 left them alone:

- `highlightRange` addresses characters of the original file; a citation carries UTF-8 byte offsets into the extracted text. Those coincide only where extraction is the identity and the text is ASCII.
- `highlightPath` wants a node — a dot path or an XPath — and no citation carries one.
- Marking a passage on a PDF page needs its bounding boxes, which the citation does not carry either.

So this is a mapping problem before it is a rendering one: what a citation would have to record, or what the viewer would have to recover, to point into the original rather than into the text read out of it. Worth deciding per format — the tree viewers and the PDF are separate answers, and PDF overlays are their own slice.

naingthet:
#847 landed the part of this the data supports, and settled the mapping question for the other two.

- **Text and Markdown originals: done.** `PlainTextParser` claims every `text/*` type and returns the bytes decoded and otherwise untouched, so a citation's byte offsets into the extracted text are byte offsets into the file. The original view marks the passage from the same span the extracted text uses — split around a `<pre>` where the source is drawn as written, and placed in the parsed tree by offset arithmetic where it is rendered as Markdown (clipped onto the chunk and the block, shifted by the hard breaks the renderer inserts, then applied to the text nodes whose recorded positions the range covers). A block whose LaTeX the pipeline rewrites, and text inside a code fence or a formula, go unmarked rather than marked by guesswork. `highlightRange` is gone: it addressed characters of the raw file, which nothing produces.
- **JSON and XML: not possible, tracked in #846.** No citation carries a node path anywhere in the chain, and one inferred from the text a range covers would point at the wrong node whenever a value repeats. The viewers keep their working `highlightPath`, and `DocumentDetails` now records why nothing feeds it.
[...]

PR #866, opened 2026-07-29
feat(retrieval): record where on the page a passage sits

### LINKED ISSUE #865: Retrieval: record where on the page a passage sits, not just which page
#843 made a citation's page real by emitting a `SourceRegion` per page from liteparse's page map. `SourceLocation::Page` carries an optional `bounds` alongside the page number, and nothing populates it yet.

#812 needs those bounds to mark a passage on a PDF page, and explicitly leaves PDF overlays as their own slice. This is the parser half of that: what a citation would have to record for the viewer to draw anything.

## The mapping is available and measurably good

liteparse's `ProjectedLine` carries the line's text and its `bbox` in viewport space, alongside `page_width`/`page_height` to normalize against. The lines are not offset-mapped into the emitted Markdown — the emitter rewrites and reflows — so the spans have to be recovered by scanning each page's Markdown for its lines with a monotone cursor.

Probed against real documents to see whether that holds up:

- a 20-page contract: 593/608 lines matched in order (97.5%)
- a 3-page report: 78/90 (86.7%)
- **zero** lines matched out of order in either, so a forward-only cursor never picks the wrong occurrence

The misses are systematic rather than random: private-use bullet glyphs the emitter rewrites to `-`, horizontal rules, and table rows reflowed into pipe syntax. All are cases where the emitted text genuinely differs from the line, so they should fall back rather than be forced.

That is exactly what `Option<PageBounds>` is for: a line that matches gets geometry, and text that does not still gets its page.

## Regions must tile, not just mark

Emitting regions only for matched lines would regress #843 — a passage landing in an unmatched gap would lose its page entirely. Each page's span should be fully tiled: line regions where the scan matched, page-only regions covering the gaps between them.

## Cap: fine regions can invalidate an evidence item

`RetrievalEvidenceInput::MAX_SOURCE_REGIONS` is 128, and exceeding it makes `validate_evidence_item` reject the evidence. Chunks are 1200 chars by default, so a page of short table cells can clip to more than 128 line regions where the page-granular map produced one or two.

So `Document::source_regions_for` should bound what it returns: when clipping a span yields more regions than an evidence item may carry, coalesce consecutive regions sharing a page into one page-only region. Geometry is the part worth losing — the page is not.

Follows #843. Unblocks #812.

naingthet:
#843 made a citation's page real. `SourceLocation::Page` has carried an optional `bounds` since then with nothing to populate it. This fills it in, so a citation can say *where* on the page it points, which is what #812 needs to draw a highlight.

## Recovering the position

liteparse knows where each line sits on the page but does not offset-map its lines into the Markdown it emits — the emitter rewrites bullets, rules and tables — so line spans have to be recovered by scanning each page's Markdown with a forward-only cursor.

Forward-only is the load-bearing part: a repeated line (a page header, a "Total" row) must not claim an earlier occurrence, or the highlight lands in the wrong place on the page.

I measured this against real documents before building on it:

| document | lines placed | out of order |
|---|---|---|
| 20-page contract | 593/608 (97.5%) | 0 |
| 3-page report | 78/90 (86.7%) | 0 |

The misses are systematic rather than random — private-use bullet glyphs the emitter rewrites to `-`, horizontal rules, table rows reflowed into pipe syntax. All are cases where the emitted text genuinely differs from the line, so they fall back instead of being forced into a match.

## Pages still tile

Emitting regions only where the scan matched would have regressed #843: a passage landing in an unmatched gap would lose its page entirely. Each page's span is fully tiled instead — placed lines carry their rectangle, and everything between them is covered by a region naming only the page. So geometry is the only thing a failed match costs.

## Two caps, for two different failures

- **Evidence validity.** A chunk clipping to more than `MAX_SOURCE_REGIONS` (128) regions does not get truncated — `validate_evidence_item` rejects it outright. A 1200-char chunk over a page of short table cells can reach that where the page-granular map produced one region. `Document::source_regions_for` now coalesces by page when a clip would overrun the cap.
- **Stored size.** A page that positions more regions than is useful keeps its page-level region and drops its geometry, so a long document of dense tables does not carry tens of thousands of rectangles that are read and rewritten on every chunking pass.

Both give up geometry and keep the page, which is the right trade in each case.

## Coordinates

Rectangles are normalized to the page box so a viewer can draw them at any zoom without knowing the page dimensions the parser saw. One overhanging the page is trimmed to it rather than dropped — the text really is there. One that cannot be placed at all (no page extent, non-finite coordinates, no area once normalized) is left off rather than guessed at.

## Tests

- Real two-page PDF through PDFium: positioned regions exist, every rectangle lies within its page, and text drawn near the top of the page does not land in the bottom half — the flipped-axis bug, which is exactly the kind a viewer would render and a unit test would miss.
- Tiling: located lines carry bounds while unmatched text still carries its page, and the regions cover the page with no gaps or overlaps.
- A repeated line matches forward rather than reclaiming the earlier occurrence.
- A page too dense to position keeps its page and drops its geometry.
- Clipping stays within the evidence cap and keeps its pages, and keeps geometry when it fits.
- Normalization: a known rectangle maps to known fractions, an overhanging one is trimmed to the page edge, and unplaceable ones are dropped.

Parser fingerprints bumped again so documents reparse with positions.

Closes #865

### DIFF
--- a/crates/openwave-core/src/lib.rs
+++ b/crates/openwave-core/src/lib.rs
@@ -139,7 +139,7 @@ pub use model::{
     ToolCallRecord, ToolCallResolution, ToolCallStatus, TurnAgentRunWait, TurnAgentRunWaitSet,
     TurnAgentRunWaitStatus, TurnCheckpointProgress, TurnClientWait, TurnClientWaitStatus,
     TurnFailureReceipt, TurnFailureRetry, TurnRun, TurnRunStatus, TurnSteer, TurnSteerStatus,
-    MAX_ATTACHMENT_REVISION, MAX_MESSAGE_ATTACHMENTS, MAX_ROOT_ATTACHMENTS,
+    MAX_ATTACHMENT_REVISION, MAX_MESSAGE_ATTACHMENTS, MAX_ROOT_ATTACHMENTS, PAGE_BOUNDS_SCALE,
 };
 pub use preview::{
     format_bytes, ResultEntry, ResultEntryKind, ResultFailure, ToolActionPreview,
--- a/crates/openwave-retrieval/src/document.rs
+++ b/crates/openwave-retrieval/src/document.rs
@@ -6,8 +6,8 @@
 //! from the vector index (offsets in the store, text rehydrated from the source),
 //! and it gives every answer a precise, verifiable pointer back into the source.
 
-pub use openwave_core::{ByteSpan, SourceLocation, SourceRegion};
-use openwave_core::{ChatId, DocumentGeneration, ProjectId};
+pub use openwave_core::{ByteSpan, PageBounds, SourceLocation, SourceRegion, PAGE_BOUNDS_SCALE};
+use openwave_core::{ChatId, DocumentGeneration, ProjectId, RetrievalEvidenceInput};
 use serde::{Deserialize, Serialize};
 use std::borrow::Cow;
 
@@ -173,9 +173,19 @@ impl Document {
     }
 
     /// Return source regions intersecting `span`, clipped to its boundaries.
+    ///
+    /// Never returns more regions than a piece of evidence is allowed to carry.
+    /// A parser that resolves positions within a page emits many small regions,
+    /// and a span over dense text — a page of short table cells — can overlap
+    /// more of them than [`RetrievalEvidenceInput::MAX_SOURCE_REGIONS`] permits,
+    /// which would make the evidence invalid rather than merely coarse. When
+    /// that happens the regions are coalesced by page: the page a passage came
+    /// from is the part worth keeping, and the geometry within it is the part
+    /// worth losing.
     #[must_use]
     pub fn source_regions_for(&self, span: ByteSpan) -> Vec<SourceRegion> {
-        self.source_regions
+        let clipped: Vec<SourceRegion> = self
+            .source_regions
             .iter()
             .filter_map(|region| {
                 let start = region.span.start.max(span.start);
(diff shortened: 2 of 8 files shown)


naingthet:
Audited this against the Brightwave implementation and changed one thing.

**Granularity — changed.** BW's unit is the block: one `content` row per Reducto/Textract element, with that element's box. I had one region per line. Pushed a commit that positions blocks instead, taking the block's rectangle as the union of the lines located inside it. So a citation clipping the middle of a paragraph now marks the paragraph, which is what BW does. Side benefit: roughly an order of magnitude fewer regions on an ordinary page, so both caps below sit far away from real documents.

**Confirmed matching, having checked rather than assumed:**

- Coordinates normalized 0–1 against the page box, top-left origin — BW passes Textract/Reducto boxes through unmodified and the viewer multiplies by `100%`. Mine is the same contract, scaled to fixed point.
- A region carrying a page but no box is first-class in BW too: `content_type` is `text` vs `bounded_text`, and the non-PDF path still sets `page`. So the page-only fill is BW-shaped, not an invention.
- Page numbers one-based.
- Regions ordered and non-overlapping per document — BW enforces it with a GiST exclude constraint, we enforce it in `validate_source_regions`.
- Box is all-or-nothing — BW's `content__bbox__is_valid` check constraint; `Option<PageBounds>` gets it structurally.
- Gaps between regions are legal in both: BW's `join_by` separators are exactly that.

**Deliberate divergences, all forced by a different source of boxes:**

- BW's processor returns a box for *every* element, so every PDF row is `bounded_text`. liteparse does not offset-map lines into emitted Markdown, so some blocks cannot be placed and fall back to page-only. Unavoidable, and the reason the fallback had to exist.
- The two caps are ours. BW needs neither: it stores rows rather than a JSON blob, and has no per-evidence region cap. Ours exist because `MAX_SOURCE_REGIONS` rejects rather than truncates.
- We reject zero-area boxes; BW stores them. A box with no area cannot be drawn, so page-only is the better answer.
- BW returns boxes sorted by `(page, top, left)`; ours come back in text order. Only matters for which box a viewer scrolls to, and text order is arguably the better pick there.

Still divergent and pre-existing, not from this PR: BW keeps `element_type`/`section_header` on the content row, whereas we carry `heading_path` on the chunk instead.

Issue #888, opened 2026-07-29
chore(execution): confirm Daytona egress semantics live before treating it as a boundary


naingthet:
Out-of-band verification split from #819 (egress apply, PR #879). The egress config now applies to Daytona sandbox creation, but Daytona's enforcement is reported as `Unconfirmed` in the UI because two behaviors are unverified against the live Daytona API — this needs a Daytona account to test:

1. **Empty-but-present allowlist = deny-all?** We emit `networkAllowList`/`domainAllowList` present-but-empty on the axis with no entries, intending 'allow nothing on that axis'. Confirm Daytona reads an empty-present allowlist as deny-all, NOT as 'no restriction on that axis' (which would silently leave that axis open). Test: create a Daytona sandbox with a domain-only policy (empty CIDR list) and verify raw-IP egress is actually blocked.
2. **Vendor-curated always-reachable services.** Daytona keeps package registries, git hosting (github.com/gitlab.com), container registries, and AI APIs reachable regardless of policy. Confirm the exact always-allowed set so the `gaps` list the UI shows is accurate, and so we know whether any is a general-purpose exfil channel (git hosting is).

Only after this is confirmed can Daytona's status move from `Unconfirmed` toward `AppliedWithGaps`/`Boundary` per what the live behavior actually is. Until then the honest `Unconfirmed` projection stands. Needs the user's Daytona account / a live test run.

naingthet:
## Live confirmation — ran against a real Daytona account (2 test sandboxes, created + destroyed, account left clean)

Region `us`, snapshot `daytonaio/sandbox:0.8.0`, `sandboxClass: container`. Egress probed from inside via the toolbox `process/execute` API. HTTP code `000` = connection blocked/failed; `403` = the egress proxy explicitly forbade the request (traffic never reached the target).

### Test 1 — domain-only policy (`domainAllowList: "example.com"`, `networkAllowList: ""` — the exact empty-present shape the adapter emits)
| Destination | Result |
|---|---|
| `https://example.com` (listed) | **200 — reachable** |
| `https://example.org` (unlisted domain) | **000 — blocked** (DNS didn't even resolve) |
| `https://1.1.1.1` (raw IP) | **000 — blocked at connect** |
| `http://1.1.1.1` (raw IP) | **403 — proxy forbids** |
| github.com / pypi.org / registry.npmjs.org / crates.io / api.anthropic.com / api.openai.com | **all 000 — blocked** |

[...]

PR #893, opened 2026-07-29
fix(execution): correct Daytona egress enforcement — strict per-sandbox boundary (tier 3+)

### LINKED ISSUE #892: fix(execution): correct Daytona egress enforcement model per live #888 findings
See #888. Correct DaytonaExecutionProvider::egress_enforcement() — remove phantom curated-service exceptions (live test showed per-sandbox policy blocks everything unlisted incl github/pypi); mark empty-present=deny-all confirmed; move Daytona from Unconfirmed toward Boundary WITH a tier-3+ qualifier (per-sandbox override needs tier 3+; can't be known statically, so model honestly/fail-safe). Keep the #879 status-driven-by-model invariant. Update UI + docs.

naingthet:
## Summary

A live test against a real Daytona account (see #888) established that a per-sandbox Daytona egress policy is a **strict, externally enforced allowlist**, not the leaky surface this code assumed:

- Only listed domains are reachable. Raw IPs, unlisted domains, and unlisted-domain DNS are all blocked.
- The package registries, public git hosting, container registries, and AI APIs that the code declared as always-reachable general-purpose exceptions are **all blocked** under a per-sandbox policy — there is no such carve-out.
- An empty-but-present `networkAllowList` is read as **deny-all** on the IP axis (previously flagged NEEDS-LIVE-CONFIRMATION; now confirmed).

So Daytona per-sandbox egress is a genuine boundary, in fact **stronger than E2B** (which limits domain rules to HTTP/HTTPS ports and leaves DNS open). The prior model wrongly declared phantom curated-service exceptions, forcing `is_credential_boundary()` to `false`.

## Changes

- **`DaytonaExecutionProvider::egress_enforcement()`** — removed the phantom general-purpose curated-service exceptions. With no general-purpose holes, `is_credential_boundary()` now returns `true` for Daytona.
- **Adapter comments** — the empty-present `networkAllowList` = deny-all behavior is marked live-confirmed.
- **Enforcement reporting** — added an `EgressEnforcementStatus::ConditionalBoundary` plus an optional `requirement` note on the row. Daytona now reports as a boundary **conditional on org tier 3+** with the requirement surfaced inline, never an unconditional green boundary.
- **UI** (`CodeExecutionPanel.tsx`) — Daytona renders with an informational-tone badge (`Boundary — conditional`) and an inline `Requires Daytona org tier 3+` line explaining that on a lower tier the per-sandbox override is refused and the account default applies.
- **Docs** — `code-execution.md` and `sandbox-providers.md` describe Daytona as a strict boundary conditional on tier 3+, E2B as applied-with-gaps, and local as the only unconditional boundary.

## How the tier-3+ conditionality is modeled

**Static conditional note, not a runtime signal.** The enforcement disclosure is computed statically in `config_info` without any live Daytona call, and it renders before any sandbox is ever created — so a runtime echo from sandbox creation cannot honestly flow into it without a stateful, drift-prone tier cache. The account tier cannot be known statically, so the honest, fail-safe representation is a distinct `ConditionalBoundary` status carrying the `Daytona org tier 3+` requirement. This neither over-claims (not an unconditional boundary) nor under-claims (no longer buried as `Unconfirmed` now that #888 confirmed the mechanism), and the distinct status token prevents the UI from rendering a plain green boundary.

The #879 invariant holds: status is derived from the model (`is_credential_boundary()` + the exception set); the tier requirement is an additional host-known precondition passed into the projection alongside `confirmed`, exactly like `confirmed` itself.

**No egress-applying behavior changed** — the #879 create-body wiring is untouched; this is about the enforcement model/reporting being accurate.

## Verification

- `cargo test -p openwave-server -p openwave-code-execution --locked` — pass (regression test updated: Daytona is `ConditionalBoundary`, no phantom gaps, requirement surfaced; the route integration test updated to match).
- Wire types regenerated (`UPDATE_WIRE_TYPES=1`); `requirement?: string` and the new status member land in `wire.ts`.
- `tsc --noEmit` clean; `pnpm test` (vitest) 627 pass, including a new panel test asserting Daytona reads as conditional-with-tier-3+ and never as an unconditional boundary.
- `clippy --all-targets -D warnings` clean; `fmt` clean.

Closes #892

### DIFF
--- a/crates/openwave-code-execution/src/daytona.rs
+++ b/crates/openwave-code-execution/src/daytona.rs
@@ -3,9 +3,7 @@ use std::time::{Duration, Instant};
 
 use async_trait::async_trait;
 use openwave_core::SecretProvider;
-use openwave_egress::{
-    EgressEnforcement, EgressPolicy, EnforcementException, ExceptionReach, ExceptionScope,
-};
+use openwave_egress::{EgressEnforcement, EgressPolicy};
 use reqwest::{Client, Response, StatusCode, Url};
 use serde::{Deserialize, Serialize};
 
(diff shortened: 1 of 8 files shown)


naingthet:
## Additional live evidence — protocol/port completeness (validates `is_credential_boundary = true`)

Before trusting the boundary flag, I ran one more live probe (raw sockets, arbitrary ports, UDP) from a domain-only sandbox, because my earlier #888 test only used curl (TCP 80/443) and a credential boundary must hold across protocols/ports. Results:

| Probe | Result |
|---|---|
| TCP raw-IP `1.1.1.1:443` | connect **OPEN** — but this is the egress **proxy** accepting the connection; `curl https://1.1.1.1` → 000, i.e. data to the non-allowlisted destination is blocked past the handshake |
| TCP raw-IP `1.1.1.1:22`, `8.8.8.8:53` | **blocked** |
| TCP allowlisted `example.com:443` | **OPEN** |
| TCP allowlisted `example.com:22` | **blocked** (allowlist only applies to the proxied ports) |
| UDP `8.8.8.8:53` | **blocked (EPERM)** |

**Mechanism:** ports 80/443 are allowlist-enforced by an SNI/Host proxy; every other TCP port and all UDP egress are dropped at the netfilter level. Data egress to any non-allowlisted destination is blocked across all protocols/ports. This is protocol-complete for the credential-exfil threat model, so `is_credential_boundary = true` (tier 3+) is justified.

**One trap worth a code comment:** a raw TCP connect to `<any-ip>:443` SUCCEEDS (it hits the proxy), so a naive "is the network open?" reachability check would misread it as an open sandbox. Any future verification of Daytona egress must probe DATA egress (an actual HTTP request to an unlisted host → blocked), not a bare TCP connect. Worth noting near the enforcement declaration so nobody later "disproves" the boundary by observing the 443 connect succeed.

Issue #763, opened 2026-07-28
Managed-mode deferred minors (pruned): 4 real, rest cut


abaas-bw:
Deferred review minors from the gateway-authoritative managed-mode epic (#742) and its slices. **Pruned 2026-07-29** — the epic is merged, most bullets are done or noise. What remains below is a ruthless cut: 4 real items, 2 optional, and a considered-and-cut list kept only so nothing is silently forgotten.

## Worth doing (one tight PR)

- [ ] **Shell hooks run behind the managed sign-in gate.** While the gate is up on an unauthenticated managed profile, `useShellShortcuts` (Cmd+N → `client.createChat()`) and `useChatPromptWatcher` (10s poll) still run — the app does real work behind a screen that's meant to be a hard stop. Move those hooks inside the gated subtree. `AppShell.tsx` (~:84/:89 install; gate at ~:294). *The closest thing to a real bug in this list.*
- [ ] **No end-to-end test that a misconfigured policy blocks a real turn.** `misconfigured` is a fail-closed security state with only unit + DOM coverage. Add one behavior test: server resolves misconfigured → gate blocks → a turn cannot run. (The "drive a real turn" category CLAUDE.md prizes over intermediate unit tests.)
- [ ] **Userinfo accepted in the gateway *value*.** The deep-link parser rejects `user:pw@` in the link but not in the `gateway=` value, and the doc comment claims it's refused — the doc lies. One-line parser fix in `deep_link.rs`'s `provision_link` + correct the comment.
- [ ] **`cargo tauri dev` re-points the real `openwave://` scheme.** A dev run persistently re-registers the scheme to the dev binary (bit us during the deep-link work). Register a distinct `openwave-dev` scheme in debug builds. `deep_link.rs` (~:70).

## Optional (take only if bundling is cheap)

- [ ] **`pairing.log` has no size cap** on a remotely-triggerable write (`deep_link.rs` ~:388). One-line size cap.
- [ ] **Stalled pending sign-in blocks the managed app up to 5 minutes** (`SIGN_IN_TIMEOUT`) with only "open the sign-in page again" — no "start over." Add a cancel/restart affordance.

## Considered and cut (won't-fix — weighed, not forgotten)

- `mcp:<slug>` access-token prune in the keychain blob — never observable, capped by endpoints ever mounted, no leak (resource keys only reach Debug).
- `assert!` in the fake-gateway test axum handler — test ergonomics for a failure nobody is debugging.
[...]

Issue #742, opened 2026-07-28
epic: gateway-authoritative managed mode (single build)


abaas-bw:
## Vision

An organization provisions OpenWave entirely through its model gateway: sign in
to the gateway, download OpenWave, and everything — models, credentials, and
app integrations — arrives through the gateway session. In this managed
experience there is no BYOK and no hand-configured MCP servers.

## Architecture

**One build, one release channel.** Managed mode is runtime policy, not a
build flavor. At startup the app resolves policy in precedence order:

1. **OS-managed policy** (MDM): macOS managed preferences for the bundle id,
   Windows registry policy keys, Linux `/etc/openwave/`. Strictly enforced —
   on an MDM-supervised device the user cannot remove it.
2. **Sticky provisioned state**: written when the app is paired via an
   `openwave://provision?gateway=…` deep link from the gateway's page —
   **Download → Install → Connect**. Persisted in the profile store, not
   user-toggleable in the UI, and it survives relaunch and sign-out (session
   state in the keychain and policy state in the store are separate layers
   with separate lifecycles).[^1]
3. **Default**: the fully open experience (BYOK providers, manual MCP servers),
   exactly as today.

[^1]: A technical user can wipe this state, but that only yields the standard
    open build: no org credential exists on the device, and gateway-provisioned
    models and apps stay entitlement-checked, revocable, and audited
    server-side. The residual exposure is the same as any unmanaged install —
    addressed by the MDM tier and by gateway-attested MCP endpoints, not by
    this flag.

When managed policy is active: the app requires gateway sign-in (login-first
gate), the gateway base URL is locked, BYOK provider surfaces and credential
write paths are disabled server-side (including env-var fallbacks and the
legacy key migration), and manual MCP configuration is disabled. Entitled apps
mount automatically from the signed-in session.

**Boundary honesty:** local lockdown is a product boundary, not a security
boundary. Org resources stay protected server-side by gateway OAuth and
entitlements; the client never holds an org API key. Strict tamper-resistance
is claimed only on MDM-managed devices, where the OS enforces the policy.

## Slices

Each slice gets its own issue and lands as a focused PR.

- [x] #625 — mount gateway MCP endpoints from the signed-in session — merged
      in #748. **Prerequisite for the rest**: managed mode removes manual MCP
      configuration, so without turnkey mounting the entitled-apps list would
      be visible but unusable.
- [x] #750 — managed-mode policy state: resolution order above, persistence,
[...]

PR #901, opened 2026-07-29
feat(desktop): lock manual MCP configuration and reduce settings under managed policy

### LINKED ISSUE #895: Desktop: MCP lockdown and settings surface reduction under managed policy
Part of #742 — the final lockdown slice. When the resolved policy is managed:

- Manual MCP configuration is disabled: `PUT /mcp/servers` refuses
  definitions with `command` or `url` transports (409 `managed_profile`,
  matching the provider guards); gateway-endpoint mounts remain the
  sanctioned path. The `OPENWAVE_MCP_CONFIG` boot file is inert under
  managed policy. Persisted manual servers degrade visibly rather than
  silently vanishing (decide and document: disabled-with-diagnostic).
- Settings surface reduction: the Providers and manual-MCP editing surfaces
  hide or become read-only; the Model Gateway panel becomes the default
  settings section for managed profiles. Unmanaged profiles unchanged.
- Inherited parked items (from #763 / the #798 scope notes):
  - `GET /gateway/status` becomes policy-aware (today it reads
    configured/enabled/base_url from the stored row; a pure-MDM profile
    renders "not configured" while everything works).
  - `model_roles::usable_policy` threads the shared OsPolicySource and is
    managed-aware, so utility-model work (chat titling) resolves to a
    gateway model instead of silently failing against a locked BYOK
    provider.
  - Web-search and code-execution credentials under managed: decide and
    document (default: remain open, documented in the epic as deliberate —
    the gateway provisions no replacement for them today).

Unmanaged profiles must be byte-for-byte unaffected.

abaas-bw:
The final lockdown slice of #742: on a managed profile the manual MCP paths close, the settings surface stops offering what the server refuses, and the two items parked out of #763 / #798 land with it. Unmanaged profiles are unchanged.

## Manual MCP configuration is locked

`PUT /mcp/servers` refuses a body that **adds or edits** a `command` or `url` server with the same stable `managed_profile` 409 the provider guards use, before anything is validated or connected — so a refused write leaves the configuration exactly as it was. Gateway-endpoint mounts remain the sanctioned path.

The guard is on additions and edits rather than on every manual definition in the body because the Model Gateway panel mounts an endpoint by saving the *complete* server list: refusing any body that mentions a pre-existing manual server would make a profile's own history block the one path it is supposed to use. Removing a manual server is a candidate without it, so nothing is trapped in the configuration either.

**Persisted manual servers: disabled with a diagnostic, not deleted.** They stay in the stored configuration untouched — the definition keeps its `enabled` flag — and the runtime forces them down at connect/publish time with `"Disabled by managed policy. Gateway-managed MCP endpoints remain available."`. Deleting them would be data loss on a policy that can be lifted; leaving them running would be the lockdown failing. Supervision skips them and an explicit reconnect refuses, so nothing quietly walks them back up.

The `OPENWAVE_MCP_CONFIG` boot file is inert on a managed profile (with a `tracing::warn!` as the operator's diagnostic). It is a host-environment channel, which is exactly what the lockdown exists to close, so it is ignored outright rather than partially honored.

## Inherited parked items

- **`GET /gateway/status` is policy-aware.** When managed, `configured` / `enabled` / `base_url` come from the resolved policy, mirroring `connection()`. A pure-MDM profile — which may have no stored provider row at all — no longer renders "not configured" while sign-in, routing, and mounts all work, and the stored row can no longer read the gateway as off. A misconfigured managed policy (no URL) reads unconfigured, honestly.
- **`model_roles` threads the shared `OsPolicySource`** (the `NoOsPolicy` there was documented-deferred; this un-defers it, through `AppState` and the turn worker). A managed profile's curated role defaults all name BYOK providers the policy locked out, so the ordered walk is over the gateway's entitled models instead, in the order the gateway listed them — there is no cost signal to rank them by, and entitlement is the gateway's to describe. Utility work (chat titling) resolves to a gateway model instead of failing silently, and still degrades to `None` exactly as a no-model-configured install does when the gateway offers none. The `chat` role is untouched: it has no default list on purpose, and the managed substitution only applies to roles that walk one.

**Web search and code execution stay open under managed policy, deliberately:** the gateway provisions no replacement for them today, so locking them would remove capability rather than relocate it. The epic documents this.

## Renderer

Minimal visibility/read-only gating, not a redesign. `ManagedGate` publishes the policy it already resolved through a small context, so the panels read one answer instead of each re-fetching `/policy`. Under managed: settings opens on the Model Gateway section, the Providers section leaves the rail and its route renders a read-only explanation (a deep link must land on something legible), and the MCP panel becomes a read-only list — gateway mounts with their health, locked manual servers with the server's own reason, no editors or save/add controls. The gateway panel keeps its mount toggles. `GatewayStatus` semantics changed but its shape did not, so no wire types moved.

## Tests

- `a_managed_profile_refuses_manual_mcp_servers_and_accepts_gateway_mounts` — the whole write path over the real route: stdio, HTTP, and an edit of an existing manual server each refuse with `managed_profile` and change nothing; the gateway mount is accepted, carries the untouched manual definition along forced-disabled, and removing it still works.
- `managed_policy_forces_manual_servers_down_and_ignores_the_boot_file` — the runtime boundary: persisted manual servers disabled with the diagnostic and their definition intact, tools out of the registry, reconnect refused, gateway mount still resolving; and a managed profile whose only configuration is the boot file ends up with nothing configured or persisted.
- `a_pure_mdm_profile_reads_configured_from_policy_without_a_stored_row` — the status regression, with the fixture asserting there is genuinely no stored row.
- `a_managed_profile_resolves_the_utility_role_to_a_gateway_model` — the BYOK default before, nothing once managed with no session, the gateway model once signed in and synced.
- UI: the MCP and Providers panels' managed rendering (DOM), and the managed section list / default settings path.

No tests were removed.

## Verification

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, `cargo test --workspace`, `cargo check --workspace --locked` (no lock diff); under `crates/openwave-desktop/ui`: `pnpm install --frozen-lockfile`, `tsc`, `vitest run` (630 passed), and a production build.

Closes #895


abaas-bw:
Pushed 7d0c38a addressing the review.

**CRITICAL — a running manual server survived the policy flip.** The decision was live, the effect was not: a manual server connected while the profile was open kept its client and kept serving tools for the process lifetime, because the supervisor only *excluded* locked servers from probing and nothing else rewrote the registry. `McpRuntime::enforce_manual_lockdown` now does the teardown — drops the client, clears the prefetched views, publishes `Disabled` with the managed diagnostic, bumps the epoch so an in-flight reconnect lands stale and abandons its result, and republishes the tool registry. The supervisor runs it before each sweep, so an MDM push takes effect within one health interval; `pair_with_gateway` runs it inline, so the deep-link path has no window at all.

Pairing needed a runtime handle it did not have. Rather than a process-global, `PairingHandle` (store + live `McpRuntime`, from `Server::pairing_handle()`) is what the shell's deep-link task now carries — it cannot write policy without being able to apply it. The desktop's `PairingStore` watch channel carries the handle instead of the bare store.

**MAJOR — one-shot policy read in the renderer.** The gate now watches `/policy` on the same `SESSION_WATCH_MS` cadence it already runs for the session, and transitions managed state live. An unchanged answer is a no-op (`samePolicy`) so the poll does not re-render the app every five seconds; an error *response* fails closed to the blocked screen exactly as the first read does, while a transport failure leaves the last answer standing.

**Minors, all taken.** `reconnect_if_epoch` re-checks the lockdown against the definition as it stands after the per-server lock. The manual-addition check moved inside the mutation lock that commits the replacement (`replace_under_policy`), so the verdict and the commit see one definition set — the route no longer has to argue about the window. `status()` resolves policy once and passes it to a new `connection_for`. Managed role defaults now prefer the smallest entitled context window rather than first-listed: the curated lists these stand in for name each provider's cheapest row, and context window is the only price proxy the gateway gives us.

**Tests** (+4 Rust, +2 renderer; none removed):
- `a_running_manual_server_is_torn_down_when_policy_flips_managed` — a genuinely healthy HTTP server, `provision` in-process, then: tools gone from the registry, `Disabled` + managed diagnostic, `tool_count` 0, and a second sweep is a no-op.
- The route test now also refuses a rename and an enabled-toggle of an untouched manual server, pinning the whole-definition equality the guard rests on.
- The utility-model test lists a flagship first and asserts the walk still lands on the cheap model.
- `pairing mid-session flips the app to managed without a restart` renders a probe component that reads `useManagedPolicy()` — the same value the panels gate on — so the context publication is asserted rather than assumed; `a mid-session sign-out on a newly managed profile raises the gate` covers the gate coming up live.
- Nits: the read-only MCP list keys by name (the editable one still keys by index on purpose — names change under the cursor); the settings redirect renders a placeholder instead of `null`.

Deferred as you bucketed them: the boot-file-laundering baseline and the gateway-mount-only boot-file asymmetry go to #763.

Gates: fmt, clippy `-D warnings`, `cargo test --workspace` (1578 passed), `cargo check --locked` (no lock diff); UI `tsc`, vitest 632 passed, production build.

Issue #874, opened 2026-07-29
feat(server): drive a sandbox-resident run — admission, attach, exit criteria


naingthet:
Slice C of #823. Depends on Slices A and B.

Wire the host to drive a background agent run INSIDE a container: attached-only admission (local containers have no external lifetime cap), the DurableOperationStore (#858) backing the protocol host's operation log, the reverse-RPC model-inference path (host is model proxy while attached), the resumable event stream with committed-cursor drain, and teardown obligations.

Exit criteria (local-container-testable, from the delivery sequence): unplanned disconnect + reattachment; host restart reconciling the existing container rather than starting a second attempt; duplicate event delivery discarded by sequence number; reattachment draining buffered results before deadline evaluation; crash-between-provision-and-handle-commit reclaimed by the tag sweep; teardown driven to completion. The block_in_place multi-thread precondition from #870 binds here. Closes #823.

naingthet:
Consolidating: PR #919 landed the core drive path and the strongest exit criteria (exactly-once model proxying, duplicate-event discard by cursor, teardown on every terminal path, lease liveness, tag-sweep reclaim at the backend). Every remaining exit criterion here — host-restart reconciliation, reattachment draining buffered results, durable provisioning intent and teardown obligations, the reaper — is enumerated in #920, which also gates production routing. Tracking the remainder in one place; closing this in favor of #920.

PR #919, opened 2026-07-29
feat(server): drive a sandbox-resident agent run on a local container

naingthet:
The integrative slice for sandbox-resident runs: get one background agent run to execute **inside a container**, host-driven and attached-only.

> **Not production-wired.** Nothing routes background children to the `container` execution location yet — only this PR's own tests admit there. Routing is **gated on #920**, specifically the two blockers named there: the local-container transport is currently **unauthenticated** (any local process can connect to the published loopback port, complete the version-only handshake, and hijack the run's model channel), and there is **no host-side cap on the number of inference operations** a run may request (`MAX_STEPS` is enforced inside the untrusted container). Both need their own slices; neither is a regression of anything shipped, because nothing routes here.

## What the host does

1. **Admits** a background run at a new `container` execution location. `AgentRunExecutionLocation::Container` joins the tier/location split; `admit_sandbox_container_agent_run` writes it, and a widening migration relaxes the execution-location CHECK on both SQLite and PostgreSQL. The in-process scheduler's claim already filters to `in_process`, so it leaves container runs alone; `claim_container_agent_run` picks the exact run up under a bounded lease. Container runs admit with a **single attempt**.
2. **Provisions** a container through the `SandboxBackend`, committing a host-minted correlation tag before the create so the backend's orphan sweep can reclaim a container whose run never committed a handle. The run's delegated task rides the provisioning request and is delivered into the container by environment name (interim, until the run-init frame lands — the field says so).
3. **Connects** to the container's mapped loopback address with `WireClient::connect`, doing the version handshake.
4. **Drives** the run: backs the protocol host's operation log with the crash-safe `DurableOperationStore`; answers the sandbox's reverse-RPC **model-inference** calls with the host's own `ProviderResolver`, under the **same model-registry policy** the in-process worker applies (resolved once before any container exists, failed closed for an absent or unregistered model) — so no model credential lives in the container and the sandbox names no model, provider, or endpoint; drains the event stream, acknowledging by cursor; and commits the result through the run tier's **fenced** result path. It **heartbeats the lease** for the life of the drive and is **bounded by the run's absolute deadline**.
5. **Tears down** the container idempotently on every terminal path.

## Correctness details worth reviewing

- **Lease liveness.** A container run routinely outlives one lease period. The driver heartbeats (~15s against a 60s lease), and the in-process **lease reaper now skips container runs** — lease expiry means "the in-process worker died", and a sandbox-resident run has none. The deadline scan (no such exemption) remains the backstop. Without both, a run was failed out from under a container that was still working and still spending.
- **Terminating without a result.** The in-container agent ends without a `Result` on ordinary paths (step limit, failed model step) and the supervisor keeps serving afterwards — so a driver watching only for a result would wait on the open socket forever, never tear down, and leak the container. The event stream now carries a terminal `Failed` payload beside `Result`; the agent emits one on every non-result exit and the driver terminalizes on either.
- **Migration rollback** refuses up front while any `container` row remains, rather than failing partway through the SQLite rebuild (stranding the scratch table) or dropping the PostgreSQL constraint without re-adding it.

## Tests and CI

The non-Docker tests drive the **real** in-container agent loop (`openwave-sandbox-agent`) over a real loopback TCP socket, with only Docker and the host model mocked. The mock backend starts the agent on **the task the driver delivered**, the same relationship Docker has to the container's environment — so a driver that failed to deliver the task would run the image's default and fail the assertion. Coverage: end-to-end result committed exactly once with each model step proxied exactly once (op-log exactly-once) and the real task reaching the container; loop-ends-without-a-result terminalizing and tearing down; the lease heartbeat observed across several lease periods; the reaper exemption; admission routing; unreachable-container terminal failure with teardown. All timeout-wrapped, multi-thread runtime.

**CI:** a new post-merge `sandbox-resident container e2e` lane builds the agent image and drives a real container via `--ignored`, and **asserts the Docker daemon is present** rather than skipping silently into a green result. It is in the required aggregate gate, with its post-merge-only exemption asserted there (same treatment as the durable vector lane, and for the same reason — it is expensive). On pull requests the loopback tests are the coverage.

## Exit criteria: landed vs deferred

**Landed:** admission routing; reverse-RPC model inference recorded exactly once with a re-issue replaying the recorded response; teardown on every terminal path including loop-end-without-result; one-attempt terminal failure; duplicate-event discard by cursor; tag-sweep-reclaim (at the backend, whose tag the driver stamps); lease liveness and deadline bounding.

**Deferred to #920** (which now gates routing): transport authentication; per-run inference spend budget; durable provisioning-intent/handle + host-restart reconciliation; durable teardown obligation and the sweep that drives it (`reclaim_orphans` has no caller yet); the deadline reaper for a container whose driver died; late-result-as-evidence; cancellation (attached + cascade); the no-worker-lease persisted shape; the run-init frame; concurrency limits for container runs.

## Verified

`cargo test --locked` on openwave-core (567), openwave-server (456 + 5), openwave-sandbox-protocol, openwave-sandbox-agent; `clippy --all-targets` clean on all four; `fmt`; `cargo check -p openwave-core --no-default-features --features postgres --locked`; wire-types currency (`wire.ts` regenerated for the new location variant); `cargo check --workspace --locked` with no lockfile drift. The Docker e2e runs in the new CI lane post-merge.

Part of #874.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -454,6 +454,59 @@ jobs:
           cargo test -p openwave-core --no-default-features --features postgres
           --test postgres_turn_state --locked
 
+  sandbox-container:
+    name: sandbox-resident container e2e
+    needs: changes
+    # Builds the sandbox-agent image (compiling the agent crate's slice of the
+    # workspace inside Docker) and drives a real container end to end. That is
+    # expensive, and the container runtime is a detected capability rather than a
+    # hard dependency, so — like the durable vector lane — pay for it once per
+    # merge instead of once per push to every open PR. The rest of the stack is
+    # covered on every PR by the loopback tests, which drive the same driver
+    # against the real agent over a socket.
+    if: ${{ needs.changes.outputs.workspace == 'true' && github.event_name != 'pull_request' }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    env:
+      CARGO_INCREMENTAL: 0
+      CARGO_PROFILE_DEV_DEBUG: 0
+      CARGO_PROFILE_TEST_DEBUG: 0
+      SCCACHE_GHA_ENABLED: "true"
+      SCCACHE_GHA_RW_MODE: READ_ONLY
+      RUSTC_WRAPPER: sccache
+      RUSTFLAGS: -C link-arg=-fuse-ld=mold
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+      - name: Install toolchain (honors rust-toolchain.toml)
+        run: rustup show
+      - name: Set up compiler cache
+        uses: mozilla-actions/sccache-action@9e7fa8a12102821edf02ca5dbea1acd0f89a2696 # v0.0.10
+        with:
+          version: v0.16.0
+      - name: Install headless system deps
+        run: |
+          sudo apt-get update
+          sudo apt-get install -y --no-install-recommends mold
+      - uses: Swatinem/rust-cache@e18b497796c12c097a38f9edb9d0641fb99eee32 # v2
+        with:
+          shared-key: cargo-registry-v3-${{ hashFiles('Cargo.lock') }}
+          add-job-id-key: "false"
+          add-rust-environment-hash-key: "false"
+          cache-bin: false
+          cache-targets: false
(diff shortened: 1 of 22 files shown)


naingthet:
Deferred exit criteria are tracked in #920.

PR #921, opened 2026-07-29
feat(web-search): native page fetch policy and readability extraction engine

naingthet:
The agent can search the web but cannot open any result it finds. This is
the first of three slices that close that gap; it adds the engine only —
no tool, no trait change, no server wiring, so nothing is model-reachable
yet.

## What's here

**`fetch_policy.rs`** — a pure, always-compiled admission policy for
model-supplied URLs. HTTPS only, no userinfo, port 443 only, fragments
stripped, 2 KiB URL cap. Every IP literal and every DNS-resolved address
is checked against a deny list built on `openwave-egress`'s `CidrBlock`:
loopback, unspecified, RFC1918, link-local v4 (which covers the cloud
metadata endpoint) and v6, CGNAT, ULA, multicast, broadcast, reserved.
IPv4-mapped IPv6 and NAT64 addresses are judged by their embedded v4, so
`::ffff:10.0.0.1` cannot walk through the v6 path.

**`native.rs`** — the fetcher and readability engine, behind a new
`extract-native` feature. Redirects are followed manually, up to five
hops, and every hop re-enters the full policy with fresh DNS resolution;
connections are pinned to the vetted addresses so a second DNS answer
cannot land somewhere else. No cookies, no ambient credentials, an
honest User-Agent, a 4 MiB streamed response cap, and a content-type
allowlist. Extraction runs dom_smoothie in markdown mode under a
24k-character head-and-tail budget, and returns a typed
`NoReadableContent` rather than a plausible-looking fragment when a page
is a JavaScript shell or consent wall — callers must be able to tell
"extracted" from "nothing there".

## Why the policy is this paranoid

Every existing egress path in this crate is bound to one vendor's HTTPS
authority before credentials attach. This is the first code that fetches
an arbitrary URL chosen by the model, from the host process, so the
address-space and redirect checks are the boundary rather than a
belt-and-braces layer.

## Tests

Six of the twenty-one cover the invariants that are easy to reverse by
accident: denied address space including the IPv4-mapped and metadata
forms, redirect hops toward private space refused mid-chain, scheme and
userinfo and port rejects, budget truncation, and the JS-shell case. The
end-to-end extraction test drives a fixture page through the injected
transport; no test touches the network.

## Notes for review

- New transitive deps `cssparser` and `selectors` are MPL-2.0. That's
  file-level copyleft and link-compatible with our Apache-2.0, but it is
  the first MPL code in the workspace, so flagging it deliberately.
- Nothing enables `extract-native` yet, so its gated tests do not run in
  CI until the next slice wires the tool up — the same gap `http` had
  before it had a consumer. Worth landing that slice promptly.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6800,6 +6800,7 @@ version = "0.0.0"
 dependencies = [
  "async-trait",
  "chrono",
+ "dom_query 0.28.0",
  "dom_smoothie",
  "futures",
  "openwave-core",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1835,7 +1835,20 @@ version = "0.36.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "dae61cf9c0abb83bd659dab65b7e4e38d8236824c85f0f804f173567bda257d2"
 dependencies = [
- "cssparser-macros",
+ "cssparser-macros 0.6.1",
+ "dtoa-short",
+ "itoa",
+ "phf 0.13.1",
+ "smallvec",
+]
+
+[[package]]
+name = "cssparser"
+version = "0.37.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8c9cdaae01d5ed7882b04d795e7f752f46ff52d2fa3b50a20d28c464510bba98"
+dependencies = [
+ "cssparser-macros 0.7.0",
  "dtoa-short",
  "itoa",
  "phf 0.13.1",
@@ -1847,6 +1860,16 @@ name = "cssparser-macros"
 version = "0.6.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "13b588ba4ac1a99f7f2964d24b3d896ddc6bf847ee3855dbd4366f058cfcd331"
+dependencies = [
+ "quote",
+ "syn 2.0.118",
+]
+
+[[package]]
+name = "cssparser-macros"
+version = "0.7.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "10a2a99df6e410a8ff4245aa2006499ea662245f967cc7c0a38c83ef8eb44dbf"
 dependencies = [
  "quote",
  "syn 2.0.118",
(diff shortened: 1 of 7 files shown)


naingthet:
Pushed a hardening commit (47c7327) after an adversarial review of the fetch path. Summary of what it found, since it changes how the first commit should be read:

**The pinning was not actually in force.** The client was built without `no_proxy()`, and reqwest honors system and environment proxy configuration by default. When a proxy is in play the connector dials the proxy, so the addresses we resolved, vetted, and pinned were never consulted — meaning on any host with `HTTPS_PROXY` set (routine on managed desktops) the deny list and per-hop revetting were advisory, rebinding was live again, and reqwest turns proxy userinfo into an `Authorization` header on a request documented to carry no ambient credentials.

**A page could abort the process.** The markdown serializer recurses per nested list level, so markup well inside both the element cap and the body cap could overflow the stack — not catchable in Rust, it aborts, which takes the desktop app with it. Indentation was also quadratic in depth. Now the document is parsed once, refused past a depth cap, with the element cap lowered to match, and the parse moved off the runtime worker.

**Extraction had no overall deadline** — the timeout was per request and DNS was untimed, so a black-holed nameserver across five hops could hold blocking-pool threads for minutes on one tool call.

**Errors described the network.** A denied address was reported with the address in it, and "does not resolve" was distinguishable from "resolves into denied space", which together made this a name-existence and internal-address scanner. Both collapse to one opaque refusal now.

Also in: the reserved and test ranges the deny list was missing, judging the IPv4 embedded in 6to4 and Teredo addresses rather than only mapped and NAT64 forms, and control/bidi stripping on extracted text to match what the search path already did.

**On the tests.** The review's most useful finding is that the pinning tests were not testing pinning — the fake transport never reaches the real fetcher, so deleting `resolve_to_addrs` failed nothing, and the same was true of `https_only`, the redirect policy, and the response cap. Pinning, the redirect bound, the depth cap, and the resolver deadline now have tests that were each confirmed to fail with their defense removed. The pinning test lives in its own integration binary because it sets proxy environment variables.

One known gap: the streamed response cap has no socket-level test, because `https_only` would require a TLS server whose certificate the production client trusts, and injecting a test root into the shipped builder is a worse trade than testing the accumulator directly.

PR #931, opened 2026-07-29
feat(desktop): honest gateway mount rows — revoked, stale, and failed states

### LINKED ISSUE #899: Gateway panel: mount rows for revoked entitlements, staleness, and retry
Promoted from #763 (from the #748 review pair). (1) The endpoint list derives only from currently entitled apps, so a mount whose grant was revoked loses its row/toggle while supervision keeps retrying it. (2) The mount list is fetched once per sign-in transition — later degradation keeps a stale healthy line, and a fetch failure silently disables every toggle with no error or retry. One UX slice covering visibility, refresh, and honest failure states.

abaas-bw:
The gateway panel's MCP endpoint list derived only from currently entitled apps, so a mount whose entitlement was revoked lost its row and toggle here while supervision kept retrying it — it stayed visible only in the MCP servers panel. The mount list was also fetched once per sign-in transition: later degradation kept a stale healthy line, and a fetch failure set the list to null, silently disabling every toggle with no error and no way to retry.

## What changed

- **Union rows.** Endpoint rows are now the union of the entitled apps' slugs and the configured gateway mounts. A configured-but-no-longer-entitled mount keeps its row, rendered with an explanatory line ("No longer granted to your teams. Switch off to unmount it.") in place of the health line, and its unmount toggle keeps working.
- **Refresh cadence.** The mount list re-reads every 15s while the panel is visible. It is a local supervisor read (`listMcpServers`), not gateway traffic, so the cadence is cheap; health lines no longer rot at whatever they said on first load.
- **Honest failure.** A failed list read keeps the last-known rows and surfaces a named error with a Retry button inside the MCP endpoints block. Retry refetches immediately and restarts the cadence. Toggles are only disabled while the configured state is genuinely unknown (nothing loaded yet), and that state is now visible rather than silent.

Managed/unmanaged neutrality: no policy logic touched; the panel behaves identically in both modes.

## Tests

Extended `GatewayPanel.dom.test.tsx`:

- `keeps a row and unmount toggle for a mount whose entitlement was revoked` — no entitled apps, one configured gateway mount; asserts the row, the explanatory line, the checked toggle, and that toggling off puts an empty server list.
- `surfaces a failed server-list fetch as a retryable error, not dead toggles` — first `listMcpServers` rejects; asserts the named error and disabled toggle, then that Retry clears the error and re-enables the toggle.

Gates run locally: `pnpm install --frozen-lockfile`, `tsc`, full vitest suite (643 passing), production build. UI-only change; no Rust touched.

Closes #899

### DIFF
--- a/crates/openwave-desktop/ui/src/settings/GatewayPanel.dom.test.tsx
+++ b/crates/openwave-desktop/ui/src/settings/GatewayPanel.dom.test.tsx
@@ -268,6 +268,91 @@ describe("GatewayPanel", () => {
     );
   });
 
+  it("keeps a row and unmount toggle for a mount whose entitlement was revoked", async () => {
+    const revokedMount = {
+      name: "revoked-tools",
+      command: null,
+      args: [],
+      env: {},
+      env_from: [],
+      cwd: null,
+      url: null,
+      bearer_token_env: null,
+      gateway_endpoint: "revoked-tools",
+      request_timeout_ms: 60_000,
+      enabled: true,
+      health: "reconnecting",
+      tool_count: 0,
+      diagnostic: null,
+    };
+    const putMcpServers = vi.fn().mockResolvedValue({ servers: [] });
+    const client = api({
+      getGatewayStatus: vi.fn().mockResolvedValue(signedIn),
+      // No entitled app references the mounted slug any more.
+      getGatewayApps: vi.fn().mockResolvedValue({ supported: true, apps: [] }),
+      listMcpServers: vi.fn().mockResolvedValue({ servers: [revokedMount] }),
+      putMcpServers,
+    });
+    const user = userEvent.setup();
+    render(<GatewayPanel client={client} onChanged={() => undefined} />);
+
+    // The configured mount keeps its row, with an explanation instead of a
+    // health line.
+    expect(await screen.findByText("revoked-tools")).toBeInTheDocument();
+    expect(
+      screen.getByText(/No longer granted to your teams/),
+    ).toBeInTheDocument();
+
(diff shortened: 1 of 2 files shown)


abaas-bw:
Addressed the review in 8763a29:

- **Major 1** — the MCP endpoints block is no longer nested inside the apps-gated section. It renders whenever signed in and (union non-empty OR mount-list error pending), so a failed `getGatewayApps` or a pre-apps gateway can't hide rows, error, or Retry. With entitlements unreadable, rows degrade to configured mounts with a "Couldn't read your entitlements" note, and the revoked claim is only made after a successful apps read (`entitlementsKnown`). The "Connected apps" list keeps its original gating.
- **Major 2** — new test `keeps last-known rows and live toggles when a later refresh fails`: first read resolves with a healthy mount, the 15s interval read rejects, and the row, its health line, and its checked+enabled toggle must survive next to the error. Re-adding `setMcpServers(null)` to the catch now fails the suite.
- **Minors** — monotonic request-id guard (`mountsRequestRef`) so an in-flight interval read can't clobber a toggle write's fresher list or resolve out of order; `mcpServers === null` rows are labeled "Mount state unknown." instead of passing unknown off as unmounted; error-class prefix stripped from the rendered message.
- **Nits** — the "No connected apps" empty copy is suppressed above a populated (revoked-only) mount list; the revoked-row test asserts the health line is absent; the nested ternary flattened into a `rowNote` helper; shared `gatewayMount`/`incidentApps` test fixtures.

Gates re-run: tsc, vitest (645 passing, 12 in this file), production build.

PR #933, opened 2026-07-29
ci: type-check the Windows target

### LINKED ISSUE #897: CI: add a Windows-target cargo check lane
Promoted from #763 (from the #796 review). No CI lane compiles cfg(windows) code — runners are ubuntu/macos only, so RegistryPolicySource and any future Windows-gated code is type-checked by nothing and can merge broken. A cargo check --target x86_64-pc-windows-msvc step (cross-check, no tests) closes it.

abaas-bw:
No CI lane compiled `cfg(windows)` code — every runner is ubuntu or macos — so `RegistryPolicySource` (winreg, `managed_policy.rs`) and the other Windows-gated paths in openwave-server, openwave-core, openwave-host-broker, and openwave-desktop were type-checked by nothing and could merge broken.

This adds a `windows check` lane: `cargo check --workspace --locked --target x86_64-pc-windows-msvc` on `windows-latest`. Check only, no tests — it proves the Windows target type-checks, not that it behaves.

**Why a native Windows runner instead of a cheaper Linux cross-check:** `aws-lc-sys`, which nearly every workspace crate reaches, compiles C against Windows SDK headers (`windows.h`) that the hosted Linux and macOS images don't carry — a Linux cross-check fails in its build script before reaching any of our code. Scoping the lane to fewer crates doesn't help either, since the crates that carry the Windows-gated code (openwave-server, openwave-core) are inside the aws-lc-sys dependency cone.

Gate wiring, consistent with the existing lanes:

- Gated on the change-scope detector's `rust` output (same scope as clippy — the lane covers every crate, including openwave-desktop), and the detector's comment updated to name it.
- Required in the `Require every Rust lane` aggregate exactly like fmt/clippy/desktop: `success` when `rust == true`, `skipped` when `rust == false`, so it can't silently stop gating.
- Same sccache setup as the other Rust lanes (READ_ONLY on PRs, READ_WRITE on main). The cargo download cache uses its own `cargo-registry-windows-v1` key with this lane as its single main-branch writer, so it never races the `test` lane's `cargo-registry-v3` save.
- Covered by the existing workflow-level concurrency group; no new group needed.

Local verification (no Windows machine available): `actionlint` clean; the msvc cross-check reproduces the aws-lc-sys failure documented above; a full-workspace `cargo check --workspace --locked --target x86_64-pc-windows-gnu` with a mingw toolchain — the closest local stand-in that type-checks the same `cfg(windows)` Rust code — passes on this branch. The msvc lane itself runs on this PR (the ci.yml edit forces full validation), which is the authoritative proof.

One finding the check already surfaced: `turn_worker.rs:2251` warns `unused_mut` on Windows (the `mut` on `DirBuilder` is only used under `cfg(unix)`). Harmless for a warnings-tolerant check lane, left untouched here; it would matter only if a Windows clippy lane is ever added.

Closes #897

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -142,8 +142,9 @@ jobs:
             exit 0
           fi
 
-          # `rust` gates the lanes that cover every crate: rustfmt, clippy, and
-          # the openwave-desktop tests. `workspace` gates the lanes that cover
+          # `rust` gates the lanes that cover every crate: rustfmt, clippy, the
+          # Windows type check, and the openwave-desktop tests. `workspace`
+          # gates the lanes that cover
           # everything *except* openwave-desktop, so it can stay false when a
           # change cannot reach them. `workspace` implies `rust`; the gate below
           # asserts that.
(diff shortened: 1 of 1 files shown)


abaas-bw:
Parked in draft by decision (2026-07-29): the lane costs ~11 minutes per PR run, which is too heavy to require right now relative to the small cfg(windows) surface it protects. The branch stays current as the ready-made fix if/when Windows surface grows or a cheaper shape is found (e.g. scoping the check to the crates carrying windows-gated code, or running it on merge-to-main only rather than PRs — likely the right compromise). Issue #897 stays open tracking the coverage gap.

abaas-bw:
Sizing note for whenever this is revived, since the ~11 min was measured cold:

- **Merge-to-main-only** is the cheapest way to keep full coverage: the `vectors` lane already establishes the exact pattern (`github.event_name != 'pull_request'` plus an asserted `skipped` exemption in the aggregate gate), so the cost drops from every PR push to once per merge, at the price of post-merge signal.
- **Crate scoping saves less than it looks.** The cfg(windows) code lives in openwave-server, openwave-core, openwave-host-broker, and openwave-desktop — but openwave-server's dependency tree spans most of the workspace and aws-lc-sys dominates the cold compile, so `-p` scoping only meaningfully helps if openwave-desktop is dropped (cutting the Tauri/WebView2 tree, at the cost of the receipt_store.rs Windows path). `-p openwave-server -p openwave-core -p openwave-host-broker` is the sensible reduced scope if that trade is acceptable.
- The measured run was fully cold; sccache saves on main would trim steady-state somewhat, though check-mode metadata compiles aren't sccache-cacheable, so don't expect the Linux lanes' warm-cache ratio.

PR #956, opened 2026-07-29
feat(desktop): managed profiles pick from entitled models only

### LINKED ISSUE #946: Desktop: managed Models page — entitled models only, self-healing defaults
Part of #742 (gateway-native UX phase). Under managed policy the Models/roles page currently shows BYOK providers as '<provider> — unavailable' and role defaults can point at providers that can never serve them (requires refresh + manual re-selection to recover — see Brandon's field report).

Under managed: no provider dropdown (there is exactly one provider), a flat per-role list of the gateway's entitled models, role defaults that auto-resolve to a gateway-served model when the stored selection is not gateway-served (consistent with the picker's existing silent re-route and #901's utility-role handling), and live updates when policy flips or the model sync completes — no refresh needed. Unmanaged profiles byte-for-byte unchanged.

The 'unavailable rows remain visible for clarity' rationale only holds when the user can fix the unavailability; under managed they cannot.

abaas-bw:
Under managed policy the Models/roles settings page rendered every BYOK provider as "<provider> — unavailable", and role defaults could point at providers the policy can never let serve them — Brandon's field report needed a refresh, new defaults, and a picker reset before the app worked.

## What changed

**Managed shape of the Models page.** Under a managed profile the page drops the provider dropdown entirely (there is exactly one provider) and offers each role one flat list of the gateway's entitled models. The "unavailable rows remain visible for clarity" rationale only holds when the reader can fix the unavailability; under managed they cannot, so locked-out BYOK rows no longer appear at all.

**Self-healing defaults, resolved server-side.** A stored default the gateway does not serve renders as — and resolves to — the automatic pick rather than a dead selection. The resolution locus is the server: `resolved_role_key` in `routes.rs` is the single answer for what a role's default means (the settings page and the composer both label "automatic" with it), so the re-route lives there rather than being reimplemented in a client. The `utility` role already walked the gateway's entitled models cheapest-first under managed (#901); the `chat` role now joins it, landing on the first entitled model in the gateway's own order — the row the composer picker offers first — when its stored or boot default is locked out. The stored pin is left in place, so a profile returned to the open experience gets its selection back.

**Live updates.** The panel re-reads the catalog when the managed policy flips and runs a modest 5s watch while managed, so a completed model sync appears without a manual refresh — the same cadence `ManagedGate` and the gateway panel already use. Unmanaged profiles keep the read-once behavior and render byte-for-byte unchanged.

## Tests

- `model_roles_resolve_at_read_time_and_honor_an_explicit_pin` (extended, `openwave-server`): after a managed flip with a BYOK chat pin carried in from before it, the roles read reports the pin as stored but resolves chat to the gateway's first listed model and utility to its cheapest — extending the existing #901 coverage rather than duplicating it.
- `ModelsPanel under managed policy > offers one flat entitled list per role and reads a dead pin as automatic` (DOM): no provider combobox, entitled-only options, dead pin shown as "Automatic — <resolved>", selecting an entitled model PUTs its key.
- `ModelsPanel under managed policy > shows a completed model sync without a manual refresh` (DOM): the empty "not synced yet" state is replaced by the entitled list on the watch's next tick.

Verified locally with `cargo check -p openwave-server --locked`, `cargo clippy -p openwave-server --all-targets --locked`, the model-roles server tests, `tsc`, and the touched vitest files.

Closes #946

### DIFF
--- a/crates/openwave-desktop/ui/src/settings/ModelsPanel.dom.test.tsx
+++ b/crates/openwave-desktop/ui/src/settings/ModelsPanel.dom.test.tsx
@@ -1,8 +1,8 @@
 // @vitest-environment jsdom
 
-import { render, screen, waitFor } from "@testing-library/react";
+import { act, cleanup, render, screen, waitFor } from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
-import { describe, expect, it, vi } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 import type { ApiClient, ModelInfo, ModelRoleInfo } from "../api";
 import { ModelsPanel } from "./ModelsPanel";
 
@@ -53,6 +53,11 @@ const roles: ModelRoleInfo[] = [
   { role: "chat", selection: "gpt-4o", resolved_key: "openai::gpt-4o" },
   { role: "utility", selection: null, resolved_key: "openai::gpt-4o-mini" },
 ];
+
+afterEach(() => {
+  cleanup();
+  vi.useRealTimers();
+});
 
 describe("ModelsPanel", () => {
   it("picks a model per role and names what automatic resolves to", async () => {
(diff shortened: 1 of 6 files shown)


abaas-bw:
Addressed the review in 229449d:

- **Critical (display/accept divergence):** the managed chat re-route now lives in one shared function, `model_roles::effective_chat_policy`, called by both `resolved_role_key` (the label) and the turn-accept freeze in `send_message`. The routes.rs doc contract — the label a client shows for "default" is what the next turn actually gets — holds again. The roles test's managed tail now POSTs a message against the stale BYOK default and asserts the turn is accepted and frozen on `model_gateway::gateway-flagship`; reverting the accept-path change makes that send 409 and the test fail (verified by mutation).
- **Dead pin clearable:** with a dead pin the Select trigger holds no value (the automatic label renders as placeholder), and the automatic entry — "Automatic — Gateway Flagship (clears your previous Claude Opus 4.8 pin)" — is a real selection that PUTs null.
- **Preserved-pin contract stated:** the row says the previous provider selection is kept and restored if the device leaves managed mode, and that picking a model replaces it.
- Minors: unmanaged early return pinned by an assertion (dead default still resolves to its own key), entitled-list order sourced once in `model_roles` beside `gateway_defaults`, catalog watch at 15s with an in-flight guard, signed-out gateway reads "Sign in to your gateway to choose models." with the pickers disabled, duplicated chat hint collapsed onto an optional `managedHint`.

PR #957, opened 2026-07-29
feat(desktop): mount gateway endpoints from the MCP servers page

### LINKED ISSUE #947: Desktop: gateway MCP endpoint mounts move to the MCP servers page
Part of #742 (gateway-native UX phase). Mount toggles for gateway endpoints currently live in the Model Gateway settings panel; server health for those mounts lives on the MCP servers page. Consolidate: the MCP servers page gains a gateway-endpoints section carrying the mount toggles (consistent with #901's rules — gateway_endpoint definitions are exactly the writes managed mode allows), transferring #931's union/revoked/staleness/retry row logic wholesale. The Model Gateway panel keeps the Connected apps list as informational, pointing at the MCP page to mount. Open (unpaired) profiles see no gateway section on either page.

abaas-bw:
Mount toggles for entitled gateway endpoints lived in the Model Gateway panel, while the health of what they mounted lived on the MCP servers page. Deciding whether a mount was worth keeping meant reading two sections, and mounting had no view of what mounting had already produced.

The toggles move to a **Gateway endpoints** section at the top of the MCP servers page. The row logic transfers as it was: the union of entitled slugs and configured `gateway_endpoint` servers, the revoked-grant row that keeps its unmount toggle, the 15s health refresh, the retryable list failure that keeps last-known rows instead of dead toggles, the request-id guard against a slow read clobbering a fresher list, and the derived mount name that fits the 32-byte namespace.

The section reads the gateway status and the entitled apps itself (same supported/failure handling as the panel had) and renders nothing at all without a signed-in session, so an unpaired profile sees no gateway surface on either page. It keeps its own view of the configured servers, separate from the editable list around it, so a background refresh never replaces a form part-way through an edit; the list a mount write returns is adopted into that editable list only when nothing is unsaved.

Under managed policy the toggles stay live while every manual server above is read-only. No server change was needed: `manual_additions` refuses manual definitions that are *added or changed*, and a mount write sends the existing manual definitions unchanged alongside the new `gateway_endpoint` one, so it is admitted from this page exactly as it was from the gateway panel. Its doc comment, which named the gateway panel as the caller, now names this page.

The Model Gateway panel keeps **Connected apps** as an informational list — name, kind, and endpoint slugs, no toggles — with a button that opens the MCP servers section. The panel takes an `onOpenMcpSettings` callback rather than reaching for the router itself, matching how the other settings panels take their effects as props.

### Tests

Moved from `GatewayPanel.dom.test.tsx` to `McpPanel.test.tsx`, adapted to the new host (six tests): mount with a session-bound definition, derived mount name for a long slug, revoked-entitlement row with a working unmount toggle, retryable server-list failure, keep-last-known rows when a later refresh fails, and configured mounts surviving an unreadable entitlements read. Assertions that could now match the same server twice — the mount row and that server's own card further down the page — are scoped to the row via a `mountRow` helper, which localizes a regression better than the bare page query did.

Added: the gateway panel no longer renders mount toggles (asserted against a client that reports a configured mount) and its button routes to the MCP servers page; the endpoints section is absent without a gateway session even when a gateway mount is configured; and the section's toggle still writes on a managed profile.

Two assertions did not survive the move and are not worth reconstructing:

- `expect(screen.queryByText(/No connected apps are granted/)).not.toBeInTheDocument()` in the revoked-row test asserted that the gateway panel's empty apps copy did not sit above a populated mount list. Those two lists are no longer on the same page, so it now asserts nothing about the code.
- `expect(screen.findByText("revoked-tools"))` and the equivalent slug lookups were replaced by the row-scoped queries above; matching the bare slug on the MCP page is ambiguous — it is also the server's name in the list below — and a passing match would no longer tell us which surface rendered it.

Verified locally: `tsc`, the full vitest suite (652 tests), and a production `vite build`.

Closes #947

### DIFF
--- a/crates/openwave-desktop/ui/src/settings/GatewayPanel.dom.test.tsx
+++ b/crates/openwave-desktop/ui/src/settings/GatewayPanel.dom.test.tsx
@@ -31,8 +31,6 @@ function api(overrides: Partial<Record<keyof ApiClient, unknown>> = {}) {
     gatewaySignOut: vi.fn().mockResolvedValue(signedOut),
     syncGatewayModels: vi.fn().mockResolvedValue(signedIn),
     getGatewayApps: vi.fn().mockResolvedValue({ supported: true, apps: [] }),
-    listMcpServers: vi.fn().mockResolvedValue({ servers: [] }),
-    putMcpServers: vi.fn().mockResolvedValue({ servers: [] }),
     putProvider: vi.fn().mockResolvedValue({}),
     ...overrides,
   } as unknown as ApiClient;
@@ -208,6 +206,10 @@ describe("GatewayPanel", () => {
     );
     expect(await screen.findByText("Signed in")).toBeInTheDocument();
     expect(screen.queryByText("Connected apps")).not.toBeInTheDocument();
+    // The route to mounting still shows: older gateways mount by slug too.
+    expect(
+      screen.getByRole("button", { name: /Mount endpoints in MCP servers/ }),
+    ).toBeInTheDocument();
   });
 
   it("keeps connected apps informational and sends mounting to the MCP page", async () => {
@@ -222,28 +224,6 @@ describe("GatewayPanel", () => {
             app_kind: "rest_api",
             enabled: true,
             mcp_endpoint_slugs: ["example-security-tools"],
-          },
-        ],
-      }),
-      // A configured mount for that endpoint exists; this panel still shows
-      // no toggle for it — mounting lives on the MCP servers page now.
-      listMcpServers: vi.fn().mockResolvedValue({
-        servers: [
-          {
-            name: "example-security-tools",
-            command: null,
-            args: [],
-            env: {},
-            env_from: [],
-            cwd: null,
-            url: null,
-            bearer_token_env: null,
-            gateway_endpoint: "example-security-tools",
-            request_timeout_ms: 60_000,
-            enabled: true,
-            health: "healthy",
-            tool_count: 3,
-            diagnostic: null,
           },
         ],
       }),
--- a/crates/openwave-desktop/ui/src/settings/GatewayPanel.dom.test.tsx
+++ b/crates/openwave-desktop/ui/src/settings/GatewayPanel.dom.test.tsx
@@ -22,40 +22,6 @@ const signedIn: GatewayStatus = {
   model_count: 2,
 };
 
-/** A configured gateway mount as `listMcpServers` reports it. */
-function gatewayMount(slug: string, overrides: Record<string, unknown> = {}) {
-  return {
-    name: slug,
-    command: null,
-    args: [],
-    env: {},
-    env_from: [],
-    cwd: null,
-    url: null,
-    bearer_token_env: null,
-    gateway_endpoint: slug,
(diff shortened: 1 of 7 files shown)


abaas-bw:
Addressed the review in 76eb4fb — the top two majors structurally, per the suggestion:

**One list (majors 1+2).** `GatewayEndpoints` is now purely presentational; `McpPanel` owns the single server list, one read effect (initial load, Retry, and the 15s cadence while a gateway session exists — the doubled `listMcpServers` reads are gone), and the mount writes. Every fresh list lands through one reconcile: wholesale when nothing is unsaved, otherwise merged around the draft — gateway mounts follow the saved configuration by `gateway_endpoint` (so a Save after a mid-edit mount carries the mount instead of reverting it), manual rows keep their unsaved edits, and edited mount rows refresh only their health. `dirty` is mirrored in a ref, so a slow mount write or background read judges the present, not its captured render.

**Signed-out mounts (major 3).** The section stays visible with configured mounts — toggles off, a sign-in note — and disappears only on an unpaired profile with nothing to show, so the read-only copy never points at a section that isn't there.

**Docs (major 4).** `docs/mcp-servers.md` now says Settings → MCP servers → Gateway endpoints.

Minors: Retry clears the panel's action error too; `setMounted` failures surface `errorMessage()` and the retry tests re-pin the underlying message; the signpost renders whenever signed in (no `apps.supported` gate); the positional "Disabled below." copy is now position-free. New/upgraded tests: **"keeps a mount made during unsaved edits through the next save"** (pins the reconcile) and **"mounts on a managed profile, round-tripping a manual server unchanged"** (pins the admission contract with an exact-object match). Nits: dead `listMcpServers`/`putMcpServers` fixtures dropped from the gateway tests. Noting, per review: `mountName` truncates with `String.slice` (UTF-16 code units) against a limit documented in bytes — endpoint slugs are ASCII today, left as is.

Verified: `tsc`, full vitest suite (654), production build.

abaas-bw:
Final round addressed in 26c8386:

1. **Gateway-card Remove → Unmount.** Went with the setMounted route rather than hiding it: the card keeps a removal affordance, but it is labeled **Unmount** and goes through the same immediate mount write as the section's toggle — mounts have one owner, and the draft never carries a gateway deletion the reconcile would undo.
2. **Loading gated on the request id.** A superseded first read no longer clears `loading`; only the read that lands does, so no empty-list flash from a late-arriving invalidated read.
3. **Stale-capture test, reworked to the exercisable race.** The mount write disables the form while in flight (`working`), so the completion a reader can actually type against is the background list read. New test **"a background refresh landing mid-edit keeps the unsaved draft"**: defers the 15s cadence read, types during the flight, resolves, asserts the draft survives. Verified by mutation — forcing the reconcile to ignore the ref fails this test (and the dirty-save one); restoring passes.
4. Nits: `save()`/`reconnect()` catches use `errorMessage()`; the reconcile's ref-read carries a one-line soundness comment against GatewayPanel's stated convention (read-only, so a StrictMode double-invoke computes the same list twice); the apps-less signpost sits in a `SettingsSection`.

Verified: `tsc`, the two touched files 5× green, full suite 655/655. Not merging.

Issue #1024, opened 2026-07-29
Deep-link pairing: drop the restart prompt — restart automatically, and only when embedder enforcement is outstanding


abaas-bw:
## Problem

After a first provision, a second modal asks the user to restart now or later. The restart exists for exactly one enforcement gap: the embeddings client is boot-scoped (`resolve_embedder` — the vector index is dimension-bound to it), so a profile that booted with a BYOK OpenAI embedder keeps it live until relaunch. Everything else about managed enforcement applies without a restart.

## Change

Remove the dialog. After a pairing that newly provisions the profile:

- If the boot-resolved embedder was the offline one (fresh installs, OpenAI disabled or keyless — the common onboarding case), do nothing: there is no gap, and no restart happens at all.
- If the boot-resolved embedder was BYOK OpenAI, restart automatically (no prompt) to close the gap immediately. The server reports whether embedder enforcement is outstanding as part of the pairing outcome — the shell no longer guesses off `newly_provisioned` alone.

An idempotent re-pair continues to change nothing and never restarts.

With #1026 (provision commits on successful sign-in), the restart decision runs right after sign-in completes — a natural transition moment, and the relaunch comes back signed in and fully enforced.

abaas-bw:
Closing as obsolete rather than done: the restart's only justification was the boot-scoped embedder, and #1034 deleted the embedders entirely. After #1026 (provision commits on sign-in) there is no restart anywhere in the pairing flow and no enforcement gap left open.

PR #1032, opened 2026-07-29
feat(pairing): the sign-in screen is the consent — provision commits on successful sign-in

### LINKED ISSUE #1026: Deep-link pairing: the sign-in screen is the consent — provision commits on successful sign-in
## Problem

Pairing consent is a native OS dialog floating over the normal, unmanaged app UI, followed (on first provision) by a second restart dialog. Product direction: the user should never see the unmanaged app during onboarding, and the flow should have the fewest possible screens — ideally the login screen is the only thing they see.

## Change

Collapse consent into the sign-in screen. When a validated `openwave://provision` link arrives:

- The shell records the pending gateway URL (shell-validated, held natively — the renderer never supplies it). Nothing is written to policy.
- The app window surfaces covered by a full-window gate-style screen: "Sign in to <origin>", with the consequence stated (this device becomes managed by that gateway) and a Not now affordance. The unmanaged UI is never visible behind it.
- Clicking Sign in launches the OAuth flow against the pending gateway. This click is the consent: the flow must never auto-launch on link arrival, because a hostile gateway can auto-complete OAuth server-side with zero interaction.
- On successful sign-in, the server commits the provision (existing probe + provision path, using the natively held URL) and stores the session. On failure or cancel, nothing is written and the device stays unmanaged.
- Not now dismisses the surface and clears the pending pairing.

Conflict handling is unchanged: a link for a different gateway on an already-provisioned device gets the visible refuse-forever explanation (#932 decision).

## Why this is safe

Today's dialog gates the policy write on a confirmation click; this design gates it on a sign-in click at a surface naming the same origin, plus a successful authentication against that gateway. A drive-by page can at most pop open a login screen the user ignores; it can never durably manage the device. The write path stays out of renderer control: the pending URL is set only by a shell-validated deep link, and the renderer's signal is the sign-in it can already perform.

## Notes

- Supersedes the two-dialog flow from #795's UX (the security property is preserved, re-homed).
- Restart handling after commit is #1024 (conditional, automatic).
- The surface needs the title-bar drag region from #1025.

abaas-bw:
Closes #1026

Stacked on #1030 (the consent surface reuses its `WindowDragStrip`); GitHub will retarget this onto `main` when that merges.

## What changes

Clicking `openwave://provision?gateway=…` now lands the user directly on a full-window sign-in screen naming the gateway — no native dialogs, and the unmanaged app is never visible behind a pairing prompt. The flow:

1. **Shell** (`deep_link.rs`): a valid link is *registered*, not honored — the validated, normalized gateway URL is parked process-ephemerally via the new `register_pending_pairing`. Nothing durable is written and the gateway is not probed (probing on link arrival would hand any web page a localhost port-scan primitive). Both dialogs are gone; the only remaining native dialog is the conflict refusal (link for a gateway other than the managing one), which stays refuse-forever per #932.
2. **Gate** (`ManagedGate.tsx`): `GET /policy` now carries `pending_gateway_url` while the profile is unmanaged. The gate covers the window with the consent copy ("it will manage this device… cannot be undone from within OpenWave"), a **Sign in and connect** button, and **Not now** (a renderer-safe dismiss route — declining changes nothing durable).
3. **Commit** (`gateway_runtime.rs` / `pairing.rs`): `begin_sign_in` targets the pending gateway on an unmanaged profile. The OAuth exchange task, on success, runs `commit_signed_in_pairing` — re-resolve policy under the pairing lock (an MDM push that claimed the profile mid-flow wins and the pairing is refused), provision, apply MCP lockdown — and only then stores the session, so no session ever lands on a profile the gateway doesn't manage. Dismissals and re-registrations bump the sign-in generation, so a stale exchange abandons rather than committing a replaced pairing.

## Security framing

The old native dialog gated the policy write on a confirmation click. This gates it on a sign-in click at a surface naming the same origin, *plus* a completed authentication against that gateway — strictly more user intent than before. A drive-by page can at most pop open a sign-in screen the user ignores. The OAuth flow is never auto-launched on link arrival (a hostile gateway can auto-complete OAuth server-side; the in-app click is the consent gesture). The write path remains unreachable from any HTTP route; the renderer's only new power is dismissing a prompt.

## Restart

The "Pairing complete — restart" dialog is gone, and — since #1034 deleted the boot-scoped embedders — nothing about managed enforcement needs a restart at all. #1024/#1033 (conditional restart) were closed as obsolete.

## Tests

- `pairing.rs`: registration writes nothing durable and probes nothing / latest-link-wins / dismissal; typed conflict refusal parks nothing; commit provisions + stale snapshot never honored; commit refuses a profile claimed mid-flow; commit takes down running manual MCP servers. The old `pair_with_gateway` tests were reworked onto the new seams rather than deleted; the probe-failure case is gone with the probe itself.
- `gateway_runtime.rs`: unmanaged sign-in targets the pending gateway, starting commits nothing, dismissal restores the refusal (this is the write-path-guard test that replaced `deep_link.rs`'s `only_a_confirmed_pairing_reaches_the_write_path`).
- `ManagedGate.dom.test.tsx`: pending pairing gates the app behind the consent sign-in; Not now declines and returns the open product.
- Wire bindings regenerated (`pending_gateway_url` optional on `ManagedPolicy`).

## Not verified here

The full OAuth round-trip through a real browser and gateway, and a real `openwave://` link on macOS — both need the running app. Suggested smoke test: `openwave://provision?gateway=http://127.0.0.1:28081` against the dev gateway, checking (1) no dialogs, straight to the consent sign-in; (2) Not now returns the open app with nothing written; (3) completing sign-in flips the profile managed and lands signed in; (4) a second link for a different gateway shows the refusal dialog.


abaas-bw:
Heads-up for sequencing: this PR rewrites regions that two other in-flight drafts touch — #1029 edits `deep_link.rs` (userinfo-in-gateway refusal, dev scheme, log cap) and `ManagedGate.tsx`/`AppShell.tsx` (Start over, `GatedShellHooks`), and #1031 edits `gateway_runtime.rs` (`store_session` revoke-before-overwrite, retire rename). The scopes are complementary — no duplicated work — but whichever lands second needs a mechanical rebase. Notably #1029's "Start over" affordance and #1031's revoke-before-overwrite compose cleanly with the pending-pairing flow here (a started-over pending flow is invalidated by the same generation counter; a pairing commit's `store_session` gets the targeted revoke for free).

PR #1033, opened 2026-07-29
feat(pairing): restart automatically after a commit only when the BYOK embedder is live

### LINKED ISSUE #1024: Deep-link pairing: drop the restart prompt — restart automatically, and only when embedder enforcement is outstanding

abaas-bw:
Closes #1024

Stacked on #1032; GitHub will retarget when it merges.

The removed "Pairing complete — restart now/later" dialog guarded exactly one enforcement gap: `resolve_embedder` is boot-scoped (the vector index is dimension-bound to the embedder), so a profile that booted unmanaged with a live BYOK OpenAI embedder keeps embedding documents through it after a pairing commits. Everything else about managed enforcement applies live.

This closes that gap without any prompt, and only when it exists:

- `resolve_embedder` now also reports whether it picked the BYOK embedder; `bind_inner` stamps that onto the gateway runtime.
- A pairing commit that finds the flag set flips a `watch::Sender` the `Server` exposes as `enforcement_restart_watch()`.
- The desktop shell watches it after boot and calls `app.restart()` with a `pairing.log` line. The restart lands immediately after the sign-in that committed the pairing — a natural transition moment with nothing in flight, and the relaunch comes back signed in and fully enforced.
- Fresh installs (the normal deep-link onboarding case) boot with the offline embedder and never see a restart.

Tests: the runtime commit requests a restart exactly when the flag is set (`a_commit_requests_a_restart_only_when_the_byok_embedder_is_live`), and the `resolve_embedder` contract test now pins the reported pick in all three arms.

Not verified here: the actual `app.restart()` relaunch, which needs the running app — the suggested smoke test from #1032 extends naturally: pair on a profile that had OpenAI enabled+keyed and watch it relaunch straight into the signed-in managed state.

### DIFF
--- a/crates/openwave-desktop/src/deep_link.rs
+++ b/crates/openwave-desktop/src/deep_link.rs
@@ -253,7 +253,7 @@ async fn wait_pairing_handle(app: &tauri::AppHandle) -> Result<PairingHandle, St
 /// launches, `pairing.log` under app-data for GUI launches. Only the gateway
 /// origin is ever named — never the full link or URL — and gateway errors
 /// already strip URLs and token material.
-fn log_pairing(app: &tauri::AppHandle, message: &str) {
+pub(crate) fn log_pairing(app: &tauri::AppHandle, message: &str) {
     use std::io::Write;
     eprintln!("openwave-desktop: gateway pairing: {message}");
     let Ok(dir) = crate::data_dir(app) else {
--- a/crates/openwave-desktop/src/lib.rs
+++ b/crates/openwave-desktop/src/lib.rs
@@ -290,6 +290,22 @@ async fn boot_server(
         .initialize_store(server.store())?;
     // Unblock any pairing task parked on a deep link that arrived pre-boot.
     let _ = store_tx.send(Some(server.pairing_handle()));
+    // A pairing commit that finds this launch's BYOK embedder still live
+    // requests the restart that applies the embedder lockdown; fresh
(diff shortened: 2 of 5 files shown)


abaas-bw:
Obsolete: #1034 deleted the vector store and embedders, so the boot-scoped BYOK embedder this restart existed to swap out no longer exists. Nothing about managed enforcement needs a restart anymore — the answer to "is restart absolutely required?" is now simply no, with no conditional machinery.

PR #1113, opened 2026-07-30
fix(host-broker): stop mirroring read grants into write grants on migration

### LINKED ISSUE #1071: fix(host-broker): stop mirroring read grants into write grants on migration
The v2 to v3 state migration copies every `ReadFiles` grant into a `WriteFiles` grant and reuses the original consent record (`crates/openwave-host-broker/src/broker/state_file.rs:150-166`).

Anyone who attached a folder under read-only semantics silently gains write authority on upgrade, with an audit trail asserting they approved it. The consent record is the part that makes this bad: it is not merely a wider grant, it is a wider grant claiming provenance it does not have.

Granting write on new folder attachments is intended product behavior. Retroactively upgrading existing grants is not.

## Fix

Migrate read grants as read-only. New attachments carry write; existing ones acquire it through an explicit action or a one-time re-consent.

## Tests

One test: a v2 fixture with a read grant migrates to v3 with no write grant.

naingthet:
The version 2 to version 3 state migration copied every `ReadFiles` grant into a
`WriteFiles` grant and reused the original consent record for it. Someone who
attached a folder while the broker only understood reads came back after an
upgrade with the agent able to write into it, and with a consent record on file
asserting they had approved exactly that.

Version 2 grants now load as they were persisted. The consent record is the part
that mattered: the only consent available to attach to a widened grant is the one
the user gave for reading, so there is no honest way to mint the write grant here.
Write authority on an existing root has to come from a fresh consent.

Newly attached folders are unaffected — `prepare_registration` still mints read
and write grants together when a folder is picked, which is the intended product
behavior.

One test: a version 2 fixture carrying a read grant loads with the read grant
intact and no write grant.

Note for follow-up: roots registered before version 3 now load read-only, so a
user who had been writing to one since upgrading will find writes refused until
they re-attach the folder. That is the correct authority state, but the
re-consent path is worth designing rather than leaving as a surprise.

Closes #1071

### DIFF
--- a/crates/openwave-host-broker/src/broker/state_file.rs
+++ b/crates/openwave-host-broker/src/broker/state_file.rs
@@ -147,27 +147,14 @@ impl StateFile {
                 return Err(invalid_data("duplicate persisted operation identity").into());
             }
         }
-        let mut grants = persisted.grants;
-        if persisted.version == 2 {
-            let write_grants = grants
-                .iter()
-                .filter(|grant| grant.capability() == crate::Capability::ReadFiles)
-                .map(|grant| {
-                    Grant::from_consent(
-                        crate::GrantId::new(),
-                        grant.subject(),
-                        crate::Capability::WriteFiles,
-                        grant.scope().clone(),
-                        grant.consent().clone(),
-                    )
-                    .map_err(BrokerError::from)
-                })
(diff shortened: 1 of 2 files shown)


naingthet:
Filed the re-consent affordance as 1114 so the refusal path doesn't ship without a remedy.

Issue #1088, opened 2026-07-30
fix(core): advertise tools in a stable order


naingthet:
Found in a cross-reference audit against the Alpha implementation.

`ToolRegistry` iterates in `HashMap` order (`crates/openwave-core/src/agent.rs:213`), so the advertised tool block reorders between turns. That defeats prompt-prefix caching and makes runs harder to reproduce. Only the MCP server surface sorts (`openwave-mcp/src/server.rs:287`).

While here: advertised schemas strip `default` (`openwave-core/src/tool.rs:138`) while tools rely on `#[serde(default)]`, so the model is never told what omitting a field means.

PR #1153, opened 2026-07-30
feat(router): cache Anthropic prompt prefixes

### LINKED ISSUE #1087: feat(router): cache Anthropic prompt prefixes
Found in a cross-reference audit against the Alpha implementation.

There is no `cache_control` anywhere in the workspace. An agentic turn makes up to 16 model calls, each re-sending the full system prompt, tool specs, and transcript at uncached input price.

The `Usage` struct already carries `cache_read_input_tokens` and `cache_creation_input_tokens` (`crates/openwave-core/src/event.rs:349-351`), and `context.rs:151-154` has a comment about avoiding cache invalidation — the reasoning is there, the cache never was.

Alpha places ephemeral breakpoints on the last system block and the last tool.

Needs stable tool ordering to actually hit; see the companion issue.

## Tests

None. The usage fields simply stop reading zero.

naingthet:
An agentic turn makes up to 16 model calls, and every one of them re-sent the
system prompt, the whole tool advertisement, and the entire growing transcript
at full input price. Nothing in the workspace set `cache_control`, so the
`cache_read_input_tokens` / `cache_creation_input_tokens` fields on `Usage` have
always read zero.

The Anthropic adapter now places three ephemeral breakpoints, which is the whole
change:

- **Last tool definition.** Tools render first, so this caches the tool segment
  on its own. It is the only breakpoint that applies when a request carries no
  system prompt, and it still hits when two chats share a tool set but differ in
  their system prompt.
- **The system prompt.** Sent as a one-element block array rather than a bare
  string so it can carry the marker. Tools render before `system`, so this
  breakpoint covers tools *and* system; both breakpoints live inside the same
  prefix, so the tokens are written once either way.
- **The last content block of the transcript**, moved to the tail on every call.

Three, against a hard cap of four.

## Why the transcript breakpoint moves rather than anchors

A write costs about 1.25x uncached input and a read about 0.1x, so a breakpoint
placed where the prefix changes every call is strictly worse than no caching.
The tail is the one place in an agentic loop where that is not true: each step
appends an assistant message and its tool results and re-sends everything, so
the tail is exactly the boundary between what the previous call already wrote
and what this call adds. Each step writes only its own delta and reads
everything before it. The write premium is paid once per token and recovered on
the second read; in a 16-call turn most deltas are read many times over.

An anchored breakpoint, as in the reference implementation, would cache a fixed
early prefix and pay full price for all the growth above it. It is also the
worst place to anchor in this codebase specifically: context fitting mutates the
*head* of the transcript — it truncates there, replaces it with a checkpoint
summary, evicts old images to text stand-ins — while leaving the tail untouched.
An anchor would sit precisely where the invalidation happens.

## What has to hold for any of this to hit

Caching is a prefix match, so the stability of everything above a breakpoint is
what decides whether it earns its price:

- **Tool order.** `ToolRegistry::specs_for_foreground` iterates a `HashMap`, so
  the advertised order is currently random per process. That invalidates the
  tool and system breakpoints on every call and would make this change a net
  loss on its own. #1088 makes the order deterministic; **these two pair, and
  the caching only starts paying once that lands.**
- **The system prompt** is `AgentConfig::system_prompt`, fixed for a chat's
  configuration and unchanged across a turn's calls.
- **Minimum cacheable prefix.** Anthropic enforces a per-model floor — 512
  tokens on Opus 5, 1024 on Opus 4.8 and the Sonnet 5/4.6 line, higher on some
  older models. A short prompt with few tools may not clear it, but an
  under-length breakpoint is *ignored*, not charged, so there is nothing to gate
  on and no downside to placing it.

TTL is the 5-minute default. A turn's calls land seconds apart, so the 1-hour
TTL would double the write price to buy a window that is never needed.

One test covers this: the request JSON is a wire contract, and a regression here
is invisible except as a bill.

Closes #1087

### DIFF
--- a/crates/openwave-router/src/anthropic.rs
+++ b/crates/openwave-router/src/anthropic.rs
@@ -176,7 +176,7 @@ impl ModelProvider for AnthropicProvider {
 /// becomes a top-level field, and only `user`/`assistant` roles are valid on
 /// messages.
 fn build_request_json(req: &ChatRequest) -> Result<Value> {
-    let messages = req
+    let mut messages = req
         .messages
         .iter()
         .map(|message| {
@@ -186,6 +186,7 @@ fn build_request_json(req: &ChatRequest) -> Result<Value> {
             }))
         })
         .collect::<Result<Vec<_>>>()?;
+    mark_cacheable_transcript_tail(&mut messages);
 
     let mut body = json!({
         "model": req.model,
@@ -199,7 +200,13 @@ fn build_request_json(req: &ChatRequest) -> Result<Value> {
     // policy and make durable model attribution ambiguous.
 
     if let Some(system) = &req.system {
-        body["system"] = json!(system);
+        // Block form, not a bare string, so the prefix through the system
+        // prompt carries a cache breakpoint.
+        body["system"] = json!([{
+            "type": "text",
+            "text": system,
+            "cache_control": ephemeral_cache_control(),
+        }]);
     }
     if !req.tools.is_empty() {
         body["tools"] = Value::Array(
@@ -256,6 +263,9 @@ fn build_request_json(req: &ChatRequest) -> Result<Value> {
             )))
         }
     }
+    // After the whole tool array is settled, including a structured-output tool
+    // appended above.
+    mark_last_tool_cacheable(&mut body);
     if let Some(temperature) = req.temperature {
         body["temperature"] = json!(temperature);
     }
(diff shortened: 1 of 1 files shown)


naingthet:
Holding auto-merge until #1088 lands.

Tools render before `system` on the wire, so while `ToolRegistry` still iterates a `HashMap` the tool block reorders every call and invalidates the whole prefix — all three breakpoints included. Merged on its own this change pays the 1.25x cache-write premium on every request and reads nothing back, which is strictly worse than no caching at all.

Nothing wrong with the change; it just has a hard dependency on a stable prefix. Re-enabling auto-merge once #1088 is in.

Issue #1223, opened 2026-07-30
MCP App views hydrate empty after journal replay: history cards carry a synthetic call id the payload route rejects


abaas-bw:
Found during the connected-apps / MCP Apps smoke test against a local model-gateway (follow-ups context: #624).

During a live turn the MCP App card fetches `GET /chats/{chat}/calls/{call}/mcp-app-payload` with the real call id and gets 200; the sandboxed view renders. When the turn completes and the transcript rehydrates from the durable snapshot, `hydrateTranscriptHistory` mints `tool-history:${started_at}:${index}` ids (`TranscriptHistory.ts:101`) because `ChatToolActivitySnapshot` deliberately withholds the canonical call id, and `ChatTranscriptPresentation.ts` sets `callId: entry.id`. The remounted card's payload fetch then 400s (`Cannot parse call_id ... UUID parsing failed`), the failure is swallowed (the known #624 catch), and the view sits at its own loading state ("Waiting for a governed API call…") indefinitely while looking healthy. Every historical/reloaded MCP App card is affected — the working view exists only for the seconds between tool completion and transcript rehydration.

Fix: carry the real call id in `ChatToolActivitySnapshot` (the builder already has `call.id`; the payload route already exposes per-call data keyed on exactly this id to the same authenticated renderer, so this widens no boundary), regenerate wire.ts, and use it for the tool entry's `callId` in transcript hydration.

abaas-bw:
Verified end to end with the fix on this branch (PR #1228): reloaded the chat, the history entry carried the canonical call id, `GET /chats/{chat}/calls/{call}/mcp-app-payload` returned 200, and the gateway's audit-receipt view rendered fully inside the sandboxed frame — `listTeams · HTTP 200 · application/json`, the 'Governed by Model Gateway · credential injected server-side · call recorded to audit' strip, response headers, and the body tree. The matching gateway-side audit row is in `tool_call_events` (upstream 200, succeeded, attributed to the signed-in admin). Screenshot + full verification notes: https://claude.ai/code/artifact/3ce4b4cf-8a74-4e3d-b1e5-772c0e67ba2b (session artifact). Repro needed model-gateway#179 (steerable dev-fixture tool calls) since the fixture provider otherwise never emits tool_use.

abaas-bw:
Fixed by #1228 (merged 2026-07-30). Rehydrated MCP App cards now resolve their payload by the canonical call id; verified live with the audit-receipt view rendering after reload.

Issue #624, opened 2026-07-27
MCP Apps and gateway integration deferred follow-ups (pruned): 6 real, 3 optional, rest cut


abaas-bw:
Non-blocking findings from the review passes over the seven integration PRs
(#599, #604, #607, #612, #621, #622, #623). **Pruned 2026-07-29** — every
finding was verified against the code, not just re-read. What remains is a
ruthless cut: 6 real items, 3 optional, one superseded, and a
considered-and-cut list kept only so nothing is silently forgotten.

## Worth doing

- [ ] **A server request arriving mid-SSE-stream is answered only after the
  final response — and never at all on timeout.** The HTTP transport queues
  server→client requests and replies after the stream ends
  (`crates/openwave-mcp/src/http.rs` ~:168-196); the stdio path answers
  inline. Worse than originally reported: the whole call sits under the 60s
  `request_timeout`, and a timeout *drops* the future, so the queued
  fail-closed `method_not_found`/pong reply is never sent — a server that
  blocks on it becomes a 60-second hang plus "external server timed out"
  instead of a fast, clean error. The "reply POSTs cannot deadlock" comment
  at ~:191 doesn't hold (reqwest pools a second connection; the stream owns
  the `Response`, not `&self`), so replying inline compiles as-is. *The one
  real bug in this list.*
- [ ] **Theme changes never reach mounted app views.**
  `ui/notifications/host-context-changed` appears nowhere in the tree;
  `hostContext` is only sent in the `ui/initialize` reply
  (`McpAppBridge.ts` ~:99). The app itself observes OS theme flips
  (`theme.ts` ~:65), so every mounted view visibly keeps the stale theme
  until card remount — a light iframe in a dark transcript. Post the
  notification off `themeMode`/`matchMedia`.
- [ ] **The view payload fetch is one bare fetch with a swallowed catch.**
  `McpAppCard.tsx` ~:78-91 — a transient failure is terminal for the mount:
  the frame renders, never receives `tool-input`/`tool-result`, and sits in
  its own loading state forever while looking healthy. One retry or a
  visible "couldn't load data — retry" affordance.
- [ ] **Gateway model-sync errors reach Settings as a Rust `Debug` dump
  naming a provider the user doesn't have.** `gateway_runtime.rs` ~:362
  formats with `{error:?}`, so an admin clicking Refresh models can see
  `gateway model sync rejected: ServerError { status: 400, ... "openai_compatible supports at most 64 custom models" }`
[...]

PR #1228, opened 2026-07-30
fix(desktop): carry the canonical call id through transcript history so MCP App views rehydrate

abaas-bw:
Found during the connected-apps / MCP Apps end-to-end smoke test against a local model-gateway (deferred-items context: #624; the swallowed payload-fetch catch listed there is what made this silent).

## What breaks

During a live turn the MCP App card fetches `GET /chats/{chat}/calls/{call}/mcp-app-payload` with the real call id (200, view renders). On turn completion the transcript rehydrates from the durable snapshot, and history entries invented a snapshot-local id — `tool-history:${started_at}:${index}` — because `ChatToolActivitySnapshot` deliberately withheld the canonical call id. The remounted card's payload fetch then 400s (`call_id` must be a UUID), the failure is swallowed, and the sandboxed view sits at its own loading state ("Waiting for a governed API call…") while looking healthy. Every historical / reloaded MCP App view is affected; the working view exists only between tool completion and rehydration.

## Fix

- Expose the canonical `CallId` on `ChatToolActivitySnapshot`. The withholding stance predates MCP Apps: the payload route already keys renderer-readable data on exactly this id for the same authenticated client, so this widens no boundary — history cards just stop inventing an identity the server can only reject.
- Regenerate `wire.ts` (`UPDATE_WIRE_TYPES=1 cargo test -p openwave-server`).
- Use it as the tool entry's `id` and `callId` in `hydrateTranscriptHistory` / `presentChatTranscript` — which also makes history keys stable across snapshots instead of index-derived.

## Tests

- `TranscriptHistory.test.ts` asserts the hydrated tool entry carries the canonical id/callId (was pinned to the synthetic `tool-history:` shape).
- `ChatTranscriptPresentation.test.ts` asserts the presented tool message's `callId` is the wire `call_id` — the exact contract the MCP App card depends on.
- Full UI suite green (640 tests); `cargo test -p openwave-core --lib` green (514).

Verified end to end against a local gateway: with this fix a reloaded chat's MCP App card fetches its payload with the real id and the governed tool audit receipt view renders with data (screenshot in #1223 thread).

### DIFF
--- a/crates/openwave-core/src/db/ops/conversation.rs
+++ b/crates/openwave-core/src/db/ops/conversation.rs
@@ -952,6 +952,7 @@ fn tool_activity_from_call(
         ),
     };
     ChatToolActivitySnapshot {
+        call_id: call.id,
         tool: crate::RendererToolName::from(call.name.as_str()),
         action: crate::preview::ToolActionPreview::build(&call.name, &call.arguments),
         result,
--- a/crates/openwave-core/src/storage.rs
+++ b/crates/openwave-core/src/storage.rs
@@ -198,6 +198,16 @@ pub enum ChatToolActivityStatus {
 /// result it can carry is one a tool explicitly projects through a closed type.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, ts_rs::TS)]
 pub struct ChatToolActivitySnapshot {
+    /// Canonical call id, the same [`crate::id::CallId`] the live event stream
+    /// carried for this call.
+    ///
+    /// History withholds arbitrary call detail, but not this identity: the MCP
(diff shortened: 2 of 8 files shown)


abaas-bw:
CI caught the guard for the old contract: `transcript_tool_activity_is_allowlisted_and_redacts_canonical_tool_data` asserts the snapshot carries no canonical call id — exactly the decision this PR revisits. Amended deliberately rather than worked around: the call id moved from the redaction list to a positive assertion on the folded card (it's the payload key a rehydrated MCP App view resolves by), and every other canonical field — tool name, arguments, results, provider identity, error detail — stays asserted-hidden. Conversations module green locally (19/19).

PR #1500, opened 2026-08-04
fix(router): use OpenAI Responses for native models

naingthet:
## Summary

- add a native OpenAI Responses API adapter for the curated `openai` provider
- preserve reasoning effort alongside function tools instead of disabling reasoning to satisfy Chat Completions
- keep `openai_compatible` on the widely-supported `/v1/chat/completions` protocol
- normalize Responses text, reasoning summaries, tool calls, usage, incomplete responses, refusals, and failures into OpenWave provider events
- carry strict JSON Schema output, explicit tool choice, image inputs, and tool-result history over the Responses wire
- discard orphaned historical function-call pairs before request submission

## Root cause

Native OpenAI models were routed through the shared Chat Completions adapter. GPT-5.6 rejects a Chat Completions request that combines function tools with an active `reasoning_effort`, so switching a tool-enabled chat from Anthropic to OpenAI failed before streaming.

The earlier version of this PR worked around that by sending `reasoning_effort: "none"` whenever tools were advertised. That removed the 400 but also silently removed the user-selected reasoning capability from ordinary agent turns.

Alpha already handles native OpenAI through `/v1/responses`, where reasoning and function tools are supported together. This change adopts the same provider-boundary architecture in OpenWave: native OpenAI gets a Responses adapter, while third-party OpenAI-compatible endpoints retain Chat Completions.

## Provider contract coverage

- native `/v1/responses` endpoint selection
- compatible `/v1/chat/completions` endpoint preservation
- reasoning effort plus `summary: auto`
- streamed text and reasoning summaries
- streamed and completed function-call arguments without duplication
- function call/output history and orphan sanitization
- strict tools, tool choice, and JSON Schema output
- hydrated image inputs
- fresh/cached/output token normalization
- max-output termination, content-filter refusal, provider failure, and premature stream failure

## Verification

- `cargo test -p openwave-router --locked` (97 passed)
- `cargo check -p openwave-router --no-default-features --features openai --locked`
- `cargo check -p openwave-router --no-default-features --features openai-compat --locked`
- `cargo check -p openwave-server --locked`
- `git diff --check`

`cargo clippy -p openwave-router --all-targets --locked -- -D warnings` is currently blocked by unrelated `openwave-core` warnings already present on `main` (unused variables and dead code).

Live provider verification was not run because this worktree has no OpenAI API key.

### DIFF
--- a/crates/openwave-router/Cargo.toml
+++ b/crates/openwave-router/Cargo.toml
@@ -11,10 +11,12 @@ authors.workspace = true
 workspace = true
 
 [features]
-default = ["anthropic", "openai-compat", "gemini"]
+default = ["anthropic", "openai", "openai-compat", "gemini"]
 # The Anthropic (Messages API) provider adapter.
 anthropic = ["_http"]
-# OpenAI-compatible Chat Completions adapter (OpenAI, OpenRouter, vLLM, …).
+# Native OpenAI Responses API adapter.
+openai = ["_http"]
+# OpenAI-compatible Chat Completions adapter (OpenRouter, vLLM, …).
 openai-compat = ["_http"]
 # Google Gemini Developer API adapter (native GenerateContent protocol).
 gemini = ["_http", "dep:ring", "dep:uuid"]
--- a/crates/openwave-router/src/lib.rs
+++ b/crates/openwave-router/src/lib.rs
@@ -20,6 +20,9 @@ pub mod anthropic;
 #[cfg(feature = "openai-compat")]
 pub mod openai_compat;
 
+#[cfg(feature = "openai")]
+pub mod openai;
+
 #[cfg(feature = "gemini")]
 pub mod gemini;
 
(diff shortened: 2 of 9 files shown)


naingthet:
Heads-up on a conflict: #1492 (merged an hour ago) touched the same effort serialization from the opposite direction — it stopped emitting `reasoning_effort: "none"` for GPT-5.6 on the theory that the explicit `none` was being rejected, while this PR forces `none` whenever tools are advertised. Both theories are unverified against the live API (no keys in either worktree), and merged as-is this partially reverts #1492's behavior for tool-bearing requests.

Since #1492 also landed provider-400 body excerpts in the turn error, the cheapest way to settle this: run one turn against `gpt-5.6-sol` in the app and read the exact complaint — it will say precisely which parameter combination Chat Completions rejects. Suggest holding this until that error text confirms which shape is right (or rebasing and noting the interaction explicitly if you've already seen the live error).

PR #1521, opened 2026-08-04
perf(core): compile tool schema validators at registration

### LINKED ISSUE #1509: Validate tool arguments against the advertised input schema
Tool calls are executed with whatever arguments the model produced; the input_schema each tool advertises is never enforced at the call boundary. For built-in tools the argument structs give partial protection via deserialization, but externally-sourced tools (MCP servers) execute with arguments no schema ever checked, and schema violations from the model surface as confusing downstream tool failures instead of a clean typed refusal.

As MCP servers become a primary integration surface, this graduates from robustness gap to correctness/safety issue: a server's advertised schema is the contract the approval preview and the user reasoned about; arguments outside it should not execute.

Scope: validate arguments against the advertised JSON schema at the registry call boundary (jsonschema is already a workspace dependency), fail the tool call with a typed, model-readable error naming the violating field, and cover the MCP client path first. Bound validation cost and depth for adversarial schemas. One behavior test: an out-of-schema call is refused with the structured error and never reaches the tool.

naingthet:
Closes #1509

## Summary

- compile each tool input-schema validator once when the tool is registered
- expose schema mismatch checks through the shared registry so the existing agent and MCP-server dispatch refusals reuse the compiled validator
- keep invalid or draft-incompatible externally supplied schemas fail-open so a misconfigured tool is not made permanently unusable
- pin Draft 2020-12 for OpenWave's compact schemars-generated schemas and document the compatibility tradeoff for common draft-07 MCP schemas

Issue #1509 overstated the gap: schema validation already existed at both dispatch points. What was missing was compile-once registration and a unified registry accessor; this PR preserves the existing refusal behavior while removing per-call validator compilation.

## Verification

- `cargo fmt --all`
- `cargo test -p openwave-core -p openwave-mcp --locked`

### DIFF
--- a/crates/openwave-core/src/agent.rs
+++ b/crates/openwave-core/src/agent.rs
@@ -119,20 +119,51 @@ pub struct ToolRegistry {
 }
 
 #[derive(Clone)]
+struct RegisteredSpec {
+    spec: ToolSpec,
+    validator: Option<jsonschema::Validator>,
+}
+
+impl RegisteredSpec {
+    fn new(spec: ToolSpec) -> Self {
+        // ToolSpec schemas are draft 2020-12, including schemars-generated
+        // schemas whose compact provider form deliberately omits `$schema`.
+        // Invalid tool-authored schemas fail open: they are registration bugs,
+        // and refusing every call would make an existing tool unusable.
+        let validator = jsonschema::options()
+            .with_draft(jsonschema::Draft::Draft202012)
+            .build(&spec.input_schema)
+            .ok();
+        Self { spec, validator }
+    }
+
+    fn mismatch(&self, arguments: &Value) -> Option<String> {
+        self.validator
+            .as_ref()?
+            .validate(arguments)
+            .err()
+            .map(|error| error.to_string())
+    }
+}
+
+#[derive(Clone)]
 enum RegisteredTool {
-    Server(Arc<dyn Tool>),
+    Server {
+        tool: Arc<dyn Tool>,
+        registered: RegisteredSpec,
+    },
     Client {
-        spec: ToolSpec,
+        registered: RegisteredSpec,
         validate_arguments: Option<fn(&Value) -> bool>,
         class: ApprovalClass,
     },
     ForegroundClient {
-        spec: ToolSpec,
+        registered: RegisteredSpec,
         validate_arguments: fn(&Value) -> bool,
         class: ApprovalClass,
     },
     ForegroundOrchestration {
-        spec: ToolSpec,
+        registered: RegisteredSpec,
         kind: ForegroundOrchestrationKind,
         class: ApprovalClass,
     },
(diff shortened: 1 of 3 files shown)


naingthet:
Four things before this can land, one of which changes how we should think about #1509 itself.

**1. The premise of #1509 was stale, and the PR body inherits it.** Schema validation was already running at both dispatch points before this change — `agent.rs` called `crate::tool::schema_mismatch(&spec.input_schema, &arguments)` at the refusal point, and `mcp/src/server.rs` called it too. This PR is a compile-once-at-registration refactor plus a unified registry accessor, not the introduction of a missing control. That is still worth landing (compiling a validator per call on every tool call is real waste), but please retitle and rewrite the body to describe what it actually does. `Closes #1509` should stay, with a line in the body noting the issue overstated the gap and what was actually missing.

**2. Forcing Draft 2020-12 is a behavior change on schemas we do not author.** The old code used `jsonschema::validator_for`, which honors a schema's own `$schema` and picks the draft accordingly. The new `RegisteredSpec::new` pins `Draft202012` unconditionally. That is right for our own schemars-generated specs, but external MCP servers register tool specs through this same `register_client` path — `turn_worker.rs:2545-2572` exercises exactly that with a spec explicitly labelled an untrusted remote schema — and draft-07 is still the most common draft in the wild. A draft-07 schema using tuple-form `items` means something different under 2020-12.

The failure direction looks safe (a schema that won't compile under 2020-12 hits `.ok()` and fails open to today's behavior rather than falsely rejecting a valid call), so this is not a blocker, but it should be deliberate rather than incidental: either keep `validator_for`'s auto-detection for externally-supplied specs, or pin 2020-12 and say in the comment that external draft-07 schemas degrade to unvalidated rather than mis-validated. Right now the comment explains why `$schema` is absent from our own specs and doesn't mention external ones at all.

**3. A deliberate design decision lost its only coverage.** `schema_mismatch_names_the_violation_and_fails_open_on_a_non_schema` was deleted and nothing replaces its fail-open half. Fail-open on an invalid schema is a real decision — it is the difference between one tool being misconfigured and that tool being permanently unusable — and it is now enforced by an `.ok()` with no test behind it. Restore that case against the new registry API.

**4. The new test overclaims and reaches into internals.** `registry_compiles_draft_2020_12_schemas_once_for_every_tool_surface` covers the server and client surfaces, not the foreground-client or orchestration ones its name promises. And `schema_is_compiled` is a `#[cfg(test)]` accessor that asserts a private field is populated, which is close to the "asserts the code exists" pattern CLAUDE.md says to keep out. The behavior worth pinning is that a mismatched call is refused on each surface that can receive one — assert that through `schema_mismatch`, and either widen the test to the surfaces it names or rename it to what it covers.

naingthet:
APPROVED — all four points addressed.

The title and body now describe what the change is (compile-once at registration) and state plainly that #1509 overstated the gap. The draft comment at `agent.rs:126` now names the real tradeoff: external MCP schemas share this path, draft-07 is common, and its tuple-form `items` means something different under 2020-12 — so incompatible schemas fail compilation and stay unvalidated rather than being validated under the wrong draft. That is the right call and it is now legible to whoever hits it next.

`registry_fails_open_when_a_tool_schema_does_not_compile` restores the coverage that was lost, and the renamed test asserts refusal through `schema_mismatch` with the `cfg(test)` internals accessor deleted, so the assertion is about behavior rather than about a field being populated.

Merge when the lanes are green.

PR #1522, opened 2026-08-04
fix(router): translate tool schemas into Gemini's accepted subset

### LINKED ISSUE #1516: Gemini rejects tool schemas: input_schema is sent verbatim into a restricted OpenAPI subset
Any chat on a Gemini model fails with `invalid_request` as soon as the full foreground tool set is advertised — including a bare "hi" — because the request is rejected before any turn work happens.

```
gemini returned 400: Invalid JSON payload received.
Unknown name "type" at 'tools[0].function_declarations[1].parameters.properties[0].value': Proto field is not repeating, cannot start list.
Invalid JSON payload received. Unknown name "uniqueItems" at 'tools[0]…
```

`crates/openwave-router/src/gemini.rs:377` assigns `"parameters": tool.input_schema` verbatim. Gemini's `FunctionDeclaration.parameters` is not JSON Schema — it is a protobuf `Schema` message accepting a restricted OpenAPI 3.0 subset, so any keyword outside that subset is a hard 400 rather than an ignored field. Two distinct violations are visible in the one response:

- **`uniqueItems`** is not in the subset. We emit it deliberately on array arguments (`crates/openwave-core/src/agent_tools.rs:138`, `crates/openwave-core/src/client_tools.rs:95`).
- **`type` as a list.** The proto's `type` is a single enum, so a JSON Schema union type (`["string","null"]`, which is what an optional field serializes to) cannot be parsed at all.

The fix is a translation layer at the Gemini adapter boundary rather than degrading the schemas we advertise to every other provider — the same specs are correct JSON Schema and other adapters accept them as-is. Note the truncation in the error text: the two named violations are the first ones the parser hit, not necessarily the complete set, so the translation needs to be driven by Gemini's documented accepted subset rather than by patching the two keywords in this message.

Found via the provider error-body excerpt added in #1492.

naingthet:
Closes #1516

## Summary

- translate JSON Schema tool parameters at the Gemini adapter boundary using an allowlist of supported Schema fields
- recursively preserve object, array, enum, range, length, pattern, default, and supported format semantics
- collapse nullable JSON Schema unions to one concrete Gemini type plus `nullable: true`
- drop unsupported constraints such as `uniqueItems`, `additionalProperties`, and JSON Schema metadata without changing the shared tool specs
- fail the turn rather than send a degraded request when a tool schema cannot be translated

## Verification

- `cargo fmt --all`
- `cargo test -p openwave-router --locked`
- `cargo test -p openwave-core --locked`

### DIFF
--- a/crates/openwave-core/src/lib.rs
+++ b/crates/openwave-core/src/lib.rs
@@ -224,7 +224,7 @@ pub use tool::{
     ToolErrorCategory, ToolOutput, ToolScratch, ToolSpec, ToolUiView,
 };
 #[cfg(feature = "tools")]
-pub use tools::{CreateAppTool, ListDir, ReadFile, WriteFile};
+pub use tools::{create_app_tool_spec, CreateAppTool, ListDir, ReadFile, WriteFile};
 pub use user_questions::{
     ask_user_questions_tool_spec, validate_ask_user_questions_arguments, AnswerUserQuestions,
     AnswerUserQuestionsRequest, AskUserQuestionsArgs, PendingUserQuestions, UserQuestion,
--- a/crates/openwave-core/src/tools/mod.rs
+++ b/crates/openwave-core/src/tools/mod.rs
@@ -17,5 +17,11 @@ pub use list_dir::ListDir;
 pub use read_file::ReadFile;
 pub use write_file::WriteFile;
 
+/// Model-facing contract for the built-in local-app publisher.
+#[must_use]
+pub fn create_app_tool_spec() -> crate::ToolSpec {
+    definitions::create_app()
+}
+
 #[cfg(test)]
 mod tests;
--- a/crates/openwave-router/Cargo.toml
+++ b/crates/openwave-router/Cargo.toml
@@ -41,4 +41,5 @@ uuid = { workspace = true, optional = true }
 
 [dev-dependencies]
 axum = { workspace = true }
+openwave-core = { path = "../openwave-core", default-features = false, features = ["tools"] }
 tokio = { workspace = true, features = ["net", "rt-multi-thread", "test-util"] }
--- a/crates/openwave-router/src/gemini.rs
+++ b/crates/openwave-router/src/gemini.rs
@@ -1047,9 +1047,15 @@ fn classify_gemini_error(
 mod tests {
     use super::*;
     use openwave_core::provider::ChatMessage;
-    use openwave_core::tool::ToolSpec;
+    use openwave_core::tool::{Tool, ToolSpec};
     use openwave_core::{
-        request_folder_access_tool_spec, spawn_sandbox_agent_tool_spec, wait_for_agents_tool_spec,
+        ask_user_questions_tool_spec, exit_plan_mode_tool_spec, import_connected_file_tool_spec,
+        list_connected_folders_tool_spec, list_folder_tool_spec, read_connected_file_tool_spec,
+        request_folder_access_tool_spec, sandbox_folder_access_proposal_tool_spec,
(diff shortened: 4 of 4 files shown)


naingthet:
The translation itself is right, and building from an allowlist is the correct shape — a new schemars keyword can no longer become an unknown protobuf field. I verified the allowlist contents against the failure: `uniqueItems` and `additionalProperties` are absent from `GEMINI_SCHEMA_KEYWORDS`, and the union-type collapse produces `{"type": "STRING", "nullable": true}` as it should.

One structural concern, and it is about the test rather than the translator.

`gemini_tool_schema` has four hard-error branches — a union of two concrete types, a non-string type name, an array with no item schema, and a schema with no type/enum/anyOf. Each turns a translation gap into an `AgentError::Provider`, which fails the whole turn. That is a better failure than a provider 400 (it names our own limitation instead of surfacing protobuf parser noise), but the blast radius is identical to the bug this PR fixes: one tool spec that trips a branch takes down every Gemini chat, including "hi", because the full tool set is advertised on every turn. The regression test covers three hand-picked specs, so it cannot see that happen.

I ran the current full set through `gemini_tool_schema` locally to check we are not shipping into that state — all fifteen public `*_tool_spec()` constructors plus the three server-registered scratch tools (`read_file`, `list_dir`, `write_file`) translate without error today. So this is not blocking. But please widen the test to enumerate every tool spec rather than three, so that adding a spec with, say, an untyped property cannot silently reintroduce the outage. If enumerating the real registry from `openwave-router` is awkward because of the dependency direction, asserting over every public `*_tool_spec()` constructor plus the server tools' `Tool::spec()` gets almost all of the value.

Worth a sentence in the PR body too: the failure mode when translation cannot proceed is a failed turn, not a degraded request. That is a deliberate choice and reviewers should see it stated.

naingthet:
The expanded test is what I wanted — all 18 core-owned specs now go through `gemini_tool_schema` in one place, so a spec that grows an unsupported keyword fails here instead of in a user's chat. That closes the main gap.

One thing left, then this is good to merge.

**Four production tool schemas are still uncovered, and the crate boundary is why.** `agent_deps` (`openwave-server/src/lib.rs:1068-1141`) registers more than this test can see:

- `ExecTool` lives in `openwave-code-execution/src/tool.rs:17`
- `ListSourcesTool`, `ReadSourceTool`, `ReadToolResultTool` are `pub(crate)` in `openwave-server/src/source_tools.rs:26`

`openwave-router` depends only on `openwave-core`, so none of those four are reachable from this test — and `exec` is both the most complex schema we ship and the one most likely to gain arguments later. A spec regression there produces exactly the original bug: one bad schema breaks every Gemini chat, for every tool.

Two small changes:

1. Add `create_app_tool_spec()` to the list here — it is in `openwave-core`, so it belongs in this test and is currently missing.
2. Add a mirror test in `openwave-server` that builds the real registered set and asserts each spec translates. It can call the translator through the router's public surface, or assert the narrower property directly if exporting it is awkward. The point is that the assertion lives where the full registration list is visible, so adding a tool to `agent_deps` without a Gemini-safe schema fails a test rather than shipping.

If (2) turns out to need an awkward export, say so and I'll take a guard test in `openwave-server` that just walks the specs for the keywords Gemini rejects — less precise, but it covers the blast radius.

naingthet:
Scoping down my previous comment: add `create_app_tool_spec()` to the existing test's list — it is in `openwave-core` and was simply missed — and skip the `openwave-server` mirror test. The 18-spec test already covers the surface that broke, and the translator hard-errors on anything unsupported, so a server-side guard is coverage we can add later if a regression ever shows up there. One line, push, done.

naingthet:
APPROVED.

PR #1523, opened 2026-08-04
feat: attribute provider failures to the model and label unverified models

### LINKED ISSUE #1511: Attribute model failures to the model, and label unverified models honestly
The curated catalog only lists models we stand behind, but the OpenAI-compatible slot lets users point OpenWave at any endpoint/model. When such a model emits malformed tool calls or rejects requests, today's failure reads as an OpenWave error — the product takes the blame for the model.

Two small changes:

1. **Selection-time labeling.** Models outside the curated catalog show an 'unverified — not tested for agent work' hint in the picker and settings, one line, not a nag.
2. **Blame-attributing errors.** Turn failures caused by the model's own output (malformed/unparseable tool calls, schema violations, refused parameter shapes) name the model in the user-facing error ('<model> produced a malformed tool call') with the bounded provider detail from #1492's error surfacing, distinct from OpenWave-side failures.

Together with the conformance suite (#1508), this keeps the escape hatch open without letting a weak model make the app look broken.

naingthet:
## Summary

- add explicit verified/unverified tiers to every curated model and mark custom endpoint models unverified
- show a subtle Unverified chip with an explanatory tooltip in the model picker
- name the selected model id and provider on transcript turn-failure notices

## Verification

- `cargo fmt --all`
- `cargo test -p openwave-server`
- `pnpm test`
- `pnpm build`

Closes #1511


naingthet:
The chip and the wire tier look right. The failure attribution does not — as written it can name the wrong model, which is the specific failure #1511 was filed to prevent.

**1. The blame label is the chat's *current* model, not the model that ran the failed turn.**

`ChatRoute.tsx` computes `modelForSelection(models, chat.model ?? defaultModelKey)` once and `MessageList` applies that same value to every `turn_failure` message in the transcript. But the transcript message carries only a category — `ChatSessionReducer.ts:477-481` and `ChatTranscriptPresentation.ts:55-58` both build it as `{ id, role: "turn_failure", category }`, with no model on it.

So: a turn fails on model A, the user switches to model B — the single most likely thing a user does after a model fails — and the historical failure notice now reads "model B". Reload the chat and every past failure in its history is stamped with whatever is selected today. That is worse than no label, because it is confidently wrong and points the user at a model that was never involved.

The fix is to record the model on the failure rather than infer it at render time: carry the model id and provider on the turn-failure event through the projection alongside `failureCategory`, and have `TurnFailureNotice` read it off the message. Live-stream failures can take it from the turn that was in flight; historical ones need it persisted.

**2. Opus 4.8 and 4.7 are `Unverified` while Opus 5 is `Verified`.** If that is a real statement about what has been exercised, fine — but it means the chip appears on first-party Anthropic rows, which reads as a warning about the model rather than about our coverage. Worth confirming it is deliberate and not an artifact of which rows happened to get tested.

**3. No test covers the attribution itself.** `ModelMenu.test.tsx` renders the chip in isolation; nothing asserts that a failure notice names the model that actually failed. Once (1) is fixed, that assertion is the one worth having — it is the whole feature.

Holding the merge on (1).

naingthet:
APPROVED — this is the right fix.

The model now rides on the failure rather than being inferred at render time. `ChatTerminalTurnSnapshot` carries it from `turn.model` for history, `turn_failed` carries it for live turns, and the `failureModel` prop threading through `ChatRoute` → `ChatView` → `MessageList` → `MessageBubbleImpl` is gone entirely, so there is no longer a path by which the currently-selected model can reach a historical notice.

Both paths have a test that would catch a regression: `ChatSessionReducer.test.ts` asserts the live reducer puts the failing model on the message, and `ChatTranscriptPresentation.test.ts` asserts a rehydrated failed turn carries `failure_model` through. Those are the two ways a failure reaches the screen, and both are pinned.

Reading the id and provider off the turn rather than looking up a `ModelInfo` is also the better choice — a failure on a model since removed from the catalog still names it instead of rendering blank.

Merge when the lanes are green.

PR #1524, opened 2026-08-04
fix(desktop): report folder exec and write reach to the model

### LINKED ISSUE #1444: Exec and write reach are invisible on the model-facing folder surface
Split out of #1212, whose folders-UI half landed in #1390/#1404/#1410. The user can now see and revoke a folder's command reach; the agent still cannot see it, and in one place is told something untrue.

Attaching a folder mints read, write, **and** exec in one ceremony (`Broker::prepare_registration` — "choosing a folder is how the user says the agent may work in it"). Three places fail to carry that through:

1. **`list_connected_folders` drops the capabilities the broker computed.** `root_capabilities()` puts every candidate through the same `authorize()` call that gates the operation and hands back a `RootAccess { capabilities }`, but `serialize_result` in `crates/openwave-desktop/src/client_execution/folder_operations.rs` emits only `root_id` and `display_name`. The agent has no way to ask which connected folders it may write to or run commands in, so it discovers the answer by being refused.

2. **`RequestFolderAccessResult::Connected.capabilities` is a hardcoded `[read_files]`.** `connected_resolution` in `crates/openwave-desktop/src/client_execution.rs` builds it from a literal while the field is documented as "Capabilities the trusted host actually granted". After a successful pick the agent is told it got read, when it got read, write, and exec.

3. **`RequestedFolderCapability` in `openwave-core` still only knows `ReadFiles`,** so the request cannot name exec even as a proposal, and the result type has no variant to report it with. `RequestFolderAccessArgs::is_well_formed` pins the list to exactly `[ReadFiles]` and there is a turn-worker test asserting `["write_files"]` is refused — deliberate, so widening the request schema is its own decision rather than a drive-by.

The cheapest coherent shape is probably to give the *result* its own capability vocabulary (the result is serialized JSON, not an advertised schema) and report the broker's real `root_capabilities`, leaving the request proposal narrow until the grant ladder actually narrows. Fixing 1 and 2 is worthwhile on its own; 3 only becomes load-bearing if attach ever stops granting all three.

naingthet:
## Summary

- report each connected folder’s broker-authorized read, write, and command capabilities in `list_connected_folders` results
- return the broker’s live capability set after folder consent instead of a hardcoded read-only result
- keep folder access requests limited to the existing read-only proposal vocabulary while using a separate result vocabulary for actual grants

## Tests

- `cargo fmt --all --check`
- `cargo test -p openwave-core --locked`
- `cargo test -p openwave-desktop --locked`
- `cargo test -p openwave-server --locked`
- `UPDATE_WIRE_TYPES=1 cargo test -p openwave-server the_generated_bindings_are_current --locked`

Removed the old connected-result unit assertion because the result now intentionally depends on live broker authorization; the new broker-to-serialization regression test covers that boundary directly.

Closes #1444

### DIFF
--- a/crates/openwave-core/src/client_tools.rs
+++ b/crates/openwave-core/src/client_tools.rs
@@ -45,6 +45,23 @@ pub enum RequestedFolderCapability {
     ReadFiles,
 }
 
+/// One folder capability the trusted host reports as currently granted.
+///
+/// This vocabulary is deliberately separate from [`RequestedFolderCapability`]:
+/// the request remains a narrow, untrusted proposal while the result reports
+/// the broker's real authorization decision.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+#[non_exhaustive]
+pub enum GrantedFolderCapability {
+    /// List directories and read files below the selected folder.
+    ReadFiles,
+    /// Create files or replace them after any required approval.
+    WriteFiles,
+    /// Expose the selected folder to model-authored commands.
+    ExecuteCommands,
+}
+
 /// Non-authoritative, well-known starting location for the native picker.
 ///
 /// This is deliberately not a free-form path. The trusted desktop decides how
@@ -113,14 +130,14 @@ pub struct RequestFolderAccessArgs {
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
 #[serde(tag = "status", rename_all = "snake_case", deny_unknown_fields)]
 pub enum RequestFolderAccessResult {
-    /// The user selected a folder and the broker reports live read access.
+    /// The user selected a folder and the broker reports its live capabilities.
     Connected {
         /// Opaque broker-local root identity, never a host path.
         root_id: uuid::Uuid,
         /// Bounded leaf name safe for display.
         display_name: String,
         /// Capabilities the trusted host actually granted.
-        capabilities: Vec<RequestedFolderCapability>,
+        capabilities: Vec<GrantedFolderCapability>,
     },
     /// The user declined or closed the picker; no access was granted.
     Declined,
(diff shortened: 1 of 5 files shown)


naingthet:
Direction is right — the model was being told read-only about folders it could actually write to and run commands in. Two things to settle before this lands.

**1. The deleted test took a security invariant with it, and the replacement does not cover the same code.**

`terminal_results_are_typed_and_path_free` asserted that `connected_resolution`'s serialized result contains no host path. The new test exercises `serialize_result` in `folder_operations.rs`, a different function in a different module. After this change nothing asserts that the folder-consent result is path-free.

The PR body's reason — the result now depends on live broker authorization — argues that the test needs a broker fixture, not that the invariant should stop being checked. `connected_resolution` is now the function that reads `RootAccess` straight from the broker and serializes it toward the model, so it is more worth pinning than before, not less. Please restore the path-free assertion against the new async shape.

**2. `Capability` is `#[non_exhaustive]`, so the `_ => None` arm silently drops future capabilities.**

`capability.rs:21` marks the enum non-exhaustive, which is why the wildcard is there. The consequence is that any capability added later is reported to the model as absent until someone remembers to extend two separate match blocks — one in `client_execution.rs`, one in `granted_folder_capabilities`. Under-reporting is the safe direction, so this is not a blocker, but it is the same "the model believes it has less reach than it does" bug this PR is fixing, deferred to the next capability. Worth collapsing the two matches into one shared conversion so there is a single place to update, and dropping a comment on the wildcard saying what it costs.

**3. New failure path on consent.** `connected_resolution` can now return `the selected folder is no longer available to this conversation` if the root is missing from `ListRoots`. Previously consent could not fail after the picker succeeded. Is that reachable as a race between attachment completing and the root appearing in the listing? If so the user sees an error immediately after granting access, which is a bad moment to fail.

naingthet:
Addressed all three review points in the latest commit:

1. Restored `terminal_results_are_typed_and_path_free` against the connected-result serialization path, including broker-derived read/write/exec capabilities.
2. Collapsed broker capability conversion into one shared helper. Its `Capability` wildcard now documents that unknown future capabilities are deliberately under-reported until the result vocabulary is extended.
3. Confirmed the post-consent miss is reachable as a race: revocation, root unavailability, or broker transport failure can occur after product attachment completes but before `ListRoots` returns. That race no longer turns successful consent into an error; it returns the pathless connected identity with an empty capability set, and a later `list_connected_folders` call reports current reach.

Re-ran `cargo test -p openwave-desktop --locked`, `cargo check --workspace --locked`, formatting, and diff checks.

naingthet:
APPROVED — path-free invariant restored against the extracted pure helper, single shared capability conversion with the non-exhaustive wildcard documented, and the post-consent race resolved by under-reporting instead of failing after consent. Arming auto-merge.

Issue #1454, opened 2026-08-04
Wire up the background agent subsystem for v1


naingthet:
Decision: background agents are v1 scope, not deferred.

The subsystem is substantially built but never wired in. The `spawn_agent` / `wait_for_agents` tools are defined in `crates/openwave-core/src/agent_tools.rs` but are **never registered** — `agent_deps()` in `crates/openwave-server/src/lib.rs:1052-1080` registers the file/exec/source/app/web tools and nothing else. Consequently:

- Every caller of `admit_sandbox_agent_run` / `accept_sandbox_agent_run_and_park_turn` is inside `#[cfg(test)]` modules (`sandbox_agent_run_worker.rs`, `sandbox_web_search_worker.rs`).
- The workers are spawned at startup (`lib.rs:1002-1005`) and poll forever against tables nothing fills.
- ~14 tables (`sandbox_agent_admission`, `sandbox_provision`, `sandbox_tool_call`(+receipt), `agent_run_inbox`, `agent_run_result`, `agent_run_cancellation`, the `turn_agent_run_wait*` family, `operation_log`, the background tier of `agent_run`/`agent_run_claim`) have no production write path.

Work to bring it live:

1. Register `spawn_agent` / `wait_for_agents` in `agent_deps()` and give them an approval class / permission-mode treatment (spawning an agent is at least `Sensitive`-adjacent — decide deliberately).
2. Drive one real end-to-end turn: spawn → child run executes in the sandbox → result consumed → parent turn resumes. The dormant `turn_agent_run_wait` legacy path should be deleted first rather than revived (see the schema-consolidation issue).
3. Surface in the UI: minimum viable is child-run status on the transcript; live inspection (#954) and mid-run steering (#951) remain follow-ups.
4. Reconcile with #953 (connected apps inside the sandbox agent) — still gated on gateway/MCP setup; not required for first wiring.
5. Schema: since these tables are now being committed to at v1, do the receipt/admission consolidations before the baseline freezes rather than after.

Related: #951, #953, #954. The scheduler-table growth bug and the pre-v1 schema consolidation are filed separately.

naingthet:
Audited this before building, and most of the premise no longer holds — the subsystem is wired and live on the default path. Recording what I found so the remaining scope is visible.

[...]

Issue #954, opened 2026-07-29
feat: live sibling/subagent inspection for background agents


naingthet:
OpenWave has durable join (wait_for_agents) but no way to observe a background agent's live status/progress while it runs. Add a read model exposing live background-run status — state, recent events/progress — so a parent (or the UI) can inspect running children without waiting for terminal results.

Scope: a renderer-safe live read model over the agent-run event stream (the resumable, sequenced stream already exists in the protocol). Read-only; no new authority. Complements the existing GET /chats/{id}/agent-runs read model noted in docs/agent-runs.md.

naingthet:
Claim released: no open PR and the working branch has been idle since the 2026-07-31 claim (no commits after 2026-07-30 anywhere matching this scope); re-claim before resuming.

naingthet:
Labeled deferred: sandbox-resident agent work is deliberately parked per the current project direction; re-claim when that track reopens.

Issue #951, opened 2026-07-29
feat: steer a live sandbox-resident agent (host to agent, mid-run)


naingthet:
Background agents are fire-and-join today (spawn then wait_for_agents). There is no way to inject guidance into a running agent. Add a steering capability: the host sends a mid-run message/instruction to a live sandbox-resident agent over the existing host-held transport (crates/openwave-sandbox-protocol), and the agent incorporates it on its next step.

Scope: a steer frame on the wire protocol, host-side API to send it, agent-side handling to fold it into the loop. Attached-only initially (needs a live connection). Depends on the wire transport (merged, #900) and production wiring.

naingthet:
Claim released: no open PR and the working branch has been idle since the 2026-07-31 claim (no commits after 2026-07-30 anywhere matching this scope); re-claim before resuming.

naingthet:
Labeled deferred: sandbox-resident agent work is deliberately parked per the current project direction; re-claim when that track reopens.

PR #1525, opened 2026-08-04
feat(desktop): show background child-agent runs in the transcript

naingthet:
## Summary

- show each background child’s delegated task beneath its existing transcript status row
- show the immutable terminal result or failure text when the child settles
- keep the existing live status refresh path and transcript card styling

This is step 3 of #1454: minimum viable child-run status in the transcript. It does not add live sibling/subagent inspection (#954), mid-run steering (#951), or any other steps from the parent issue.

## Verification

- `cargo fmt --all --check`
- `cargo test -p openwave-server --locked`
- `pnpm test`
- `pnpm build`

### DIFF
--- a/crates/openwave-core/src/db/mod.rs
+++ b/crates/openwave-core/src/db/mod.rs
@@ -1097,6 +1097,10 @@ impl Store for DbStore {
         ops::agent_run::list_agent_runs(self, chat_id).await
     }
 
+    async fn get_agent_run_result(&self, id: AgentRunId) -> Result<Option<crate::AgentRunResult>> {
+        ops::agent_run::get_agent_run_result(self, id).await
+    }
+
     async fn claim_agent_run(
         &self,
         lease_token: uuid::Uuid,
--- a/crates/openwave-core/src/db/ops/agent_run.rs
+++ b/crates/openwave-core/src/db/ops/agent_run.rs
@@ -2835,6 +2835,18 @@ pub(in crate::db) async fn list_agent_runs(
         .collect()
 }
 
+pub(in crate::db) async fn get_agent_run_result(
+    store: &DbStore,
+    id: AgentRunId,
+) -> Result<Option<AgentRunResult>> {
+    entities::agent_run_result::Entity::find_by_id(id.0)
+        .one(&store.conn)
+        .await
+        .map_err(store_err)?
+        .map(agent_run_result_from_model)
+        .transpose()
+}
+
 fn validate_request(
     id: AgentRunId,
     parent_id: Option<AgentRunId>,
--- a/crates/openwave-core/src/storage.rs
+++ b/crates/openwave-core/src/storage.rs
@@ -2265,6 +2265,11 @@ pub trait Store: Send + Sync {
         agent_run_storage_unavailable()
     }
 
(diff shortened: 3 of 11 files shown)


naingthet:
Blocking:

1. `crates/openwave-server/src/routes.rs:2409-2413` exposes `AgentRunResult.text` wholesale as `terminal_text`, and `crates/openwave-desktop/ui/src/BackgroundAgentList.tsx:294-298` renders it verbatim. For terminal failures that text is built from the raw executor diagnostic at `crates/openwave-core/src/db/ops/agent_run.rs:1659-1668`. Those diagnostics are produced from `error.to_string()` and transport/container failure details, so they can contain credentials, absolute host paths, URLs, or other operator-only data. The changed regression test demonstrates the concrete leak: `crates/openwave-server/src/tests/sandbox.rs:1177-1242` puts `Authorization: Bearer private-token` in `error_detail` and now asserts that exact token is returned by `/chats/{id}/agent-runs`; the previous assertion required that detail not cross the renderer boundary. Opening or reloading the chat then displays the bearer token in the transcript. This violates the established renderer-safe contract and must be fixed before merge: failure snapshots need a separately constructed, bounded user-safe message (or only the stable error code), never the persisted raw diagnostic.

I did not find another blocking correctness/lifecycle issue. The task and successful/cancelled terminal text are read from the durable run/result records and remain associated through the immutable run/spawn identities across polling, reload, and chat switches. The per-terminal-run result lookup is an N+1 read, but I consider that non-blocking at the current bounded child count.

naingthet:
The review above is verified and correct — holding merge on its blocking finding.

`fail_agent_run` (`agent_run.rs:1659`) persists `format!("Sandbox task failed ({error_code}): {error_detail}")` with the raw executor diagnostic, and this PR exposes that string as `terminal_text` straight to the renderer. Executor diagnostics come from `error.to_string()` on transport and container failures, which can carry auth headers, host paths, and URLs.

The tell is the test: `agent_run_snapshots_omit_persisted_raw_failure_detail` existed precisely to pin that raw detail never reaches a snapshot, and the change rewrites it to assert `Bearer private-token` **does** come through. A guard test being inverted to bless the behavior it guards against is never the fix.

Required change: for failed runs, build `terminal_text` from the stable error code only (e.g. `Sandbox task failed (sandbox_transport_failed)`), never `error_detail`. Restore the omit-raw-detail assertion to its original meaning. The success path (result text) is fine as is.

naingthet:
Addressed the blocking review in `4907c9a7`. Failed-run transcript text is now constructed only from the stable error code; persisted executor diagnostics are never returned to the renderer. The regression test again asserts that the raw bearer-bearing detail is absent. Successful and cancelled terminal receipts continue to use their bounded immutable display text.

Verified with:
- `cargo fmt --all --check`
- `cargo test -p openwave-server tests::sandbox::agent_run_snapshots_omit_persisted_raw_failure_detail --locked`
- `pnpm test -- BackgroundAgentList.test.tsx`
- `pnpm build`

naingthet:
APPROVED — failed runs now surface only the stable error code, the guard test asserts its original invariant again, and the whole-snapshot negative assertion pins it. Arming auto-merge.

PR #1526, opened 2026-08-04
fix(core): recover create_app's parked approval as the workspace kind

abaas-bw:
Found smoke-testing local apps end to end: ask chat to build an app, get the workspace consent card ("Allow OpenWave to create or modify files in this chat's workspace?"), click Approve — and the decision fails with `409: this action cannot be approved from the renderer`.

## Cause

Two kind derivations disagree about `create_app`:

- **At park time**, the kind comes from `ToolApprovalKind::for_call`, which folds any `Workspace`-class tool into `WorkspaceMayModifyFiles` — so the renderer presents the approvable workspace card.
- **At decide time**, the broker re-reads the stored call, and recovery re-derives the kind from `for_tool_name` — the name table alone (the DB column stores the legacy fold, not the kind). `create_app` was missing from that table, so the stored call recovered as `Unsupported`, whose `is_approvable()` is false → every Approve of the card 409s while Reject works.

The card was approvable on screen and unapprovable on the wire — worse than fail-closed, because the human's consent was solicited and then discarded.

## Fix

`create_app` joins the name table as `WorkspaceMayModifyFiles`, the kind it already parks and renders under. One arm plus a comment explaining the park/recover asymmetry so the next Workspace-class tool doesn't repeat this.

## Test

`a_parked_create_app_call_can_be_approved` drives the broker over the store-recovered call: the kind assertion pins the table entry, and the resolve leg pins that Approve resolves instead of refusing. Approvals suite (24) and the core approval tests (37) pass.

Note for reviewers: `write_file` was already in the table; `create_app` was the only `Workspace`-class tool missing. The generic degradation for genuinely unknown workspace tools (recover as rejectable-only) is unchanged.

### DIFF
--- a/crates/openwave-core/src/approval.rs
+++ b/crates/openwave-core/src/approval.rs
@@ -107,6 +107,11 @@ impl ToolApprovalKind {
             "web_extract" => Self::WebExtractMayFetchUrl,
             "exec" => Self::ExecMayRunNetworkedCommand,
             "write_file" => Self::WorkspaceMayModifyFiles,
+            // `create_app` writes app revisions, a `Workspace`-class effect.
+            // Recovery of a parked call re-derives the kind from this table
+            // alone, so a workspace tool missing here parks as an approvable
+            // card the renderer presents and then 409s the approval itself.
+            "create_app" => Self::WorkspaceMayModifyFiles,
             name if name.starts_with("mcp__") => Self::ExternalMcpMayCallServer,
             _ => Self::Unsupported,
         }
--- a/crates/openwave-server/src/approvals.rs
+++ b/crates/openwave-server/src/approvals.rs
@@ -1493,6 +1493,29 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn a_parked_create_app_call_can_be_approved() {
+        // `create_app` is `Workspace`-class, so in Ask mode it parks behind
+        // the workspace consent card. The card is presentable and approvable;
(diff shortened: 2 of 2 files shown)


naingthet:
Real bug, precisely diagnosed, and the writeup makes the park/recover asymmetry legible. Two things before merge.

**1. The commit carries a `Co-Authored-By` trailer that must not land.** `5cfbd537` ends with `Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>`. Squash-merge concatenates commit messages into the squash body, so this reaches `main`'s history in a public repository. Please amend it out and force-push. I already stripped the matching footer from the PR description.

**2. The fix is right, but the next Workspace-class tool will repeat this.** `create_app` is the second entry that had to be added to `for_tool_name` by hand, and nothing makes the third one fail loudly — it fails quietly, at the worst possible moment, after a human has already granted consent. The code comment helps whoever reads that function, but the person who adds a tool is not reading that function.

Worth closing structurally in this PR: a test that walks the registered tool set, filters to `ApprovalClass::Workspace`, and asserts each name recovers through `for_tool_name` to a kind whose `is_approvable()` is true. That converts "someone remembers the table" into a build failure, and it is the kind of test the repo wants — it fails only when a real defect is introduced, and the defect it catches is one we have now shipped twice.

If the registry is not reachable from `approval.rs`'s test module, an explicit list of Workspace-class tool names asserted against the table is a weaker but acceptable substitute — say so and I'll take it.

Issue #1520, opened 2026-08-04
Three different meanings of "write a file" diverge in durability, approval, and undo


naingthet:
Audited whether the bespoke `read_file`/`write_file`/`list_dir` tools should be dropped in favor of exec-only file manipulation. Conclusion: keep the direct surface (see the rationale at the bottom), but the audit surfaced a set of real divergences that are worth fixing regardless of that decision. A user says "edit a file" and means one thing; the system implements three, with different durability, approval class, and undo.

**1. Misattributed output revisions (a real bug, not just design debt).** `write_file("output/x")` mutates a scratch file but publishes nothing — output publication happens in the post-exec scan (`crates/openwave-code-execution/src/tool.rs:156-172`). So a later, unrelated `exec` picks up those earlier bytes and publishes them as a revision attributed to *its own* call and turn (`crates/openwave-server/src/code_execution.rs:1936-1966`). The version history then credits the wrong action.

**2. Provider-dependent persistence of the same command.** Local exec runs directly in chat scratch and edits ordinary files in place; managed providers stage copies in and pull back only `output/` and `preview/` (`crates/openwave-server/src/code_execution.rs:1816-1829`, `:1859-1874`). The same command leaves an intermediate edit durable on local exec and silently absent on E2B/Daytona.

**3. Approval class doesn't track risk.** A direct scratch text write is `Workspace`; any exec recipe is `Sensitive` — even `printf` into scratch with network off (`crates/openwave-core/src/tools/write_file.rs:32-34`, `crates/openwave-code-execution/src/tool.rs:117-122`). This is defensible as written (exec's surface is unbounded, and script executors are deliberately excluded from auto-judge candidacy in `crates/openwave-core/src/approval.rs:1471-1491`), but it means the cheap path and the expensive path for the same user intent sit on opposite sides of a consent boundary.

**4. Undo coverage is inconsistent.** Exec writes to granted host folders get digest preconditions, prior-byte snapshots, a journaled 20-turn undo (`crates/openwave-server/src/exec_file_snapshot.rs`, `crates/openwave-code-execution/src/overlay.rs:904-1009`). Writes to `output/` become append-only versioned outputs with restore. Plain scratch writes — by `write_file` *or* by local exec — have neither: last-writer-wins, no snapshot. That may well be the intended contract for scratch, but it is currently implicit.

[...]

PR #1534, opened 2026-08-04
fix(core): reserve output/ from write_file

naingthet:
Addresses item (1) of #1520 only — misattributed output revisions. Items (2)-(5) stay open.

## The bug

`write_file("output/x")` mutates a scratch file and publishes nothing. Output publication happens in the post-exec scan (`crates/openwave-code-execution/src/tool.rs:156-172`), which attributes every revision it records to the call and turn that owns the exec it just ran (`crates/openwave-server/src/code_execution.rs:1936-1966`). So the bytes sat in `output/` until some later, unrelated `exec` ran, and that call was credited with the revision. Version history showed the wrong action as the producer of the change.

## Resolution: (b), reserve `output/`

I went with reserving the directory rather than having `write_file` invoke the publication path itself.

Publication needs a store handle and the turn that owns the call. `ToolCtx` carries neither — it has `chat_id`, `project_id`, `call_id`, and a capability-confined scratch `Dir` — and `sync_output_directory` is driven from the server's provider wrapper, which has both. Making `write_file` publish means either threading a store into `ToolCtx` for every built-in tool or moving the tool out of `openwave-core`; both are a much larger diff than this one, in service of giving outputs a second producer whose durability, approval class (`Workspace` vs `Sensitive`), and undo behavior differ from the first. One producer is the clearer contract, and it is the one the model is already told to use: scratch is intermediate work, `output/` is what `exec` publishes.

## What changed

- `crates/openwave-core/src/tools/private_scratch.rs` — `is_published_output_path`, which skips leading `.` components so `./output/x` is caught too.
- `crates/openwave-core/src/tools/write_file.rs` — a path naming `output/` or anything under it returns a `ToolErrorCategory::InvalidArguments` failure naming the path and telling the model to save the file from an `exec` command instead. Invalid-arguments rather than the generic tool-failure category because this is the model getting the request wrong, not the product failing, and `is_product_failure()` should stay false for it.
- `crates/openwave-core/src/tools/definitions.rs` — `write_file` states the reservation; `read_file` says `output/` is readable but only `exec` writes it.
- `crates/openwave-server/src/foreground_prompt.rs` — the scratch guidance line already told the model to produce user-visible files in `output/`; it now says to do so from an `exec` command and that `write_file` cannot write there. Golden prompt identity re-pinned accordingly.

`read_file`/`list_dir` are untouched in behavior — reading and listing published outputs stays available, which matters in plan mode where `exec` is never advertised.

## Test

One: `write_file_refuses_the_published_output_directory`, covering `output/report.md`, `./output/nested/report.md`, and bare `output`, asserting the call fails, the message points at `exec`, and nothing was created on disk. No mirror or guard tests added.

## Verification

```
$ cargo fmt --all
$ cargo test -p openwave-core --locked
test result: ok. 578 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 31.44s
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (postgres_turn_state)
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (doc-tests)

$ cargo test -p openwave-server --locked
test result: ok. 587 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out; finished in 78.17s
test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 2.00s
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out (doc-tests)

$ cargo check --workspace --locked
Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 04s
```

No lockfile drift, no new dependencies. Nothing under `crates/openwave-desktop/ui/` changed, so `pnpm test`/`pnpm build` do not apply.

Not verified by driving the app: the end-to-end effect on a real turn's version history. The attribution path itself is unchanged — this removes the only way to reach it without an owning exec call.

Part of #1520.

### DIFF
--- a/crates/openwave-core/src/tools/definitions.rs
+++ b/crates/openwave-core/src/tools/definitions.rs
@@ -14,7 +14,8 @@ pub(super) const WRITE_FILE: &str = "write_file";
 pub(super) fn read_file() -> ToolSpec {
     ToolSpec::for_args::<read_file_tool::Arguments>(
         READ_FILE,
-        "Read a UTF-8 text file, relative to the private scratch directory.",
+        "Read a UTF-8 text file, relative to the private scratch directory. Files under \
+         output/ are readable here, but only exec can write them.",
     )
 }
 
@@ -29,7 +30,10 @@ pub(super) fn write_file() -> ToolSpec {
     ToolSpec::for_args::<write_file_tool::Arguments>(
         WRITE_FILE,
         "Write a UTF-8 text file into private scratch, creating parent directories. \
-         Overwrites an existing file.",
+         Overwrites an existing file. Scratch files are intermediate work, not \
+         user-visible outputs: output/ is reserved and a write there is refused, \
+         because a user-visible output is published by saving it into output/ from \
+         an exec command.",
     )
 }
 
--- a/crates/openwave-core/src/tools/private_scratch.rs
+++ b/crates/openwave-core/src/tools/private_scratch.rs
@@ -30,6 +30,21 @@ pub(super) fn relative_path(rel: &str) -> std::result::Result<PathBuf, String> {
     Ok(path.to_path_buf())
 }
 
+/// Whether a scratch-relative path names the published-output directory or
+/// anything inside it.
+///
+/// `output/` is not ordinary scratch: its bytes become durable, versioned
+/// outputs when an exec call's post-command scan publishes them, attributed to
+/// that call and its turn. A write that lands there without publishing anything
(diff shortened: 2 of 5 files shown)


naingthet:
APPROVED — reservation over publication is the right resolution while ToolCtx has no store handle; auto-merge armed.

Issue #1510, opened 2026-08-04
A single inbox for everything waiting on the user


naingthet:
Approvals, plan reviews, and agent questions currently live inside their chats: when several chats are working (especially with background agents), finding what is blocked on you means hunting through transcripts. Meanwhile every one of these is already a durable row in the journal with park/resume semantics.

Add an inbox that projects all pending attention items — tool approvals, permission-mode questions, plan-mode reviews, agent questions — into one surface: a unified query in the server, an Inbox view with an unread badge in the desktop shell, idempotent first-responder resolution routes, and a deep link that returns to the exact chat and transcript position each item paused at. The journal stays the source of truth; the inbox is a read model, not a parallel store.

Out of scope for the first slice: any delivery outside the app (notifications, messaging transports), and any change to autonomy semantics — an unattended chat still parks exactly as today; the inbox only makes the parked set visible and answerable from one place.

PR #1539, opened 2026-08-04
feat(server): unified inbox read model over parked work

naingthet:
Approvals, agent questions, plan reviews, and folder-access requests all park
as durable journal rows with the same park/resume semantics, but each one is
only visible inside the chat it stopped. With several chats working at once,
finding what is blocked on you means opening them one at a time.

This adds the server half of an inbox: one owner-scoped read over the rows that
are already parked, projected into a single ordered list.

- `Store::list_inbox_items_scoped` and `GET /inbox` return every parked item for
  the requesting principal, oldest first — the thing blocked longest is the
  thing most worth seeing, and it is what survives the 200-item cap.
- The cross-chat prompt query behind `GET /chats/pending-prompts` is refactored
  to produce rows once; the existing per-chat summary and the new inbox both
  fold from the same rows rather than re-deriving them. Pending tool approvals
  are the one kind the old query never covered, and they are added here.
- A call parked on its approval gate is listed once, as an approval: that is the
  decision actually being asked, so it claims the call even when the tool behind
  it also has a renderer prompt.
- Items stay as opaque as the existing attention summary — identity, kind, chat
  title, and when it parked. Question prose, plan text, and canonical tool
  arguments stay behind the chat the item points at. The tool name on an
  approval crosses as the closed renderer vocabulary, as it already does on the
  approval card.

There are no new tables and no new resolution routes. The inbox holds no state,
so an item leaves it exactly when its own existing route commits — which also
means resolving from the inbox is first-responder by construction: whoever
commits first wins and a second, different attempt meets the same 409 the
in-chat card would have met.

The desktop Inbox view and its deep links are a separate change stacked on this.

## Tests

One test in `openwave-server`: a parked approval, question, plan review, and
folder-access request are listed together, then disappear after each is resolved
through its own established route, and a second conflicting approval decision
answers `409`. Output write-backs travel the identical client-call branch as
folder access, so they are not walked a second time.

## Verification

```
$ cargo fmt --all
$ cargo test -p openwave-core --locked
test result: ok. 577 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 38.99s

$ cargo test -p openwave-server --locked
test result: ok. 588 passed; 0 failed; 3 ignored; 0 measured; 0 filtered out; finished in 80.19s
test result: ok. 5 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.76s

$ cargo check --workspace --locked
(clean)

$ pnpm test            # crates/openwave-desktop/ui
Test Files  109 passed (109)
     Tests  722 passed (722)

$ pnpm build           # crates/openwave-desktop/ui
✓ built in 14.73s
```

Part of #1510.


naingthet:
APPROVED — read-model-only is the right first slice; owner scope as the visibility filter and the approval-claims-the-call dedup both hold up. Auto-merge armed.

PR #1541, opened 2026-08-04
feat(desktop): unified inbox for pending approvals and questions

### LINKED ISSUE #1510: A single inbox for everything waiting on the user

naingthet:
Follows #1539, which added the cross-chat read model and `GET /inbox`. This is
the surface: one place that shows everything waiting, and a way back to the
card that is waiting.

- An **Inbox** route listing every parked item — approvals, questions, plan
  reviews, folder access, save-to-folder — with the conversation it belongs to
  and how long it has been waiting.
- An **Inbox entry on both rails**, with a count badge. Being blocked is not
  scoped to a conversation: the reader is usually inside one chat when another
  parks, so the way to the inbox exists on the chat rail too.
- **Deep links.** Opening an item navigates to `/c/{chatId}?focus={callId}` —
  layout and addressing state live in the URL here, so the parked call rides
  beside the panel params. On arrival the transcript looks for the card
  deciding that call, scrolls it into view, and rings it briefly. Waiting cards
  now carry `data-pending-call-id`; that attribute is the only transcript
  position the app can address, and it is exactly the one an inbox item names.
  The param is dropped once honored, so a reload does not re-scroll to a card
  that may have been answered since.
- **One cross-chat poll, not two.** The shell watcher that marked parked chats
  now reads `/inbox` and feeds both the rail's attention markers and the inbox,
  so the badge and the list cannot disagree. A side effect worth naming: a chat
  parked on a tool approval now gets the same attention marker a parked
  question always did — the old summary read never covered approvals.

Nothing is answered from the inbox itself. An item is a pointer back to the
card that owns the decision, which is the only place with the question, the
plan, or the action under review — and keeping every answer on that path is
what keeps one resolution path instead of two.

## Tests

- One behavior test that a deep link reveals the card it named and then clears
  itself from the URL — the promise the inbox makes, end to end.
- Validator coverage for the new wire shape: an unknown kind, a tool outside
  the renderer vocabulary, and smuggled detail are all rejected.
- Existing `ChatView` tests now mount inside the test router, because the view
  reads the URL; the shell and watcher tests follow the read they now make.
  No new tests were added for the rendering of the list itself.

## Verification

```
$ cargo fmt --all --check
(clean; no Rust changed in this slice)

$ pnpm test            # crates/openwave-desktop/ui
Test Files  109 passed (109)
     Tests  725 passed (725)

$ pnpm build           # crates/openwave-desktop/ui
✓ built in 14.58s
```

Not verified by driving the running app: the reveal and the rings were checked
under jsdom, not on screen.

Closes #1510.

### DIFF
--- a/crates/openwave-desktop/ui/src/AppShell.dom.test.tsx
+++ b/crates/openwave-desktop/ui/src/AppShell.dom.test.tsx
@@ -40,9 +40,22 @@ const listChats = vi.fn(async () => chats);
 const createChat = vi.fn(async () => chats[0]);
 const listPendingUserQuestions = vi.fn(async () => [] as unknown[]);
 const listPendingFolderAccessRequests = vi.fn(async () => [] as unknown[]);
-const listPendingChatPrompts = vi.fn(async () => [] as unknown[]);
+const listInbox = vi.fn(async () => [] as unknown[]);
 const listPendingOutputWritebackRequests = vi.fn(async () => [] as unknown[]);
 const requestUserAttention = vi.fn(async () => {});
+
+/** One waiting question, as the shell's cross-chat read returns it. */
+function parked(chatId: string, callId: string) {
+  return {
+    chatId,
+    chatTitle: null,
+    turnId: "turn-1",
+    callId,
+    kind: "question" as const,
+    action: null,
+    requestedAt: "2026-08-04T00:00:00Z",
+  };
+}
 const postMessage = vi.fn(async () => {});
 // Unmanaged is the shape most of these shell tests model; the managed gate
 // has its own DOM tests, and the one shell test that flips this to managed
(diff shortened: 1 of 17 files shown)


naingthet:
APPROVED — the data-pending-call-id addressing plus the single consolidated poll is the right shape; nice touch dropping the focus param after honoring it. Auto-merge armed.

Issue #1462, opened 2026-08-04
Pre-v1 schema: squash the migration baseline and consolidate accreted tables


naingthet:
Do while pre-v1 databases are still disposable. `crates/openwave-server/src/desktop_schema.rs` resets any older schema epoch, so every change below is a rewrite-the-baseline + epoch-bump — zero data-preserving migrations. That window closes permanently at v1.0.0.

Context change: background agents are now v1 scope (#1454), so the previously-dormant tables are being **committed to**, which makes the consolidations below worth doing rather than mooting them by deletion.

Ordered plan:

1. **Squash the 53 migrations into one baseline** and bump `DESKTOP_SCHEMA_EPOCH` (`crates/openwave-core/src/db/migration.rs` is 11k lines, ~60% removable legacy scaffolding). This gates everything else.
2. **Delete the dead `turn_agent_run_wait` legacy path** (atomic-admission column, `legacy_wait`/`reused_legacy_child` guards, the always-miss lookup in `resume_turn_after_agent_run_inbox_consumption_on`) — ~800 lines across 9 files. The wait-set path (`turn_agent_run_wait_set`/`_member`) is the one #1454 revives; the legacy single-wait path should not come back.
3. **Fold `sandbox_agent_admission` into `agent_run`** — same PK, same transaction, duplicates four columns; `resolve_existing_sandbox_admission_on`'s drift check becomes structurally impossible. Add `origin_turn_id`/`delegated_root_id`/`delegated_relative_path` + CHECKs.
4. **Fold `sandbox_tool_call_receipt` into `sandbox_tool_call`** — mirrors how `tool_call` already stores `result`/`error_code`/`error_detail` inline; idempotency via the existing conditional-update + payload-comparison pattern.
5. **Merge `exec_file_snapshot` + `exec_file_rejection` into one `exec_file_change`** with a `classification` discriminator + CHECKs — the only reader already re-merges them, and the duplicated retention prune becomes one.
6. Smaller: drop `turn_agent_run_wait_set.{provider_id,history_order,arguments}` (duplicated from `tool_call`, hand-verified today); backfill-and-drop the legacy `user_question` answer columns; fold the three one-row advisory-lock tables into one `advisory_lock(name)`.
[...]

PR #1542, opened 2026-08-04
refactor(core): squash desktop schema migrations into a v1 baseline

naingthet:
Replaces the 51 accreted migrations with a single `Baseline` migration and
bumps `DESKTOP_SCHEMA_EPOCH` to 3, so any existing pre-v1 database is
rebuilt on next start. `crates/openwave-core/src/db/migration.rs` goes from
11,114 lines to 63; the schema now lives in `migration/idens.rs` (table and
column identifiers) and `migration/baseline/` split by domain — chat, turn,
agent runs, tools, content, sandbox — 6,223 lines in total.

One structural change fell out of the squash: each table's indexes are
created immediately after the table instead of in a second pass. Postgres
rejects a foreign key whose target uniqueness comes from a standalone unique
index unless that index already exists, which the old chain satisfied by
accident of interleaving. The first assembly failed on Postgres for exactly
this reason (`event` → `turn_claim (turn_id, token)`).

## Schema equivalence

Both schemas were produced by running the migrations into a scratch database
and dumping the result — the pre-squash chain from `main`, the baseline from
this branch.

**SQLite — byte-identical.** The dump covers, per object: the `sqlite_master`
SQL text (whitespace-normalized) for all 50 tables and every index, plus
`pragma_table_info` (column order, type, nullability, default, primary key),
`pragma_index_list` / `pragma_index_info` (uniqueness, origin, partiality,
key columns), `pragma_foreign_key_list` (column tuples and referential
actions), and the contents of every table so the three advisory-lock seed
rows are compared too.

```
$ diff /tmp/schema_old.txt /tmp/schema_new.txt && echo IDENTICAL
IDENTICAL
```

**Postgres 15 — semantically identical, four cosmetic differences.** Compared
with `pg_dump --schema-only` against two databases in the same cluster. The
whole diff, in categories:

- **Physical column order** on `agent_run` (`tier`, `execution_location`,
  `model`), `document.chat_id`, `output_revision.created_at`, and
  `sandbox_tool_call.retry_at`. The old chain appended these with `ALTER
  TABLE`; the baseline declares them in their logical position, which is
  also the order SQLite already had after its table rebuilds. Nothing reads
  columns positionally.
- **Auto-generated CHECK constraint names.** Constraints the old chain added
  via `ALTER TABLE ... ADD CONSTRAINT chk_*` are now inline in `CREATE
  TABLE`, so Postgres names them `<table>_check`, `<table>_<column>_check`.
  Identical predicates; nothing in the tree references a constraint name.
- `assistant_citation_light_pkey` → `assistant_citation_pkey`, a leftover
  from the citations rebuild creating a scratch table and renaming it.
- The `\restrict` token `pg_dump` emits per run.

No other lines differ: same tables, columns, types, defaults, predicates,
foreign keys, and indexes.

## Tests

Fourteen tests are removed, all of which drove `Migrator::up` to an
intermediate point, seeded rows in a schema shape that no longer exists, and
asserted the remaining steps preserved them:

- `m0006_upgrades_an_existing_store_without_losing_records`, `m0011`, `m0013`,
  `m0014`, `m0015`, `m0018`, `m0021`, `m0024`, `m0037`, plus the shared
  `create_chat_before_agent_run_split` seed helper — each replays a partial
  chain against legacy rows.
- `lightweight_citation_migration_has_a_working_down_and_reapply`,
  `retire_document_pipeline_migration_has_a_working_down_and_reapply`,
  `message_document_attachment_migration_has_a_working_down_and_reapply` —
  each exercises `down` then re-`up` of one retired migration, along with the
  `migrations_added_after` helper that existed only to skip past later steps.
- `absorption_migrates_servers_rekeys_manifests_and_drops_grants` and
  `owner_attribution_migration_backfills_existing_rows_to_the_local_owner` —
  each asserts a backfill that a fresh database never performs.

`baseline_migration_rolls_back_multi_wait_dependencies_in_fk_order` stays and
now covers the baseline's own `down`, which drops in reverse creation order.

## Verification

```
cargo fmt --all --check          # clean
cargo test -p openwave-core      # 563 passed
cargo test -p openwave-server    # 587 passed, 3 ignored (+5, +0)
cargo check --workspace --locked # clean, no lockfile drift
```

Postgres was verified locally as described above; `full-ci` is set so the
PostgreSQL turn-state lane runs the baseline against Postgres in CI too,
since that backend is where the schema construction can diverge.

## Noted, not changed

Things the squash made visible in the schema the chain produces. Reporting
rather than fixing — several are already scoped by later steps of #1462:

- `user_question` carries two parallel answer shapes; only the legacy trio is
  constrained by a CHECK (step 6 of the issue).
- `turn_agent_run_wait.atomic_admission` has a default and no constraint
  referencing it — dead, and the next PR removes it.
- `document.source_sha256` is declared as an unsized blob while every other
  digest column is `binary_len(32)`; the 32-byte length is enforced only
  inside a CHECK. `document.chat_id` has no foreign key, unlike
  `document.project_id`.
- `exec_file_snapshot.chat_id` is `ON DELETE RESTRICT` while
  `exec_file_rejection.chat_id` is `ON DELETE CASCADE`.
- `turn_run.agent_run_depth` is `NOT NULL DEFAULT 0` with `CHECK (= 0)`; it
  exists only to carry the composite foreign key into `agent_run`.
- `agent_run_result.text` and `payload_json` hold the same content.
- `assistant_citation` keeps `fk_assistant_citation_light_*` and
  `idx_assistant_citation_light_message_ordinal` names from the rebuild that
  created it under a scratch name. Renaming them would change the schema, so
  they stayed.

Part of #1462.


naingthet:
Reviewed: the equivalence evidence is exactly what this change needed — the SQLite byte-identical dump plus the categorized pg_dump diff, and the index-ordering catch on Postgres justifies the full-ci label on its own. Test removals all check out against the no-data-preserving-migrations premise. Merging manually once every full-ci lane reports (not arming auto-merge: the heavy lanes are not required checks, so --auto would merge past a red Postgres lane).

Issue #1477, opened 2026-08-04
Delegating to a background agent bypasses the chat's approval gate


naingthet:
`spawn_sandbox_agent` and `wait_for_agents` declare `ApprovalClass::Sensitive`, but nothing gates on that class today: it decides advertisement (a plan turn never sees the pair) and not admission.

The concrete consequence: in a chat whose permission mode would stop and ask before a foreground `web_search`, the model can spawn a child whose task is "search the web for X" and the same egress happens with no card. A child's tool calls are advanced by the sandbox worker under its own lease and never re-enter the parent chat's gate, so the spawn is the only point at which the reader could have been asked. Where the host routes children to a container (`OPENWAVE_CONTAINER_EXECUTION_ENABLED`), the delegated surface widens to running commands in that container.

Why it was not fixed with the class declaration: the approval gate parks a **pending server tool call** on a durable approval receipt (`request_tool_call_approval_and_append_event` reads the locked `tool_call` row and requires `execution = server`, `status = pending`). A spawn has no such record — its tool call is written already completed, in the same transaction that admits the child. Gating it needs a durable pending-spawn checkpoint: a parked turn state the reader can answer, exact-retry recovery for an interrupted decision, and admission only after approval commits.

Two candidate shapes, both wanting a deliberate product call before code:

1. **Gate the spawn.** One card per delegation, at the moment the reader can still say no. Costs a prompt per spawn in `Ask`/`Auto`, which cuts against keeping delegation cheap enough that the model uses it.
2. **Bound the child by the chat's existing consent.** Advertise a sandbox tool to a child only where the equivalent foreground call would not have parked (mode, or a standing grant covering it). No new prompts; children silently lose web search in the default mode.

Neither is obviously right, which is why this is filed rather than chosen. Whatever lands should also decide whether a delegation consent, once given, is standing-grantable per chat.

Context: #1454.

naingthet:
Scope grew with #1553: a delegated background run now executes up to 16 host commands (exec recipes) with no gate in any permission mode — the class decides advertisement, not admission, and there is no pending record to park on. Confinement itself holds (empty folder grants, no staged paths, seatbelt on macOS — the only host with local exec today, chat network policy inherited), so this is squarely the consent-boundary decision this issue tracks: either delegation is explicitly the consent event for arbitrary sandboxed execution (document it), or spawn/exec-bearing runs need a parkable approval in Ask mode.

naingthet:
[...]

PR #1553, opened 2026-08-05
feat(agents): background agents produce files and submit them as outputs

naingthet:
A background run's deliverable used to be host-synthesized text: the worker took whatever final answer the model returned and wrote it into the conversation as an output whose name the host invented (#1550 improved that name; the artifact should not have existed). An agent's output should be the files it wrote, under the names it gave them.

The two halves needed for that already existed and had never been connected. `output_scan.rs` publishes files under `output/` as outputs keyed by filename, with `RevisionProducer::Run(run_id)` already defined for exactly this case — but its only caller was the foreground exec path, because a background run's whole tool surface was `web_search` + `read_delegated_file`, gated to a single checkpoint.

## What changes

- **A real multi-step loop.** Sixteen tool calls, with `exec` parked and resumed through the same durable checkpoint mechanics `web_search` already used, on its own executor lane (`sandbox_exec_worker.rs`). Each run gets a workspace named by the run, so siblings delegated in one message never share scratch.
- **The scan runs for background work.** After each command the host scans that workspace's `output/` and publishes what it finds into the parent conversation's outputs, attributed to the run: new filename → new output, same filename again → next version.
- **A `done` tool ends the run** by submitting the filenames it wrote plus a short summary (16 names, 4,000 chars). Names resolve against the outputs the scan already published, newest live match per name; naming a file the run never wrote fails the completion like any other malformed one, bounded by the run's own attempt budget. A run that genuinely produced nothing submits nothing. `done` is terminal, so it needs no executor lane and stays advertised on the last step — a run that spent its budget writing files can still hand them over.
- **The synthesized artifact and its scaffolding come out:** `record_result_output`, `agent_result_filename`, `merge_agent_run_result`, and `OutputId::for_run` / `OutputRevisionId::for_run`.
- **The snapshot carries the files.** `produced_output` was a tier-and-status guess standing in for a real answer; it is replaced by `submitted_outputs`, and the renderer shows one pill per file that opens the output it names. The panel's prose block is the run's summary alone, since the filenames are already on screen. `exec` joins the activity vocabulary, so the timeline finally shows the work.

## Boundary

A background run's exec is confined to its own workspace: no folder grants, no staged host paths, no chat scratch. Delegation already bypasses the conversation's approval gate, so this path must not be the one that hands a delegated agent the user's files. The only thing the parent conversation contributes is its network policy.

## Notes

- Existing `Agent result *.md` and task-titled rows in a live database stay until deleted by hand; nothing here rewrites history.
- Deleted two tests: the result-filename derivation test and the output-merge test, both covering code this removes.
- Labelled `full-ci` — this touches the exec and turn-state paths.

## Verification

`cargo test -p openwave-core -p openwave-server --locked`, clippy `--all-targets`, `cargo fmt`, and in `crates/openwave-desktop/ui`: `tsc --noEmit`, `pnpm test`, `pnpm build`. Not driven in the running app — the end-to-end check worth doing is delegating a task that requires producing a document and confirming Outputs gains a row named by the file the agent wrote.

### DIFF
--- a/crates/openwave-core/src/agent_tools.rs
+++ b/crates/openwave-core/src/agent_tools.rs
@@ -30,6 +30,14 @@ pub const WEB_EXTRACT_TOOL: &str = "web_extract";
 pub const SANDBOX_WEB_SEARCH_TOOL: &str = WEB_SEARCH_TOOL;
 /// Stable name for the native-executed exact delegated file read.
 pub const SANDBOX_READ_DELEGATED_FILE_TOOL: &str = "read_delegated_file";
+/// Stable name for command execution inside a background run's own workspace.
+///
+/// Deliberately the same name a foreground turn sees: it is the same operation
+/// against the same kind of private workspace, and a model that has learned one
+/// vocabulary should not have to learn a second one when it is delegated work.
+pub const SANDBOX_EXEC_TOOL: &str = "exec";
+/// Stable name for a background run's terminal submission of its own files.
+pub const SANDBOX_DONE_TOOL: &str = "done";
 
 /// Maximum task length in Unicode scalar values advertised to a model.
 ///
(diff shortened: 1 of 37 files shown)


naingthet:
Post-merge review findings (verified against main, fast-follow in progress):

1. **Background-produced outputs are unreadable end to end.** `publish_output_directory` opens `scratch_root.join(workspace.as_str())` (code_execution.rs:1969), and for agent runs the workspace is `agent-run-{run_id}` (sandbox_exec_worker.rs:125). `sync_output_directory` writes revision bytes into that same Dir — but every reader resolves `<scratch>/<chat_id>/outputs/...` (deliverables.rs read_output_revision_bytes). So a background run's outputs render "Output content is unavailable", and when its filename matches an existing foreground output, the scan appends a head revision whose bytes are in the wrong tree — the working output becomes unreadable at head (restorable via history). The green CI didn't catch it because the new tests fabricate outputs with `create_output` and never touch the filesystem.

2. **`resolve_submitted_outputs` matches by filename only** (sandbox_agent_run_worker.rs:716-736) with no check that a revision was produced by this run, so a run that wrote nothing can claim a foreground output as its deliverable. The new submission test pins this by fixing up an output with `producing_run_id: None`.

A fix PR is being prepared for both (publish into the chat's scratch root while scanning the run's workspace; filter submission resolution by producing run; direct filename lookup instead of the capped `list_outputs` scan; plus one test that reads submitted bytes back through the real reader path).

Also noted, tracked separately: agent-run workspace directories are never cleaned up (survive chat deletion); no unique constraint on `(chat_id, filename)` makes concurrent same-filename scans fork silently; and delegation now grants a run up to 16 ungated host commands in any permission mode — that consent-boundary decision is being recorded on #1477.

PR #1556, opened 2026-08-05
fix(agents): publish a run's files under its conversation and scope submission to its own work

naingthet:
A review of #1553 after it merged turned up four defects in the path it added. All four are fixed here.

**A submitted output could not be opened.** `sync_output_directory` wrote revision bytes relative to the directory it scanned. For a foreground turn that directory is the conversation's own scratch, so the write path and the read path agreed by coincidence. A background run's workspace is keyed by the run, so its bytes landed at `scratch_root/agent-run-<id>/outputs/…` while every reader resolves `scratch_root/<chat_id>/outputs/…`. The catalog gained a row, the pill rendered, and opening, previewing, downloading, or restoring it failed. The scan now takes the workspace it reads from and the conversation scratch it publishes into as separate arguments; the foreground caller passes one directory for both.

**A run could submit a file it never wrote.** Resolution matched a filename against every live output in the conversation. An agent that produced nothing could call `done(["Q3 revenue.xlsx"], "here's the analysis")`, name a document the user's own turn produced, and have it presented as the agent's deliverable — and the same lookup was a name-existence oracle over the catalog. A name now only resolves to an output some revision of which this run published.

**A NUL byte in command output wedged the run.** The resolver refuses any receipt containing a NUL, so a command that printed one (reading a binary it had just written, say) never got a receipt at all. The call stayed `Claimed`, its executor lease expired, and the same command was re-executed every lease period until the run's deadline. The byte is stripped where the streams are clamped.

**Host error strings reached model context.** A failed output scan put the raw `CodeExecutionError` — built from a store or filesystem error — into the receipt, bypassing the coarse classification every other executor failure goes through. The model now gets the fact that outputs were not recorded; the detail is logged.

Also rejects a submission that repeats a filename at the storage boundary, which the tool boundary already refused.

## Tests

One added: publishing from a workspace that is not the conversation's scratch writes bytes where the reader looks and not beside the file that produced them. That is the defect above, and nothing else in the suite reads back what the scan writes.

The submission test now seeds two outputs — one this run published, one it did not — so the negative case covers the rule that actually matters. It previously asserted the old, wrong behavior: it seeded an output with no producing run and expected `done` to resolve it.

## Verification

`cargo test -p openwave-core -p openwave-server --locked` (1,156 tests), clippy `--all-targets` clean, `cargo fmt`. Nothing UI-facing changed. I did not drive the app; the check worth running is the one #1553 named — delegate a task requiring a document, then open the resulting row in Outputs, which is what was broken.

## Left for follow-ups

- A `done` naming an unresolvable file fails the run, and because `done` is not a checkpoint the retry replays byte-identical context and re-emits the identical call until the attempt budget is gone. A resolution failure should reach the model as something it can correct.
- Each background run creates a scratch tree that is never destroyed, so host scratch now grows with delegation volume rather than with the number of conversations.

### DIFF
--- a/crates/openwave-core/src/db/ops/agent_run.rs
+++ b/crates/openwave-core/src/db/ops/agent_run.rs
@@ -2412,6 +2412,17 @@ fn validate_agent_run_result_payload(payload: &AgentRunResultPayload) -> Result<
                     "agent-run submission names an invalid filename".into(),
                 ));
             }
+            // One pill per file: a repeated name would present the same output
+            // twice and say nothing extra.
+            let distinct = outputs
+                .iter()
+                .map(|output| output.filename.as_str())
+                .collect::<std::collections::HashSet<_>>();
+            if distinct.len() != outputs.len() {
+                return Err(AgentError::Store(
+                    "agent-run submission repeats a filename".into(),
+                ));
+            }
             Ok(())
         }
         AgentRunResultPayload::FolderAccessProposal { request } if request.is_well_formed() => {
--- a/crates/openwave-core/src/db/tests/output.rs
+++ b/crates/openwave-core/src/db/tests/output.rs
@@ -556,6 +556,7 @@ mod output_scan {
         sync_output_directory(
             store,
             scratch,
+            scratch,
             chat_id,
             call_id,
             RevisionProducer::Turn(TurnId::new()),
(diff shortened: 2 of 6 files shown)


naingthet:
Merged on green full-ci — the head-corruption path made this worth landing immediately. Follow-up stacking on this: indexed filename lookup for submission resolution plus a publisher/reader boundary test.

PR #1559, opened 2026-08-05
fix(server): resolve submitted outputs by indexed filename lookup

naingthet:
Two follow-ups to #1556, which fixed where a background run's output revisions
are published and scoped submission to the run's own work.

Submission resolved a run's filenames against `list_outputs(chat_id,
OUTPUT_LOOKUP_LIMIT)` — the newest 1000 live outputs. The producer check then
filtered that page down to outputs the run itself wrote. In a conversation
holding more than 1000 outputs, a file the run genuinely produced can sit off
the end of the page, and `done` refuses it with "which it never wrote under
output/" — a run that did its work correctly fails, and burns an attempt doing
it. The page was never the right question to ask: submission knows the exact
filename it is resolving, so it now asks the store for that filename directly.
`Store::find_outputs_by_filename` returns the conversation's live outputs
carrying one exact name, newest-updated first, which is the same ordering the
scan applies. The producer check is unchanged; only the candidate set it runs
over stops being capped.

The second change is a test across the publisher/reader boundary that #1556
moved. `a_background_runs_published_output_is_readable_from_the_chats_scratch`
runs the real publication path for an `agent-run-{id}` workspace with a file
actually written under `output/`, then reads the bytes back through
`read_output_revision_bytes` — the same reader the deliverables catalog uses —
and asserts they match. It lives in the desktop crate because that is where the
reader is, and it is the only test that drives both sides against each other:
publisher and reader each look correct in isolation, and the defect lived
exactly in their disagreement about which directory holds the bytes. Reverting
#1556's scan/publication split makes it fail with "Output content is
unavailable", which is the string a user saw on every background-produced file,
so it is guarding the seam it claims to.

No schema change — the new store method is a query over existing columns.

### DIFF
--- a/crates/openwave-core/src/db/mod.rs
+++ b/crates/openwave-core/src/db/mod.rs
@@ -746,6 +746,14 @@ impl Store for DbStore {
         ops::output::list_outputs(self, chat_id, limit).await
     }
 
+    async fn find_outputs_by_filename(
+        &self,
+        chat_id: ChatId,
+        filename: &str,
+    ) -> Result<Vec<OutputRecord>> {
+        ops::output::find_outputs_by_filename(self, chat_id, filename).await
+    }
+
     async fn list_output_revisions(&self, output_id: OutputId) -> Result<Vec<OutputRevision>> {
         ops::output::list_output_revisions(self, output_id).await
     }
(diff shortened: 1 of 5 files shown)


naingthet:
APPROVED — exactly the two salvaged deltas, and verifying the boundary test against the merged #1556 code (not your own) was the right discipline. The feature-unification edge is fine as-is: it fails at compile time, loudly. Auto-merge armed.

Issue #1555, opened 2026-08-05
Output filename identity has no uniqueness: concurrent scans fork, cross-producer writes shadow


naingthet:
Filename is the identity of an output, but there is no unique index on `(chat_id, filename)` — only `idx_output_chat_created`. Two consequences, both now reachable since background runs (#1553) publish concurrently with the foreground turn:

- Two concurrent scans creating the same filename derive distinct output ids (`OutputId::for_call_artifact(call_id, filename)`) and produce two live outputs with one name; "newest live match" resolution then forks silently.
- Absent the race, a background run writing `report.md` silently appends a revision to the user's foreground `report.md` — cross-producer merging by filename with no signal in the UI about who produced the head.

Needs a decision first: is cross-producer filename merging intended (one document, many producers — arguably the files-first model) or should producers namespace? Whichever way, concurrent same-filename creation should be made deterministic — a unique constraint with first-writer-wins retry, or serialized creation per (chat, filename).

Related fast-follow: the publication-tree fix for #1553's revision-byte location (in review) — this issue is the remaining identity-semantics question, not the storage bug.

naingthet:
Second dependent call site as of #1559: `Store::find_outputs_by_filename` returns all live rows for a name ordered `UpdatedAt desc, Id desc`, and submission resolution picks the newest the run produced. Correct under the race, but it's now two places whose behavior silently depends on the duplicate-name ordering tiebreak — worth keeping in mind when deciding the uniqueness constraint here.

naingthet:
PR #1578 (merged) closes the determinism half: creation serializes on the chat write lock with a typed `OutputFilenameTaken` conflict, backed by a partial unique index on `(chat_id, filename) WHERE deleted_at IS NULL`; a losing scan appends a revision to the winner, and `restore_output` refuses to undelete into a reclaimed name. Full CI including the Postgres lane passed.

What remains open here is only the semantics decision: is cross-producer filename merging intended (a background run's `report.md` appending to the foreground's `report.md` as one document), or should producers namespace? The determinism fix works under either answer.

naingthet:
[...]

PR #1578, opened 2026-08-05
fix(server): make same-filename output creation deterministic under concurrency

naingthet:
## The race

Filename is the identity everything outside the store addresses an output by, but nothing enforced that a conversation held only one output per name. `sync_output_directory` takes one snapshot of the conversation's live outputs and then decides, per scanned file, whether to create or revise. Two scans running at once — routine now that background runs publish into their parent conversation while the foreground turn is still writing — both see no `report.md`, both derive a distinct `OutputId::for_call_artifact(call_id, filename)`, and both create. The conversation ends up with two live outputs answering to one name, and "newest live match" resolution forks between them with nothing to tell the user which document they are looking at.

A test that runs two scans of `report.md` concurrently reproduces this on `main`: two live outputs, one name.

## Mechanism: serialize creation per conversation, first writer keeps the name

`create_output` already begins a transaction and takes the owning chat's write lock — the same lock `append_output_revision` uses to keep ordinals from colliding. Creation is therefore already serialized per conversation; all that was missing was the check. It now runs inside that transaction, after the lock: if a live output in the chat already carries the filename, the creation is refused with `AgentError::OutputFilenameTaken` naming the winner.

Serializing on the existing lock rather than catching a constraint violation and retrying keeps the whole decision in one place and needs no new transaction structure or backoff. The partial unique index on `(chat_id, filename) WHERE deleted_at IS NULL` is still there, as the schema-level statement of the invariant: no future write path can quietly reintroduce the fork, and the check above can read a single row rather than a list. Retracted outputs are excluded from the index because deleting `report.md` must leave the name free for a later one.

The scan handles the refusal by doing what it would have done had its snapshot been current: it re-reads the winning output and appends its bytes as a revision, publishing them under the winner's id. The create and revise paths were factored into one `revise_output` helper so both reach the store the same way. The loser leaves behind one unreferenced file at its own (never-recorded) revision path in conversation scratch — inert, and only on the racing path.

## Restore

Excluding retracted outputs from the index gives restore a new failure mode: undeleting an output whose filename has since been reclaimed would violate the index. `restore_output` now takes the chat write lock and reports the same `OutputFilenameTaken` conflict, so the caller gets an actionable message instead of a raw constraint error.

## Schema

Baseline edit, so `DESKTOP_SCHEMA_EPOCH` moves 5 -> 6 and pre-v1 local databases are rebuilt, per the pre-v1 lifecycle.

## Still open

This is only the half of #1555 that is wanted under either answer to its product question. **Whether cross-producer filename merging is intended** — one document with many producers, versus producers namespacing their outputs — is untouched and still open. A background run writing `report.md` still appends to the user's foreground `report.md`; that is now deterministic rather than forked, but the UI still says nothing about who produced the head revision. Part of #1555.

## Tests

One added: `concurrent_scans_of_one_filename_land_on_a_single_output` runs two barrier-synchronized scans of the same filename from separate workspaces into one conversation and asserts the conversation holds exactly one live `report.md` with two revisions, that both scans addressed it, and that both revisions' bytes are readable where the record says they are. Verified red on the pre-fix code (two live outputs), green after.

One modified: `output_writeback_filename_resolves_to_the_newest_live_output` seeded two live outputs with one filename to prove filename resolution picks the live record. That state is no longer constructible, so it retracts the first output instead and keeps the same assertion.

## Verification

`cargo test -p openwave-core --locked --lib`, `cargo test -p openwave-server --locked --lib`, `cargo check --workspace --locked --all-targets`, `cargo clippy -p openwave-core -p openwave-server --locked --all-targets`, `cargo fmt --check` — all clean. The partial unique index follows the same `and_where` form as the existing ones in the baseline, so it renders for both backends, but I have not exercised it against PostgreSQL locally.

### DIFF
--- a/crates/openwave-core/src/agent.rs
+++ b/crates/openwave-core/src/agent.rs
@@ -5671,8 +5671,11 @@ mod tests {
     #[tokio::test]
     async fn output_writeback_filename_resolves_to_the_newest_live_output() {
         let (_db, store, chat, turn_id, lease_token) = output_writeback_fixture().await;
+        // A retracted `report.md` frees the name for a later one. Only one live
+        // output can hold it, and that is the one the checkpoint must resolve.
         let older = Utc::now() - chrono::Duration::minutes(10);
-        create_named_output(&store, chat.id, "report.md", older).await;
+        let retracted = create_named_output(&store, chat.id, "report.md", older).await;
+        store.delete_output(retracted, older).await.unwrap();
         let newest = create_named_output(&store, chat.id, "report.md", Utc::now()).await;
 
         let root_id = uuid::Uuid::new_v4();
--- a/crates/openwave-core/src/db/migration/baseline/content.rs
+++ b/crates/openwave-core/src/db/migration/baseline/content.rs
@@ -135,13 +135,27 @@ pub(super) fn output_table() -> TableCreateStatement {
 }
 
 pub(super) fn output_indexes() -> Vec<IndexCreateStatement> {
-    vec![Index::create()
-        .name("idx_output_chat_created")
-        .table(Output::Table)
-        .col(Output::ChatId)
-        .col(Output::CreatedAt)
-        .col(Output::Id)
-        .to_owned()]
+    vec![
+        Index::create()
+            .name("idx_output_chat_created")
+            .table(Output::Table)
+            .col(Output::ChatId)
+            .col(Output::CreatedAt)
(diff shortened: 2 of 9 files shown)


naingthet:
Labelled `full-ci`: the new partial unique index is the one thing I cannot verify locally against PostgreSQL, and the turn-state lane is where it gets exercised.

Issue #1604, opened 2026-08-05
feat(desktop): render a background agent's live progress in the transcript and panel


naingthet:
The server side of #954 landed: `GET /chats/{chat_id}/agent-runs/{run_id}/progress` returns a background run's ordered, resumable progress lines, and `api.ts` has the typed client (`listAgentRunProgress`) and defensive parser (`parseAgentRunProgress`) for it. Nothing renders it yet.

Scope of this slice:

- The transcript's child-run status rows (`BackgroundAgentList.tsx`, wired from `MessageList.tsx`) show a static state string for a running child. Show the run's most recent progress line instead, so the row says what the child is doing rather than only that it is doing something.
- The background-agent activity panel already polls per-run activity history alongside its Stop control. Add the progress stream beside it, polled with the cursor the panel last held so a poll returns only new lines.

Notes for whoever picks this up:

- Poll with `nextSequence` from the previous page. The server returns nothing for a cursor that is already current, so the poll is cheap; do not re-read from zero.
- Retention bounds the stream server-side, so a panel that was closed for a long time may find the oldest lines gone. Treat a gap as normal, not as an error.
- The text is model-authored prose. Render it the way the transcript already renders model-influenced text, with the same escaping and length discipline.

Refs #954.

PR #1605, opened 2026-08-05
feat(server): live progress read model for background agent runs

### LINKED ISSUE #954: feat: live sibling/subagent inspection for background agents

naingthet:
A background run is observable only at its edges today: the state it is in (`GET /chats/{id}/agent-runs`), the checkpoint it currently sits on (the snapshot's `activity` and the activity history route), and the result it eventually submits. Between those there is nothing to read, so a parent or the desktop watching a long child sees a static "running" row until the run terminates. Durable join tells you when a child is done; nothing tells you what it is doing.

This adds the stream in between.

## The read model

`GET /chats/{chat_id}/agent-runs/{run_id}/progress?after_sequence=N&limit=M` returns a page of the bounded lines a background run published, each stamped with a monotonic per-run sequence, plus the cursor to resume from. A reader polls with the cursor it last held and receives only what has arrived since. A missing, wrong-chat, or foreground run answers `404` without revealing whether an unrelated run identifier exists — the same posture the activity history route already takes.

## Where the lines come from

Two producers, matching the two execution adapters:

- A container-located run's progress comes from the sandbox protocol's own resumable, sequenced event stream. The driver was already draining it and committing its cursor; `EventPayload::Progress` fell through a `_ => {}` and was discarded.
- An in-process run's progress comes from the text its model produced before checkpointing on a tool — the run's own account of what it is about to do. `complete_sandbox_task` was dropping it on the tool-use path.

Each line carries its producer's identity for that line: a protocol event sequence, or the durable checkpoint the narration belongs to. A unique `(run_id, source_key)` index makes republishing a no-op, so a reattached container redelivering a batch and a worker retrying an ambiguous commit both leave one line rather than two.

## Posture

The stream is deliberately outside the correctness contract. Nothing reads it to decide what a run may do next. The append takes no lease and commits separately from the transition that produced the line — it runs only after that checkpoint has already committed under a proven lease — and a failure to publish is reported and dropped rather than allowed to disturb a transition that succeeded. Retention bounds the table to a fixed number of lines per run, so a child that narrates for an hour costs a bounded number of rows and a reader that stops polling may find the oldest lines trimmed; order is the guarantee, not completeness.

Only the run's own prose crosses the boundary — the same class of text the terminal result already carries. Tool arguments, queries, results, folder and file identities, host paths, provider identities, executor leases, and raw diagnostics all stay server-side, exactly as they do for the activity projections.

## What is not here

The desktop consumption point is split out as #1604, so this PR is the server contract plus the typed client and defensive parser the UI slice will build on. The transcript rows and the background-agent panel still render state and activity only.

## Tests

Two, both behavioral:

- `agent_run_progress_is_ordered_resumable_and_bound_to_its_chat` drives the route: an empty run reads as an empty page whose cursor does not advance, republishing a producer key leaves one line, resuming from a cursor returns only what followed it, and the wrong chat answers `404`.
- The existing `checkpoints_web_search_after_a_text_preamble_and_rebuilds_its_receipt` worker test gains assertions that the preamble it already produces is published as progress and that reading past the cursor returns nothing — the write path, exercised where a real run produces it, rather than a second fixture.

Schema epoch bumped for the new table.

Closes #954.

### DIFF
--- a/crates/openwave-core/src/db/entities.rs
+++ b/crates/openwave-core/src/db/entities.rs
@@ -703,6 +703,27 @@ pub mod agent_run_result {
     impl ActiveModelBehavior for ActiveModel {}
 }
 
+pub mod agent_run_progress {
+    use sea_orm::entity::prelude::*;
+
+    #[derive(Clone, Debug, PartialEq, DeriveEntityModel)]
+    #[sea_orm(table_name = "agent_run_progress")]
+    pub struct Model {
+        #[sea_orm(primary_key, auto_increment = false)]
+        pub agent_run_id: Uuid,
+        #[sea_orm(primary_key, auto_increment = false)]
+        pub sequence: i64,
+        pub source_key: String,
+        pub text: String,
+        pub created_at: DateTimeUtc,
+    }
+
+    #[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
+    pub enum Relation {}
+
+    impl ActiveModelBehavior for ActiveModel {}
+}
+
 pub mod agent_run_cancellation {
     use sea_orm::entity::prelude::*;
 
--- a/crates/openwave-core/src/db/migration/baseline/agent_run.rs
+++ b/crates/openwave-core/src/db/migration/baseline/agent_run.rs
@@ -472,6 +472,73 @@ pub(super) fn agent_run_result_indexes() -> Vec<IndexCreateStatement> {
         .to_owned()]
 }
 
+/// The bounded, ordered progress stream one background run publishes while it
+/// works.
+///
+/// This is observation, not correctness state: nothing reads it to decide what
+/// a run may do next, and losing a line only leaves a gap in what an observer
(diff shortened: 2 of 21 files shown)


naingthet:
Labelled full-ci for the Postgres turn-state lane: the new table's migration and the progress ops (char_length check constraint, bare ON CONFLICT DO NOTHING, exec_without_returning) were only exercised against SQLite locally, and the Postgres store runs the same sea-query statements. The lane is the only pre-merge coverage of that backend.

PR #1609, opened 2026-08-05
feat(desktop): restyle background-agent detail with markdown result and tool timeline

naingthet:
This restyles the background-agent detail panel to use the same visual language as foreground tool calls. The task header now includes icon-backed status badges and history counts, activity renders as a collapsible icon rail with outcome states, and terminal results render as markdown beneath a divider instead of raw text in a muted box.

The activity rail defaults open while work is active and closed after the run settles, while preserving an explicit user toggle. A focused DOM test covers that transition.

Typed per-step previews such as command text and exit codes require additional server data and remain a follow-up server slice.

Verification:
- `pnpm exec tsc --noEmit`
- `pnpm exec vitest run src/AgentActivityTimeline.dom.test.tsx src/BackgroundAgentList.test.tsx src/AgentRunDisplay.test.ts src/MessageMarkdown.test.tsx src/MessageMarkdown.dom.test.tsx`
- `pnpm build`

### DIFF
--- a/crates/openwave-desktop/ui/src/AgentActivityTimeline.dom.test.tsx
+++ b/crates/openwave-desktop/ui/src/AgentActivityTimeline.dom.test.tsx
@@ -1,0 +1,60 @@
+// @vitest-environment jsdom
+import { cleanup, fireEvent, render, screen } from "@testing-library/react";
+import { afterEach, describe, expect, it } from "vitest";
+
+import {
+  AgentActivityTimeline,
+  type AgentActivityState,
+} from "./AgentActivityTimeline";
+
+afterEach(cleanup);
+
+const state: AgentActivityState = {
+  loading: false,
+  error: false,
+  loaded: true,
+  items: [
+    {
+      kind: "read_delegated_file",
+      outcome: "completed",
+      at: "2026-08-05T18:36:00Z",
+    },
+    {
+      kind: "exec",
+      outcome: "failed",
+      at: "2026-08-05T18:37:00Z",
+    },
+    {
+      kind: "web_search",
+      outcome: "completed",
+      at: "2026-08-05T18:38:00Z",
+    },
+  ],
+};
+
+describe("AgentActivityTimeline", () => {
+  it("follows the run by default and preserves an explicit user toggle", () => {
+    const { rerender } = render(
+      <AgentActivityTimeline state={state} active />,
+    );
+
+    const summary = screen.getByRole("button", {
+      name: "Ran 3 tool calls · 1 failed",
+    });
+    expect(summary.getAttribute("aria-expanded")).toBe("true");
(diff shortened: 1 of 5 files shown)


naingthet:
Needs changes.

1. **must-fix — `crates/openwave-desktop/ui/src/BackgroundAgentList.tsx:122`, `crates/openwave-desktop/ui/src/BackgroundAgentList.tsx:140`, `crates/openwave-desktop/ui/src/AgentActivityTimeline.tsx:86`** The user-toggle guarantee does not hold in the inline consumer across the status transitions driven by polling. Each row lives beneath a keyed status-group parent, so moving a run from Running to Completed unmounts and recreates `BackgroundAgentRow`; that resets both the outer `expanded` state and the timeline `userToggled` state. The new DOM test rerenders one persistent timeline instance, so it cannot catch the real consumer transition. Preserve the row/toggle preference across group moves and cover that transition at the consumer level.

2. **must-fix — `crates/openwave-desktop/ui/src/AgentActivityTimeline.tsx:190`** A durable `waiting` activity is mapped to the `running` tone. That gives accepted/retry-wait steps a spinning status glyph and pulsing title while the text says they are waiting, which misrepresents the state. Waiting needs a non-running presentation, such as the shared clock treatment.

3. **nit — `crates/openwave-desktop/ui/src/AgentActivityTimeline.tsx:129`, `crates/openwave-desktop/ui/src/AgentActivityTimeline.tsx:180`** The rail trigger always uses the settled, past-tense “Ran N tool calls” summary, even when its latest row is waiting or running. The live-state design calls for the current activity in that trigger and already carries the aggregate count in the header; reserve the past-tense count for terminal runs.

Verification passed: `pnpm exec tsc --noEmit`; the five focused Vitest files (29 tests); and `pnpm build`. No generated wire files changed.

naingthet:
Addressed all three findings in 9844a1ad:

- Kept the outer row and inner activity-rail preferences in `BackgroundAgentList`, keyed by run ID, so they survive Running → Completed status-group remounts. Added consumer-level DOM coverage for that polling transition.
- Rendered waiting history with the shared clock treatment and no running pulse or spinner.
- Changed live rail triggers to show the current run activity, while terminal runs retain the aggregate past-tense tool-call summary.

Verification is green: `pnpm exec tsc --noEmit`, six focused Vitest files (30 tests), and `pnpm build`.

naingthet:
Resolved — looks good to merge. `9844a1ad` hoists the outer-row and activity-rail preferences by run ID, so they survive the Running → Completed status-group remount. `BackgroundAgentList.dom.test.tsx` exercises that real consumer transition and would fail on the previous code when the remounted row lost its open state. Waiting entries now use the clock without running animation, and live triggers show the current activity while terminal triggers use the aggregate summary. Typecheck, all six focused Vitest files (30 tests), and `pnpm build` pass.

PR #1619, opened 2026-08-05
feat(server): typed previews in background-agent activity history

naingthet:
## Summary

Adds an optional tagged `detail` preview to background-agent activity history so clients can render the command argv and exit code, public-web query, or delegated-file display name for each durable step. The projection reuses the foreground action-preview sanitization for commands and searches, and reads exec exit status from the immutable sandbox receipt.

The preview stays headline-level by design: stdout, stderr, search results, raw arguments, host paths, opaque root identities, provider identities, executor leases, and diagnostics remain server-side. Captured output needs its own retention contract and is intentionally outside this change.

Activity history is still rebuilt on read from existing sandbox call and receipt rows; no new persisted history shape or migration is introduced. `detail` is optional and omitted when it cannot be derived, so older three-field entries continue to deserialize and older clients can ignore the additive field. The generated TypeScript wire contract and agent-run documentation are updated accordingly.

## Verification

- `cargo check -p openwave-core -p openwave-server --locked`
- `cargo test -p openwave-core preview::tests --locked`
- `cargo test -p openwave-server tests::sandbox --locked`
- `cargo test -p openwave-server wire_types --locked`

### DIFF
--- a/crates/openwave-core/src/lib.rs
+++ b/crates/openwave-core/src/lib.rs
@@ -184,8 +184,9 @@ pub use plan_mode::{
     MAX_PLAN_FEEDBACK_CHARS, MAX_PLAN_TITLE_CHARS, MIN_PLAN_CONTENT_CHARS,
 };
 pub use preview::{
-    format_bytes, ExecDegradation, ResultEntry, ResultEntryKind, ResultFailure, ToolActionPreview,
-    ToolResultPreview, MAX_RESULT_ENTRIES, MAX_RESULT_ENTRY_CHARS,
+    format_bytes, AgentActivityDetail, ExecDegradation, ResultEntry, ResultEntryKind,
+    ResultFailure, ToolActionPreview, ToolResultPreview, MAX_RESULT_ENTRIES,
+    MAX_RESULT_ENTRY_CHARS,
 };
 pub use provider::{
     ChatMessage, ChatRequest, ContentBlock, MessageReasoning, ModelProvider, ProviderEvent,
--- a/crates/openwave-core/src/preview.rs
+++ b/crates/openwave-core/src/preview.rs
@@ -1,8 +1,10 @@
 //! Closed renderer projections of what a tool will do, and what it did.
 //!
-//! The renderer boundary carries no tool arguments and no tool output. These
-//! types are the deliberate exceptions: a tool may opt in to showing a human
-//! the action under review and the result it produced, field by field.
+//! The renderer boundary carries neither raw argument objects nor raw tool
+//! output. These types are the deliberate exceptions: a tool may opt in to
+//! showing a human selected fields, one by one. Model-authored strings remain
+//! untrusted text and may repeat information the model already saw; the
+//! boundary guarantee is about which stored and host-owned fields are copied.
 //!
 //! They are not passthroughs. Each variant enumerates exactly what a person
 //! needs in order to consent to an action or to understand its outcome, values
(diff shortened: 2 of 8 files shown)


naingthet:
Needs changes.

1. **must-fix — the absolute information-boundary claim is not enforceable with these source fields** (`crates/openwave-core/src/preview.rs:70-88`, `docs/agent-runs.md:327-333`). `command`, every `args` element, and `query` are arbitrary model-authored strings. The child has already seen earlier exec and search results, so it can copy stdout, stderr, a search result, an opaque identity, a diagnostic, or a host-looking path into a later argument or query; the clamp only bounds that content. For example, an exec argument containing `/Users/alice/private.txt` crosses unchanged. Either the contract must be narrowed to say that no stored result or host-owned field is projected directly, with allowed model text explicitly outside that guarantee, or the projection must stop admitting arbitrary strings that can carry excluded data.

2. **must-fix — the advertised single-line sanitizer still permits visual line breaks and text-direction spoofing** (`crates/openwave-core/src/preview.rs:95-101`, `crates/openwave-core/src/preview.rs:1002-1014`). Rust `char::is_control` removes LF, CR, and ESC, but it does not remove U+2028 LINE SEPARATOR, U+2029 PARAGRAPH SEPARATOR, or bidi formatting characters such as U+202E. Those characters survive in command, argument, query, and file detail even though the comments promise a bounded single-line value that cannot forge card structure. The boundary predicate should reject Unicode separators and directional formatting too, and a hostile detail case should lock that behavior down.

3. **must-fix — the file projection bypasses the canonical path validators** (`crates/openwave-core/src/preview.rs:85-100`). The function says malformed arguments produce no detail and the `File` variant is never a host path, but the connected-folder branches read `arguments.path` directly and only clamp it. `{"path":"/Users/alice/private.txt"}`, traversal, or a payload missing the required root identity all produce `File` detail. The current history route whitelist makes these branches unreachable today, but this exported helper embeds an unsafe contract for its next caller. Remove the unused branches or decode and validate the exact tool arguments before projecting a display value.

4. **must-fix — the new boundary behavior has no focused hostile or receipt-format test** (`crates/openwave-server/src/tests/sandbox.rs:236-300`, `crates/openwave-core/src/preview.rs:119-139`). The server test uses benign argv/query values and manually writes the expected `exit: 17` receipt. No new `AgentActivityDetail` unit test proves argument-count and field clamps, Unicode rejection, malformed/absolute file paths, first-line-only exit parsing, or absence for `exit: signal` and setup failures. This is exactly the kind of renderer wire contract where one compact table-driven test earns its cost. The compatibility case without `detail` is covered, and the route test does cover recovery of a numeric exit code from a durable receipt.

5. **nit — the checked-in desktop contract description is stale, and the current parser discards the new field** (`crates/openwave-desktop/ui/src/api.ts:403-410`, `crates/openwave-desktop/ui/src/api.ts:2339-2361`). The comment still says history contains only kind, outcome, and timestamp and never queries or paths. `parseAgentActivityHistory` also reconstructs every entry without `detail`. If this is intentionally a server-only slice, the comment should say that the desktop ignores the additive field pending renderer work; otherwise the parser needs a closed, bounded, kind-matched detail validator before preserving it.

Verification passed: `cargo check -p openwave-core -p openwave-server --locked`, `cargo test -p openwave-core preview:: --locked`, `cargo test -p openwave-server tests::sandbox --locked`, and `cargo test -p openwave-server wire_types --locked`. The outer `detail` and nested `exit_code` are correctly `serde(default)`, omitted on `None`, and optional in generated TypeScript; the `wire.ts` diff is limited to the documented shapes/comments. `with_exec_result` reads only the first receipt line and never echoes receipt text. Delegated-file admissions are UTF-8 strings and canonical path validation rejects dot segments and trailing slashes before normal admission; invalid legacy values degrade to absent or a bounded leaf. Receipt lookup is terminal-exec-only, and the maximum 16 sandbox calls makes the sequential lookups bounded rather than an unbounded long-run N+1. No PostgreSQL-divergent turn-state path is changed, so the missing `full-ci` label is not a concern on that basis.

naingthet:
Addressed all review findings in `12c6f173`:

1. Narrowed the boundary contract in the Rust docs, generated wire docs, route docs, and `docs/agent-runs.md`. Command/argv/query and progress prose are now explicitly described as model-authored text that may repeat anything the child already saw. The enforceable guarantee is provenance-based: stored results and host-only fields are not copied directly, apart from the typed numeric exit code and the admitted delegated-file leaf name.
2. Hardened the shared single-line preview sanitizer to remove U+2028/U+2029, U+202A-U+202E, and U+2066-U+2069 in addition to control characters. This deliberately hardens both foreground approval previews and background activity details because they share the same clamp and exactness predicate.
3. Removed the unreachable connected-folder branches and generic exported file helper. `AgentActivityDetail::build` now admits only exec and web-search projections; delegated-file identity remains a separate admission-derived leaf-name path.
4. Added one compact table-driven core test covering controls, Unicode separators, bidi formatting, absolute-looking model argv, rejection of connected-folder path projection, field/argument clamps, first-line-only numeric exit parsing, and absence for signal/setup-failure receipts.
5. Updated the `api.ts` contract comment to state that the desktop intentionally discards additive `detail` until the renderer slice adds a closed, bounded, kind-matched validator; the parser remains unchanged.

Rebased onto current `origin/main` (including #1609) and re-ran:

- `cargo check -p openwave-core -p openwave-server --locked`
- `cargo test -p openwave-core preview:: --locked` (22 passed)
- `cargo test -p openwave-server tests::sandbox --locked` (21 passed)
- `cargo test -p openwave-server wire_types --locked` (10 passed)

naingthet:
Needs one small change before merge. The implementation resolves the five substantive findings, and all four requested Cargo gates pass. The remaining gap is exact regression coverage in `crates/openwave-core/src/preview.rs:1185-1247`: the table exercises U+2028, U+2029, U+202E, and U+2067, but it would still pass if U+202A-U+202D or U+2066/U+2068/U+2069 stopped being filtered. Please enumerate every requested code point in U+2028/U+2029/U+202A-U+202E/U+2066-U+2069 and assert the shared sanitizer through both the background detail projection and foreground exactness path. The provenance-based contract, removal of connected-folder file projection, receipt parsing cases, compatibility annotations, and desktop follow-up documentation all re-verify cleanly.

naingthet:
Updated in `c34f05ae`. The regression table now enumerates U+2028, U+2029, every code point from U+202A through U+202E, and every code point from U+2066 through U+2069 individually. Each row asserts sanitized background detail output and foreground exactness rejection. `cargo test -p openwave-core preview:: --locked` passes all 22 tests.

PR #1623, opened 2026-08-05
feat(desktop): show typed activity previews in the background-agent timeline

naingthet:
A background run's activity timeline showed only the generic label for each
step, so five commands in a row all read "Ran a command" and the reader had no
way to tell them apart. The wire already carries an optional typed `detail` per
entry — a bounded exec command/args (plus an exit code when the receipt
recorded one), a search query, or a delegated file's leaf name — which the
desktop was discarding pending a validator.

This adds that validator and renders the result.

`parseAgentActivityHistory` now keeps `detail`, but only when it can vouch for
it, and the check is closed on every axis: the tag must be one of the three
known kinds, it must match the entry's own activity kind (an exec headline only
on an exec step, a search headline only on `web_search`, a file headline only on
the kinds that name one file or folder), keys outside the admitted set are
rejected, string fields are length-bounded and free of control characters, and
an exit code must be an integer in signed 32-bit range. Anything else drops the
headline and keeps the entry, so a step never disappears from the timeline and a
mismatched payload can never describe an action other than the one that ran.
The detail is model-authored text on a surface that otherwise renders a closed
vocabulary, which is why it gets this treatment rather than a shape check.

In the timeline, each row now carries a truncated muted headline next to its
label: monospace for a command — formatted by the same argument-vector helper
the foreground approval and result cards use, now shared out of `ToolPreview` —
and plain text for a query or file name. A failed command with a recorded exit
code surfaces `Exit N` in destructive text, matching how the foreground status
badge prefers the specific exit over the generic outcome word.

Deferred: the expandable command/output cards from the panel design need
server-side output retention for background runs, which does not exist — no
output is stored for a delegated run's calls, so there is nothing to put behind
a tab yet.

Verified with `tsc --noEmit`, the touched vitest files, and `pnpm build`.

### DIFF
--- a/crates/openwave-desktop/ui/src/AgentActivityTimeline.dom.test.tsx
+++ b/crates/openwave-desktop/ui/src/AgentActivityTimeline.dom.test.tsx
@@ -72,13 +72,12 @@ describe("AgentActivityTimeline", () => {
   it("distinguishes a failed command by its own headline and exit status", () => {
     render(<AgentActivityTimeline state={state} active={false} expanded />);
 
-    const commandRow = within(screen.getByRole("list"))
-      .getByText("pip install matplotlib")
-      .closest("li");
-    expect(commandRow).toBeTruthy();
-    expect(
-      commandRow?.querySelector(".font-mono")?.textContent,
-    ).toBe("pip install matplotlib");
-    expect(within(commandRow!).getByText("Exit 1")).toBeTruthy();
+    const commandRow = within(screen.getByRole("list")).getAllByRole(
+      "listitem",
+    )[1]!;
+    expect(commandRow.querySelector(".font-mono")?.textContent).toBe(
+      "pip install matplotlib",
+    );
+    expect(within(commandRow).getByText("Exit 1")).toBeTruthy();
   });
 });
--- a/crates/openwave-desktop/ui/src/AgentActivityTimeline.dom.test.tsx
+++ b/crates/openwave-desktop/ui/src/AgentActivityTimeline.dom.test.tsx
@@ -23,6 +23,12 @@ const state: AgentActivityState = {
       kind: "exec",
       outcome: "failed",
       at: "2026-08-05T18:37:00Z",
+      detail: {
+        kind: "exec",
+        command: "pip",
+        args: ["install", "matplotlib"],
+        exit_code: 1,
+      },
     },
     {
       kind: "web_search",
@@ -61,5 +67,18 @@ describe("AgentActivityTimeline", () => {
     });
     expect(settledTrigger.getAttribute("aria-expanded")).toBe("false");
     expect(screen.queryByRole("list")).toBeNull();
+  });
+
+  it("distinguishes a failed command by its own headline and exit status", () => {
+    render(<AgentActivityTimeline state={state} active={false} expanded />);
+
+    const commandRow = within(screen.getByRole("list"))
+      .getByText("pip install matplotlib")
+      .closest("li");
+    expect(commandRow).toBeTruthy();
+    expect(
+      commandRow?.querySelector(".font-mono")?.textContent,
+    ).toBe("pip install matplotlib");
+    expect(within(commandRow!).getByText("Exit 1")).toBeTruthy();
   });
 });
--- a/crates/openwave-desktop/ui/src/AgentActivityTimeline.tsx
+++ b/crates/openwave-desktop/ui/src/AgentActivityTimeline.tsx
@@ -4,6 +4,7 @@ import { ChevronRight } from "lucide-react";
 import type { AgentActivityHistoryEntry } from "./api";
 import { agentActivityHistoryLabel } from "./AgentRunDisplay";
 import { ToolIcon } from "./ToolIcon";
+import { execCommandHeadline } from "./ToolPreview";
 import { ToolStatusIcon, type ToolTone } from "./ToolStatusIcon";
 import { Button } from "@/components/ui/button";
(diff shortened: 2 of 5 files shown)


naingthet:
Needs changes — one small must-fix in the validator, the rest are nits. The shape is right: the detail is validated per-tag, cross-checked against the entry's own kind, and a failing detail costs the headline rather than the row. `tsc`, the four touched suites (73 tests), and `pnpm build` are all green locally.

1. **must-fix** — `crates/openwave-desktop/ui/src/api.ts:2315` `bounded` rejects only C0/C1 (`< 32`, `127..159`), but the projection that produces this field strips more than that. `preview_formatting_character` in `crates/openwave-core/src/preview.rs` also removes `U+2028`/`U+2029` and the bidi controls `U+202A..U+202E` / `U+2066..U+2069`, with the comment that this clamp "is shared by foreground approval previews and background activity details, so extending it hardens both renderer surfaces together." The client-side check is the renderer's own trust boundary for this text, and a bidi override in a command headline does exactly what the control-character rejection is there to prevent — rewrites how the line reads. Mirroring the server's set is a one-line change to the predicate and keeps the two definitions from drifting apart.

2. **nit** — `api.ts:2398`: `Math.abs(value.exit_code) < 2 ** 31` excludes `i32::MIN` (-2147483648), which the server's `parse::<i32>()` can produce. A half-open range (`>= -(2 ** 31) && < 2 ** 31`) matches the wire type exactly. Worth deciding the failure mode too: an out-of-range code currently drops the whole detail, so the reader loses the command headline over a field that is only decoration next to it — dropping just `exit_code` would be the narrower response.

3. **nit** — `api.ts:2360`: `ACTIVITY_FILE_DETAIL_KINDS` admits four kinds, but `crates/openwave-server/src/routes.rs:2407` only ever pairs a `file` detail with the delegated-read tool. The other three are speculative, and the point of the cross-check is that it is exactly as wide as what the server projects. Narrowing to `read_delegated_file` closes the matrix now and forces a deliberate edit when a folder tool starts naming a file.

4. **nit** — `api.ts:2312`: the comment justifies letting empty strings through with "an empty command argument is a real element of an argument vector," but `clamped_list`/`clamp` in `preview.rs` drop any argument that clamps to empty, so one never arrives on this wire. The relaxation is harmless; the stated reason is not the real one.

5. **nit** — `AgentActivityTimeline.tsx:189`: `text-destructive` sits in the same row as the failed status icon's `text-critical` (line 166). These resolve to different values in dark mode (`styles.css:138` vs `:152`), so the badge and the icon it belongs to are two different reds. `text-critical` keeps the row consistent.

6. **nit** — `AgentActivityTimeline.tsx:172`: with `shrink-0` on the label, its own `truncate` can no longer do anything — a non-shrinking flex item never drops below its content width — and every sibling except the headline is now `shrink-0` too. On a narrow panel a long label plus `Exit N` plus the timestamp will overflow the row rather than elide anything. Fine if that is the deliberate priority (label and exit status always readable, headline gives way), but the dead `truncate` reads as if it still protects the row.

7. **nit** — `AgentActivityTimeline.tsx:173`: a running exec has a detail, so the label pulses while the command beside it stays static. Putting the tone on the row, or on both spans, avoids the half-animated line.

8. **nit** — `AgentActivityTimeline.dom.test.tsx:76-81`: the `getByText` lookup and the `.font-mono` `textContent` check assert the same string twice; the second one exists to claim the command renders monospace, which the first already located. Keeping the class assertion alone says that in one step. The added case is otherwise well aimed — it would catch the headline disappearing or landing on the wrong row.

The `parseAgentActivityHistory` case earns its place: five payloads through one closed validator, each one a distinct way the check can fail, asserted on the whole projection rather than a field. Two shapes it does not reach and a future refactor of the argument loop could regress: a non-string element inside `args`, and a fractional `exit_code`. Cheap to fold into the same array if you want them.

`execCommandHeadline` is a clean extraction — the foreground path is byte-identical and `ToolPreview`/`ToolCallCard` still cover the quoting. Scope is clean: nothing outside `crates/openwave-desktop/ui`, generated wire types untouched.

naingthet:
Pushed. Took the must-fix and all six nits, plus the two extra parser cases.

1. `bounded` now rejects the same set the projection's `preview_formatting_character` strips — C0/C1, `U+2028`/`U+2029`, `U+202A..U+202E`, `U+2066..U+2069` — with a comment pointing at the server clamp so the two stay together.
2. Exit-code validation moved into its own helper over the half-open i32 range, so `i32::MIN` is admitted and an unusable code drops only itself. The command headline survives, which is the part worth showing.
3. `ACTIVITY_FILE_DETAIL_KINDS` narrowed to `read_delegated_file`, with a comment saying widening follows the server rather than anticipating it.
4. Rewrote the empty-string justification: nothing on this wire is expected to be empty, so the relaxation is only about not failing a payload over a field whose emptiness proves nothing.
5. `Exit N` uses `text-critical`, matching the failed icon beside it.
6. Dropped `shrink-0` from the label, so it shrinks and elides on a narrow panel instead of pushing the row over. The dead `truncate` now does what it says.
7. Took it — the pulse is on both spans, so a running step animates as one line and the timestamp stays still.
8. Took it — the row is located by index and the monospace assertion carries the string once.

Extra cases folded into the same array: a non-string element in `args` (drops the detail, since a headline built from it would misdescribe what ran), a fractional `exit_code` (keeps the command, drops the code), and a bidi override in an argument.

`tsc --noEmit`, the four suites (73 tests), and `pnpm build` are green.

PR #1664, opened 2026-08-06
feat(core): load user-authored plugins from the data directory

### LINKED ISSUE #1599: feat(core): load user-authored plugins from the data directory
Part of #1570, deferred out of #1572 (noted in PR #1582).

`Config::exec_plugins_dir` and `user_plugins_dir()` exist, but only the built-in `plugins/` tree loads. Wire user plugins end-to-end: load `{data_dir}/plugins/<name>/PLUGIN.md` through the same strict parser, with the same rules skills already enforce — built-in names reserved, no shadowing, a user plugin cannot claim a skill a built-in plugin owns, skip-with-warning posture. User plugins surface in `GET /plugins` and the Plugins UI attributed as the user's own, with the same enable flags (`plugins.enable_state` already records flags for not-currently-present names, so state survives removal/reinstall).

Small slice: loader merge rules + tests, catalog origin attribution. No UI changes needed beyond what #1597 already renders for user-origin content.

naingthet:
`Config::user_plugins_dir()` was already derived and the directory already
created, but only the bundled `plugins/` tree ever loaded — a `PLUGIN.md` the
user wrote in `{data_dir}/plugins/<name>/` was silently ignored.

User bundles now load through the same strict manifest parser as built-ins
(`load_plugins` takes the origin; parsing itself is identical for both, so a
user bundle is held to exactly the built-in grammar) and merge behind them via
a new `merged_plugins`, mirroring `merged_skills`/`merged_prompts`.

Merge rules, all skip-with-warning so one bad bundle never fails the catalog:

- a built-in bundle's name is reserved — a user bundle claiming it is dropped
  rather than shadowing curated grouping;
- a skill a built-in bundle already claims cannot be re-claimed, and neither
  can a prompt;
- membership resolves against the *merged* skill and prompt sets, so a user
  bundle can group the skills the user wrote; a member that belongs to no
  built-in bundle is fair game;
- within the user tree itself, `load_plugins`' existing single-claim rule
  applies unchanged (name order, deterministic).

Membership is resolved against *installed* rather than enabled skills: a bundle
whose member is switched off is still the bundle you installed, and dropping it
would take the switch that turns the member back on off the surface with it.

`installed_plugins()` and the prompt-composition `plugin_catalog()` now both go
through the merge, so a user bundle groups its skills in the operating prompt
and its flag gates its members exactly as a built-in one does — `owning_plugin`
resolves against the merged set instead of the built-in slice.

Enable state needed no change: flags are recorded against the name, so a user
bundle switched off, uninstalled, and reinstalled comes back off.

`PluginInfo` gains a host-derived `origin` (there is no manifest key for it),
regenerated into `wire.ts`. The library splits the bundle list the same way it
already splits skills, so bundles the user wrote appear under "Your plugins".

Tests: two added. One in the loader for the merge/reserved-name/re-claim rules
and origin attribution — each is a one-line invariant that a refactor could
reverse without any other test noticing. One in the server for the `GET
/plugins` origin contract and enable-state survival across
remove-and-reinstall, which is the behavior a user would notice breaking and
which no unit test can reach. Three UI fixtures gained the new required field;
none removed.

Verified locally: `cargo check --locked --all-targets` and `cargo clippy` on the
touched crates, the full `openwave-code-execution` suite, the server's plugin
tests, `cargo fmt`, and the UI's `tsc` + `pnpm test`. Not verified: driving the
running app.

Closes #1599

### DIFF
--- a/crates/openwave-code-execution/src/lib.rs
+++ b/crates/openwave-code-execution/src/lib.rs
@@ -57,8 +57,8 @@ pub use package_cache::{
 };
 pub use plugins::{
     derived_capabilities, is_valid_plugin_name, is_valid_plugin_router_preamble, load_plugins,
-    parse_plugin_manifest, LoadedPlugin, PluginCapability, PluginCategory, PluginPackage,
-    PluginParseError, PLUGIN_MANIFEST_FILE,
+    merged_plugins, parse_plugin_manifest, LoadedPlugin, PluginCapability, PluginCategory,
+    PluginOrigin, PluginPackage, PluginParseError, PLUGIN_MANIFEST_FILE,
 };
 pub use preview::{scan_preview_directory, PreviewScan};
 pub use prompts::{
--- a/crates/openwave-code-execution/src/plugins.rs
+++ b/crates/openwave-code-execution/src/plugins.rs
@@ -28,6 +28,10 @@
 //! a plugin naming a skill that isn't there is skipped whole rather than
 //! advertising a group the catalog cannot render.
 //!
+//! User-authored bundles under the data directory load through that same
+//! parser and merge behind the built-in tree — see [`merged_plugins`] for the
+//! precedence rules, which mirror the ones user skills already follow.
+//!
 //! Capability badges follow the same posture from the other direction: they
 //! are derived from what the member skills actually declare, and the closed
 //! key set means a manifest cannot state them at all.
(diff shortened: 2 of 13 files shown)


naingthet:
Superseded: #1687 was stacked on this branch and its squash-merge carried this PR's entire diff to main (loader merge rules, PluginOrigin, Your-plugins UI section all present on main). Closing rather than rebasing an empty diff. Closes #1599's scope — closing that issue manually since the auto-close keyword rode this PR.

PR #1684, opened 2026-08-06
fix(core): enforce tool schemas at the registry boundary

### LINKED ISSUE #1677: feat(server): validate MCP tool arguments against advertised input schemas at the mount boundary
Advertised `input_schema` on MCP-mounted tools is never checked against incoming arguments — the model can send arguments that violate the schema the server declared, and they pass straight through. This is a live gap for servers mounted today and a hard prerequisite for third-party MCPB bundles (#1575).

Scope: validate arguments against the tool's declared JSON Schema at the mount boundary before dispatch; on mismatch, return a typed tool error to the model (so it can self-correct) rather than forwarding. Cover the built-in registry path too if it shares the gap. Tests: the contract that a violating call is rejected with the typed error and a conforming call passes through unchanged.

Part of the #1570 trust groundwork; sequenced ahead of #1575.

naingthet:
Closes #1677

## What changed

`ToolRegistry` now wraps every server-side executor with the JSON Schema validator compiled from the exact `ToolSpec` stored at registration. The agent and inbound MCP server still validate before their approval gates; the registry wrapper closes the remaining direct-dispatch path used by mounted MCP proxies and other built-in registry consumers.

A violating call returns a typed `invalid_arguments` tool result with the schema needed to retry. The underlying tool is never entered, so an MCP proxy cannot send the call to the remote server. Conforming arguments are passed through unchanged.

Schema compilation defaults undeclared schemas to draft 2020-12 and honors supported explicitly declared drafts. Unknown extension keywords produce a warning and are ignored while supported constraints remain active. Invalid schemas and unsupported `$schema` dialects also warn, but deliberately fail open: a server with an exotic or broken schema remains usable rather than having every call rejected.

The existing pinned `jsonschema = "=0.49.2"` dependency is reused. `tracing = "=0.1.44"` was already resolved in the lockfile; the lock change only records it as a direct core dependency for the fail-open warnings.

## Tests

- `cargo test -p openwave-core --locked` — 570 passed, 1 real-keychain test ignored
- `cargo test -p openwave-mcp --locked` — 52 passed, plus 1 doc-test
- `cargo check -p openwave-core -p openwave-mcp --locked`
- `cargo clippy -p openwave-core -p openwave-mcp --all-targets --locked -- -D warnings`
- `GGML_CCACHE=OFF cargo check --workspace --locked`
- `cargo fmt --all --check`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4280,6 +4280,7 @@ dependencies = [
  "tempfile",
  "thiserror 2.0.19",
  "tokio",
+ "tracing",
  "ts-rs",
  "uuid",
  "windows-sys 0.61.2",
--- a/crates/openwave-core/Cargo.toml
+++ b/crates/openwave-core/Cargo.toml
@@ -48,6 +48,7 @@ schemars = { workspace = true }
 jsonschema = { workspace = true }
 sha2 = { workspace = true }
 thiserror = { workspace = true }
+tracing = "=0.1.44"
 uuid = { workspace = true, features = ["v5"] }
 chrono = { workspace = true }
 ts-rs = { workspace = true }
--- a/crates/openwave-core/src/agent.rs
+++ b/crates/openwave-core/src/agent.rs
@@ -16,11 +16,12 @@
 //! - context reduction is deterministic floor+restore (no LLM summarization);
 //!   retries with progressive reduction on provider prompt-too-long errors.
 
-use std::collections::{BTreeMap, HashMap, HashSet};
+use std::collections::{BTreeMap, BTreeSet, HashMap, HashSet};
 use std::sync::atomic::{AtomicI32, Ordering};
 use std::sync::Arc;
 use std::time::Duration;
 
+use async_trait::async_trait;
 use chrono::Utc;
 use futures::channel::{mpsc::UnboundedSender, oneshot};
 use futures::future::{self, Either};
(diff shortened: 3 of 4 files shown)


naingthet:
Reviewed with the security hat on — this passes.

What I checked:
- **No bypass path**: the wrapper is installed inside `ToolRegistry::register`, so every server-side executor — including mounted MCP proxies and any direct registry consumer — goes through validation before dispatch. The pre-approval-gate validation in the agent and MCP server stays intact, so a violating call still never reaches a consent prompt.
- **Fail-open is scoped correctly**: unknown `$schema` dialects, uncompilable schemas, and unknown extension keywords all warn-and-continue rather than bricking a working server, while supported constraints remain enforced (both tests pin that with the `x-fixture-constraint` extension). This is the right trade for third-party servers and it's documented in the code where it happens.
- **Draft handling**: honoring an explicitly declared supported dialect instead of force-applying 2020-12 fixes the draft-07 tuple-`items` mis-validation the old comment worried about — strictly better than the previous pin-and-refuse.
- **No pathological-schema DoS at call time**: compilation happens once at registration; per-call work is a compiled-validator check.

Two non-blocking observations, fine to leave as-is: the mismatch message echoes the full `input_schema` (already in the model's prompt via the tool spec, so no new exposure — just worth remembering if specs ever get huge), and the validator now lives in both the wrapper and the registry entry (small duplicate memory, correct behavior).

Issue #1576, opened 2026-08-05
feat: plugin marketplace — install from git and public skill indexes


naingthet:
Part of #1570. Deferred until #1572–#1574 land; instruction-only content first.

Distribution without a server, in line with the local-first posture:
- **Install source = a git repo or archive** pinned by tag or commit SHA; installs copy into a local cache and never load from the untrusted source path; strict parse or reject the whole plugin; built-in names reserved against impersonation.
- **Import from the public skill ecosystem**: with Agent Skills spec conformance (#1571), standard-format skills from public indexes install as single-skill plugins directly — the catalog problem is solved upstream; our work is curation, trust, and UX.
- **Versioning**: manifest version field, else pinned SHA as identity; explicit update action, no background auto-update initially.
- **Trust tiers**: instruction-only plugins install with a summary confirmation; anything capability-bearing follows the #1575 per-component approval flow.

Open questions: whether to ship a curated in-app directory of vetted plugins (parallel to the curated MCP server list, #1512), and whether user-contributed submissions get a review pipeline or stay install-by-URL.

naingthet:
Scoping decision for this slice: keep it narrow and be explicit about what's unsupported.

- **First-class**: instruction-only, sandbox-compatible skills (Agent Skills format). Install from git/archive pinned by SHA into the user plugins dir (#1599's loader path), strict parse or reject whole.
- **Included with a disclaimer**: skills that assume host tooling or network (Claude Code-style `gh`/deploy/web skills). The importer runs a static compatibility check — declared deps vs the sandbox image closure, `scripts/` presence — and stamps a visible compatibility badge instead of letting them degrade silently at runtime.
- **Out of scope, on the roadmap**: converting foreign plugin formats' other member types (hooks, slash commands, agent definitions). The importer extracts what maps (skills) and visibly lists what it skipped — no converter.
- **Curation**: seed a small vetted in-app directory (~10 skills actually tested in the sandbox); everything else stays install-by-URL. No submission/review pipeline until there's demand.

MCP-server bundles remain #1575's tier and are the most-wanted deferred piece — sequencing note there.

naingthet:
Slice 1 (install from a pinned git URL, compatibility stamp, visible skips) landed in #1687. Remaining scope per the scoping comment above: the small curated in-app directory and any install UX beyond the endpoint. Un-claiming; the remainder is unscheduled.

naingthet:
[...]

PR #1687, opened 2026-08-06
feat(server): install plugins from a pinned git URL

naingthet:
Stacked on #1664, which supplies the user plugin loader and `{data_dir}/plugins` merge rules. Review that PR first; this PR still targets `main` and can be rebased once #1664 lands.

Adds `POST /plugins/install` for instruction-only plugins pinned by a tag or full commit SHA. Git repository URLs are fetched through their source archive without executing git; direct zip, tar, and tar.gz URLs are also accepted. Fetches are HTTPS-only, bounded, redirect-vetted, DNS-pinned, and refused when they resolve into private or otherwise denied address space.

The importer unpacks into memory, strict-parses every `PLUGIN.md` and `SKILL.md` it finds with the existing parsers, and publishes only validated manifests and one-level `scripts/` files into the data directory. Runtime loading always goes back through the copied user skill/plugin trees. A bad member rejects the whole import before publication. After copying, the existing merged loaders decide reserved-name and member-ownership conflicts; a rejected merge rolls back the newly created files rather than duplicating those rules.

A repository containing one Agent Skills-format `SKILL.md` is wrapped as a one-skill plugin. Unsupported repository members are returned in the install response instead of being converted or silently ignored.

Imported plugins receive a host-owned compatibility stamp. Exact Python pins are compared with the prepared sandbox image closure, helper scripts are disclosed as opaque assumptions, and `GET /plugins` now reports `compatible`, `limited`, or `unchecked` with typed reasons. Hand-authored plugins without an importer stamp remain visible as unchecked.

Tests cover the routed single-skill install, whole-plugin rejection when one skill fails parsing, and compatibility persistence from the on-disk stamp through `GET /plugins`.

Verified locally:
- `cargo fmt --all --check`
- `cargo check -p openwave-code-execution -p openwave-server --locked`
- `cargo clippy -p openwave-code-execution -p openwave-server --all-targets --locked -- -D warnings`
- `cargo test -p openwave-server plugin --locked`
- `cargo test -p openwave-code-execution plugins::tests --locked`
- `cargo test -p openwave-server wire_types::tests::the_generated_bindings_are_current --locked`

The UI TypeScript compiler was not run because this worktree has no installed UI dependencies; the generated wire staleness check passed.

Part of #1576.


naingthet:
Reviewed with an SSRF/supply-chain focus — this passes.

What I checked:
- **Fetch path**: reuses `admit_fetch_url`/`admit_fetch_address` from the web-search crate (the already-hardened egress gate) rather than rolling its own; redirects are disabled in the client and re-vetted per hop including fresh DNS admission; resolved addresses are pinned via `resolve_to_addrs`, closing the DNS-rebinding TOCTOU; HTTPS-only, `no_proxy`, bounded time, bounded bytes on both the wire and each decompressed file. This is the right shape end to end.
- **Archive handling**: traversal-proof (`enclosed_name` + component normalization rejecting parent/root/prefix), symlinks and special files rejected outright, duplicate paths rejected, entry-count and unpacked-size caps against archive bombs, non-portable names rejected.
- **Trust boundary**: the untrusted tree is never loaded at runtime — only parser-validated manifests and one-level scripts are published, the canonical regenerated manifest is re-parsed before a byte is copied, and a rejected merge rolls back. Foreign plugin members (`.claude-plugin`, `agents`, `commands`, `hooks`, `mcp*`) are skipped *visibly* in the response, which is exactly the agreed posture.
- **Compatibility stamp**: host-owned, typed reasons, hand-authored plugins stay `unchecked` — consistent with the derived-honesty invariant.

Non-blocking notes, fine as follow-ups: the archive-URL revision check is string containment (a UX pin, not a cryptographic one — worth a docs line so nobody mistakes it for integrity verification); non-GitHub forges use the `/archive/` layout so GitLab's `/-/archive/` form will 404 (degrades safely, worth a follow-up issue when demand appears); and the `unused-mut` comment above `skipped.sort_by` reads stale — drop it in any later touch.

PR #1709, opened 2026-08-06
feat(plugins): install Agent Plugins manifests (plugin.json)

naingthet:
Teaches the plugin importer to read packages published in the [Agent Plugins](https://agent-plugins.org/) standard format (v1.0.0), so a plugin authored against the public spec installs into OpenWave without being rewritten by hand.

## What changed

Format is decided by archive content, not by the request: a root `plugin.json` takes the new standard path, otherwise the existing `PLUGIN.md` path, otherwise the existing lone-`SKILL.md` path.

`plugin.json` is validated as the closed schema [§5](https://agent-plugins.org/specification) defines. `$schema` is required and is how a client selects its validation rules — locally; the parser never fetches a schema while loading — so an identifier this client does not implement rejects the package instead of being guessed at. The two violations the spec grades as non-fatal, an unknown top-level field and a non-object `extensions`, are reported and ignored; every other schema violation refuses the import. Optional metadata is checked by JSON type only, so a non-SemVer `version`, an unparseable URL, or a non-SPDX `license` are all valid manifests here — being stricter than the format would reject packages that are perfectly conformant.

Skills are discovered structurally at the fixed location (§7.1): every immediate child directory of `skills/` holding a regular `SKILL.md`, with no recursion below that. A nonconforming skill is skipped and surfaced through the existing `skipped` disclosure rather than sinking the package or its siblings.

That last point is a deliberate difference between the two formats, documented at the module head: a `PLUGIN.md` *names* its members, so a member that does not parse means the manifest describes something that is not there and the whole import is still refused. A standard package *discovers* its skills, and the spec requires per-component failure grading.

OpenWave's own data rides in `extensions` under `io.brightwave.openwave` (`category`, `router-preamble`), converted into the canonical `PLUGIN.md` the loaders already read. Other namespaces are passed over without inspecting their contents, as §8 requires. A value we cannot use inside our own namespace is reported and ignored, not fatal — the package is still correct for every other client.

The install stamp now records the source format and the spec version it targeted. The field is defaulted, so stamps written before this change keep validating rather than degrading to `unchecked`.

## Name reconciliation

The spec's package-name grammar admits `.`; OpenWave's plugin slug does not. Since the slug is also the installed directory name, the `/plugins/enabled` toggle key, and what the UI addresses a bundle by, widening it would have to move the loader, the routes, and the UI in one change. **This PR refuses a dotted package** with an explicit error naming the limitation, rather than silently renaming it. Everything else in the two grammars already agrees (both forbid `--`, require alphanumeric ends, and cap at 64 characters), so only dotted names are affected. Widening the slug is a reasonable follow-up if real packages need it.

`mcp.json` is reported as a skipped component here; retaining it is the next PR in the stack.

## Tests

Three end-to-end installs through the real route, extending the existing fixtures: a standard package with a nonconforming skill (installs, bad skill reported skipped), the two refusals above (unsupported `$schema`, dotted name), and one asserting an unknown top-level field is reported-not-fatal while our extension round-trips into the installed manifest. Two unit tests in the parser cover the violation grading and the extension namespace rules. No tests were deleted.

`cargo fmt`, clippy `-D warnings` on both touched crates, `cargo test -p openwave-code-execution`, `cargo test -p openwave-server --lib tests::configuration`, and `cargo check --workspace --locked` (no lockfile change) all pass locally.

### DIFF
--- a/crates/openwave-code-execution/src/agent_plugins.rs
+++ b/crates/openwave-code-execution/src/agent_plugins.rs
@@ -1,0 +1,388 @@
+//! The Agent Plugins packaging format (<https://agent-plugins.org>), v1.0.0.
+//!
+//! A plugin published in that format is a directory whose root carries a
+//! `plugin.json` manifest, with components discovered at fixed locations —
+//! skills under `skills/`, one directory each. This module is the reader for
+//! that manifest: it validates the closed schema the specification defines and
+//! hands back the few facts OpenWave's own [`crate::plugins`] representation
+//! needs, so the importer can convert a standard package into the internal
+//! `PLUGIN.md` shape the loaders already understand.
+//!
+//! Two properties of the specification drive the shape of this code:
+//!
+//! * **`$schema` selects the validation rules.** It is required, and a client
+//!   that does not implement the identifier it names must reject the plugin
+//!   rather than guess. Schemas are never fetched while loading.
+//! * **Failures are graded.** Only two manifest violations are non-fatal —
+//!   unknown top-level fields, and an `extensions` value that is not an
+//!   object — and both are reported and ignored. Everything else rejects the
+//!   plugin whole, so a package never loads in a shape its author did not
+//!   describe.
+//!
+//! Client-specific data rides in `extensions` under reverse-domain
+//! namespaces. OpenWave reads [`OPENWAVE_EXTENSION_NAMESPACE`] and ignores
+//! every other namespace without inspecting it, which is what the
+//! specification requires of a client that does not implement one. Because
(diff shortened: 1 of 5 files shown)


naingthet:
Review fix: the standard-format scan matched any retained file whose basename was `plugin.json`, including one under a skill's `scripts/` — which would misroute a `PLUGIN.md` bundle into the standard path (and reject outright if two such files existed). Both the retention rule and the scan now go through one `is_package_root_file` helper that only accepts the positions a package root can occupy and never a path inside `scripts/`. The existing compatibility test's archive gained a `scripts/plugin.json`; it fails on the old scan and passes now.

PR #1710, opened 2026-08-06
feat(plugins): retain bundled MCP server config from Agent Plugins

naingthet:
Stacked on #1709. A package in the [Agent Plugins](https://agent-plugins.org/) format may ship an `mcp.json` at its root declaring the MCP servers it brings ([§7.2](https://agent-plugins.org/specification)). The importer dropped that file; it now validates it, retains a canonical copy in `plugins/<name>/`, and the loader reads it back so the catalog reflects what a plugin actually carries.

Nothing is connected or spawned here — runtime wiring is a later change. What lands is the validated, retained configuration and the badge derived from it.

## Validation

The three transports are a closed union selected by `type`:

- **stdio** — `command` is a single executable token: a bare program name the platform resolves, or a `./`-relative path inside the plugin. No placeholder is expanded in a command, so one appearing there would be launched literally and is refused instead. `cwd` must be `./`-relative or rooted at `${PLUGIN_ROOT}` / `${PLUGIN_DATA}`, and containment is checked on the path that remains after the placeholder — which is what expansion will produce. An `env` entry naming either reserved variable invalidates its server, since the client sets both last and the entry is describing an effect it cannot have.
- **streamable-http** and legacy **sse** — an absolute HTTP(S) URL with no userinfo and no fragment, HTTPS except for `localhost` and loopback literals. Headers are literal, never expanded, must be valid HTTP field names, and a case-insensitive duplicate invalidates the entry.

`env` values and header values are visible package data, not a secret mechanism — the spec is explicit about this. They are never copied into a reason string, an API response, or a log line.

The `${PLUGIN_ROOT}` / `${PLUGIN_DATA}` expansion rule (single non-recursive textual pass, only in `args`, `env` values, and `cwd`) matters when a server is launched, which is not this PR; it is documented on the types so the connect path inherits it.

## Failure grading

Per §7.2.2: an unreadable file, a `$schema` this client does not implement or whose version disagrees with the manifest's, or a top level that is not exactly `$schema` plus `mcpServers` disables MCP for that plugin alone. An unknown field, unknown `type`, or cross-variant field drops only the entry carrying it. Both go through the existing `skipped` disclosure (`mcp.json` for the whole file, `mcp.json#<server>` per entry), so an install response says what will not connect and why. `mcp/` and `mcp-servers/` — the Claude Code convention directories, not part of this standard — keep being reported as foreign members.

The retained file is regenerated from the entries that passed rather than copied through, and the loader re-parses it with the same parser, so a hand-edited copy cannot widen what a plugin claims.

## The badge

`PluginCapability::Mcp` finally has a deriving source: the loader counts the entries in the retained configuration into a host-derived `PluginPackage::mcp_servers`, and `derived_capabilities` reads that. Capabilities stay derived and unstatable by a manifest — `capabilities_derive_from_member_deps_and_never_from_the_manifest` is extended rather than replaced, and still asserts that a manifest declaring its own badges is rejected.

`url` is added to `openwave-code-execution` at the `=2.5.8` the workspace already resolves; hand-rolling URL parsing on a security boundary is not worth saving an edge in the lockfile. No version in `Cargo.lock` changes.

## Tests

Two end-to-end installs through the real route: one with a valid stdio entry, a valid streamable-http entry, and an entry with an unknown field (plugin installs, only the valid two are written back, `mcp` badge derived, bad entry reported skipped), and one whose `mcp.json` fails at the top level (plugin and its skills install, no configuration retained, no badge, disablement reported). One parser unit test covers per-entry rejection for each rule above, whole-file rejection, and the canonical round-trip. No tests were deleted.

`cargo fmt`, clippy `-D warnings` on both touched crates, `cargo test -p openwave-code-execution`, `cargo test -p openwave-server --lib tests::configuration`, and `cargo check --workspace --locked` pass locally.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4262,6 +4262,7 @@ dependencies = [
  "tracing",
  "trash",
  "ts-rs",
+ "url",
  "uuid",
 ]
 
--- a/crates/openwave-code-execution/Cargo.toml
+++ b/crates/openwave-code-execution/Cargo.toml
@@ -28,6 +28,7 @@ thiserror = { workspace = true }
 tokio = { workspace = true, features = ["io-util", "net", "process", "rt", "sync", "time"] }
 tracing = "=0.1.44"
 trash = { workspace = true }
+url = "=2.5.8"
 uuid = { workspace = true }
 
 [target.'cfg(unix)'.dependencies]
--- a/crates/openwave-code-execution/src/agent_plugins.rs
+++ b/crates/openwave-code-execution/src/agent_plugins.rs
@@ -286,6 +286,562 @@ pub fn parse_agent_plugin_manifest(
     })
 }
 
+/// The bundled MCP server configuration file, at the package root.
+pub const AGENT_PLUGIN_MCP_FILE: &str = "mcp.json";
+
+/// The only `mcp.json` `$schema` identifier this client accepts. Both schemas
+/// share the specification version, and a file whose version disagrees with the
+/// manifest's disables MCP for that plugin alone.
+pub const AGENT_PLUGIN_MCP_SCHEMA_ID: &str =
+    "https://agent-plugins.org/schemas/1.0.0/mcp.schema.json";
+
+/// The absolute, resolved plugin root, provided to a launched server.
+pub const PLUGIN_ROOT_VARIABLE: &str = "PLUGIN_ROOT";
+/// The client-managed writable data directory, provided to a launched server.
+pub const PLUGIN_DATA_VARIABLE: &str = "PLUGIN_DATA";
+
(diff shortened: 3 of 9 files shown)


naingthet:
Review fix: a root `mcp.json` in a `PLUGIN.md` bundle or a bare skill package was retained in memory and then silently dropped — it never reached the standard-format path and the foreign-member list only names the `mcp/` and `mcp-servers/` directories. Both non-standard paths now report it through the ordinary skip disclosure with "bundled MCP configuration is only read from Agent Plugins packages", pushed ahead of the generic per-member reason so that is what a response shows. Asserted in the existing compatibility test.

PR #1717, opened 2026-08-06
fix(desktop): accept path_prefix grant rungs when recovering pending approvals

### LINKED ISSUE #1712: fix(desktop): a pending write_file approval makes its chat unloadable
Entering or reloading a chat while a `write_file` approval is pending fails hydration entirely: the transcript renders only "Could not load this chat: Error: pending approval response contains an invalid item", no socket is opened, and the chat stays unusable until the approval resolves elsewhere.

Root cause: the server offers `{"path_prefix":{"segments":N}}` grant rungs for workspace file writes. The generated wire type includes the variant (`ui/src/generated/wire.ts` `ApprovalGrantRung`), but the hand-written runtime parser `parseApprovalGrantRung` in `ui/src/api.ts` only accepts `exact_action`, `whole_tool`, and `command_prefix`. `listPendingApprovals` treats one unparseable rung as a fatally invalid response, and `ChatRoute` hydration turns that into the error state above.

The live socket path does not validate rungs, so approval cards work during streaming — the failure only appears on chat entry/reload, which in Ask mode is easy to hit constantly.

Reproduced against a dev `openwave serve` with the browser UI: captured payload

```json
{"action":"write_file","approval":"workspace_may_modify_files","grant_rungs":[{"path_prefix":{"segments":2}},{"path_prefix":{"segments":1}}], ...}
```

Fix: accept the `path_prefix` variant (positive integer `segments`) in `parseApprovalGrantRung`, mirroring the wire type.

abaas-bw:
## Summary

Entering or reloading a chat while a `write_file` approval was pending rendered only "Could not load this chat: pending approval response contains an invalid item" — no transcript, no event socket — until the approval resolved from somewhere else. In Ask mode this state is easy to sit in for most of a turn.

The server offers `{"path_prefix":{"segments":N}}` grant rungs on workspace file writes, and the generated wire type (`ApprovalGrantRung`) includes the variant, but the hand-written `parseApprovalGrantRung` only accepted `exact_action`, `whole_tool`, and `command_prefix`. `listPendingApprovals` treats one unparseable rung as a fatally invalid response, and chat hydration turns that into the error state. The approval card's grant ladder already handles `path_prefix`; only the recovery parser lagged the wire type.

Reproduced live against a dev `openwave serve` (captured payload in #1712). Closes #1712.

## Testing

- New regression test mirroring the captured server payload (two `path_prefix` rungs on a `write_file` approval), plus a rejection case for non-positive `segments`.
- `pnpm vitest run src/api.test.ts src/ChatApprovalHydration.test.ts` — 51 passed.

### DIFF
--- a/crates/openwave-desktop/ui/src/api.test.ts
+++ b/crates/openwave-desktop/ui/src/api.test.ts
@@ -362,6 +362,34 @@ describe("pending approval recovery", () => {
     });
   });
 
+  // A pending write_file approval offers path_prefix rungs. Rejecting the
+  // variant made every chat with one pending unloadable (#1712), because one
+  // invalid rung fails the whole hydration response.
+  it("accepts the path_prefix rungs a write_file approval offers", () => {
+    expect(
+      parsePendingToolApproval({
+        ...safe,
+        action: "write_file",
+        approval: "workspace_may_modify_files",
+        class: "workspace",
+        preview: { tool: "write_file", path: "data/parser.js" },
+        grant_rungs: [
+          { path_prefix: { segments: 2 } },
+          { path_prefix: { segments: 1 } },
+        ],
+      })?.grantRungs,
+    ).toEqual([
+      { path_prefix: { segments: 2 } },
+      { path_prefix: { segments: 1 } },
+    ]);
+    expect(
+      parsePendingToolApproval({
+        ...safe,
+        grant_rungs: [{ path_prefix: { segments: 0 } }],
+      }),
+    ).toBeNull();
+  });
+
   it("drops a preview it cannot fully validate rather than half-rendering it", () => {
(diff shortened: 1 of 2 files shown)


abaas-bw:
Browser-verified against the same live scenario that revealed the bug, with an A/B on identical server state (one `openwave serve`, two vite UIs):

- Set up a chat in Ask mode with a pending `write_file` approval carrying `path_prefix` grant rungs (verified via `GET /chats/{id}/approvals`).
- **Unfixed UI (main)**: reload renders only "Could not load this chat: Error: pending approval response contains an invalid item".
- **This branch**: same reload renders the transcript and the approval card with its grant options; no error notice anywhere in the DOM.

PR #1718, opened 2026-08-06
fix(desktop): keep tool phases together across assistant bubbles that render nothing

### LINKED ISSUE #1713: fix(desktop): invisible assistant bubbles split tool activity groups during streaming
During a streaming turn, consecutive tool calls that should present as one activity phase render as several adjacent groups with nothing visible between them. When the turn completes, terminal hydration collapses them into one group ("Ran a command and 6 other tasks"), so the transcript visibly reshuffles at the boundary.

Mechanism (observed in the DOM during a live turn): `groupMessageItems` in `ui/src/MessageList.tsx` breaks an activity phase at any non-activity message — including assistant messages that render as `null` (no text, no sources, no reasoning). The streaming reducer produces such bubbles between tool calls (empty bubbles opened by `turn_started`/resume boundaries, and reasoning-only bubbles whose accordion sits between batches), so a contiguous run of calls is silently split. The hydrated snapshot has no such bubbles, so the groups merge on swap.

Contained fix: when deciding phase boundaries, skip assistant messages that would render nothing — so streamed grouping matches what hydration will later produce. (Reasoning-only bubbles that *do* render still legitimately split a phase; the settled-state disagreement about where reasoning lives is tracked separately.)

Reproduced in the browser UI rig with adjacent `G[Updated a file and 2 other tasks] / G[Run a command and 2 other tasks] / G[Ran a command]` groups merging into `G[Ran a command and 6 other tasks]` at turn end.

abaas-bw:
## Summary

During streaming, consecutive tool calls could render as several adjacent activity groups with nothing visible between them, then merge into one group when the turn completed — the transcript visibly reshuffled at the boundary (observed live in the browser-UI rig: `Updated a file and 2 other tasks` / `Run a command and 2 other tasks` / `Ran a command` collapsing into `Ran a command and 6 other tasks`; DOM dumps in #1713).

Cause: `groupMessageItems` ends an activity phase at any non-activity message, including assistant bubbles that render `null` (no text, no sources, no reasoning) — which the live reducer produces at turn-start and interrupt/resume boundaries. The hydrated snapshot has no such entries, so the streamed splits merge on swap.

Phase collection now swallows an invisible assistant bubble when activity continues past it. A trailing empty bubble (the response about to stream) deliberately stays outside the phase so its first characters don't move a group boundary. Purely a renderer change; hydrated transcripts are unaffected (they contain no empty bubbles).

Closes #1713.

## Testing

- New regression test: `[tool, empty assistant, tool, empty assistant]` renders one group ("Read a file and 1 other task"), trailing bubble outside.
- `pnpm vitest run` on the four MessageList/transcript suites — 73 passed.

### DIFF
--- a/crates/openwave-desktop/ui/src/MessageList.test.tsx
+++ b/crates/openwave-desktop/ui/src/MessageList.test.tsx
@@ -137,16 +137,18 @@ describe("MessageBubble", () => {
   });
 
   // The live reducer opens assistant bubbles that render nothing (turn-start
-  // and resume boundaries). One between two calls split a phase the hydrated
-  // transcript later merges, so the groups visibly reshuffled at turn end
-  // (#1713). A trailing empty bubble is the response about to stream and must
-  // stay outside the phase.
-  it("keeps one phase across an assistant bubble that renders nothing", () => {
+  // and resume boundaries), and approval resume cycles can stack several in a
+  // row. Any such run between two calls split a phase the hydrated transcript
+  // later merges, so the groups visibly reshuffled at turn end (#1713). A
+  // trailing empty bubble is the response about to stream and must stay
+  // outside the phase.
+  it("keeps one phase across assistant bubbles that render nothing", () => {
     const messages: ChatMessage[] = [
       { id: "tool-1", role: "tool", callId: "call-1", name: "web_search", status: "completed" },
       { id: "assistant-1", role: "assistant", text: "", sources: [] },
-      { id: "tool-2", role: "tool", callId: "call-2", name: "read_file", status: "completed" },
       { id: "assistant-2", role: "assistant", text: "", sources: [] },
+      { id: "tool-2", role: "tool", callId: "call-2", name: "read_file", status: "completed" },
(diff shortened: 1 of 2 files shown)


abaas-bw:
Browser-verified with the same Ask-mode flow that revealed the bug (multi-file-write turns, approving each call, sampling the transcript DOM during streaming):

- Pre-fix signature (captured this morning on main): multiple *labeled* rail groups adjacent with nothing visible between them, merging into one group at turn end.
- On this branch, across four separate multi-write turns: the calls accumulate in **one** labeled group through every approval resume boundary ("Updated a file" → "and 1 other task" → "and 2 other tasks"), and the settled/reloaded state keeps the same single group — no reshuffle at completion.
- One false alarm worth recording: a DOM sampler that counts top-level non-message elements will see the `AssistantWorkingIndicator` beside the group while the turn is busy and flag "adjacent groups". React-fiber inspection confirmed the message array groups as one phase; the indicator is the trailing working status, correct behavior.

Second commit generalizes the swallow from a single invisible bubble to a run of them — approval resume cycles can plausibly stack more than one; covered by the updated regression test.

PR #1739, opened 2026-08-06
feat(server): wire the Docker execution provider into selection

naingthet:
Stacked on #1737 (base is that branch; retarget to `main` once it lands).

## What

Makes the container backend selectable: it joins `EXECUTION_PROVIDERS`, gets an arm in `resolve()`, and reports readiness through the same single availability decision every other provider uses. The desktop picker, the per-provider availability list, and the status headline are all driven by that list, so they pick it up without further wiring.

Readiness comes from the backend's own probe rather than a credential check — the two reasons it can be unavailable are "no runtime installed" and "runtime installed, daemon not answering", and the panel's sentences say which fix applies. It needs no credential, so it does not appear among the key slots, and it is never a default selection.

## Egress — the part to read

The configured egress policy is **not** passed to this adapter. Nothing in it would enforce one, and accepting the policy would imply otherwise. The enforcement disclosure gets a new status, `not_enforced`, rather than reusing `unconfirmed`: that one means a policy *was* sent and its effect is unproven, which is a different claim.

Concretely: a conversation whose network is set to off or package-managers-only gets no enforcement when it runs on this backend. The settings row states that in the strongest tone the list has (critical badge, "This backend applies no network restriction, so a conversation's network setting does not reach it"), and the docs carry the same statement. The chat default is `Open`, so this is not a silent downgrade of the common case, but it is a real gap for anyone who narrowed it.

The obvious next slice — `--network none` for an `Off` policy, and a proxy or allowlist beyond that — is deliberately not in this change, and the pooled session already fingerprints its egress shape, so adding it later replaces stale containers rather than reusing them.

## Detached admission

The container exec backend is not the sandbox-agent container tier; running `docker exec` establishes none of the preconditions a detached background run needs. Its admission row therefore lists every precondition as missing, exactly like E2B's and Daytona's. The two tiers sharing a runtime (and an image) is precisely why the docs now spell out that they are unrelated.

## Tests

One assertion added to the existing enforcement-honesty test: the Docker row must be `not_enforced` with its gap stated and no requirement. That is the guard against the tempting future edit — passing the configured egress into the adapter and letting the row read as enforced without the container enforcing anything. One existing wire-shape assertion updated for the fourth provider row.

Verified locally: `cargo check --workspace --locked` (no lockfile change), the full `openwave-server` lib suite, clippy on `openwave-code-execution` and `openwave-server` with `-D warnings`, and `tsc` plus `vitest` (813 tests) in the desktop UI.

### DIFF
--- a/crates/openwave-desktop/ui/src/generated/wire.ts
+++ b/crates/openwave-desktop/ui/src/generated/wire.ts
@@ -926,7 +926,7 @@ export type EgressConfig = { "mode": "open" } | { "mode": "allowlist", domains: 
  * says a vendor's mechanism leaves a general-purpose destination reachable,
  * the surface must not present it as a full boundary.
  */
-export type EgressEnforcementStatus = "boundary" | "conditional_boundary" | "applied_with_gaps" | "unconfirmed";
+export type EgressEnforcementStatus = "boundary" | "conditional_boundary" | "applied_with_gaps" | "unconfirmed" | "not_enforced";
 
 /**
  * The execution backend that ran a command, as a closed vocabulary.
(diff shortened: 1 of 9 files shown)


naingthet:
Deferred follow-up for the egress gap: #1740.

PR #1790, opened 2026-08-07
feat(gateway): route models by synced protocol

### LINKED ISSUE #1651: Surface a gateway's OpenAI-protocol models
`sync_models` deliberately syncs only the `anthropic_messages` subset of a gateway's entitled models, because gateway inference runs over the Anthropic-compatible surface. Models a deployment serves only over an OpenAI-compatible protocol never reach the picker.

Surfacing them means routing gateway inference per model protocol: sync the protocol field with each entitled model and have the gateway route pick the matching adapter/surface. Router + adapter work; separate decision from the capability-inheritance slice (#1648).

naingthet:
## Summary

- sync the complete gateway entitlement list and persist each model’s inference protocol
- keep the public `model_gateway::<id>` namespace while selecting an Anthropic Messages or OpenAI Chat Completions adapter per model
- route OpenAI-protocol models through `/compat/openai/v1/chat/completions` with the same rotating `llm` bearer and conversation attribution used by the Anthropic gateway route
- preserve pre-protocol snapshots as Anthropic-only and keep unsupported future protocols out of the routable snapshot

## Why

Gateway model sync previously requested only the `anthropic_messages` subset, so models available exclusively through the gateway’s OpenAI-compatible surface never reached the picker. A single concrete gateway adapter also could not safely serve models with different wire protocols.

## Verification

- `cargo fmt --all --check`
- `cargo clippy -p openwave-router -p openwave-connectors -p openwave-server --all-targets --locked -- -D warnings`
- `cargo test -p openwave-router --locked` (119 unit tests + 4 replay contracts)
- `cargo test -p openwave-connectors --locked` (20 tests)
- `cargo test -p openwave-server gateway_runtime::tests --locked` (21 tests)
- `cargo test -p openwave-server a_managed_profile_offers_only_the_gateway_route --locked`
- `cargo test -p openwave-server a_pre_protocol_gateway_snapshot_keeps_its_anthropic_route --locked`

No tests were removed. The running desktop app was not driven against a live gateway; request routing, credentials, headers, sync, and compatibility behavior are covered by in-process boundary tests.

Closes #1651

### DIFF
--- a/crates/openwave-connectors/src/gateway.rs
+++ b/crates/openwave-connectors/src/gateway.rs
@@ -149,6 +149,14 @@ pub struct GatewayIdentity {
 #[derive(Debug, Clone, PartialEq, Eq, Deserialize)]
 pub struct GatewayModel {
     pub id: String,
+    /// The gateway inference protocol this model is served through.
+    ///
+    /// Older gateways exposed only Anthropic Messages and omitted the field,
+    /// so absence preserves that route. Newer deployments report their exact
+    /// protocol (`anthropic_messages` or `openai_chat_completions`) so clients
+    /// can select the matching compatibility surface.
+    #[serde(default = "default_gateway_model_protocol")]
+    pub protocol: String,
     /// The provider-side id this deployment routes to, when it differs from
     /// `id` — a deployment alias such as `us.anthropic.claude-opus-5` behind
     /// the gateway id `anthropic-us-claude-opus-5`. Optional because gateways
@@ -160,6 +168,10 @@ pub struct GatewayModel {
     pub max_output_tokens: Option<i64>,
     pub supports_tools: bool,
     pub supports_vision: bool,
+}
+
+fn default_gateway_model_protocol() -> String {
+    "anthropic_messages".to_string()
 }
 
 /// One entitled connected app from `/api/v1/cli/apps`: the apps the user
@@ -454,7 +466,8 @@ impl GatewayAuth {
     }
 
     /// The models the authenticated user may invoke, optionally filtered to
-    /// one inference protocol (`anthropic` / `openai`).
+    /// one inference protocol (`anthropic_messages` /
+    /// `openai_chat_completions`).
     pub async fn models(
         &self,
         access_token: &str,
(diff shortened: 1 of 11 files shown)


naingthet (commented):
Review outcome: 1 actionable finding.

- **Medium — opt the gateway OpenAI route into streamed usage** (`crates/openwave-router/src/router.rs:374-376`, `crates/openwave-router/src/openai_compat.rs:121-126`). `ModelGatewayOpenai` is constructed with `OpenAiCompatProvider::compatible`, so its `provider_id` remains `openai_compatible`; `stream()` only adds `stream_options: {"include_usage": true}` for the `openai` id. Under the Chat Completions streaming contract, the `usage` field is only present when that option is requested, so standards-compliant gateway responses will complete normally while every OpenAI-protocol gateway turn records zero input/output tokens in OpenWave. Add an explicit gateway usage opt-in (without enabling it for arbitrary compatible endpoints) and extend the captured-request test to assert the request body.

Validated exact head `a7938c863d5d0b9f19ea7870957bd2da57f3e7fc` against current `origin/main`; focused routing, dynamic-token/header, sync/persistence, and legacy-snapshot tests pass. I found no other actionable problems in protocol selection, token refresh/attribution, persisted backward compatibility, cross-provider flattening, or error sanitization.

naingthet:
Resolved review 4883856652 in `33c604e3`: `ModelGatewayOpenai` now explicitly opts into Chat Completions streamed usage, while arbitrary `openai_compatible` routes continue to omit `stream_options`. The gateway captured-request test now runs through the composite router and asserts `stream_options.include_usage == true` for the gateway request and absence for the direct compatible request. Validated with router tests (119 unit + 4 replay contracts), 21 gateway runtime tests, focused request capture, `cargo fmt --all --check`, `cargo check -p openwave-router --locked`, and combined router/connectors/server clippy with warnings denied.

naingthet (commented):
CLEAN — no actionable findings.

Round-2 independent review covered exact PR head 33c604e35e4eb2fb97954d52aff40960e5c2c2b9 against current origin/main 9163076feff7dd033bc532cb397730cc4e59b8c1. The follow-up correctly makes streamed usage an explicit adapter capability: ModelGatewayOpenai opts into stream_options.include_usage, arbitrary openai_compatible endpoints remain opted out, and the captured-request regression now exercises the composite router while asserting the gateway path, rotating bearer, conversation header, and request body.

I re-inspected the full PR for per-model protocol sync and routing, public model_gateway namespace selection, token/conversation attribution, snapshot backward compatibility, flatten-on-switch behavior, error sanitization, and test value. Verification passed: clean synthetic merge tree; cargo fmt --all --check; all openwave-router tests (119 unit + 4 replay contracts); all openwave-connectors tests; all 21 gateway_runtime tests; and the pre-protocol snapshot compatibility test.

PR #1829, opened 2026-08-08
fix(server): answer a sandbox call made after its budget is spent

naingthet:
Three background agent runs in a local build failed with `sandbox_execution_failed` / "sandbox tool checkpoint has no budget to park into", each exhausting all five attempts. Their journals are identical in shape: exactly 16 work rows (`MAX_SANDBOX_TOOL_CALLS`) spent producing a DOCX, a PDF, and a PPTX, then one more work-tool call after that tool had been withdrawn from the request.

## What went wrong

`park_sandbox_tool_calls` assumed withdrawal was a guarantee — "a step whose very first call has nothing left to spend has ignored a request that offered it no such tool at all" — and failed the run. It isn't a guarantee. Nothing in the request says a tool used to be there, and a model that has called `exec` on every step of a long transcript reads its sudden absence as an oversight rather than a rule.

The failure was also maximally expensive. The transcript is rebuilt from durable rows, so every retry replayed the same history into the same wall; all five attempts burned deterministically, and the files the run had already written under `output/` went with it.

## The fix

Two changes, one preventive and one structural:

- **Say it in words.** A request assembled while a budget is spent now carries a notice saying so and naming `done` as the move left. It is derived from the run's own row counts, so a replayed claim rebuilds an identical request.
- **Keep a row back.** `SANDBOX_TOOL_CALL_REFUSAL_RESERVE` gives each budget one durable row past what it may spend on executed calls. It is never advertised and never executes; the only thing written into it is a `tool_budget_exhausted` rejection, which the next step reads as an ordinary tool error. The store's own bound moves with it so the two cannot disagree.

The budget is unchanged as a bound — the over-budget call does not run, and no work is added. What changes is that the run survives to submit what it produced. Failing is still the outcome when the reserve is spent too, which now means the model ignored a refusal it could read.

## Tests

- New `a_call_made_after_the_budget_is_spent_is_refused_rather_than_failing_the_run` drives the production shape: budget spent, work tools gone from the request, model calls one anyway. Asserts the run is not failed, the receipt carries `tool_budget_exhausted`, the exhausted request states the budget is gone, and the next step completes the run.
- `a_sandbox_run_may_checkpoint_repeatedly_up_to_a_bounded_chain` now drives the bound including the reserve; the park past it is still refused.

Full `openwave-core` and `openwave-server` suites pass locally, plus clippy and rustfmt. No lockfile movement.

I could not reproduce this against a live provider — the three journals in the local database are the evidence, and the test encodes their shape.


naingthet:
Labelled `full-ci`: this changes the durable budget check in `park_agent_run_for_sandbox_tool_calls`, and the PostgreSQL turn-state lane is the only coverage of that path I cannot run locally.

Issue #1073, opened 2026-07-30
epic: agents edit your files safely — staged writes, digest preconditions, and undo


naingthet:
Parent for the write-safety work. Related: #408 (Cowork-style product gaps), #939 (one vocabulary for the broker boundary and the ask decision).

## Problem

Attaching a folder currently gives exec unrestricted recursive write over it. `prepare_registration` mints `WriteFiles` unconditionally from the folder picker (`crates/openwave-host-broker/src/broker.rs:705-711`), and that grant lands on `file-write*` in the Seatbelt profile (`crates/openwave-code-execution/src/local.rs:970-974, 997`).

Letting the agent edit real files is the intended product direction — that is what a desktop coworker is for. What is missing is everything that makes it safe to do so: no snapshot, no undo, no conflict detection, no record of what changed. A bad overwrite in `~/Documents` is simply gone.

Meanwhile the broker protocol has a *narrow* write path — `create_new`, digest verified, approval per replacement (`broker.rs:1665-1735`) — and the deliverable write-back demands revision + sha256 (`crates/openwave-desktop/src/client_execution/output_writeback.rs:370-372`). So the same `WriteFiles` capability means two very different things depending on which door you come through, and exec comes through the unguarded one.

## Design

Exec writes into granted roots land in a per-turn overlay rather than directly on the user files. Reads are read-through (overlay first, then the real root) so the agent iterates normally against what it just wrote. At turn end the staged set is materialized: each overwrite checks the digest the agent read (stale means reject, never clobber) and snapshots the prior bytes into the blob store with a journaled undo record. The user gets a change summary with one-click undo.

Scratch and workspace writes stay direct and unstaged. This applies only to granted roots.

**Writes are not gated by approval in the default path.** Reversibility and visibility carry the safety instead. A prompt that appears every turn gets clicked through and protects nobody, and it would make OpenWave more cumbersome than the products we are matching. The approval card survives only for writes landing outside any granted root, and for users who opt into Ask mode.

## Prior art

A real Cowork file-editing session confirms the shape and shows the failure mode to beat. Its agent works in a read-only copy-in directory plus a writable staging directory, with no route to the device from the shell — the only write path is a structured commit carrying `devicePath`, `expectedMtimeMs`, and a staged file id, returning per-file `written` / `rejected` arrays.

Two things to take from it:

[...]

PR #1848, opened 2026-08-10
feat(desktop): edit text outputs as append-only user revisions

### LINKED ISSUE #1741: feat(desktop): edit text outputs as append-only user revisions
## Problem

OpenWave outputs are durable and versioned, but a user cannot make a small correction to a generated text output in the desktop. The available loop is to ask the agent to regenerate it or export it and leave OpenWave's revision history behind.

For launch, a report should be finishable in place without weakening the append-only output contract.

## Scope

Add direct editing for the current revision of Markdown and plain-text outputs. Saving must publish a new, user-authored output revision; it must never mutate the stored bytes of an existing revision.

## Acceptance criteria

- The output detail surface offers **Edit** for supported UTF-8 Markdown and plain-text outputs.
- The editor starts from an identified current revision and warns before discarding unsaved changes.
- Save carries an expected-current-revision precondition. If another revision became current, the save is rejected with a refresh/reconcile path rather than silently overwriting it.
- A successful save appends a new revision under the same output identity and marks its producer as the user, distinct from agent/turn attribution.
- Version history, restore, reopen, export, and existing revision caps continue to work for user-authored revisions.
- Content and size validation is explicit and consistent with the output store's existing bounds.
- A boundary-level test covers conflict rejection and proves the previous revision bytes remain immutable.

## Non-goals

- Rich-text/WYSIWYG editing.
- Editing spreadsheets, Office documents, PDFs, images, or other binary outputs.
- Editing a file after it has been exported into a connected folder.
- Real-time or multi-user collaborative editing.

## Related work

- #1073 established append-only/reversible file-write safety.
- #1520 documents the existing output publication and revision-attribution boundaries.
- `docs/deliverables.md` defines the private output and explicit export contract this should preserve.

naingthet:
Closes #1741

A generated report could not be finished in place: the only ways to fix a typo were to ask the agent to regenerate the file or to export it and leave OpenWave's version history behind. Markdown and plain-text outputs now have an **Edit** affordance in the output detail panel, and saving publishes a new version of the same output.

## Append-only, like everything else about an output

Save never touches the bytes of an existing revision. It publishes the edited text at the new revision's own write-once path and appends it as the head, so the version the editor was opened on stays readable at its own id, restore can walk back to it, and the 100-revision cap applies unchanged. This is the same shape restore already had — restoring v2 while at v5 produces v6 — and it reuses the same publication helper.

## The precondition

The renderer sends the revision it was editing, and the store appends only while that revision is still current. The check lives in the store transaction that holds the conversation's write lock (`append_output_revision_from`), not in the caller, because a background run or a foreground turn can publish between reading the head and appending to it. When the head has moved, nothing is written and the caller gets `AgentError::OutputRevisionConflict` naming the revision that won.

The desktop projects that as a value rather than an error — a conflict is an expected product state with a reconcile path, not a failure — so the panel shows a banner offering the latest version, with the reader's own text still on screen to copy from. This generalizes the revision+digest precondition the connected-folder write-back already binds, per the direction in #1520 and #1073, rather than adding a second concurrency mechanism.

Saving content the head already carries is a no-op, checked before the precondition so an ambiguous save can be retried once its first attempt committed. Revision identity derives from the (base revision, content digest) pair for the same reason.

## Attribution

`RevisionProducer` gains a `User` variant. It records exactly what a user-authored revision has always recorded — neither a turn nor a run — so no stored row changes and the version-history projection already maps both-absent to `user`. Naming it makes the third case explicit at the call site instead of leaving it as the meaning of two `None`s; restore now says so too.

An edit leaves the same host note a restore does, so the model's next turn re-reads the file instead of publishing over text the user just rewrote.

## Bounds and scope

Content is validated against the text deliverable's existing rules — non-empty, UTF-8, no NUL, at most 512 KiB — shared with the `output/` scan. Only `text/markdown` and `text/plain` are editable: a plain text box is a faithful editor for prose and a hazard for CSV, JSON, charts, and HTML. Rich text is a non-goal; the editor is a textarea. The panel also hides Edit when the preview is truncated, since editing a partial preview would silently drop the rest of the file, and it asks before discarding an unsaved draft.

## Tests

One boundary test (`a_stale_edit_is_refused_and_leaves_every_earlier_revision_intact`): an agent publishes v3 while the editor sits on v2, the save is refused with the conflicting revision named, and every revision's stored bytes are compared before and after to prove nothing was rewritten — then the reconciled save appends v4 with the earlier bytes still intact, and its retry appends nothing.

One projection test pins the save result's wire shape, which the renderer branches on.

No `full-ci` label: nothing here touches a ts-rs wire type, and `ui/src/generated/wire.ts` is unchanged — the outputs IPC types are hand-written on both sides.

## Not verified

I did not drive the running app; the editor's behavior is covered by types, the existing panel tests, and the native tests above rather than by a manual pass.


naingthet:
The desktop UI lane fails on merged-main code, not on this branch: `main` at 81c7dddf does not typecheck because #1834 added `ModelInfo.recommended` and `Settings.model_visibility_overrides` without updating the fixtures that build those types. Reproduced from a clean checkout of main; filed as #1852. Every other lane here is green, and `pnpm build` passes on this branch against its own base.

PR #1853, opened 2026-08-10
feat(desktop): collapsible provider cards with per-model visibility

naingthet:
Settings → Providers gets the collapsible cards and per-model checklist from
decision record [0003 — Curated Model Visibility](docs/decisions/0003-curated-model-visibility.md),
built on the wire groundwork from #1834.

## What this does

- **Collapsible cards.** Every provider card collapses to one header row:
  chevron, provider mark, a summary, and a status pill. Expanding reveals the
  credential and configuration fields that were previously always on screen.
  Built bespoke rather than adding an accordion dependency; the chevron's
  rotation is dropped under `prefers-reduced-motion`. Expansion state persists
  per provider in `localStorage`, best-effort, following the sidebar's collapse
  preference. Cards open collapsed.
- **Header summary and status.** A connected provider reads "x of y models
  shown" (y = its catalog rows, x = the effectively visible ones); one without a
  credential reads "y models" and carries a **Set up** button in the collapsed
  header that opens the card and focuses its credential field.
- **Per-model checklist** under "Models in the picker": one row per catalog
  model — a checkbox and its display name — checked when effectively visible,
  plus a "Reset to defaults" link. The curated flag still decides what is
  checked by default; the panel does not editorialize about it. A footer note says hidden
  models stay usable in chats that already selected them.
- **Deep links.** `/settings/providers?provider=<kind>` opens that card and
  scrolls it into view; `&focus=credential` also focuses the credential input.
  Validated at the route, so an unknown value reaches the panel as nothing.
- **Effective visibility** lives in `src/modelVisibility.ts` — `isModelVisible`
  is the recommended flag flipped by an override — so the picker (separate PR)
  and this panel agree by construction.

## Write path

`model_visibility_overrides` is deviations-only and `PUT /settings` replaces it
wholesale, so every write sends the complete desired map: the panel starts from
the map it last read, adds or removes exactly the toggled model's entry, and
removes rather than rewrites the entry when a toggle returns a model to its
catalog default. Reset filters out exactly the keys of that provider's catalog
models and PUTs the remainder. Checkbox state is reconciled from the PUT
response; a failure toasts through `friendlyErrorMessage` and leaves the last
persisted state on screen.

Providers with no static catalog rows — the gateway, an unconfigured
OpenAI-compatible endpoint — keep their existing card contents and get neither
the checklist nor the "x of y" summary.

## Managed profiles

The managed branch is untouched: it still renders the static "configured by your
organization" message instead of editors. The checklist is not reachable there,
and deliberately so rather than awkwardly — a managed profile's models all come
from the gateway, whose card has no catalog rows to list, and the Providers
section is dropped from a managed rail entirely.

## Tests

- `ProvidersPanel.dom.test.tsx`: three added — a toggle writes the complete
  deviations-only map (another provider's override survives, and a model
  returned to its default loses its key rather than gaining a redundant one);
  reset removes exactly the card's own provider keys; a deep link expands the
  card it names and no other.
- `ProvidersPanel.dom.test.tsx`: existing tests now render through a
  `renderPanel` helper that opens the cards first, because a collapsed card
  shows no fields — no assertions changed.
- `ProvidersPanel.dom.test.tsx`: the reset test now matches the link by its
  "Reset to defaults" label, following the copy change away from "recommended"
  in user-facing text.
- `sections.test.ts`: one added for the providers search validator, covering
  that a non-string `provider` or an unknown `focus` reaches the panel as
  nothing.

## Unverified / incidental

- Rebased onto `main` after #1856 landed the same UI test-fixture fixes; this
  branch no longer carries them. The base is still red on `tsc` for a different
  reason — two parallel PRs each added `recommended` to the same fixture
  objects, so four test files now have duplicate keys (TS1117). #1862 fixes
  that; nothing here touches those files.
- Verified against a locally deduplicated base: `tsc --noEmit` clean, the full
  `pnpm test` suite (880 passing, 128 files), and `pnpm build`. Every remaining
  `tsc` error on this branch is one of the inherited TS1117 duplicates.
- The running app was not driven, so scroll-into-view and focus behavior on a
  real deep link are unconfirmed beyond the DOM tests.


naingthet:
Heads-up: main's desktop UI lane has been red since the last two fixture-extending merges crossed (duplicate object keys, TS1117). #1862 dedupes just those keys to unblock the lane — it overlaps your files but should rebase trivially, and none of your scope is touched. Your PR shows as conflicting against current main, so it needs a rebase before its auto-merge can fire either way.

Issue #1868, opened 2026-08-10
server+cli: attachments and deliverable/output export as shared server routes


naingthet:
Implements part 5 of decision record 5 (docs/decisions/0005, PR #1864 — don't start before it merges).

File attachment and deliverable/output handling are Tauri commands today (`crates/openwave-desktop/src/attachments.rs`, `image_attachments.rs`, `deliverables.rs` — the latter reads `openwave-core::Store` directly), so neither is reachable headlessly even though storage is core/server-side.

Scope:
- Attachment: CLI takes file-path arguments and feeds the same document-ingest routes the desktop UI uses.
- Deliverables/outputs: listing, reading, revision history, restore, and export-to-path become server routes plus CLI subcommands. The desktop's Tauri handlers are then rewritten as callers of the same new routes (native save dialogs stay desktop-side; the write-to-path logic moves) — this is real desktop migration work, not just CLI addition.

Validation: one end-to-end — a headless turn produces an output, the CLI lists its revisions and exports it to a path, and the desktop sees the same history against the same data dir.

PR #1871, opened 2026-08-10
test(desktop): enforce the native-only tauri command allowlist

### LINKED ISSUE #1870: ci: parity gate — tauri command allowlist check
Implements the enforcement mechanism from part 1 of decision record 5 (docs/decisions/0005, PR #1864 — don't start before it merges).

The record closes the set of legitimate native-only Tauri commands (OS pickers, keychain/code-signing plumbing, auto-updater, window/attention, office-converter and Node installers, local voice, debug-bundle save dialogs). Everything else must land as server routes. Without enforcement this drifts silently — the ~34 existing `#[tauri::command]` handlers in `crates/openwave-desktop/src/*.rs` accreted exactly that way.

Scope: a check (test or CI step in the Rust lane) that the set of `#[tauri::command]` functions is a subset of an explicit, in-repo allowlist file, so a new desktop-only feature fails CI until the list is consciously amended or the feature gets routes. Seed the allowlist from the current inventory; each entry gets a one-line justification. Keep it cheap — a grep-shaped unit test is fine; don't build a proc-macro registry for this.

Validation: adding an unlisted `#[tauri::command]` fails; the failure message names the allowlist and the record.

naingthet:
Part 1 of decision record 5's enforcement: a check that the set of
`#[tauri::command]` handlers stays a subset of an explicit, in-repo
allowlist, so a new desktop-only feature fails the Rust lane until it is
either given server routes or the list is consciously amended.

**The gate.** `crates/openwave-desktop/native-only-commands.txt` lists every
command as `name: justification`, grouped by which native capability it
needs, with a header explaining the rule. `command_parity.rs` (a
`#[cfg(test)]` module in the desktop crate) walks `src/` textually for
`#[tauri::command]`, resolves the following function name, and fails in both
directions: an unlisted command, and an allowlist entry whose command is
gone. Both failure messages name the allowlist file and the record, so the
fix path needs no searching. The scan tolerates `fn`/`pub fn`/`pub(crate)
fn`, `async`, and doc comments or attributes between the attribute and the
signature. No proc macro, no registry, no CI workflow change — it runs in the
existing desktop test lane.

Verified by adding a throwaway command in an unrelated module: the test fails
and names it with its file.

**Observed debt.** Seeding the list surfaced commands that do not fit the
record's closed native-only set. They are allowlisted with a `TODO(record 5)`
justification rather than refactored here — migrating them is the record's
item 4/5 work, not this change:

- Output/deliverable handling, all `Store`-direct (record item 5 moves these
  to routes): `list_deliverables`, `read_deliverable`, `read_deliverable_file`,
  `list_output_revisions`, `read_output_revision`, `restore_output_revision`,
  `save_output_revision`, `delete_output`, `restore_output`.
- Host-broker reads and mutations with no native dialog (record item 4 gives
  folders a headless consent path): `list_connected_folders`,
  `list_approved_folders`, `list_capability_consents`,
  `revoke_capability_consent`, `disconnect_folder`, `forget_folder`,
  `purge_deleted_conversation_subject`.
- `publish_chat_image`, which publishes pasted image bytes from the renderer
  with no native shell involvement.

Everything else is justified against the closed set: OS pickers and save
dialogs, folder consent dialogs, host-side execution of client-executed tool
calls, the office-converter and Node installers, the auto-updater, native
window attention, and the `server_info` bootstrap that hands the renderer the
address and token every HTTP route is reached through.

Closes #1870

### DIFF
--- a/crates/openwave-desktop/native-only-commands.txt
+++ b/crates/openwave-desktop/native-only-commands.txt
@@ -1,0 +1,77 @@
+# Native-only Tauri commands.
+#
+# Decision record 5 (docs/decisions/0005-cli-headless-feature-parity.md) makes
+# the server HTTP/WS API the canonical product surface: a feature has
+# CLI/headless parity when it is reachable through `openwave-server` routes.
+# A `#[tauri::command]` is reserved for what genuinely needs the native shell —
+# OS file/folder pickers, keychain and code-signing plumbing, the auto-updater,
+# native window/attention control, the office-converter and Node runtime
+# installers, local voice capture, and debug-bundle save dialogs. Anything else
+# arriving as a Tauri command is a parity bug.
+#
+# Every `#[tauri::command]` in `crates/openwave-desktop/src/` must appear here
+# with a one-line justification, and every line here must still name a command
+# in the source. `command_parity::allowlist_covers_every_tauri_command` enforces
+# both directions.
+#
+# Adding a line is a conscious amendment, not a formality: if the command is not
+# native-only, it belongs behind a server route instead. Lines marked
+# `TODO(record 5)` are commands that already existed when the gate landed and do
+# not fit the closed native-only set — they are recorded debt to migrate, not
+# precedent for new ones.
+#
+# Format: one `name: justification` per line; `#` starts a comment.
+
+# Shell bootstrap and native window control
+server_info: hands the renderer the embedded server's address and per-launch token — the handshake that makes every HTTP route reachable, so it cannot itself be one.
+request_user_attention: native window attention hint when a parked question needs the user.
+
+# OS pickers, drops, and save dialogs
+attach_chat_files: opens the OS file picker; only the native shell can turn a user's file choice into readable paths.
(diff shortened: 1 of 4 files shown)


naingthet:
Reviewed the full diff. The gate does what record 5 asks: both directions enforced (unlisted command, stale entry), duplicate detection, and the failure text points at the allowlist and the record so the fix path is obvious. The debt section is honest — 17 entries, none refactored here, which is the right scope split.

Two notes for the stack:

- #1868 (in flight) migrates the nine store-direct deliverable/output commands this file records as debt. Whichever of these two PRs lands second must drop the corresponding TODO lines, or the stale-entry check fails at integration. Flagging now so it isn't a surprise.
- On the judgment calls: keeping `resolve_folder_access_request` / `resolve_output_writeback_request` native-only is right — they are the host-broker boundary itself, not a missing route. `copy_chat_debug_bundle` riding the clipboard user gesture is also fine as listed.

PR #1873, opened 2026-08-10
feat(cli): unattended print-mode protocol with permission modes and NDJSON driving

### LINKED ISSUE #1865: cli: print-mode unattended protocol (permission mode flag, NDJSON driving, typed exits)
Implements part 2 of decision record 5 (docs/decisions/0005, PR #1864 — the record's merge accepts the design; don't start before it lands).

This is the load-bearing slice for the goal of coding agents driving OpenWave headlessly to test behavior. Today `openwave -p` auto-rejects every approval (`crates/openwave-cli/src/print.rs`, `REJECTION_REASON`) and **cancels the turn** on `PlanProposed` / `UserQuestionsAsked` (`print.rs:161-169`), so any scenario touching plans, questions, or ask-mode approvals cannot run unattended.

Scope:
- `--permission-mode ask|auto|allow|plan` on `-p`, mapping to the existing per-chat permission modes (`PATCH /chats/{id}`) — no new permission vocabulary.
- NDJSON driving: with `--output-format json`, emit approval requests, plan proposals, and user questions as events; accept decision lines on stdin (answer via the existing `POST .../approvals/{call_id}`, `.../plans/pending/decision`, `.../questions/pending/answer` routes the TUI already calls in `crates/openwave-cli/src/api/client.rs`).
- Undriven defaults, stated and typed: in ask mode with no driver, approvals stay rejected; plans/questions terminate the turn with a distinct exit code and machine-readable reason — never a silent cancel.
- The NDJSON protocol is a wire contract; document it in `docs-site/content/docs/headless.mdx`, including the dev-harness recipe (isolated `OPENWAVE_DATA_DIR`, `OPENWAVE_KEYCHAIN_MOCK` for debug builds, allow mode).

Validation per the record: a driven session answers a plan proposal over stdin and the turn continues — assert the decision came from the driver's stdin line, not from policy; an undriven ask-mode question run exits with the documented code and reason payload.

naingthet:
Part 2 of decision record 5: `openwave -p` becomes a protocol an agent or script can drive, instead of a demo that dies at the first interaction point. Stacked on #1864 (the record) — base is `adr/cli-headless-parity`.

Before this, print mode auto-rejected every approval and *cancelled the turn* on a plan proposal or a question, with nothing machine-readable saying so. Any scenario touching plans, questions, or ask-mode approvals was unreachable headlessly.

## Permission mode

`-p` takes `--permission-mode ask|auto|allow|plan`, applied to the chat via the existing permission-mode patch before the turn is posted (so a run that asks for `allow` and cannot get it fails rather than quietly running as `ask`). No new permission vocabulary: these are the chat's own four modes.

## Driving protocol

Driving is on when `--output-format json` is set **and** stdin is not a terminal. A terminal on stdin means a person ran the command by hand and no decision lines are coming; an input stream that ends without answering is treated the same as no driver, which is what keeps `… --output-format json < /dev/null` from blocking forever.

Requests go out on stdout, mixed into the NDJSON event stream. Every CLI-authored line carries `"openwave":"v1"` — that tag is what distinguishes a control event from a journal frame (`seq`/`event`) and is how the protocol gets versioned:

```json
{"openwave":"v1","type":"approval_request","call_id":"…","action":"exec","approval":"…","grant_rungs":[…],"preview":{…}}
{"openwave":"v1","type":"plan_proposal","call_id":"…","title":"…","plan":"…"}
{"openwave":"v1","type":"questions_asked","call_id":"…","questions":[{"id":"…","question":"…","options":[…],"question_type":"single","allow_free_form":true}]}
{"openwave":"v1","type":"error","message":"…"}
{"openwave":"v1","type":"halted","reason":"plan_undriven","exit_code":3,"call_id":"…","message":"…"}
```

Decision lines come back on stdin, one JSON object per line, and drive the same approval/plan/question routes the TUI calls:

```json
{"type":"approval","call_id":"…","decision":"approve","grant":"exact_action"}
{"type":"plan","call_id":"…","decision":"reject","feedback":"split it in two"}
{"type":"questions","call_id":"…","answers":[{"question_id":"target","selected_option_ids":["staging"],"custom_answer":"canary first"}]}
```

`call_id` is optional (only one interaction is ever outstanding) but naming a *different* call is an error, not a redirect. Unknown fields are ignored so a newer driver can send more; an unknown `type` or verdict is refused with an `error` event and the next line is read, so a malformed line never decides anything by itself.

## Undriven defaults

Stated, and typed:

- Approvals are still rejected with `non-interactive print mode`, so the model can pick another route instead of hanging.
- A plan or a question ends the run with a `halted` object naming the reason — to stdout under JSON output, to stderr under text output — and the turn is cancelled. Never a silent cancel.

## Exit codes

| Code | Meaning |
| --- | --- |
| `0` | The turn completed. |
| `1` | The turn failed, was refused, or was cancelled. |
| `2` | Usage error. |
| `3` | Parked on a plan or question with no driver (`plan_undriven` / `questions_undriven`). |
| `4` | A decision was made but could not be applied (`decision_failed`). |
| `130` | Interrupted. |

`3` and `4` are new. Both undriven reasons share `3` — it is one fact ("nobody was there to answer") and the `reason` field separates them.

## Tests

Two, both at the protocol seam:

- `a_driven_plan_proposal_is_settled_by_the_stdin_line` — the proposal is emitted with its content, a mismatched line is reported rather than applied, and the decision the run carries out is the one the *line* asked for. It rejects with feedback on purpose: a build that answered plans from policy could only accept or halt, so neither could produce that decision. The turn is not cancelled.
- `an_unanswered_question_halts_with_its_reason` — an unattached driver and one whose input ends both yield no decision, and the resulting halt carries `reason: questions_undriven` and `exit_code: 3`. The reason payload is asserted, not just the code.

Plus two small parser tests covering the contract's refusal rules (wrong kind, wrong call).

What is **not** covered here: a full process-level run that reaches a real plan proposal needs a stub model provider, and the only ones that exist live inside `openwave-server`'s private test module. Record 5's end-to-end validation (`allow`-mode turn on a fresh data dir, journal shows the completed turn) needs that harness and is not in this PR.

## Docs

`docs-site/content/docs/headless.mdx` documents the flags, both directions of the protocol, the undriven defaults, the exit-code table, and a dev-harness recipe (isolated `OPENWAVE_DATA_DIR`, `OPENWAVE_KEYCHAIN_MOCK` with its debug-only caveat, `--permission-mode allow`). `OPENWAVE_KEYCHAIN_MOCK` is now in the environment table.

Closes #1865

### DIFF
--- a/crates/openwave-cli/src/main.rs
+++ b/crates/openwave-cli/src/main.rs
@@ -23,9 +23,12 @@
 //! fresh chat, and `--chat` resumes one by id.
 //!
 //! `openwave -p "<prompt>"` runs one turn without a terminal: same engine, no
-//! interaction. stdout carries the assistant's text (or, with
+//! terminal. stdout carries the assistant's text (or, with
 //! `--output-format json`, the turn's event stream as NDJSON), and the exit
-//! status says whether the turn completed.
+//! status says whether the turn completed. `--permission-mode` sets the chat's
+//! permission mode for the run, and under `--output-format json` a driving
+//! process answers approvals, plans, and questions over stdin — see
+//! [`print::protocol`].
 
 use std::ffi::OsStr;
 use std::path::PathBuf;
@@ -42,7 +45,8 @@ use print::OutputFormat;
 
 const USAGE: &str = "usage: openwave serve\n       openwave mcp <workspace>\n       openwave \
                      rehome-secrets\n       openwave tui [--chat <id> | --new]\n       openwave -p \
-                     <prompt> [--chat <id>] [--output-format text|json]";
+                     <prompt> [--chat <id>] [--output-format text|json]\n                  \
+                     [--permission-mode ask|auto|allow|plan]";
 
 #[tokio::main]
 async fn main() {
@@ -121,6 +125,7 @@ async fn run() -> Result<i32> {
             };
             let mut chat = None;
             let mut format = OutputFormat::Text;
+            let mut permission_mode = None;
             while let Some(flag) = args.next() {
                 if flag == OsStr::new("--chat") {
                     let Some(id) = args.next() else {
@@ -138,11 +143,23 @@ async fn run() -> Result<i32> {
                         Some(value) => format = value,
                         None => usage_error("--output-format expects text or json"),
                     }
+                } else if flag == OsStr::new("--permission-mode") {
+                    let Some(value) = args.next() else {
+                        usage_error("--permission-mode requires ask, auto, allow, or plan");
+                    };
+                    // The wire tokens are the chat's own permission-mode
+                    // vocabulary; the CLI adds nothing to it.
+                    match value.to_string_lossy().as_ref() {
+                        mode @ ("ask" | "auto" | "allow" | "plan") => {
+                            permission_mode = Some(mode.to_owned());
+                        }
+                        _ => usage_error("--permission-mode expects ask, auto, allow, or plan"),
+                    }
                 } else {
                     usage_error(&format!("unknown print-mode argument {flag:?}"));
                 }
             }
-            print::run(prompt, chat, format).await
+            print::run(prompt, chat, format, permission_mode).await
         }
         Some(other) => {
             usage_error(&format!("unknown command {other:?}"));
--- a/crates/openwave-cli/src/print.rs
+++ b/crates/openwave-cli/src/print.rs
@@ -1,4 +1,4 @@
-//! `openwave -p "<prompt>"` — one non-interactive turn.
+//! `openwave -p "<prompt>"` — one unattended turn.
 //!
 //! Boots the same in-process server `openwave serve` runs and drives one turn
 //! over the loopback API, then exits with a status that says how the turn ended.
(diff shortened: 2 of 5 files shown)


naingthet:
Reviewed the full diff — this is the wire contract for record 5 part 2, so I read it as one.

What holds up well:
- The `openwave: v1` tag on every CLI-authored line is the right disambiguator against journal frames, and gives us versioning without guessing later.
- Refusing (with an `error` event) rather than approximately applying malformed/mismatched decision lines, and treating a named-but-different `call_id` as an error instead of a redirect.
- EOF/closed-stdin collapsing to the undriven policy — `< /dev/null` behaves exactly as documented, and once the channel closes the driver never re-blocks.
- `halted` carrying reason + exit code in both output formats, and the parked turn always being cancelled so it can't outlive the process.
- Approval-apply failures continue the turn while plan/answer-apply failures halt with code 4 — the asymmetry is correct and the comment explains it.

Accepted trade-offs worth stating: in undriven JSON mode the `approval_request` events are never emitted (only the stderr notice) — fine, since a consumer that wants them attaches a driver; and both undriven reasons sharing exit 3 with `reason` separating them is the right economy.

The known gap is the record's end-to-end validation (allow-mode turn, fresh data dir, journal shows completion) — blocked on a reusable stub model provider. Filing a follow-up issue so it doesn't evaporate when this closes #1865.

PR #1878, opened 2026-08-10
feat(cli): provider, model, and settings subcommands over the server routes

### LINKED ISSUE #1866: cli: setup subcommands for providers, models, settings, and MCP config
Implements part 3 of decision record 5 (docs/decisions/0005, PR #1864 — don't start before it merges).

There are no CLI subcommands for setup; the server routes exist and work (`PUT /providers/{kind}`, `/providers/{kind}/credential`, `/settings/*`, `/web-search/credentials/*`, `/code-execution/credentials/*`, `/mcp/servers` — route table in `crates/openwave-server/src/lib.rs`), but only raw curl against `openwave serve` reaches them.

Scope: `openwave provider|model|settings|mcp` subcommand families as **thin clients** of the existing HTTP routes — no logic of their own. Credential values keep flowing to the OS keychain exactly as today; headless changes the entry path, not storage. Note the CLI hand-rolls arg parsing in `crates/openwave-cli/src/main.rs` (no clap in the workspace); either extend that pattern or make a deliberate, pinned dependency decision.

Validation: one smoke test that a credential set via CLI is immediately usable by a turn. Do not re-walk each route through the CLI — that's duplicate coverage under the testing policy.

naingthet:
Third PR in the CLI/headless parity stack (record 5 → #1864, print protocol → #1873), implementing part 3 of `docs/decisions/0005-cli-headless-feature-parity.md`.

Setup was the remaining part of running OpenWave headlessly that needed raw HTTP. The routes exist and work; only curl against `openwave serve` reached them. This adds subcommand families that are thin clients of those same routes — each boots the in-process server the way print mode does, makes one call, and renders the answer.

```text
openwave provider list
openwave provider set-key <kind> [--from-env <var>]
openwave provider remove-key <kind>
openwave model list
openwave model roles
openwave model select <key|auto> [--role <role>]
openwave settings show
openwave settings web-search select <provider|off>
openwave settings web-search set-key <provider> [--from-env <var>]
openwave settings web-search remove-key <provider>
openwave settings exec select <provider|off>
openwave settings exec set-key <provider> [--from-env <var>]
openwave settings exec remove-key <provider>
openwave mcp-server list
openwave mcp-server add <name> (--command <cmd> [--arg <a>]… | --url <url> | --gateway-endpoint <slug>)
openwave mcp-server remove <name>
```

All of them take `--output-format text|json`, matching print mode's flag; `json` prints the route's response object on one line.

Routes behind each family:

| Command | Route |
| --- | --- |
| `provider list` | `GET /providers` |
| `provider set-key` | `PUT /providers/{kind}` (credential + `enabled: true`) |
| `provider remove-key` | `DELETE /providers/{kind}/credential` |
| `model list`, `model roles` | `GET /models` |
| `model select` | `PUT /models/roles/{role}` |
| `settings show` | `GET /settings`, `/web-search`, `/web-search/credentials`, `/code-execution`, `/code-execution/credentials` |
| `settings web-search select` | `PUT /web-search` |
| `settings web-search set-key`/`remove-key` | `PUT`/`DELETE /web-search/credentials/{provider}` |
| `settings exec select` | `PUT /code-execution` |
| `settings exec set-key`/`remove-key` | `PUT`/`DELETE /code-execution/credentials/{provider}` |
| `mcp-server list` | `GET /mcp/servers` |
| `mcp-server add`/`remove` | `GET` then `PUT /mcp/servers` |

## Secrets

A key is read from stdin (default) or from the environment variable named by `--from-env`. Never from an argument: argv is readable by every process on the machine, and a key pasted into a shell command also lands in its history. The choice is stated in `--help` output and in the docs page. Nothing here prints a stored key back, and storage is unchanged — headless changes the entry path, not the custody.

## MCP servers

`/mcp/servers` is mutable, but only as a complete replacement of the *user-configured* set, so `add`/`remove` read the current list, drop the live projection fields (health, tool counts) and any plugin-sourced entries the route refuses, edit, and put the whole set back. Removing a plugin-sourced server is refused with a message pointing at the plugin. Secret material is named rather than given: `--env-from` forwards a parent variable, `--bearer-token-env` names the variable holding a remote server's token, both resolved at connect time by the server.

## The keychain mock

`keyring`'s mock keeps the password on the entry object with no store behind it, so a value written through one `Entry` was invisible to the next one — a mocked process could store a credential and never read it back. That is precisely what a headless setup run does, and it made the smoke test below impossible to write. `KeychainSecretProvider` now keeps its own process-global map when mocked. Debug-only behavior, unchanged in release builds.

## Tests

One test, per the record's validation note (these are thin clients; re-walking each route through the CLI would be duplicate coverage of the server's own tests):

- `setup::tests::a_key_set_through_the_cli_shows_up_as_a_stored_credential` — stores an Anthropic key through the CLI's own command path against a real server, then asserts the provider listing reports it as credentialed and enabled, and that removing it takes it back out.
- `keychain::tests::the_mock_stores_what_it_is_given` replaces `operations_succeed_and_missing_reads_as_none`, which could only assert that a write did not error because the mock had nowhere to put it.

## Deliberately not covered

Route coverage is pragmatic, not exhaustive: no CLI surface for the ChatGPT OAuth sign-in flow (it needs a browser), voice transcription, gateway model sync, plugins, compaction tuning, model visibility overrides, egress policy, or E2B/Daytona template and snapshot fields. Provider config beyond the API key — base URLs, Vertex/Bedrock regions, custom model lists — is also left to the app for now; the credential is what blocks a headless start.

## Verified

`cargo fmt --all`, `cargo clippy -p openwave-cli --all-targets -D warnings`, `cargo test -p openwave-cli --locked` and the `openwave-core` keychain tests, all clean, no lockfile change. Also driven by hand against a throwaway profile: provider list/set-key, model list/roles, settings show, exec/web-search selection, and adding two MCP servers then removing one (the other survived the replacement).

Closes #1866

### DIFF
--- a/crates/openwave-cli/src/api/client.rs
+++ b/crates/openwave-cli/src/api/client.rs
@@ -13,7 +13,8 @@ use tokio_tungstenite::{connect_async, MaybeTlsStream, WebSocketStream};
 
 use super::wire::{
     AgentActivityItem, AgentRunSnapshot, ChatSummary, GrantRung, InboxItem, ModelCatalog,
-    PendingApprovalSnapshot, PendingPlan, PendingQuestions, Transcript,
+    PendingApprovalSnapshot, PendingPlan, PendingQuestions, ProviderInfo, ProvidersList,
+    Transcript,
 };
 
 /// The chat event stream once the upgrade completes.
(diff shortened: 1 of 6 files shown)


naingthet:
Reviewed the diff. The thin-client discipline held — setup.rs decides nothing, every command is one route call, and the secret path (stdin/--from-env, empty refused, trimmed, never argv) is exactly right and correctly explained in the usage text.

The openwave-core change is justified and well-contained: keyring's mock stores per-Entry so set-then-get across entries silently returns nothing — the process-global map behind the debug-only mock is what makes any set-then-use flow (including the required smoke test) actually exercisable. Release path untouched, and keeping the keyring default-builder redirect as a second fence against a mocked process touching the real keychain is a good instinct.

Two small things, neither blocking:
- The USAGE block omits `--gateway-endpoint` from `mcp-server add`, though the parser and docs support it — worth a one-line fix whenever this branch is next touched.
- The mcp-server add/remove read-modify-write over the full-replacement route can race a concurrent editor (desktop open at the same time). Acceptable for a setup command; noting it so it's a known trade, not an oversight.

PR #1880, opened 2026-08-10
feat(server): serve deliverables and attachments over shared routes for headless parity

### LINKED ISSUE #1868: server+cli: attachments and deliverable/output export as shared server routes

naingthet:
Part 5 of decision record 5. Conversation outputs and file attachment were the
last user-facing features that only existed as Tauri commands: `deliverables.rs`
read `openwave-core::Store` directly, so nothing outside the desktop shell could
see, read, revise, or export a file a turn produced. The storage was always
core-side — only the transport was native.

## The route surface

All on the primary bearer, alongside the rest of the chat API:

| Route | |
| --- | --- |
| `GET /chats/{chat}/outputs` | the conversation's live outputs |
| `GET /chats/{chat}/outputs/{output}` | bounded text preview of the current revision |
| `DELETE /chats/{chat}/outputs/{output}` | soft-delete |
| `POST /chats/{chat}/outputs/{output}/restore` | undo that delete |
| `GET /chats/{chat}/outputs/{output}/content[?revision_id=]` | one revision's complete bytes |
| `GET /chats/{chat}/outputs/{output}/revisions` | version history |
| `POST /chats/{chat}/outputs/{output}/revisions` | publish a user edit (`expectedRevisionId` precondition) |
| `GET /chats/{chat}/outputs/{output}/revisions/{revision}` | preview one exact version |
| `POST /chats/{chat}/outputs/{output}/revisions/{revision}/restore` | restore an earlier version |

Attachment needed no new routes: `POST /chats/{chat}/documents/raw` and
`POST /chats/{chat}/attachments/images` already exist and, in a headless
embedding, already sit on the primary bearer. What was missing was a client, so
`openwave attach <chat> <file>` reads the file, sniffs what it actually is, and
posts it to whichever of the two endpoints fits — the same split the desktop's
one attach gesture makes.

Content sniffing moved from `openwave-desktop::media_type` to
`openwave-server::media_type` so both shells decide a media type the same way;
the desktop's three call sites now go through it and `infer` moves with it (no
lockfile version changes, only dependency edges).

## What moved out of the desktop

Ten Tauri commands are gone — `list_deliverables`, `read_deliverable`,
`read_deliverable_file`, `list_output_revisions`, `read_output_revision`,
`restore_output_revision`, `save_output_revision`, `delete_output`,
`restore_output` — and the webview calls the routes instead, exactly as it does
for chats, documents, and everything else.

`export_deliverable` stays: an OS save dialog is on the record's closed
native-only list. It is now a route caller like `export_library_document`
already was — it fetches the preview for identity, fetches the bytes from the
content route, opens the dialog, revalidates that the revision is still current,
and writes. Its durable receipt still keys an ambiguous retry to one exact
terminal outcome and never repeats a dispatched write.

Reading a revision out of private scratch is now
`openwave-server::output_files`, shared by the routes and by the
connected-folder write-back path that legitimately stays in the shell.

## Deliberately unchanged

- **Payload shapes**, camelCase included. This is a transport move; renaming
  fields at the same time would hide one change inside another. The renderer's
  hand-written validators are the same ones, now narrowing HTTP responses.
  `readDeliverableFile` is the one exception: the content route serves raw bytes
  rather than a base64 field, so the validator checks a `Uint8Array` and the
  base64 decoder is gone.
- **Restore is append-only.** Restoring publishes a *new* head revision carrying
  the earlier version's content. Nothing is rewound and no revision loses its
  id or its bytes; the test asserts a superseded revision still serves its own
  content after a restore.
- **Export-to-path is client-side.** The route serves bytes; the caller decides
  where they land. That is the only shape that serves both a save dialog and a
  CLI path argument.
- **No persisted or schema change.**

## CLI

```
openwave output list <chat>
openwave output show <chat> <output> [--revision <id>]
openwave output revisions <chat> <output>
openwave output export <chat> <output> <path> [--revision <id>]
openwave attach <chat> <file>
```

Hand-rolled argument parsing in the existing style, no new dependencies.
Documented in `docs-site/content/docs/headless.mdx`.

## Tests

One end-to-end test,
`tests::outputs::a_published_output_is_listed_read_revised_restored_and_exported_over_http`:
a background run publishes a file, and the catalog, preview, edit, conflict,
history, restore, delete, undo, and exported bytes all come back over HTTP with
no shell involved. It absorbs the desktop test that used to drive the run
publisher against the scratch reader — the gap that is invisible to either half
alone — and the wire-shape assertions the old unit tests made against
hand-built structs, which are now made against the server's actual responses.

The desktop keeps the tests about the thing it still does: export receipt
recovery, atomic replace with preserved permissions, symlinked destinations, and
the renderer-safety of the export request and result. The scratch-read scoping
and symlink tests moved to `output_files` with the code.

## Verification

`cargo fmt --all`; clippy clean with `-D warnings` across the workspace and the
desktop crate; `cargo test --workspace --exclude openwave-desktop` and
`cargo test -p openwave-desktop` both green; `pnpm test` (885) and `pnpm build`
green. No lockfile version changes.

Single-package invocations (`cargo test -p openwave-server`,
`cargo check -p openwave-cli`) fail on this tree for unrelated reasons that
predate this branch — they resolve `openwave-core` with fewer features and lose
`compact_now`/`append_chat_event` — so verification used the workspace form CI
runs.

Not verified: live drag-and-drop attachment and the native save dialog, which
need the running app.

Closes #1868

---

**Rebased onto `main`** after the CLI chain (#1873, #1878, #1885) and the parity
gate (#1871) landed; the base is now `main` rather than the record branch.

Three conflicts, all resolved by keeping main's structure:

- `crates/openwave-cli/src/main.rs` — the module list, `USAGE`, and the helper
  block below `run()`. Main's versions won everywhere; `output`/`attach` were
  added to the multi-line usage block and `output_command` now sits alongside
  the setup parsers.
- `crates/openwave-cli/src/api/client.rs` — the `super::wire` import list only;
  both sides' names merged.
- `docs-site/content/docs/headless.mdx` — the command block; both sets listed.

`print.rs` was untouched by this branch, so main's rewritten driver protocol
came through unchanged.

**Attach mode**: `output` and `attach` take `--server <url>` /
`--server-token-env <var>` like every other client command. They open a
`connect::Session` and hand its client to an `execute` split, the same shape
`setup::run` uses, so the module's own `with_client` helper is gone.

**Allowlist**: the nine `TODO(record 5)` lines this PR discharges are removed
from `crates/openwave-desktop/native-only-commands.txt` — `list_deliverables`,
`read_deliverable`, `read_deliverable_file`, `list_output_revisions`,
`read_output_revision`, `restore_output_revision`, `save_output_revision`,
`delete_output`, `restore_output`. `export_deliverable` keeps its entry (native
save dialog) and `publish_chat_image` keeps its TODO, since this PR does not
touch it. `command_parity::allowlist_covers_every_tauri_command` passes.

Re-verified after the rebase: `cargo fmt --all`; clippy `-D warnings` clean on
the workspace and on the desktop crate; `cargo test --workspace --exclude
openwave-desktop --locked` 1930 passing; `cargo test -p openwave-desktop
--locked` 113 passing; `tsc --noEmit`, `pnpm test` (893), and `pnpm build`
green. Lockfile still carries only dependency edges, no version changes.


naingthet:
Reviewed the diff end to end. This is the record-5 part-5 shape done right:

- Every handler binds to the chat first (`require_chat`, plus the explicit `chat_id` filter on the soft-delete restore path where the live-only lookup can't apply). The content route serving raw bytes with `Content-Disposition: attachment` and never inviting the browser to render an agent-controlled media type is the correct paranoia.
- Append-only restore and edit semantics survived the transport move intact, and the `expectedRevisionId` precondition returning conflict-as-value (with the current revision to reload) preserves the editor contract exactly.
- Keeping export-to-path client-side while the route serves bytes is the right split for both the native dialog and a CLI writing to an argument — and thinning `export_deliverable` while keeping its receipt/retry semantics was the careful version of that.
- Deleting the nine Store-direct handlers outright (rather than proxying them) and rewriting `deliverables.ts` onto the routes is exactly the convergence the record wants. The module-level `{baseUrl, token}` is a fair mirror of the process-global invoke bridge it replaced; fine as flagged.
- Dependency moves are clean: `infer` migrated desktop→server at the same `=0.22.0` pin, `cap-fs-ext` via workspace, lockfile edges only.

Integration notes for the stack:
- #1871's allowlist records these nine commands as TODO debt; whichever of #1871/this merges second must drop those lines or the stale-entry check fires (already flagged on #1871).
- The pre-existing single-package `cargo test -p openwave-server` feature-resolution breakage you confirmed on the base branch is worth its own small issue — workspace-form verification matches CI, so it doesn't block this PR.

Unverified by design (per repo policy): live drag-and-drop and the native save dialog — hand-over noted.

naingthet:
Rebased onto current `main`, which now carries the deployment-plane authorization split (`0004-self-host-deployment-plane-authorization`) as well as #1885's `--server` flags and #1888's queued turns. The output routes are registered on the plain member-plane `api` router, unchanged in path, handler, and body limit — they serve owner-scoped chat data, so nothing about them belongs behind `require_admin`, and `deployment_api` comes out of the rebase untouched. I added two representative entries for them to `member_plane_routes()` in the conformance suite so the classification is pinned rather than implied.

Two things I decided while resolving. The CLI usage string had been rewritten on `main` into the multi-line form with the provider/model/settings families, so the `output` and `attach` lines moved into that layout rather than the older single-line string; the same applies to the command list in `docs/headless.mdx`. And `output`/`attach` do not take `--server` — they embed a server the way they did before #1885 landed, which matches the docs' statement that `tui`, `-p`, and the setup commands are the ones that attach. Teaching them to attach is a real follow-up, but it is a behavior change rather than a conflict resolution, so I left it alone here.

`cargo check`, `cargo test -p openwave-server`, and `cargo fmt --check` are green locally.

PR #1881, opened 2026-08-10
feat(cli): operator-provisioned folder consent for headless installs

### LINKED ISSUE #1867: cli: operator folder consent (connect/list/disconnect via ConsentMethod::OperatorConfig)
Implements part 4 of decision record 5 (docs/decisions/0005, PR #1864 — don't start before it merges).

Connected folders are brokered by `openwave-host-broker` behind a native picker/dialog; there is no headless consent path. The broker already models the intended mechanism: `ConsentMethod::OperatorConfig` in `crates/openwave-host-broker/src/capability.rs` ("a local operator deliberately provisioned a headless installation") — it is just not wired to anything.

Scope:
- `openwave folder connect <path>` / `list` / `disconnect`, recording grants via `OperatorConfig`.
- Explicit provisioning only: the CLI never grants a folder because an agent asked mid-turn. Mid-turn folder requests in headless mode fail with the same typed refusal an undecided desktop prompt produces.
- Grants made this way must be visible and revocable from the desktop's connected-folders UI against the same data dir.

This is a security-boundary change (a new provenance of standing grant): review with a strong model, and treat audit/revocation/display as first-class.

Validation per the record: an OperatorConfig grant is visible and revocable in the desktop UI; a mid-turn headless folder request produces the typed refusal, never a grant.

naingthet:
Implements part 4 of decision record 5: connected folders get the headless
consent path the broker already anticipated.

Stacked on #1880 (`feat/headless-outputs-attachments`); merge that first.

`openwave folder connect <path> --chat <id>` records standing consent for one
host folder; `openwave folder list [--chat <id>]` reports every grant with its
provenance; `openwave folder disconnect <path-or-root-id> --chat <id>` takes it
back. Hand-rolled argument parsing, no new third-party dependencies.

## Security invariants and how each is enforced

**The grant is the desktop's grant, not a parallel one.** `connect` calls the
broker's own `RegisterRoot` against the same `host-broker-state.json` and the
same policy the desktop's sidecar builds (host protected locations plus the
profile's private directory), with `ConsentMethod::OperatorConfig`. The broker
mints the same `ListRoots`/`ReadFiles`/`WriteFiles` grants — plus
`ExecuteCommands` when the same host probe the desktop uses says local exec is
available — so a folder connected here carries exactly the reach the picker
would have given it on that machine. The CLI cannot stamp any other consent
method: `prepare_registration` accepts only `FolderPicker` and
`OperatorConfig`, and attach accepts nothing that describes a migration.

**The CLI never grants a folder because an agent asked, and no driver can
either.** These are standalone provisioning commands; there is no flag, and no
code path, from a mid-turn request to a grant. `openwave -p` refuses a parked
`request_folder_access` with `RequestFolderAccessResult::Declined` —
byte-identical to what an undecided desktop prompt produces — claiming the call
only while it is still unclaimed and still that tool, and never reaching any
other resolution.

Critically, this sits *outside* the stdin driving protocol #1873 introduced.
Approvals, plans, and questions become an `Interaction` and go through
`settle()`, where a driver may answer them. A folder request does not: the
`ToolCallStarted` arm calls `decline_folder_request` directly, that function
takes no `Driver` argument, and `Interaction` is a closed three-variant enum
with no folder case — so no request event is emitted for a folder and
`parse_decision` rejects any line that tries to answer one. A driven session
declines exactly like an undriven one. Folder access is host-machine consent,
and `openwave folder connect` is the only thing that gives it.

**Attach mode cannot decline, and does not pretend to.** `Session` now carries
the client-executor token, and it is `Some` only when embedding. That
credential is the native-only boundary — it asserts "I am the trusted surface
for this server" — and `--server` conveys only a bearer, with no flag to hand
the executor token to an arbitrary attaching client. An attached run therefore
emits a notice naming the call and leaves it for the server's own client
executor, which on a desktop is the process that will show the user a picker;
cancelling somebody else's turn over it would be worse than saying so. The
limitation is real and stated below.

**`openwave folder` refuses `--server`.** It writes this machine's broker state
and this profile's data directory; accepting the flag would claim the grant
lands somewhere it does not. It follows `serve`/`mcp`/`rehome-secrets` with
`server_flags.refuse("folder")`.

**Product state and host authority converge or neither stands.** `connect`
commits product intent first, carries the change id as the broker mutation's
idempotency identity, and reads the durable receipt back before finishing —
the desktop reconciler's sequence. If the product attachment cannot be
committed, the host approval is withdrawn rather than left as standing reach no
surface can show; if the withdrawal also fails, the command says so and points
at `folder list`.

**Two owners of one capability store is refused, not merged.** Both the
server's `openwave.lock` and the broker's `host-broker.lock` are exclusive, so
these commands fail with a stated reason while OpenWave is running rather than
writing behind it.

**Nothing host-shaped reaches the terminal.** Broker error text is already free
of absolute paths and raw OS errors. Folder display names are host filenames,
so the listing bounds them and replaces control characters, tabs, and
bidirectional overrides — a directory cannot invent a column or a row in
tab-separated output.

**Audit is the existing one.** Opening the broker installs the same
`JsonlAuditSink` under the data directory, so registration, attach, detach, and
revocation land in the audit log the desktop's sidecar writes, with no second
audit system and no gap.

## Path handling

The path argument is canonicalized before anything is touched: a nonexistent
path is rejected with a plain message, a non-directory is rejected, and a
symlink resolves to the directory it names. That is deliberately the same
answer the broker's `RootPolicy::open_root` produces — it canonicalizes and
then pins a descriptor — so the CLI adds an earlier, clearer error rather than
a different policy. `disconnect` takes a root id exactly, or a path matched by
the leaf name the broker reports; the broker never exposes a root's absolute
path, so an ambiguous name is refused with the candidate ids rather than
guessed at.

## Cross-shell visibility

No UI change was needed. The desktop's Permissions surface already renders
`operator_config` as "by operator configuration", and the connected-folders
panel is a grouping of the same consent statements. An operator grant appears
there, and revoking it there removes the reach — verified against the store, not
in the running app.

## Scope decision worth review

`disconnect` detaches the folder from the chat and then withdraws the host
approval when that chat's subject is the only holder of root-scoped grants on
it, leaving it in place otherwise and saying so. Detach-only would leave a
headless installation able to create standing grants it could never withdraw;
unconditional revocation would take a folder away from a sibling chat in the
same project.

## Tests

- `openwave-cli/tests/folder.rs::an_operator_grant_is_one_record_both_shells_read_and_either_side_can_revoke`
  — connect through a symlink, assert the folder is named for its target, then
  assert the CLI listing and the broker query behind `list_capability_consents`
  describe the same grant with operator provenance; disconnect from the CLI and
  a fresh grant revoked from the desktop side each disappear from both.
- `openwave-cli/tests/folder.rs::folder_commands_refuse_to_pretend_they_target_another_server`
  — `folder list --server …` exits with a usage error rather than silently
  provisioning locally.
- `print::protocol::tests::no_decision_line_can_grant_a_folder` — folder-shaped
  and approval-shaped decision lines are all refused by `parse_decision`. It
  fails first, on purpose, if a folder verb is ever added to the protocol.
- `openwave-server` `a_headless_folder_request_resumes_the_turn_with_the_declined_result`
  — a real turn parks `request_folder_access`, the refusal sequence print mode
  performs is applied, and the model is handed `{"status":"declined"}` while the
  turn resumes.
- `folder::tests::mutating_commands_require_an_explicit_chat` — the mutating
  commands refuse to run without a conversation, and a flag-shaped argument is
  never swallowed as a path.

## Verified / not verified

`cargo fmt --all`, `cargo clippy -p openwave-cli -p openwave-host-broker
--all-targets -D warnings`, `cargo test -p openwave-cli -p
openwave-host-broker --locked`, the new server test, and `cargo check -p
openwave-desktop --locked` all pass on this base. The lockfile changes only by
the new workspace path dependencies and `uuid`; no version moved.

Not verified: seeing the grant in the running desktop app. The assertions are
against the same store query the folders and Permissions panels read, not the
rendered UI.

Known limitation, stated rather than papered over: an attached (`--server`) run
against a server that has no client executor of its own can wait on a folder
request, because this process holds no executor credential and declining on that
server's behalf is not its call. That is the attached server's missing executor,
not a decision the CLI can make for it.

Closes #1867

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4375,7 +4375,9 @@ dependencies = [
  "chrono",
  "crossterm",
  "futures",
+ "openwave-code-execution",
  "openwave-core",
+ "openwave-host-broker",
  "openwave-mcp",
  "openwave-server",
  "pulldown-cmark",
@@ -4389,6 +4391,7 @@ dependencies = [
  "tokio",
  "tokio-tungstenite 0.30.0",
  "unicode-width",
+ "uuid",
 ]
 
 [[package]]
--- a/crates/openwave-cli/Cargo.toml
+++ b/crates/openwave-cli/Cargo.toml
@@ -20,6 +20,12 @@ path = "src/main.rs"
 # built-in file tools for `openwave mcp`; server persistence features still come
 # through `openwave-server`. This avoids directly enabling unrelated defaults.
 openwave-core = { path = "../openwave-core", default-features = false, features = ["tools"] }
+# `openwave folder …` decides local command reach exactly as the desktop does,
+# so an operator-provisioned folder carries the same grants on the same host.
+openwave-code-execution = { path = "../openwave-code-execution" }
+# The capability store behind connected folders. `openwave folder …` opens it
+# in-process the way the desktop's sidecar opens it, against the same data dir.
+openwave-host-broker = { path = "../openwave-host-broker" }
 openwave-mcp = { path = "../openwave-mcp" }
 openwave-server = { path = "../openwave-server" }
 # Timestamps on transcript entries and inbox rows.
@@ -43,6 +49,7 @@ tokio-tungstenite = { workspace = true }
 ratatui-textarea = "=0.9.2"
 # Display-width clipping for lines rendered into fixed scrollback/panel buffers.
 unicode-width = "=0.2.0"
+uuid = { workspace = true }
 
 [dev-dependencies]
 tempfile = { workspace = true }
--- a/crates/openwave-cli/src/api/client.rs
+++ b/crates/openwave-cli/src/api/client.rs
@@ -784,6 +784,103 @@ impl Client {
     }
 }
 
+/// The second per-launch credential, presented on the client-executor routes.
+///
+/// These routes exist for the trusted surface that executes a client-owned tool
+/// call — in the desktop, the native shell that opens the folder picker. A
+/// headless run has no such surface, which is exactly why print mode needs
+/// them: it must answer a parked call rather than leave the turn hanging on a
+/// consent flow that can never appear.
+const CLIENT_EXECUTOR_HEADER: &str = "x-openwave-client-executor";
+
+/// One parked client-executed tool call, as much as the CLI acts on.
+#[derive(Debug, Clone, Deserialize)]
+pub struct PendingClientCall {
+    pub id: CallId,
+    pub name: String,
+    #[serde(default)]
+    pub client_executor_id: Option<uuid::Uuid>,
+}
+
+impl Client {
+    /// Every client-owned tool call currently parked on this chat.
+    pub async fn pending_client_executions(
+        &self,
+        executor_token: &str,
+        chat: ChatId,
+    ) -> Result<Vec<PendingClientCall>> {
+        let response = self
+            .http
+            .get(format!(
+                "{}/chats/{chat}/client-executions/pending/raw",
+                self.base
+            ))
+            .header(CLIENT_EXECUTOR_HEADER, executor_token)
+            .send()
+            .await
+            .map_err(request_error)?;
+        Self::expect_success(response)
+            .await?
+            .json::<Vec<PendingClientCall>>()
+            .await
+            .map_err(request_error)
+    }
+
+    /// Take ownership of one parked call so it can be resolved.
+    pub async fn claim_client_execution(
+        &self,
+        executor_token: &str,
+        chat: ChatId,
+        call_id: CallId,
+        executor_id: uuid::Uuid,
+        lease_token: uuid::Uuid,
+    ) -> Result<()> {
+        let response = self
+            .http
+            .post(format!(
+                "{}/chats/{chat}/client-executions/{call_id}/claim",
+                self.base
+            ))
+            .header(CLIENT_EXECUTOR_HEADER, executor_token)
+            .json(&serde_json::json!({
+                "executor_id": executor_id,
+                "lease_token": lease_token,
+            }))
+            .send()
+            .await
+            .map_err(request_error)?;
+        Self::expect_success(response).await?;
+        Ok(())
+    }
+
+    /// Report the terminal, model-facing result of one claimed call.
+    pub async fn resolve_client_execution(
+        &self,
+        executor_token: &str,
+        chat: ChatId,
+        call_id: CallId,
+        lease_token: uuid::Uuid,
+        result: &str,
+    ) -> Result<()> {
+        let response = self
+            .http
+            .post(format!(
+                "{}/chats/{chat}/client-executions/{call_id}/resolve",
+                self.base
+            ))
+            .header(CLIENT_EXECUTOR_HEADER, executor_token)
+            .json(&serde_json::json!({
+                "lease_token": lease_token,
+                "resolution": { "status": "completed", "result": result },
+            }))
+            .send()
+            .await
+            .map_err(request_error)?;
+        Self::expect_success(response).await?;
+        Ok(())
+    }
+}
+
 /// reqwest's `Display` already strips nothing for loopback URLs, but keep the
 /// mapping in one place.
 fn request_error(error: reqwest::Error) -> AgentError {
--- a/crates/openwave-cli/src/connect.rs
+++ b/crates/openwave-cli/src/connect.rs
@@ -109,6 +109,7 @@ fn base_url(value: &str) -> Result<String> {
 pub struct Session {
     client: Client,
     serve: Option<tokio::task::JoinHandle<Result<()>>>,
+    client_executor_token: Option<String>,
 }
 
 impl Session {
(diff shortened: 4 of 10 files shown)


naingthet:
Security review — read the full diff line by line, since this PR creates a new provenance of standing host reach.

**The record's invariants hold, verifiably:**
- The only grant-creating path is `connect` → `RegisterRoot` with `ConsentMethod::OperatorConfig`, a standalone operator command that requires an explicit `--chat`. There is no flag, mode, or code path by which a mid-turn request produces a grant.
- Print mode's refusal is the contract's own `Declined` variant, byte-identical to a closed desktop prompt — the model can't distinguish a headless run from a user who declined, so there is no prompt-shaped retry to fish for. The decline path only touches calls that are still named `request_folder_access` and still unclaimed; anything owned by another executor is left strictly alone.
- Failed product commit ⇒ the broker approval is withdrawn, and if even that fails the operator is pointed at `folder list` — no silent standing reach without a surface that shows it.
- Reading the durable receipt back instead of trusting the control response, and carrying the change id as idempotency identity, matches the desktop reconciler exactly. The interrupted-run story (stable 0600 executor identity, settle-pending-before-new-work, create_new race handling) is sound.
- `safe_label` neutralizing control chars and bidi overrides in the tab-separated listing is a detail I was going to ask for; glad it's already here.

**Design calls, endorsed:**
- `disconnect` withdrawing the host approval only when this subject is the sole root-scoped grant holder is the right conservative default, and saying plainly when the approval is left standing avoids the worse failure (invisible reach).
- Refusing to run while OpenWave holds the locks is correct — sequential cross-shell visibility is the supported arrangement, and the tests prove the shared-record contract that matters.

**Stack integration (important):** this PR edits the pre-#1873 `print.rs`; PR #1873 rewrote that file around the driving protocol. Whichever merges second carries a real conflict, and the resolution has a security constraint: the folder decline must survive into the driven world **as a decline** — a stdin driver must never gain a way to answer a folder request with a grant. Flagging here and on #1873's chain so the rebase is done with that rule in view.

**Known remaining gap** (out of this PR's scope): the other client-executed folder tools (`list_folder`, `read_connected_file`, `import_connected_file`, `write_output_to_connected_folder`) still have no headless executor, so a turn *using* an operator-connected folder will park. Filing the follow-up issue now.

naingthet:
Rebased onto current `main` (through #1888). Conflicts were confined to the CLI crate, all of them collisions with the CLI stack that landed in the meantime:

- `main.rs` — the crate docs, the module list, and `USAGE` had all been rewritten by the `--server`/`--permission-mode` work. Kept those verbatim and added the `folder` module, its three usage lines, and its doc paragraph alongside.
- `print.rs` — print mode now boots through `connect::Session` and settles parked interactions through a driver, so the folder refusal had to be re-seated on that shape rather than on the old inline server boot. The refusal itself is unchanged: a `request_folder_access` call is claimed only while unclaimed and still that tool, and answered with `RequestFolderAccessResult::Declined`.

One adaptation the rebase forced. The client-executor token used to be read straight off the embedded server; with `--server` there may not be an embedded server at all, and that token is minted per launch and never leaves the process that owns it. `Session` now carries it as an `Option`, and in attach mode print mode says the request belongs to the attached server's own client instead of trying to claim a call it cannot authenticate for. Embedded runs — the headless case this is for — are unaffected.

**Plane classification:** nothing here lands on either HTTP plane. The consent surface is entirely CLI-side provisioning: `folder connect|list|disconnect` talk to the host broker directly, before or outside a running server, and add no routes. The only server-side change is a lifecycle test. So the member/deployment split in decision record 0004 does not apply, and `deployment_api`, its layer, and `deployment_plane_routes()` in the conformance test are byte-identical to `main`. Worth stating plainly because these grants *are* host-filesystem reach — if this surface ever grows an HTTP route that creates or modifies a folder grant at deployment scope, that route belongs behind the admin gate.

Verified on the new base: `cargo check -p openwave-cli -p openwave-server --locked`, `cargo test -p openwave-cli -p openwave-server --locked`, and `cargo fmt --all -- --check` all pass. One unrelated flake on a first run — `one_server_process_owns_a_desktop_data_directory` lost a data-directory lock race under parallel load — passed on its own and on a clean re-run of the full suite.

naingthet:
Re-reviewed the rebased `print.rs` integration personally — the security invariant survived the rewrite, and it's enforced structurally, not by discipline:

- `decline_folder_request` is called from the `ToolCallStarted` arm, outside `settle()`, and takes no `Driver` — the type system itself prevents consulting the driving protocol. `Interaction` remains a closed three-variant enum, so no folder request event can ever be emitted for a driver, and `parse_decision` has no folder vocabulary.
- The new `no_decision_line_can_grant_a_folder` test is the right tripwire, including the approval-shaped-line case — that's the repurposing attack someone would actually try.
- The attach-mode call is honest and correct: `client_executor_token()` exists only when embedding, `--server` conveys only a bearer, and no flag hands the executor credential to an attaching client — that credential is the native-trust boundary, and keeping it unexposeable is better than any documentation of when not to use it. Leaving the parked call for the attached server's own executor, with a notice naming it, is the honest behavior.
- Reporting-only failure handling (never cancelling the turn over a folder request) is the right severity.

Stacked correctly as the last PR of the chain; ready to merge after #1880, pending green lanes.

PR #1885, opened 2026-08-10
feat(cli): attach to a running server and guard the data dir against double embedding

### LINKED ISSUE #1869: cli: attach mode (--server URL) and same-data-dir embed guard
Implements part 6 of decision record 5 (docs/decisions/0005, PR #1864 — don't start before it merges).

The CLI always embeds a server in-process (`-p`, `tui`, `serve` all boot via `openwave_server::bind_configured`). Two processes embedding servers over one data dir is unsupported but currently undetected — nothing stops `openwave serve` and the desktop app racing the same SQLite database beyond WAL locking.

Scope:
- `--server <url>` / `OPENWAVE_SERVER_URL` + token: act as a pure HTTP/WS client of an already-running `openwave serve` or desktop instance. The loopback client in `crates/openwave-cli/src/api/client.rs` already speaks the full protocol; this is mostly plumbing the base URL/token instead of booting in-process.
- Embed guard: when the CLI would embed a server over a data dir that another live server owns, refuse with a clear message pointing at attach mode, rather than corrupting quietly. (A lock/pid marker in the data dir is the obvious shape; keep it honest across crashes.)

Validation: attach mode runs a turn against a desktop-owned instance; the guard test is the crash-recovery case — a stale marker from a killed process must not permanently brick the data dir.

naingthet:
Part 6 of decision record 5 — the last of the four stacked CLI/headless PRs (#1864 → #1873 → #1878 → this).

## Attach mode

`--server <url>`, or `OPENWAVE_SERVER_URL`, makes `tui`, `-p`, and every setup command a pure HTTP+WS client of a server that is already running instead of embedding one. The flag wins over the variable. The loopback client already spoke the entire protocol, so this is a base URL and a token in place of a `bind_configured` boot: a new `connect::Session` either binds the engine in-process (unchanged default, and it still owns the accept loop for the command's lifetime) or builds a client against a remote base. An attached command reads no data directory, writes no log file, and touches no keychain.

The token comes from `OPENWAVE_SERVER_TOKEN`, or from the variable `--server-token-env <var>` names. It is never an argument — argv is readable by every process on the machine and lands in shell history, and the token is full authority over the profile. `serve`, `mcp`, and `rehome-secrets` reject the flag; the first two are servers and the third rewrites local keychain items. They ignore the environment variable, so a shell that exports it for its `-p` runs can still start a daemon.

The flags are lifted out of argv in a pre-pass rather than added to each parser. They apply to every client command, and no command takes a value that begins with `--` (each parser rejects one), so nothing else can legitimately be spelled this way.

## Embed guard

`openwave-server` has claimed its data directory since #144 — an advisory lock on `openwave.lock` acquired in `bind_inner`, which every `bind*` entry point goes through, so `openwave serve`, the CLI's embedded boots, and the desktop's `boot_server` are all already covered. What was missing was the message: it said a directory was owned without saying what to do about it. It now names attach mode, and the desktop needs no change to honor it.

The lock is the right mechanism rather than a PID file precisely because it is honest across a crash: the kernel releases it when the file descriptor closes, which happens on a clean exit and a `SIGKILL` alike, so a leftover marker file never has to be reasoned about or cleaned up. `std::fs::File::try_lock` is `flock` on Unix and `LockFileEx` on Windows — no new dependency, and `Cargo.lock` is unchanged.

## Tests

Three at the process boundary in `crates/openwave-cli/tests/serve.rs`, which is where the interesting behavior actually lives — two real processes:

- `a_setup_command_attaches_to_a_running_server` — spawns `serve`, then runs `provider list --server <url>` against it with the token in the environment, and reads the route's answer. The attaching process is pointed at the *same* `OPENWAVE_DATA_DIR` on purpose: a client that quietly embedded a server of its own would hit the guard and fail rather than pass by accident.
- `a_second_process_embedding_the_same_data_directory_is_refused` — the other half; asserts the refusal names `--server`.
- `a_killed_server_leaves_its_data_directory_usable` — the crash case. `SIGKILL` the daemon (no unwinding, no destructors, no shutdown path), confirm the lock file is still on disk, and start a second server on the directory successfully.

`one_server_process_owns_a_desktop_data_directory` in `openwave-server` keeps its in-process claim/refuse/reclaim coverage and now also asserts the refusal points at attach mode.

One unit test in `connect` covers base-URL normalization and the refusals that happen before any request is built.

## Docs

`docs-site/content/docs/headless.mdx` gains an "Attaching to a running server" section — the one-server-per-directory rule, the flags, and how to get a token — plus `OPENWAVE_SERVER_URL` and `OPENWAVE_SERVER_TOKEN` in the environment table.

On tokens the docs say what is true rather than what would be convenient: `openwave serve` prints one, and there is a snippet that captures it. **The desktop's is not reachable** — it mints a per-launch token and hands it straight to its webview, and nothing displays or writes it out, so attaching to a running desktop instance is not possible today. The documented answer is to quit the app and run against its data directory. The existing callout suggesting `OPENWAVE_DATA_DIR` be pointed at the desktop's directory now says the app has to be closed first, which is what the guard enforces.

Closes #1869

### DIFF
--- a/crates/openwave-cli/src/api/client.rs
+++ b/crates/openwave-cli/src/api/client.rs
@@ -42,7 +42,17 @@ struct ChatCreated {
 }
 
 impl Client {
+    /// A client for a server bound in this process.
     pub fn new(addr: SocketAddr, token: &str) -> Result<Self> {
+        Self::attach(format!("http://{addr}"), token)
+    }
+
+    /// A client for a server running somewhere else, named by its base URL.
+    ///
+    /// `base` is already normalized (scheme present, no trailing slash) — see
+    /// [`crate::connect`], which is the only thing that builds one from user
+    /// input.
+    pub fn attach(base: String, token: &str) -> Result<Self> {
         let mut headers = reqwest::header::HeaderMap::new();
         let mut value = reqwest::header::HeaderValue::from_str(&format!("Bearer {token}"))
             .map_err(|error| AgentError::msg(format!("invalid server token: {error}")))?;
@@ -56,7 +66,7 @@ impl Client {
             })?;
         Ok(Self {
             http,
-            base: format!("http://{addr}"),
+            base,
             token: token.to_owned(),
         })
     }
--- a/crates/openwave-cli/src/connect.rs
+++ b/crates/openwave-cli/src/connect.rs
@@ -1,0 +1,184 @@
+//! Attach or embed — how a command reaches a server.
+//!
+//! Every client command (`tui`, `-p`, and the setup families) needs an
+//! `openwave-server` to talk to, and there are two honest ways to get one. By
+//! default the CLI **embeds**: it binds the server in-process over its own data
+//! directory, which is the right shape for a script or an agent trying things
+//! out in isolation. With `--server <url>` (or `OPENWAVE_SERVER_URL`) it
+//! **attaches** instead, becoming a pure HTTP+WS client of an already-running
+//! `openwave serve` — the same client the desktop webview is.
+//!
+//! Attaching is what a second process on one data directory must do. A data
+//! directory belongs to exactly one server process (`openwave-server` holds an
+//! advisory lock on it for the life of the process), so pointing a second
+//! embedding CLI at a directory the desktop or a running daemon already owns is
+//! refused rather than allowed to race the database.
+//!
+//! The token never rides argv. It comes from `OPENWAVE_SERVER_TOKEN`, or from
+//! the variable `--server-token-env <var>` names — a command line is readable
+//! by every process on the machine and lands in shell history, and a per-launch
+//! bearer token is full authority over the profile.
+
+use openwave_core::{AgentError, Result};
+
+use crate::api::client::Client;
+
+/// Names the server to attach to instead of embedding one.
+pub const SERVER_URL_ENV: &str = "OPENWAVE_SERVER_URL";
+/// Holds the bearer token for that server, unless `--server-token-env` names
+/// another variable.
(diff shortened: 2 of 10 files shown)


naingthet:
Reviewed the full diff. Finding that `InstanceLock` already guarded the data dir — and that the real gap was the error message and the missing attach path — was the right call over building a second lock; kernel-dropped flock is crash-honest in exactly the way the issue demanded, and the SIGKILL test proves it rather than asserting it.

What holds up:
- `Session`'s embed/attach split unifies what print/tui/setup each hand-rolled, and attach mode touching nothing local (no data dir, no logs, no keychain) is the correct definition of 'pure client'.
- Token discipline carried through: never argv, `--server-token-env` for indirection, and the refusal message teaches all three escapes (attach, quit, move OPENWAVE_DATA_DIR).
- Strict base-URL validation (scheme required, no query/fragment) before the ws derivation is built from it.
- The attach test given the same OPENWAVE_DATA_DIR is nicely discriminating — a client that silently embedded would hit the guard and fail.
- Honest usage doc for the desktop case: attaching to a running desktop instance is not possible today because nothing surfaces its token — stated as fact, not promised away. If we ever want it, that's a small desktop feature (surface/copy the token), not a CLI change.

Noted trade-offs, all acceptable: the http→ws rewrite in attach mode is untested (flagged in the PR), https bases are normalized but never driven against TLS, and the Windows `LockFileEx` path has no CI signal while that lane is parked — per repo policy it still got the cross-platform care, it just runs nothing.

One pathological edge worth knowing, not fixing: the pre-pass lifts `--server` from any argv position, so a `-p` prompt that is literally the string `--server` would be consumed as the flag. Nobody types that; documenting it here is enough.

PR #1912, opened 2026-08-10
test(release): enforce updater broker transition

### LINKED ISSUE #1911: test(release): enforce reversible updater broker quiescence
Update the trusted release-policy test to require the reversible broker admission barrier used during desktop update installation. This is a prerequisite for #1907 because the policy lane executes the test from `main`.

Deliverable: focused test-only PR with adversarial review.

naingthet:
## Summary
- preserve the existing shutdown-before-install release invariant during migration
- admit the stronger reversible broker barrier only when all quiesce, resume, and pinned-session controls are present
- provide a trusted-main policy prerequisite for #1907

After #1907 merges, the legacy branch will be removed in a follow-up tightening PR.

## Verification
- `git diff --check`

Closes #1911

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -42,6 +42,36 @@ updates:
           - minor
           - patch
 
+  # Public documentation is a deployed Next.js application with its own
+  # production dependency graph and lockfile, not prose-only content.
+  - package-ecosystem: npm
+    directory: "/docs-site"
+    schedule:
+      interval: weekly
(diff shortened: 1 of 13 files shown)


naingthet:
Rebased onto the #1904 stack and tightened per the PR's own follow-up plan: with #1907 merged, the legacy shutdown-before-install branch is no longer admitted — the reversible broker barrier is the only accepted transition, and the legacy form moved into the rejected mutation cases.

PR #1942, opened 2026-08-11
fix(shell-policy): refuse paths the analyzer cannot resolve

naingthet:
The analyzer literalizes a word without expanding it: a parameter expansion or
command substitution contributes its *source text* to the token, and glob
metacharacters survive verbatim. Every path check downstream — the
sensitive-path deny floor for redirect writes, and the `Ask` tiers for
arguments and read redirects — then substring-matches those literals. So on the
parsed path a token is only a *spelling* of the path that will be opened, and a
spelling the shell resolves later matches none of the sensitive markers.

That is enough to walk past the floor with any grant covering a benign program,
or with a folder-scoped "act without asking" rule, and it reaches every operand
position. `cat .en?` names the credential file that `cat .env` is refused for.
`awk '{print}' $(cat p)` prints whatever path the agent wrote into a scratch
file a moment earlier — no environment variable, no cooperating prior shell,
both halves written by the same author. `F=~/.ssh/id_rsa; cat $F` splits the
path across an assignment that is never checked at all, because a
pure-assignment command executes nothing and so is never recorded as a leaf.

Scope worth stating plainly: `analyze_shell_command` has no production callers
today. The shipped product routes everything through `analyze_argv`, where
operands are already resolved and none of this applies. This hardens the
shell-line path against the day something does call it, which is also why the
false-positive side below is treated as seriously as the bypass side — a policy
that prompts on ordinary work teaches people to click through prompts.

## What now refuses

Three checks, all applied only when the leaves came from a command line.

**A command substitution in any operand, whoever runs it.** Its text is
computed by a command that already ran, so nothing about the resulting path is
visible here. Unlike a variable it needs no cooperating environment, and
enumerating the programs it matters for is hopeless — any program that opens a
file will do. The one exemption is stated over the *flag* rather than over a
program list: `-j`, `-P`, `--jobs`, `--parallel` and `--max-procs` take a
number, so `make -j$(nproc)` runs, while `sort -o$(cat p) data` is refused
because `-o` names a file that gets overwritten. Keying this on the flag rather
than on which program it belongs to is what stops `jq -f$(cat p)` and
`git -C$(cat p)` from riding along.

**A parameter expansion**, when the token is already path-shaped or when the
program's operands are paths.

**A glob in an operand the program opens**, when the token is rooted outside
the working tree (`/…`, `~…`) or could expand onto one of the sensitive markers.

The last one is the substance of this round. Its four previous versions each
guessed at the *shape* of a disguise — a bracket in a suspicious position, a
leading dot, a metacharacter in a non-final segment — and each guess was
defeated by spelling the same path a different way, because the shapes were
invented alongside the marker list instead of derived from it. The markers are
not all dotfiles and not all whole segments: an extension (`.pem`), a bare
filename (`id_rsa`, `authorized_keys`) and a directory two segments above the
glob all slipped past a dotfile-shaped rule.

The check now asks the question the deny floor itself asks. The token's
metacharacters are read as wildcards — `?` matches one character, `*` a run,
`[…]` the class — and each marker is tested for reachability under the same
segment-normalized comparison `hits_sensitive` uses.

The one judgement call is which wildcards count as evidence that the token is
aimed broadly rather than at one file, and only `*` is: it swallows a run of
unknown length, so `src/*` can expand onto `src/etc/passwd` and `*.rs` onto
`id_rsa.rs`. Refusing those would refuse most globs ever written while catching
nobody, since anyone who wanted that file would just write it. A `?` or a
bracket class is the opposite — a precise one-character disguise, standing in
for a character the author already knows. So a marker counts as reached when
the token spells it with any number of single-character wildcards but more
literal characters than `*` covers. `.e??`, `.???/id_???` and `id_rs?` are
refused; `ls *.rs` and `grep -r foo src/*` are not. Specificity is not a ratio,
which is what an earlier scoring rule assumed — it counted `?` against the
token and so let half a marker be wildcarded for free.

What this deliberately leaves uncovered is a `*` covering most of a marker,
which is the working-directory question this predicate cannot answer anyway,
and the `cd`-then-glob gap noted at the end. The gain is that a marker added to
the list later is covered the day it is added, with no second place to update.

Two role fixes the matcher depends on, since a check that never runs on an
operand protects nothing:

- `grep`, `rg` and `wc` open their operands as files and were not on the
  path-operand list, so every rule here could be walked around by typing a
  different program name — `grep '' .en?` reads exactly what `cat .en?` reads.
  They are on it now. `find` stays off deliberately: its operands are
  predicates, and `-name` is matched against basenames.
- `awk`/`sed`/`grep`/`rg` take a script or pattern as their first operand, so
  that slot is skipped — `awk '{print $1}' data.txt` is ordinary work and the
  `$1` is a field reference. Working out *which* operand that is means tracking
  flags that take a separated value (`-e`, `-f`, `-F`, `-v`, `-m`, and the long
  forms). Without that, `sed -e 's/.*//' file.txt` shifted one slot along and
  got the filename skipped and the regex checked as a path. It also means
  counting the words the parser removes: `k=v` is stripped out of `argv`, but
  the program still counts it, so `grep a=b file` and `awk -v x=1 '{print}' f`
  each shifted the real operand out of every check until word positions were
  tracked alongside `argv`.
- `sed -i ''` is not assumed to be BSD. BSD reads that empty argument as a
  backup suffix and GNU reads it as the script, and nothing here can tell which
  will run; letting it spend the script slot is right under both, because it
  leaves the file in a path-checked slot either way.

**Assignment values** now go through the same `hits_sensitive` / `climbs_out`
checks an argument gets, which kills the assign-then-use family at its source.

## What deliberately still runs

Pinned in the positive corpus, so a later widening has to break a test that
says why: `find . -name '*.py'`, `ls *.rs`, `grep -r foo src/*`,
`grep -r foo src/**/*.rs`, `ls src/[a-z]*.rs`,
`ls report[1].pdf`, `cat file-[1].txt`, `grep 'a?b' file.txt`,
`grep -E 'a[b]c' file`, `grep '[n]ginx' access.log`, `grep '.*' file.txt`,
`grep '.*foo' file.txt`, `sed 's/.*//' file.txt`, `sed -e 's/.*//' file.txt`,
`sed -i '' 's/.*//' file.txt`, `awk -F , '{print $1}' data.txt`,
`awk '/.*x/{print}' data.txt`, `awk '{print $1}' data.txt`,
`sed -n '1,5p' file.txt`, `make -j$(nproc)`, `echo $USER`,
`CFLAGS=-I../include make`.

One case still asks and is worth flagging rather than hiding:
`PYTHONPATH=../src pytest`. A search-path assignment pointing outside the
granted folder is a normal monorepo idiom, but it is also exactly what a
folder-scoped grant is about, and the value is the only place that path is
visible in plain text. Distinguishing a search path from an output path by
variable name would exempt the one variable class that can change which code a
vetted program loads, so the prompt stands. Easy to revisit if it proves
annoying in practice.

## What now prompts that did not

A command substitution counts wherever it appears, so a date stamped into a
filename prompts: `tar -czf backup-$(date +%F).tgz src`,
`echo "done at $(date)"`, `git commit -m "$(cat msg)"`. Nothing on this path
can tell `$(date +%F)` from a path the caller wrote into a scratch file a
moment earlier, and narrowing by program name is the maintenance liability
`COUNT_FLAGS` exists to avoid. Likewise a glob rooted outside the working tree:
`ls ~/Downloads/*.pdf`, `cat ~/notes/*.md`, `tail -n 20 /var/log/*.log`.

These follow from the design rather than from a rule that could be tightened,
so they stand — but the first is an everyday idiom, and it is pinned in the
Ask corpus so its cost stays visible instead of being rediscovered.

## Tiers

Unresolvable redirect *writes* `Ask` rather than `Deny`. "I cannot resolve
this" is a weaker signal than "this is a known-sensitive path", and `Deny`
cannot be granted around at all; `echo done > $LOGFILE` is not
`> ~/.ssh/authorized_keys`. Named sensitive writes keep `Deny`. Everything else
this adds lands on `Ask`, matching the tier a named sensitive path in an
argument already gets. After the demotion,
`F=~/.ssh/authorized_keys; echo x >> $F` is refused as
`assignment names a sensitive path` — the assignment check is now the
load-bearing one for that command, with the redirect check behind it for a
value the analyzer never sees.

`PREFIX=../out make` asks, for the same reason `make ../out` would.
`CFLAGS=-I../include make` does not, because `-I../include` is a flag with a
path glued to it rather than a path; the asymmetry is called out in a comment
at the check.

`analyze_argv` is exempt from all of it. Its callers hand over an executable and
an argument vector that go straight to `execve`, so a `*` or a `$` there is a
literal the program receives, not something a shell will resolve afterwards.
Refusing it would reject ordinary operands while protecting nothing. A test
pins this.

## Not fixed here

`cd ~/.ssh && cat *` is allowed both before and after this change: the `.ssh/`
marker requires a trailing slash, so the bare `~/.ssh` operand misses it, and a
relative glob is allowed. That is a pre-existing gap in the marker list plus the
absence of any cwd tracking, and it also means a relative glob's reach is only
bounded by the working directory *until a `cd` moves it*. The doc comment on
the glob rule now says so explicitly rather than claiming a containment the
analyzer does not have.

## Verification

`cargo fmt --all --check`, `cargo check -p openwave-shell-policy --locked`,
`cargo clippy -p openwave-shell-policy --all-targets --locked -- -D warnings`,
`cargo test -p openwave-shell-policy --locked` (16 passed),
`cargo check --workspace --locked`, `cargo test -p openwave-core --locked approval`
(37 passed) and `cargo test -p openwave-server --locked approval` (40 passed)
are all clean.

### DIFF
--- a/crates/openwave-shell-policy/src/lib.rs
+++ b/crates/openwave-shell-policy/src/lib.rs
@@ -244,17 +244,21 @@ enum Expansion {
 /// `$HOME/.ssh/authorized_keys` and `/et[c]/shadow` both miss every marker
 /// while naming a file the floor exists to protect.
 ///
-/// The shape rules keep it narrow. An expansion only matters in something
-/// already shaped like a path, so `echo $USER` is untouched. A glob matters
-/// when the token is rooted outside the working tree (`/…`, `~…`), climbs
-/// (`..`), or hides a character inside a directory name (`[`, `?` with a
-/// separator) — which leaves the everyday relative globs (`*.py`, `src/*`,
-/// `src/**/*.rs`) exactly as they were, since they can only expand to
-/// non-hidden entries under a directory the grant already covers.
-fn unvettable_path_argument(token: &str) -> bool {
+/// The shape rules keep it narrow. An expansion matters in something already
+/// shaped like a path, or in an operand of a program whose operands *are*
+/// paths — so `cat $F` is caught while `echo $USER` and `make -j$(nproc)` are
+/// untouched. A glob matters when the token is rooted outside the working tree
+/// (`/…`, `~…`), climbs (`..`), hides a character inside a directory name
+/// (`[`, `?` with a separator), or spells a literal with a one-character
+/// bracket group — which leaves the everyday relative globs (`*.py`, `src/*`,
+/// `src/**/*.rs`) and ordinary regex operands exactly as they were.
+fn unvettable_path_argument(token: &str, program_takes_paths: bool) -> bool {
     let expansion = token.contains('$') || token.contains('`');
     let path_shaped = token.contains('/') || token.starts_with('~');
-    if expansion && path_shaped {
+    if expansion && (path_shaped || program_takes_paths) {
+        return true;
+    }
+    if spells_a_literal(token) {
         return true;
     }
     if !token.contains(['*', '?', '[']) {
@@ -264,6 +268,20 @@ fn unvettable_path_argument(token: &str) -> bool {
         || token.starts_with('~')
         || token.contains("..")
         || (path_shaped && token.contains(['[', '?']))
+}
+
+/// A bracket group that can match exactly one character, as in `~/.bashr[c]`.
+///
+/// That is not globbing, it is spelling a literal out of reach of a substring
+/// check: the token names one specific file while matching no marker. Real
+/// patterns quantify over more than one character (`[a-z]`, `[abc]`, `[^x]`),
+/// so this catches the disguise without touching them.
+fn spells_a_literal(token: &str) -> bool {
+    let bytes = token.as_bytes();
+    bytes
+        .iter()
+        .enumerate()
+        .any(|(i, &b)| b == b'[' && bytes.get(i + 2) == Some(&b']'))
 }
 
 /// A redirect target the analyzer cannot resolve to a path.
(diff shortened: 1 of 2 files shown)


naingthet:
Parking this. It hardens an entry point with no production callers, and it has been through six rounds without converging — the cost is now clearly out of proportion to the risk it removes.

The current head is worth keeping: the rule is derived from the sensitive-marker list rather than guessing at the shape of a disguise, which is the first version of this that stands on a principle instead of a pattern. Whoever picks it up should start by re-running the probes rather than re-reading the diff; every round that shipped broken shipped because it read correctly.

Two known gaps stay documented in the body: `cd ~/.ssh && cat *` needs working-directory tracking this predicate cannot do, and `PYTHONPATH=../src pytest` prompts.

naingthet:
Retracting the parking note above — this is back in review, not shelved. Round 6 is pushed and under an independent review that has not reported yet.

PR #1944, opened 2026-08-11
fix(cli): surface the print-mode failures that left a run parked in silence

naingthet:
Three print-mode failures shared a shape: something went wrong and the run said nothing about it. A caller saw a process that hung, or one that exited as if all was well.

## Ctrl-C during a driven run

Waiting for the next journal frame, and waiting for a driver to write a decision line, were both indefinite waits that did not watch for SIGINT. Pressing Ctrl-C against a driven run left the turn running server-side while the process sat there.

Both waits now race an interrupt watcher. On the first signal the run cancels the turn, emits a `halted` object with reason `interrupted`, and exits 130. The watcher stays alive afterwards: a second Ctrl-C exits the process immediately, which is what the default disposition would have done anyway, so a run that cannot cancel cleanly is still escapable.

The watcher is installed after the turn is posted, not before. Installing it earlier would swallow signals over the WebSocket handshake and the POST — neither of which watches the interrupt, and neither of which has a turn to cancel yet.

The coverage has a limit worth stating: the interrupt covers the frame wait and the driven decision wait. A `--server` host that accepts TCP and then never answers still blocks with the interrupt unwatched, because the CLI's HTTP client sets no request timeout. That is unchanged by this PR and a second Ctrl-C remains the way out of it.

## Pending-plan and pending-question lookups

When the turn parked on a plan or a question, the run fetched the pending request to show the driver what it was answering. That fetch discarded its error, so a failed lookup was indistinguishable from "nothing is pending" — the run continued past the interaction it was parked on, waited for frames that would never come, and eventually exited as though the turn had simply not proposed anything.

The two lookups now return a typed result. A genuine "nothing pending" still continues, since the interaction may have settled before the fetch landed. A transport or server failure produces a `halted` object with reason `pending_lookup_failed`, cancels the parked turn, writes the reason to stderr, and exits 4.

## Declining a folder request

`request_folder_access` cannot be answered by a driving process: the protocol's decision vocabulary is closed at approval, plan, and questions, and standing folder consent comes only from `openwave folder connect`. Print mode printed a notice that it was declining and then did nothing further, so the call stayed parked in the client-execution set and the turn waited on it indefinitely.

The refusal is now actually delivered: the run claims the parked call and resolves it with the declined outcome, the same claim-and-resolve the desktop client performs.

The delivery runs on its own task rather than in the event loop, and reports back over a channel, because the call is not claimable the instant it starts. It polls with backoff and no deadline; a wall-clock budget would have raced a step that parks its client call last. A poller is dropped as soon as the loop sees `ToolCallCompleted` for that call, so a call that ends some other way stops being waited on.

Which outcomes end the run is deliberately narrow. A call that never parks is a routine outcome — the agent can reject it at the client checkpoint and decline it itself, in which case there is nothing to deliver and the turn completes normally. Only a call observed sitting in the pending set that then cannot be refused, because the claim failed, the resolve failed, or another executor already holds it, is fatal; that produces reason `folder_decline_failed` and exit 4. Everything else is reported to stderr and the run carries on.

The exit-code table in the headless docs names the new reasons.

## Tests

One added: an unreadable pending-plan lookup must halt rather than read as "nothing pending". It points the client at a closed local port, which is the failure the old code silently absorbed.

### DIFF
--- a/crates/openwave-cli/src/print.rs
+++ b/crates/openwave-cli/src/print.rs
@@ -192,25 +192,13 @@ async fn one_turn(
     // A resumed chat replays its history first; nothing before this turn's own
     // `TurnStarted` is ours to print.
     let mut ours = false;
-    // Watch for the interrupt on its own task: a `ctrl_c()` future created per
-    // loop iteration could miss a signal that lands between two iterations.
-    let (interrupted, mut interrupt) = tokio::sync::oneshot::channel();
-    tokio::spawn(async move {
-        if tokio::signal::ctrl_c().await.is_ok() {
-            let _ = interrupted.send(());
-        }
-    });
+    let mut interrupt = Interrupt::watch();
 
     let outcome = loop {
         let frame = tokio::select! {
             frame = stream.next(client, chat) => frame?,
-            _ = &mut interrupt => {
-                printer.finish();
-                eprintln!("openwave: interrupted; cancelling the turn");
-                // Best effort: the turn may already have finished, which the
-                // server answers with a conflict.
-                let _ = client.cancel_turn(chat, turn_id).await;
-                return Ok(EXIT_INTERRUPTED);
+            () = interrupt.fired() => {
+                break halted(client, chat, turn_id, &interrupted(), &mut printer).await;
             }
         };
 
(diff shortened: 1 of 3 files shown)


naingthet:
Reworked the folder-decline change after review. The fatal branch was firing on a routine outcome — a call the agent declines at the checkpoint never parks, so the 10s poll expired and turned a completing turn into exit 4 with empty stdout, since the loop broke without draining the buffered text.

The refusal now runs on its own task rather than inline, and reports back over a channel the event loop selects on. Three consequences: the turn's output keeps streaming while a refusal waits; there is no fixed wall clock, so a call that parks late (an isolated client call is taken only after every sibling in its step is terminal) is still declined; and a call that never parks simply never reports, which leaves the run exactly as it was before — one silent, successful turn.

The outcomes are now enumerated instead of collapsed into `Err`. Only a call seen sitting in the pending set can end the run: a claim or resolve that failed, or a claim held by another executor. A poll that fails, a call that has not parked yet, and a call that parks under another tool's name are all non-fatal — none of them establishes that this run owes an answer it cannot deliver.

Also from review: the interrupt watch is installed before the message is posted, and returns only once the handler is registered, so no window is left where SIGINT would kill the process and abandon the turn it should cancel. Its doc comment no longer claims coverage it does not have — the HTTP client has no request timeout, so a call to an unresponsive `--server` host still blocks with the interrupt unwatched, and a second Ctrl-C no longer kills the process. Giving the client a request timeout is the right lever for that, but it changes every CLI surface and belongs in its own change, so the comment states the limitation instead.

PR #1951, opened 2026-08-12
fix(shell-policy): refuse a glob that spells as much of a marker as it hides

naingthet:
## What was wrong

`could_reach_sensitive` treats a token's metacharacters as wildcards and asks whether it could match one of `SENSITIVE_TARGET_MARKERS`. It scores a spelled character `+1`, a `*`-supplied one `-1`, and a single-character wildcard `0`, then asked for a **positive** total. A tie therefore read as unreachable, so these auto-ran under a folder-wide grant:

- `cat .e*` — spells `.e`, lets `*` supply `nv`
- `cp .np* /tmp/x` — spells `.np`, lets `*` supply `mrc`
- the same shape against `.p*`, `.??*`

## The change

One operator, at both the early-exit and the final check: `score > 0` becomes `score >= 0`.

The reasoning, which is now in the doc comment: a `*` that covers no more of a marker than the token already spells is not a pattern aimed at a directory. `.e*` names two of the four characters in `.env` and expands onto almost nothing else. A glob written for ordinary work spells far less of a marker than it covers, so the tie can go to refusing without catching it.

## What it costs

Nothing measurable. I ran 53 everyday glob commands (`ls *`, `cat *.rs`, `cp *.json dist/`, `rg pat crates/*/src`, `prettier --write 'src/**/*.ts'`, `tar czf out.tgz src/*`, `cat .eslintrc.json`, `ls .github/workflows/*`, and so on) under `allow_all()` with the gate at `> 0` and again at `>= 0`, and diffed the verdicts. **Zero changed.** All 16 existing tests pass unmodified.

## What is still uncovered

A `*` covering *most* of a marker — `.*` reaching `.env` — stays reachable. That is the working-directory question this predicate cannot answer, and the doc comment now names it explicitly rather than describing the old boundary.

## Tests

Two entries added to `NEVER_ALLOW`, the corpus that pins the floor at its widest grant: `cat .e*` and `cp .np* /tmp/x`. They sit with the existing `.e??` / `.???/id_???` boundary cases under a comment that now covers both halves of the rule. No tests removed.

Follow-up to #1942.

### DIFF
--- a/crates/openwave-shell-policy/src/lib.rs
+++ b/crates/openwave-shell-policy/src/lib.rs
@@ -340,13 +340,19 @@ fn substitution_is_a_count(token: &str) -> bool {
 /// `?` or a bracket class is the opposite — a precise one-character disguise,
 /// standing in for a character the author already knows. So a marker counts as
 /// reachable when the token spells it out with any number of single-character
-/// wildcards but **more literal characters than `*` covers**: `.e??` and
-/// `.???/id_???` are refused, `*.pe?` is refused for pinning `.pe`, and
-/// `src/*` is not.
+/// wildcards and **at least as many literal characters as `*` covers**: `.e??`,
+/// `.???/id_???` and `.e*` are refused, `*.pe?` is refused for pinning `.pe`,
+/// and `src/*` is not.
 ///
-/// What that leaves uncovered is a `*` covering most of a marker, which is the
-/// working-directory question this predicate cannot answer anyway — and the
-/// `cd`-then-glob gap below, which it also cannot.
+/// The tie goes to refusing because a `*` that covers no more of a marker than
+/// the token already spells is not a pattern aimed at a directory: `.e*` names
+/// two of the four characters in `.env` and expands onto almost nothing else.
+/// A glob written for ordinary work spells much less of a marker than it
+/// covers, so it stays on the safe side of the tie.
+///
+/// What that leaves uncovered is a `*` covering *most* of a marker, such as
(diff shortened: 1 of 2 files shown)


naingthet:
Closing unmerged. The premise was wrong.

I measured this as free by running 53 everyday glob commands and finding no verdict changes, but that sample only covered `prefix*` shapes. The score-zero tie is symmetric: it also fires when a `*` leads and the token spells a short literal suffix. Independent review found real regressions from Allow to Ask, including `grep -l foo src/*init*.py` and `cat **/*init*.py` (reaching `.gdbinit`), `cat src/*email*.ts` (reaching `.pem`), `grep -l x *file` (reaching `.profile`), `cat pages/login*`, and `cat config/env*.yml`. An `__init__.py` sweep is ordinary work, and two spelled characters after a star is far too weak a signal to refuse on.

The other half is that the closure buys close to nothing. `.e*` and `.np*` shut, but `.?*`, `.*`, `.s*`, `.n*` and `.a*` stay open, so anyone deliberately disguising a read of `.env` or `~/.ssh/id_rsa` deletes one character and walks through. Real cost to real work, near-zero gain against the threat.

The asymmetric variant review suggested — tie only when the star trails a spelled marker prefix — would drop most of the false positives, but it does not change the second problem. The bypass is a keystroke away either way. The honest boundary is the one #1942 already shipped, and the gap it leaves is the working-directory question this predicate cannot answer.

Refs #1942.

PR #1995, opened 2026-08-12
feat(desktop): computer use — screen capture and consent-gated app control

naingthet:
## What

First-class computer use for the macOS desktop app: screen capture and consent-gated control of native applications, per decision record `docs/decisions/0010-computer-use-screen-capture-and-app-control.md`.

Three layers with a strict trust gradient:

- **`openwave-cu-helper`** (new Swift executable, `crates/openwave-host-broker/helper/`) — a dumb per-operation executor over JSON stdio: ScreenCaptureKit capture with Set-of-Marks overlay, bounded `AXUIElement` tree reads with index-path ids + FNV-1a drift fingerprints, and CGEvent input synthesis. No policy, no persistence, 30s timeout, screenshots written to a broker-owned staging path (never on stdout).
- **Host broker** — all policy. New `CaptureScreen` / `ReadAppContent` / `ControlApp` capabilities scoped per app bundle id (+ whole-display), a hard app blocklist (exact-or-dotted-prefix), the consequential-action classifier with a deliberately narrow lexicon (commit-shaped labels trip it; navigation/dismissal words do not), single-use act-time confirmations (TOCTOU-safe: the label is re-read and must still match at act time), and append-only audit with control intent recorded durably before synthesis. Protocol 10 → 11.
- **Core + server + desktop** — ten `computer_*` client-executed tool contracts in `openwave-core`, registered in the server only for the desktop profile on macOS; a `ToolApprovalKind::ComputerMayControlApp` (approvable once, never standing-grantable, never auto-judgeable); a desktop claim/execute/resolve module with per-app consent cards that write the broker grant, a single-use confirmation surface, a screenshot→blob→transcript image path, and an always-on "controlling {app}" indicator with a Stop latch that short-circuits control before the next broker round-trip.

## Consent model (calibrated — not over-asked)

1. OS TCC (Screen Recording + Accessibility) requested together on first use.
2. A per-app grant, asked once per app via the in-chat card (once / always for this chat / always), written to the broker grant store.
3. An act-time confirmation only for consequential actions.

Within a granted app, individual clicks/keys/scrolls do not card. Reads never card once their grant exists. Plan mode refuses control tools, keeps reads. No LLM auto-approval judge for control.

## How verified

- `cargo build --workspace` clean; suites green: host-broker 146, core 603, server 779, desktop 132, UI 926 (+ tsc clean, command-parity gate passes).
- **End-to-end against a live app**: drove the real broker over its stdio protocol with the real helper — Hello advertises the CU ops (helper resolved), permission preflight, a Terminal control op refused by the blocklist (non-retryable), a ControlApp grant on Notes, an AX tree read (truncation-bounded), a capture returning 29 marks + a handoff id, single-use handoff redemption (second redeem fails `NotFound`), and an annotated PNG with the numbered mark badges rendered at the target elements (visually confirmed).

## Not in this slice / follow-ups

- The full app UI driving the agent loop against a real desktop session (the pieces are tested in isolation; a live turn is the remaining integration check).
- Per decision record: browsers are out of the starter scope; no per-app plugin skills (kept generic — see the prompt section); hand-off tier, AX-tree diffing, and app-launch affordance are candidate follow-ups.
- Helper code-signing / stable bundle path and the TCC responsible-process attribution must be validated on a signed build (decision record flags this).


naingthet:
Thanks — both bugs confirmed in code and fixed in 40829c3c.

**1. Captures never reaching the transcript.** Confirmed: `finish_capture` computed `capture_image_refs()` but the desktop's `StoredResolution::Completed` had no `images` field, so the refs were dropped before the resolve POST. Fixed by threading an `images: Option<Vec<ImageRef>>` through `StoredResolution::Completed` → the resolution wire → the store, which now projects `ToolResultPreview::ScreenCapture` with the real mark count (read from the capture's reported `rows`). Added a route test proving the desktop wire deserializes into the server's resolution shape (`completed_resolution_with_images_round_trips_from_the_desktop_wire`), since the serde field-name match across the process boundary is exactly where this class of bug lives.

**2. Confirmation eviction dead.** Confirmed: the insert into `pending_confirmations` never pushed onto `confirmation_order`, so the eviction loop popped an empty queue and the map grew unbounded. The insert now enrolls each confirmation id, so the 64-entry FIFO cap actually holds.

All lanes green (clippy `-D warnings`, rustfmt, full test suite, desktop UI, postgres, gitleaks, cargo-deny/machete). The two mid-run lane failures were CI infra (503 on setup); a rerun passed them.

The review body was truncated at ~768 chars in the API, so I only saw the two named bugs plus the "key chords / raw coordinates" consent note in the summary fragment. If the remaining items (the other 2 bugs / 3 suggestions / 1 nit) are in the truncated portion or as inline comments, point me at them — the `images`/`rows` projection path is the area I'd most want a second look on regardless.

naingthet:
All four review passes (security, Rust, Swift, frontend) are addressed. Summary of what changed since the review:

**Critical (security)**
- **Coordinate-click confinement**: clicks/scrolls now verify the target point is inside a window the granted app owns (CGWindowList ownership check), so a compromised model can no longer drive OpenWave's own consent/Stop/Resume controls or a blocked app's window via raw coordinates. Verified live: an out-of-app point refuses with `target_outside_app`.
- **Key-press gate**: chorded keys and bare Return/Delete/ForwardDelete now hold for the same single-use confirmation as clicks (the commit paths — send/delete/quit — that bypassed the element-label gate). Regression-tested.
- **Scroll/focus reclassified**: they synthesize input and move windows, so they now require the `ControlApp` grant and register `Sensitive` (plan mode refuses them), not the read grant.

**High (correctness)**
- Capture blobs are now counted as live references by the orphan sweep and servable to the renderer (`ScreenCapture` threaded into both seams; regression tests added).
- Helper: scroll Int32 crash clamped; AX messaging timeout set so a hung target app fails instead of wedging the broker; the auto-yield re-checks the foreground mid-typing-burst; non-BMP text posts as whole grapheme clusters.

**Medium / other**
- Confirmations bind to the element's fingerprint, not just the (truncated) label — a same-label element swap now voids the confirmation (regression-tested).
- The helper's blocklist mirrors across capture/read/list (not just control) with the broker's dotted-boundary semantics.
- The consent card names the bundle id (the real grant principal), not just the app's self-reported name.
- Master enablement setting (`computer_use.enabled`, default on) gates registration at boot; `computer_capture_screen` refuses with a typed error when the selected model takes no image input.
- App capture picks the display the app is actually on; the mark overlay is correct on secondary displays.
- Frontend: the consent/confirmation card query/listen race, missing approval-card presentation for `computer_may_control_app`, the missing `screen_capture` preview parser, and fire-and-forget decision invokes (now busy-guarded with an error surface).

**Filed as follow-ups** (hardening, not blockers): #2010 (audit fidelity for held actions + scroll/focus intent), #2011 (once-grant durability on crash), #2012 (yield as a structured error code), #2013 (IDE/shell-class blocklist breadth + classifier locale/label-less coverage).

All CI lanes green. Auto-merge is off — ready for your review.

naingthet:
Rebased onto main now that the Tidebreak rename (#2014) merged — no longer stacked; this targets `main` directly and applies cleanly.

The rebase followed the rename throughout: crate references (`tidebreak-core` / `-server` / `-desktop` / `-host-broker`), the helper binary (`openwave-cu-helper` → `tidebreak-cu-helper` — TCC binds grants to the code signature, so the name is a stable contract), and the self-automation blocklist (`io.brightwave.tidebreak`). One real fix fell out of it: the rebase initially kept an older `ci.yml` that predated the transitional trusted-policy self-check block, tripping `release policy`; restored to main's version. All lanes green (clippy `-D warnings`, rustfmt, full test suites, desktop UI, postgres, gitleaks, cargo-deny/machete).

Ready to merge.

Issue #535, opened 2026-07-27
Desktop: route the app and address the workspace layout in the URL


naingthet:
Part of #533 (phase A).

Navigation today is ad-hoc store state: a `Surface` union in a zustand store, persisted to localStorage, with no URL, no history, and no deep links. That cannot express "open this document at this citation", and it makes the panel layout impossible to share or restore.

Introduce a router and move navigation into the URL:

- Routes for home, a chat, and settings.
- The panel arrangement encoded as `?left=`, `?right=`, `?fullscreen=` search params with a small segment grammar (`chat`, `sources`, `sources.{documentId}`, `outputs`, `folders`, `chats`).
- `Surface` and the workspace-layout store retire; per-chat layout memory is replaced by the URL plus history.

Views are moved, not rewritten. The resizable two-slot layout stays as it is until the next slice.

PR #2007, opened 2026-08-12
feat(desktop): open the gateway's app page instead of publishing here

### LINKED ISSUE #2005: Replace the Publish dialog with a link to the app's gateway page (decision record 0011)
Decision record 0011 (docs/decisions/0011-shared-app-publishing-ownership.md): publishing is a governance action on the gateway's own web surface; no harness product surface mutates publish state.

Implementation:

- The app page's Publish affordance opens the app's page on the gateway, built from the managed policy's gateway URL, instead of raising the team-picker/preflight dialog.
- Remove the dialog, the preflight, and the picker's CLI-tier teams fetch; no OpenWave code path calls the CLI-tier publish route afterward (record 0011's validation clause — the dialog is removed, not feature-flagged off).

Blocked on brightwave-inc/model-gateway#532: the gateway shared-app page is viewer-only today, and a link pointing at a page with no publish control strands authors. The existing dialog stays until #532 lands.

abaas-bw:
Closes #2005.

Implements decision record 0011: publishing a shared app is a governance action done on the gateway's own web surface, next to the publish state, the team grants, and the revocation it belongs with.

> ✅ **Unblocked.** brightwave-inc/model-gateway#535 has merged, so the gateway's shared-app page now carries the author-facing publish affordance this links to.

## What changed

The Publish dialog, its team picker, and its preflight are **deleted**. In their place, **Publish at gateway** resolves the app's page at the deployment holding it and opens it in the browser. Removed rather than feature-flagged off — record 0011 names a flag as the plausible wrong implementation, and says the flag's removal is part of the swap, not a follow-up.

## The new route

`POST /apps/{id}/gateway-page` answers with the address of the app's page, registering the draft first if the app has never been registered.

- **Registration on the way is deliberate.** Drafts are registered lazily, on first relay or consent, so an app the author has only ever run locally has no page at the gateway yet; a link to nothing is worse than the wait. Registering is *not* publishing — it creates the author's own draft, reachable by them alone, and changes nothing about who else can open the app. The old publish route did the same thing on its way through.
- **A `POST`, not a `GET`,** because it is not a read.
- Every way it can fail to produce an address is an outcome (`no_gateway`, `not_registered`, `refused`, `unreachable`) rather than an error, the way the invoke ladder answers. Only a broken local request — no such app, or an app binding nothing at the gateway — is a `ServerError`.
- The URL is built through `Url` rather than by formatting. A gateway identifier is only required to be printable ASCII, which admits `/` and `?`; formatted into a string, one of those would silently address a different page on the same origin. There is a unit test for exactly that.

## Deletions worth flagging in review

Record 0011's validation clause is that *"OpenWave contains no code path that calls the CLI-tier publish route"*. Keeping an unreachable client would not satisfy that, so this removes the whole publish client — `GatewayAuth::publish_shared_app`, `GatewayConnection::publish_shared_app`, `decode_shared_app_publish`, `GatewayPublishOutcome`, `GatewayDraftSource::publish` and its implementations, and `GatewayPublish`. The gateway keeps its CLI-tier route as API for scripted use; that is the gateway's side of the boundary, not ours.

The teams client goes with it — `GET /gateway/teams`, `GatewayRuntime::teams`, `GatewayAuth::teams`, `GatewayTeam`, `GatewayTeams`, `GatewayTeamInfo`. Its only purpose was deciding whether to show the button. That decision is now made from the managed policy already in React context: the affordance appears for an app that actually binds something at the gateway (the same predicate registration follows) on a profile paired with one.

No `#[allow(dead_code)]` anywhere — `cargo clippy --workspace --all-targets` is clean.

## Tests

The three dialog tests are replaced by three on the affordance: that it resolves the page and opens the returned URL, that a **refusal opens nothing** (a refusal carries no address, and sending an author to a page the gateway just declined would strand them), and the two-part visibility predicate. Server-side, the route is covered for an unpaired profile (it answers, and reaches the gateway zero times — registering against a deployment this profile is not paired with would leave a draft nothing here will ever read again), for a registered app resolving to the exact page URL, and for a refusal travelling in the gateway's own words.

## Docs

`docs/local-apps.md`'s "Promotion: planned for, not built" section was stale — registration and publishing both shipped. Rewritten as "Sharing: registered here, published at the gateway", stating where publishing happens and why. `docs/gateway-boundary.md`'s wire list gains the shared-apps routes and records that the publish route is deliberately absent.

## Verification

`cargo fmt --all --check`, `cargo clippy --workspace --all-targets --locked`, `cargo test -p openwave-server`, and `pnpm test` (927 tests) all pass locally. `wire.ts` was regenerated via `UPDATE_WIRE_TYPES=1` and is committed. I have not driven the running desktop app to watch the browser hop happen.

## Second commit: a rename

`openSignInPage` opens any URL in the user's browser. Five of its six callers pass a sign-in URL; the sixth — mine — passes the gateway page an app is published from. Renamed to `openInBrowser` so the new call site does not read as something it isn't. Mechanical: one import and one call in each of four files.


abaas-bw:
Blocked on brightwave-inc/model-gateway#535 (the gateway's author-facing publish affordance). Merging this before that lands would point the Publish button at a page with no publish control. Holding until #535 is in.

abaas-bw:
Unblocked: brightwave-inc/model-gateway#535 merged, so the gateway's shared-app page now carries the author-facing publish affordance this links to. Merge order satisfied.

PR #2038, opened 2026-08-13
ci: isolate trivy from sandbox image publish credentials

naingthet:
## Why

LiteLLM's March 2026 PyPI compromise (TeamPCP) came from an unpinned Trivy running in the same CI environment as long-lived publish credentials. Tidebreak already checksum-pins the Trivy binary and does not keep a long-lived registry token, but the scanner still ran in the `build` job that holds `packages: write` and a live `GITHUB_TOKEN`. A compromised scanner on that runner can steal the token the way the March payload actually did and push while the job is still running.

## What

Split `.github/workflows/publish-sandbox-image.yml` so Trivy never holds publish rights:

- **`build`** still builds, smoke-tests, and pushes the per-arch tags. No Trivy.
- **`scan`** is `packages: read` only, with no checkout. It pulls those arch images, checksum-installs Trivy, and scans.
- **`manifest`** waits on `scan` and only then creates the user-facing `$IMAGE_TAG` manifest list.

A failed scan leaves the per-arch tags on GHCR and never creates the version tag the digest pin follows. The workflow-security tests now lock that boundary: Trivy only in `scan`, exactly two `packages: write` jobs, and `manifest` needs `scan`.

## Notes

- Not verified by running the publish workflow (main / tag / schedule / dispatch only).
- Production Apple / Tauri / AWS secrets and `E2B_API_KEY` were already isolated from this workflow and stay that way.

### DIFF
--- a/.github/workflows/publish-sandbox-image.yml
+++ b/.github/workflows/publish-sandbox-image.yml
@@ -40,11 +40,14 @@ run-name: >-
 # whole sandbox under emulation there). Image tags are treated as immutable:
 # a tag that already exists on GHCR fails the run rather than being repointed.
 #
-# Before anything is pushed, both variants pass a trivy vulnerability and
-# secret scan (pinned binary release, checksum-verified — this workflow avoids
-# third-party actions). The full all-severities report lands in the run
-# summary; the publish itself fails only on CRITICAL vulnerabilities with a
-# released fix, or on any detected secret.
+# Per-architecture tags are pushed first. Both variants then pass a trivy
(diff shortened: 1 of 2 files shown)


naingthet:
Blocked on the trusted base-branch policy job: main still requires Trivy to live in `build`. #2039 relaxes that assertion to accept either shape; once it lands I will rebase this PR.

naingthet:
Rebased onto main after #2039. Trusted policy now accepts the isolated scan job; this PR tightens the assertion to that shape only.

PR #2044, opened 2026-08-13
test: accept signing jobs that isolate installers from Apple material

naingthet:
## Why

CI runs `scripts/workflow-security.test.mjs` from the **base branch** against the PR's workflows, so a pull request cannot weaken a locked invariant and delete the assertion in the same commit. That also blocks a legitimate strengthening: the follow-up moves untrusted installers (floating pnpm, lifecycle scripts, sccache-action, rust-cache) off the desktop signing jobs after Apple/Tauri material is on disk.

## What

Teach the signing-job policy to accept either shape:

- **current:** `pnpm/action-setup` `version: 10`, `pnpm install --frozen-lockfile`, and sccache / rust-cache / UI install after Validate signing
- **next:** pinned `version: 10.18.3`, `--ignore-scripts`, those installer steps before the first Apple/Tauri secret, and `build_macos` rust-cache `save-if: false`

A signing job with no pnpm pin, no lockfile install, or an unconditional rust-cache save still fails. Existing updater-key, scan, and secret-isolation assertions are unchanged.

This is a test-only change. No workflow behavior moves until the implementation PR rebases onto it.

### DIFF
--- a/scripts/workflow-security-mutations.test.mjs
+++ b/scripts/workflow-security-mutations.test.mjs
@@ -50,6 +50,19 @@ function edit(root, path, mutate) {
   writeFileSync(target, after);
 }
 
+function editWorkflowJob(source, name, mutateJob) {
+  const marker = `  ${name}:\n`;
+  const start = source.indexOf(marker);
+  assert.notEqual(start, -1, `missing workflow job: ${name}`);
+  const remainder = source.slice(start + marker.length);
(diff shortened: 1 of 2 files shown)


naingthet:
Superseded by https://github.com/brightwave-inc/tidebreak/pull/2048, which landed first and already includes these dual-shape tests plus the workflow change. Closing rather than rebasing a conflicting precursor over merged work.

PR #2045, opened 2026-08-13
ci: publish sandbox images from main only

naingthet:
## Why

Sandbox image publish could run from any `v*` tag or any-ref `workflow_dispatch`. Unlike E2B template publish, it had no default-branch workflow gate and no `merge-base --is-ancestor` check, so a tag or dispatch of a non-main SHA could still build and push. The `build` checkout also left `persist-credentials` at its default (`true`) and then `COPY . .` with no dockerignore, so `.git/config` (and the job's `GITHUB_TOKEN`) entered the Docker context.

## What

- `resolve` fetches the default branch and refuses the run unless dispatch/schedule execute that branch's workflow, and unless `$GITHUB_SHA` is an ancestor of it (tags included).
- `build` checkout sets `persist-credentials: false`.
- `crates/tidebreak-sandbox-agent/Dockerfile.dockerignore` excludes `.git` and `.git/**` from the workspace-root build context.
- Policy tests now require that shape. A mutation that deletes the ancestor guard fails closed.

Tag push still runs this workflow from the tag (the `on.push.tags` trigger is unchanged). The ancestor check is the minimum that keeps a non-main SHA from publishing; dispatching the main workflow from a tag, the way `release.yml` does, is left for a follow-up if we want the running YAML itself to come only from main.

## Notes

- Depends on the test-only precursor [#2043](https://github.com/brightwave-inc/tidebreak/pull/2043), which is **not yet on main**. Rebase this PR after #2043 merges so the trusted policy lock matches the new shape.
- Not verified by running the publish workflow (main / tag / schedule / dispatch only).
- Production Apple / Tauri / AWS secrets and `E2B_API_KEY` were already isolated from this workflow and stay that way.

Closes nothing; no issue was filed for this session.

### DIFF
--- a/.github/workflows/publish-sandbox-image.yml
+++ b/.github/workflows/publish-sandbox-image.yml
@@ -25,6 +25,12 @@ run-name: >-
 #     vulnerability gate flush through even when nothing in-repo changes;
 #   - manual dispatch with an explicit release-style tag.
 #
+# Publication source is the default branch only. Manual dispatch and the
+# weekly schedule must run this workflow from that branch; every publish
+# (including a v* tag) refuses a SHA that is not an ancestor of it. The
+# build checkout does not persist credentials, and the image dockerignore
+# keeps `.git` out of the `COPY . .` context.
+#
 # Release publishes keep human `vX.Y.Z` tags. Scheduled and main-push rebuilds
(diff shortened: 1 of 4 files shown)


naingthet:
Rebased onto main after #2043. Kept the tightened main-only assertions; dropped the accept-either wrapper.

PR #2051, opened 2026-08-13
ci: publish sandbox images from published releases

naingthet:
## Why

`on.push.tags` loads this workflow from the tagged commit. An attacker (or an old commit) can therefore run a different YAML than the one on `main`. Desktop release already avoids that: `release` events run the default-branch file.

## What

- Trigger cadence is a **published, non-prerelease GitHub Release**, not a raw `v*` tag push.
- `resolve` reads `github.event.release.tag_name`, rejects drafts/prereleases, and still requires the tag commit to be an ancestor of the default branch.
- Main-push rebuilds, the weekly schedule, and manual dispatch are unchanged.

Depends on #2050 (trusted policy must accept this trigger). Rebase after that lands.

Not verified: a live `release.published` run.

### DIFF
--- a/.github/workflows/publish-sandbox-image.yml
+++ b/.github/workflows/publish-sandbox-image.yml
@@ -3,6 +3,8 @@ name: Publish sandbox agent images
 run-name: >-
   ${{ github.event_name == 'workflow_dispatch'
     && format('Publish sandbox images {0}', inputs.image_tag)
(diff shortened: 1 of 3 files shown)


naingthet:
Rebased onto main after #2050. Trusted policy now accepts the release-event trigger.

PR #2074, opened 2026-08-14
fix: harden soft-launch security boundaries

naingthet:
## Summary

- Move computer-use consent, consequential confirmation, resume, and local-command MCP authorization behind OS-native trust boundaries.
- Isolate DOCX rendering in a sanitized, scriptless opaque-origin iframe.
- Enforce immutable principal ownership for local apps, grants, invocation, gateway drafts, view sessions, and connected-app usage counts.
- Scope sticky chat defaults per self-host principal and reject unsafe credential-bearing provider transports.
- Harden release provenance, source-scoped SBOM generation, Docker reproducibility, workflow credential isolation, and release-built static documentation.

## Documentation publication

Decision record 0022 establishes published GitHub Releases as the docs publication boundary. A credential-free runner builds `docs-site/` from the validated release SHA with `BASE_PATH=/docs` and transfers only inert Build Output. A fresh deployment runner sparsely checks out a dedicated Vercel client package, installs its exact frozen dependency graph with lifecycle scripts disabled, deploys without assigning the stable alias, verifies pages, assets, search, sitemap, metadata, security headers, and deployment identity, then promotes.

## Compatibility

- Desktop SQLite advances its pre-v1 schema epoch.
- Existing self-host databases receive an ordered migration that adds and backfills `app.owner`; legacy rows use the non-user `local` owner.
- Credential-bearing configurable provider endpoints require HTTPS. Credentialless Ollama retains loopback HTTP support.
- Enabled local MCP commands now require an absolute executable path; disabled drafts remain editable.
- The published SBOM is explicitly source-scoped and is not represented as installer-content metadata.
- Docs source remains in `docs-site/`; `docs/` remains ordinary Markdown and decision/release documentation.

## Safety and risk

- Stop is linearized with every acting broker dispatch, including initial operations, post-consent reissues, and consequential redemption.
- Native MCP approval displays the complete bounded execution context and exact absolute executable without disclosing secret values.
- Connected-app and gateway usage counts are scoped to the authenticated owner.
- Documentation build code never shares a runner filesystem with the Vercel production credential.
- The credentialed deployment client is installed from a dedicated exact lockfile with scripts disabled.
- A failed docs build or staged smoke test cannot replace the current production deployment.
- Repository settings were not changed.

## Validation

- 11 focused computer-use tests
- 13 native server-info and approval-manifest tests
- Two-principal authenticated gateway-app route regression
- Provider transport regression tests
- Docs build, lint, typecheck, and static Build Output packaging
- Real unpromoted Vercel deployment: pages, assets, search, sitemap, canonical/OG URLs, and security headers
- 70 workflow policy and mutation tests
- `actionlint`

## Review status

All prior P1/P2 findings have been fixed. A fresh final convergence review is running against commit `99ab7125`.

### DIFF
--- a/.github/vercel-cli/package.json
+++ b/.github/vercel-cli/package.json
@@ -1,0 +1,9 @@
+{
+  "name": "tidebreak-vercel-cli",
+  "version": "0.0.0",
+  "private": true,
+  "packageManager": "pnpm@10.18.3",
+  "dependencies": {
+    "vercel": "59.0.0"
+  }
+}
--- a/.github/vercel-cli/pnpm-lock.yaml
+++ b/.github/vercel-cli/pnpm-lock.yaml
@@ -1,0 +1,3037 @@
+lockfileVersion: '9.0'
+
+settings:
+  autoInstallPeers: true
+  excludeLinksFromLockfile: false
+
+importers:
+
+  .:
+    dependencies:
+      vercel:
+        specifier: 59.0.0
+        version: 59.0.0(@emnapi/core@1.11.3)(@emnapi/runtime@1.11.3)
+
+packages:
+
+  '@bytecodealliance/preview2-shim@0.17.6':
+    resolution: {integrity: sha512-n3cM88gTen5980UOBAD6xDcNNL3ocTK8keab21bpx1ONdA+ARj7uD1qoFxOWCyKlkpSi195FH+GeAut7Oc6zZw==}
+
+  '@edge-runtime/format@2.2.1':
+    resolution: {integrity: sha512-JQTRVuiusQLNNLe2W9tnzBlV/GvSVcozLl4XZHk5swnRZ/v6jp8TqR8P7sqmJsQqblDZ3EztcWmLDbhRje/+8g==}
+    engines: {node: '>=16'}
+
+  '@edge-runtime/node-utils@2.3.0':
+    resolution: {integrity: sha512-uUtx8BFoO1hNxtHjp3eqVPC/mWImGb2exOfGjMLUoipuWgjej+f4o/VP4bUI8U40gu7Teogd5VTeZUkGvJSPOQ==}
+    engines: {node: '>=16'}
+
+  '@edge-runtime/ponyfill@2.4.2':
+    resolution: {integrity: sha512-oN17GjFr69chu6sDLvXxdhg0Qe8EZviGSuqzR9qOiKh4MhFYGdBBcqRNzdmYeAdeRzOW2mM9yil4RftUQ7sUOA==}
+    engines: {node: '>=16'}
+
+  '@edge-runtime/primitives@4.1.0':
+    resolution: {integrity: sha512-Vw0lbJ2lvRUqc7/soqygUX216Xb8T3WBZ987oywz6aJqRxcwSVWwr9e+Nqo2m9bxobA9mdbWNNoRY6S9eko1EQ==}
+    engines: {node: '>=16'}
+
+  '@edge-runtime/vm@3.2.0':
+    resolution: {integrity: sha512-0dEVyRLM/lG4gp1R/Ik5bfPl/1wX00xFwd5KcNH602tzBa09oF7pbTKETEhR1GjZ75K6OJnYFu8II2dyMhONMw==}
+    engines: {node: '>=16'}
+
+  '@emnapi/core@1.11.3':
+    resolution: {integrity: sha512-zLpS5asjEb7lq8jYLq37N6XKaE41DIexlY1rF/z4/tIl3wo13Sqm28fRyfIsKZD+NZ8mM5RoKkpW/rBcuoSZSg==}
+
+  '@emnapi/runtime@1.11.3':
+    resolution: {integrity: sha512-Xz4Tpyki7XyrpbUK1jR1AhdAdaXyhhY4lZ3neLodmhpuWfy2PAQN5B46sAiU4liOXGLkHypn/qU+jvfWSCYYLA==}
+
+  '@emnapi/wasi-threads@1.2.3':
+    resolution: {integrity: sha512-ELEBe8PsLvvJ6QMr0zLt8ffvOHW/dc1m3CEzNMg7aJUv3bMaoDtw2TXyDAwkYBuroxxuHEwhRTLJSe5sya547g==}
+
+  '@esbuild/aix-ppc64@0.27.0':
+    resolution: {integrity: sha512-KuZrd2hRjz01y5JK9mEBSD3Vj3mbCvemhT466rSuJYeE/hjuBrHfjjcjMdTm/sz7au+++sdbJZJmuBwQLuw68A==}
+    engines: {node: '>=18'}
+    cpu: [ppc64]
+    os: [aix]
+
+  '@esbuild/android-arm64@0.27.0':
+    resolution: {integrity: sha512-CC3vt4+1xZrs97/PKDkl0yN7w8edvU2vZvAFGD16n9F0Cvniy5qvzRXjfO1l94efczkkQE6g1x0i73Qf5uthOQ==}
+    engines: {node: '>=18'}
+    cpu: [arm64]
+    os: [android]
+
+  '@esbuild/android-arm@0.27.0':
+    resolution: {integrity: sha512-j67aezrPNYWJEOHUNLPj9maeJte7uSMM6gMoxfPC9hOg8N02JuQi/T7ewumf4tNvJadFkvLZMlAq73b9uwdMyQ==}
+    engines: {node: '>=18'}
+    cpu: [arm]
+    os: [android]
+
+  '@esbuild/android-x64@0.27.0':
+    resolution: {integrity: sha512-wurMkF1nmQajBO1+0CJmcN17U4BP6GqNSROP8t0X/Jiw2ltYGLHpEksp9MpoBqkrFR3kv2/te6Sha26k3+yZ9Q==}
+    engines: {node: '>=18'}
+    cpu: [x64]
+    os: [android]
+
+  '@esbuild/darwin-arm64@0.27.0':
+    resolution: {integrity: sha512-uJOQKYCcHhg07DL7i8MzjvS2LaP7W7Pn/7uA0B5S1EnqAirJtbyw4yC5jQ5qcFjHK9l6o/MX9QisBg12kNkdHg==}
+    engines: {node: '>=18'}
+    cpu: [arm64]
+    os: [darwin]
+
+  '@esbuild/darwin-x64@0.27.0':
+    resolution: {integrity: sha512-8mG6arH3yB/4ZXiEnXof5MK72dE6zM9cDvUcPtxhUZsDjESl9JipZYW60C3JGreKCEP+p8P/72r69m4AZGJd5g==}
+    engines: {node: '>=18'}
+    cpu: [x64]
+    os: [darwin]
+
+  '@esbuild/freebsd-arm64@0.27.0':
+    resolution: {integrity: sha512-9FHtyO988CwNMMOE3YIeci+UV+x5Zy8fI2qHNpsEtSF83YPBmE8UWmfYAQg6Ux7Gsmd4FejZqnEUZCMGaNQHQw==}
+    engines: {node: '>=18'}
+    cpu: [arm64]
+    os: [freebsd]
+
+  '@esbuild/freebsd-x64@0.27.0':
+    resolution: {integrity: sha512-zCMeMXI4HS/tXvJz8vWGexpZj2YVtRAihHLk1imZj4efx1BQzN76YFeKqlDr3bUWI26wHwLWPd3rwh6pe4EV7g==}
+    engines: {node: '>=18'}
+    cpu: [x64]
+    os: [freebsd]
+
+  '@esbuild/linux-arm64@0.27.0':
+    resolution: {integrity: sha512-AS18v0V+vZiLJyi/4LphvBE+OIX682Pu7ZYNsdUHyUKSoRwdnOsMf6FDekwoAFKej14WAkOef3zAORJgAtXnlQ==}
+    engines: {node: '>=18'}
+    cpu: [arm64]
+    os: [linux]
+
+  '@esbuild/linux-arm@0.27.0':
+    resolution: {integrity: sha512-t76XLQDpxgmq2cNXKTVEB7O7YMb42atj2Re2Haf45HkaUpjM2J0UuJZDuaGbPbamzZ7bawyGFUkodL+zcE+jvQ==}
+    engines: {node: '>=18'}
+    cpu: [arm]
+    os: [linux]
+
+  '@esbuild/linux-ia32@0.27.0':
+    resolution: {integrity: sha512-Mz1jxqm/kfgKkc/KLHC5qIujMvnnarD9ra1cEcrs7qshTUSksPihGrWHVG5+osAIQ68577Zpww7SGapmzSt4Nw==}
+    engines: {node: '>=18'}
+    cpu: [ia32]
+    os: [linux]
+
+  '@esbuild/linux-loong64@0.27.0':
+    resolution: {integrity: sha512-QbEREjdJeIreIAbdG2hLU1yXm1uu+LTdzoq1KCo4G4pFOLlvIspBm36QrQOar9LFduavoWX2msNFAAAY9j4BDg==}
+    engines: {node: '>=18'}
+    cpu: [loong64]
+    os: [linux]
+
+  '@esbuild/linux-mips64el@0.27.0':
+    resolution: {integrity: sha512-sJz3zRNe4tO2wxvDpH/HYJilb6+2YJxo/ZNbVdtFiKDufzWq4JmKAiHy9iGoLjAV7r/W32VgaHGkk35cUXlNOg==}
+    engines: {node: '>=18'}
+    cpu: [mips64el]
+    os: [linux]
+
+  '@esbuild/linux-ppc64@0.27.0':
+    resolution: {integrity: sha512-z9N10FBD0DCS2dmSABDBb5TLAyF1/ydVb+N4pi88T45efQ/w4ohr/F/QYCkxDPnkhkp6AIpIcQKQ8F0ANoA2JA==}
+    engines: {node: '>=18'}
+    cpu: [ppc64]
+    os: [linux]
+
+  '@esbuild/linux-riscv64@0.27.0':
+    resolution: {integrity: sha512-pQdyAIZ0BWIC5GyvVFn5awDiO14TkT/19FTmFcPdDec94KJ1uZcmFs21Fo8auMXzD4Tt+diXu1LW1gHus9fhFQ==}
+    engines: {node: '>=18'}
+    cpu: [riscv64]
+    os: [linux]
+
+  '@esbuild/linux-s390x@0.27.0':
+    resolution: {integrity: sha512-hPlRWR4eIDDEci953RI1BLZitgi5uqcsjKMxwYfmi4LcwyWo2IcRP+lThVnKjNtk90pLS8nKdroXYOqW+QQH+w==}
+    engines: {node: '>=18'}
+    cpu: [s390x]
+    os: [linux]
+
+  '@esbuild/linux-x64@0.27.0':
+    resolution: {integrity: sha512-1hBWx4OUJE2cab++aVZ7pObD6s+DK4mPGpemtnAORBvb5l/g5xFGk0vc0PjSkrDs0XaXj9yyob3d14XqvnQ4gw==}
+    engines: {node: '>=18'}
+    cpu: [x64]
+    os: [linux]
+
+  '@esbuild/netbsd-arm64@0.27.0':
+    resolution: {integrity: sha512-6m0sfQfxfQfy1qRuecMkJlf1cIzTOgyaeXaiVaaki8/v+WB+U4hc6ik15ZW6TAllRlg/WuQXxWj1jx6C+dfy3w==}
+    engines: {node: '>=18'}
+    cpu: [arm64]
+    os: [netbsd]
+
+  '@esbuild/netbsd-x64@0.27.0':
+    resolution: {integrity: sha512-xbbOdfn06FtcJ9d0ShxxvSn2iUsGd/lgPIO2V3VZIPDbEaIj1/3nBBe1AwuEZKXVXkMmpr6LUAgMkLD/4D2PPA==}
+    engines: {node: '>=18'}
+    cpu: [x64]
+    os: [netbsd]
+
+  '@esbuild/openbsd-arm64@0.27.0':
+    resolution: {integrity: sha512-fWgqR8uNbCQ/GGv0yhzttj6sU/9Z5/Sv/VGU3F5OuXK6J6SlriONKrQ7tNlwBrJZXRYk5jUhuWvF7GYzGguBZQ==}
+    engines: {node: '>=18'}
+    cpu: [arm64]
+    os: [openbsd]
+
+  '@esbuild/openbsd-x64@0.27.0':
+    resolution: {integrity: sha512-aCwlRdSNMNxkGGqQajMUza6uXzR/U0dIl1QmLjPtRbLOx3Gy3otfFu/VjATy4yQzo9yFDGTxYDo1FfAD9oRD2A==}
+    engines: {node: '>=18'}
+    cpu: [x64]
+    os: [openbsd]
+
+  '@esbuild/openharmony-arm64@0.27.0':
+    resolution: {integrity: sha512-nyvsBccxNAsNYz2jVFYwEGuRRomqZ149A39SHWk4hV0jWxKM0hjBPm3AmdxcbHiFLbBSwG6SbpIcUbXjgyECfA==}
+    engines: {node: '>=18'}
+    cpu: [arm64]
+    os: [openharmony]
+
+  '@esbuild/sunos-x64@0.27.0':
+    resolution: {integrity: sha512-Q1KY1iJafM+UX6CFEL+F4HRTgygmEW568YMqDA5UV97AuZSm21b7SXIrRJDwXWPzr8MGr75fUZPV67FdtMHlHA==}
+    engines: {node: '>=18'}
+    cpu: [x64]
+    os: [sunos]
+
+  '@esbuild/win32-arm64@0.27.0':
+    resolution: {integrity: sha512-W1eyGNi6d+8kOmZIwi/EDjrL9nxQIQ0MiGqe/AWc6+IaHloxHSGoeRgDRKHFISThLmsewZ5nHFvGFWdBYlgKPg==}
+    engines: {node: '>=18'}
+    cpu: [arm64]
+    os: [win32]
+
+  '@esbuild/win32-ia32@0.27.0':
+    resolution: {integrity: sha512-30z1aKL9h22kQhilnYkORFYt+3wp7yZsHWus+wSKAJR8JtdfI76LJ4SBdMsCopTR3z/ORqVu5L1vtnHZWVj4cQ==}
+    engines: {node: '>=18'}
+    cpu: [ia32]
+    os: [win32]
+
+  '@esbuild/win32-x64@0.27.0':
+    resolution: {integrity: sha512-aIitBcjQeyOhMTImhLZmtxfdOcuNRpwlPNmlFKPcHQYPhEssw75Cl1TSXJXpMkzaua9FUetx/4OQKq7eJul5Cg==}
+    engines: {node: '>=18'}
+    cpu: [x64]
+    os: [win32]
+
+  '@fastify/busboy@2.1.1':
+    resolution: {integrity: sha512-vBZP4NlzfOlerQTnba4aqZoMhE/a9HY7HRqoOPaETQcSQuWEIyZMHGfVu6w9wGtGK5fED5qRs2DteVCjOH60sA==}
+    engines: {node: '>=14'}
+
+  '@isaacs/balanced-match@4.0.1':
+    resolution: {integrity: sha512-yzMTt9lEb8Gv7zRioUilSglI0c0smZ9k5D65677DLWLtWJaXIS3CqcGyUFByYKlnUj6TkjLVs54fBl6+TiGQDQ==}
+    engines: {node: 20 || >=22}
+
+  '@isaacs/brace-expansion@5.0.1':
+    resolution: {integrity: sha512-WMz71T1JS624nWj2n2fnYAuPovhv7EUhk69R6i9dsVyzxt5eM3bjwvgk9L+APE1TRscGysAVMANkB0jh0LQZrQ==}
+    engines: {node: 20 || >=22}
+
+  '@isaacs/fs-minipass@4.0.1':
+    resolution: {integrity: sha512-wgm9Ehl2jpeqP3zw/7mo3kRHFp5MEDhqAdwy1fTGkHAwnkGOVsgpvQhL8B5n1qlb01jV3n/bI0ZfZp5lWA1k4w==}
+    engines: {node: '>=18.0.0'}
+
+  '@mapbox/node-pre-gyp@2.0.3':
+    resolution: {integrity: sha512-uwPAhccfFJlsfCxMYTwOdVfOz3xqyj8xYL3zJj8f0pb30tLohnnFPhLuqp4/qoEz8sNxe4SESZedcBojRefIzg==}
+    engines: {node: '>=18'}
+    hasBin: true
+
+  '@napi-rs/keyring-darwin-arm64@1.2.0':
+    resolution: {integrity: sha512-CA83rDeyONDADO25JLZsh3eHY8yTEtm/RS6ecPsY+1v+dSawzT9GywBMu2r6uOp1IEhQs/xAfxgybGAFr17lSA==}
+    engines: {node: '>= 10'}
+    cpu: [arm64]
+    os: [darwin]
+
+  '@napi-rs/keyring-darwin-x64@1.2.0':
+    resolution: {integrity: sha512-dBHjtKRCj4ByfnfqIKIJLo3wueQNJhLRyuxtX/rR4K/XtcS7VLlRD01XXizjpre54vpmObj63w+ZpHG+mGM8uA==}
+    engines: {node: '>= 10'}
+    cpu: [x64]
+    os: [darwin]
+
+  '@napi-rs/keyring-freebsd-x64@1.2.0':
+    resolution: {integrity: sha512-DPZFr11pNJSnaoh0dzSUNF+T6ORhy3CkzUT3uGixbA71cAOPJ24iG8e8QrLOkuC/StWrAku3gBnth2XMWOcR3Q==}
+    engines: {node: '>= 10'}
+    cpu: [x64]
+    os: [freebsd]
+
+  '@napi-rs/keyring-linux-arm-gnueabihf@1.2.0':
+    resolution: {integrity: sha512-8xv6DyEMlvRdqJzp4F39RLUmmTQsLcGYYv/3eIfZNZN1O5257tHxTrFYqAsny659rJJK2EKeSa7PhrSibQqRWQ==}
+    engines: {node: '>= 10'}
+    cpu: [arm]
+    os: [linux]
+
+  '@napi-rs/keyring-linux-arm64-gnu@1.2.0':
+    resolution: {integrity: sha512-Pu2V6Py+PBt7inryEecirl+t+ti8bhZphjP+W68iVaXHUxLdWmkgL9KI1VkbRHbx5k8K5Tew9OP218YfmVguIA==}
+    engines: {node: '>= 10'}
+    cpu: [arm64]
+    os: [linux]
+
+  '@napi-rs/keyring-linux-arm64-musl@1.2.0':
+    resolution: {integrity: sha512-8TDymrpC4P1a9iDEaegT7RnrkmrJN5eNZh3Im3UEV5PPYGtrb82CRxsuFohthCWQW81O483u1bu+25+XA4nKUw==}
+    engines: {node: '>= 10'}
+    cpu: [arm64]
+    os: [linux]
+
+  '@napi-rs/keyring-linux-riscv64-gnu@1.2.0':
+    resolution: {integrity: sha512-awsB5XI1MYL7fwfjMDGmKOWvNgJEO7mM7iVEMS0fO39f0kVJnOSjlu7RHcXAF0LOx+0VfF3oxbWqJmZbvRCRHw==}
+    engines: {node: '>= 10'}
+    cpu: [riscv64]
+    os: [linux]
+
+  '@napi-rs/keyring-linux-x64-gnu@1.2.0':
+    resolution: {integrity: sha512-8E+7z4tbxSJXxIBqA+vfB1CGajpCDRyTyqXkBig5NtASrv4YXcntSo96Iah2QDR5zD3dSTsmbqJudcj9rKKuHQ==}
+    engines: {node: '>= 10'}
+    cpu: [x64]
+    os: [linux]
+
+  '@napi-rs/keyring-linux-x64-musl@1.2.0':
+    resolution: {integrity: sha512-8RZ8yVEnmWr/3BxKgBSzmgntI7lNEsY7xouNfOsQkuVAiCNmxzJwETspzK3PQ2FHtDxgz5vHQDEBVGMyM4hUHA==}
+    engines: {node: '>= 10'}
+    cpu: [x64]
+    os: [linux]
+
+  '@napi-rs/keyring-win32-arm64-msvc@1.2.0':
+    resolution: {integrity: sha512-AoqaDZpQ6KPE19VBLpxyORcp+yWmHI9Xs9Oo0PJ4mfHma4nFSLVdhAubJCxdlNptHe5va7ghGCHj3L9Akiv4cQ==}
+    engines: {node: '>= 10'}
+    cpu: [arm64]
+    os: [win32]
+
+  '@napi-rs/keyring-win32-ia32-msvc@1.2.0':
+    resolution: {integrity: sha512-EYL+EEI6bCsYi3LfwcQdnX3P/R76ENKNn+3PmpGheBsUFLuh0gQuP7aMVHM4rTw6UVe+L3vCLZSptq/oeacz0A==}
+    engines: {node: '>= 10'}
+    cpu: [ia32]
+    os: [win32]
+
+  '@napi-rs/keyring-win32-x64-msvc@1.2.0':
+    resolution: {integrity: sha512-xFlx/TsmqmCwNU9v+AVnEJgoEAlBYgzFF5Ihz1rMpPAt4qQWWkMd4sCyM1gMJ1A/GnRqRegDiQpwaxGUHFtFbA==}
+    engines: {node: '>= 10'}
+    cpu: [x64]
+    os: [win32]
+
+  '@napi-rs/keyring@1.2.0':
+    resolution: {integrity: sha512-d0d4Oyxm+v980PEq1ZH2PmS6cvpMIRc17eYpiU47KgW+lzxklMu6+HOEOPmxrpnF/XQZ0+Q78I2mgMhbIIo/dg==}
+    engines: {node: '>= 10'}
+
+  '@napi-rs/wasm-runtime@1.2.3':
+    resolution: {integrity: sha512-UMduMbqO5s5zF2NkNacMT/yK5Y5QiKvWr2+50bzIIxFDwVJ2h49b+oyjaCGPhJxd2/gC2x39EHv/gHVuu36x2Q==}
+    engines: {node: ^20.19.0 || ^22.13.0 || >=23.5.0}
+    peerDependencies:
+      '@emnapi/core': ^1.7.1 || ^2.0.0-alpha.4
+      '@emnapi/runtime': ^1.7.1 || ^2.0.0-alpha.4
+
+  '@nodelib/fs.scandir@2.1.5':
+    resolution: {integrity: sha512-vq24Bq3ym5HEQm2NKCr3yXDwjc7vTsEThRDnkp2DK9p1uqLR+DHurm/NOTo0KG7HYHU7eppKZj3MyqYuMBf62g==}
+    engines: {node: '>= 8'}
+
+  '@nodelib/fs.stat@2.0.5':
+    resolution: {integrity: sha512-RkhPPp2zrqDAQA/2jNhnztcPAlv64XdhIp7a7454A5ovI7Bukxgt7MX7udwAu3zg1DcpPU0rz3VV1SeaqvY4+A==}
+    engines: {node: '>= 8'}
+
+  '@nodelib/fs.walk@1.2.8':
+    resolution: {integrity: sha512-oGB+UxlgWcgQkgwo8GcEGwemoTFt3FIO9ababBmaGwXIoBKZ+GTy0pP185beGg7Llih/NSHSV2XAs1lnznocSg==}
+    engines: {node: '>= 8'}
+
+  '@oxc-project/types@0.110.0':
+    resolution: {integrity: sha512-6Ct21OIlrEnFEJk5LT4e63pk3btsI6/TusD/GStLi7wYlGJNOl1GI9qvXAnRAxQU9zqA2Oz+UwhfTOU2rPZVow==}
+
+  '@oxc-transform/binding-android-arm-eabi@0.111.0':
+    resolution: {integrity: sha512-NdFLicvorfHYu0g2ftjVJaH7+Dz27AQUNJOq8t/ofRUoWmczOodgUCHx8C1M1htCN4ZmhS/FzfSy6yd/UngJGg==}
+    engines: {node: ^20.19.0 || >=22.12.0}
+    cpu: [arm]
+    os: [android]
+
+  '@oxc-transform/binding-android-arm64@0.111.0':
+    resolution: {integrity: sha512-J2v9ajarD2FYlhHtjbgZUFsS2Kvi27pPxDWLGCy7i8tO60xBoozX9/ktSgbiE/QsxKaUhfv4zVKppKWUo71PmQ==}
+    engines: {node: ^20.19.0 || >=22.12.0}
+    cpu: [arm64]
+    os: [android]
+
+  '@oxc-transform/binding-darwin-arm64@0.111.0':
+    resolution: {integrity: sha512-2UYmExxpXzmiHTldhNlosWqG9Nc4US51K0GB9RLcGlTE23WO33vVo1NVAKwxPE+KYuhffwDnRYTovTMUjzwvZA==}
+    engines: {node: ^20.19.0 || >=22.12.0}
+    cpu: [arm64]
+    os: [darwin]
+
+  '@oxc-transform/binding-darwin-x64@0.111.0':
+    resolution: {integrity: sha512-c4YRwfLV8Pj/ToiTCbndZaHxM2BD4W3bltr/fjXZcGypEK+U2RZFDL7tIZYT/tyneAC9hCORZKDaKhLLNuzPtA==}
+    engines: {node: ^20.19.0 || >=22.12.0}
+    cpu: [x64]
+    os: [darwin]
+
+  '@oxc-transform/binding-freebsd-x64@0.111.0':
+    resolution: {integrity: sha512-prvf32IcEuLnLZbNVomFosBu0CaZpyj3YsZ6epbOgJy8iJjfLsXBb+PrkO/NBKzjuJoJa2+u7jFKRE0KT7gSOw==}
+    engines: {node: ^20.19.0 || >=22.12.0}
+    cpu: [x64]
+    os: [freebsd]
+
+  '@oxc-transform/binding-linux-arm-gnueabihf@0.111.0':
+    resolution: {integrity: sha512-+se3579Wp7VOk8TnTZCpT+obTAyzOw2b/UuoM0+51LtbzCSfjKxd4A+o7zRl7GyPrPZvx57KdbMOC9rWB1xNrw==}
+    engines: {node: ^20.19.0 || >=22.12.0}
+    cpu: [arm]
+    os: [linux]
+
+  '@oxc-transform/binding-linux-arm-musleabihf@0.111.0':
+    resolution: {integrity: sha512-8faC99pStqaSDPK/vBgaagAHUeL0LcIzfeSjSiDTtvPGc3AwZIeqC1tx3CP15a6tWXjdgS/IUw4IjfD5HweBlg==}
+    engines: {node: ^20.19.0 || >=22.12.0}
+    cpu: [arm]
+    os: [linux]
+
+  '@oxc-transform/binding-linux-arm64-gnu@0.111.0':
+    resolution: {integrity: sha512-HtfQv8j796gzI5WR/RaP6IMwFpiL0vYeDrUA1hYhlPzTHKYan/B+NlhJkKOI1v24yAl/yEnFmb0pxIxLNqBqBA==}
+    engines: {node: ^20.19.0 || >=22.12.0}
+    cpu: [arm64]
+    os: [linux]
+
+  '@oxc-transform/binding-linux-arm64-musl@0.111.0':
+    resolution: {integrity: sha512-ARyfcMCIxVLDgLf6FQ8Oo1/TFySpnquV+vuSb4SFQZfYDqgMklzwv0NYXxWD0aB6enElyMDs6pQJBzusEKCkOg==}
+    engines: {node: ^20.19.0 || >=22.12.0}
+    cpu: [arm64]
+    os: [linux]
+
+  '@oxc-transform/binding-linux-ppc64-gnu@0.111.0':
+    resolution: {integrity: sha512-PKpVRrSvBNK3tv9vwxn7Fay+QWZmprPGlEqJcseBJllQc5mFMD4Q/w44chu5iR9ZLsDeSHzmNWrgMLo4J0sP2A==}
+    engines: {node: ^20.19.0 || >=22.12.0}
+    cpu: [ppc64]
+    os: [linux]
+
+  '@oxc-transform/binding-linux-riscv64-gnu@0.111.0':
+    resolution: {integrity: sha512-9bUml6rMgk+8GF5rvNMweFspkzSiCjqpV6HduwiUyexqfGKrmjq9IZOxxvnzkE2RGdQzP507NNDoVNYIoGQYuA==}
+    engines: {node: ^20.19.0 || >=22.12.0}
+    cpu: [riscv64]
+    os: [linux]
+
+  '@oxc-transform/binding-linux-riscv64-musl@0.111.0':
+    resolution: {integrity: sha512-tzGCohGxaeH6KRJjfYZd4mHCoGjCai6N+zZi1Oj+tSDMAAdyvs1dRzYb8PNUGnybCg3Te4M0jLPzWZaSmnKraQ==}
+    engines: {node: ^20.19.0 || >=22.12.0}
+    cpu: [riscv64]
+    os: [linux]
+
+  '@oxc-transform/binding-linux-s390x-gnu@0.111.0':
+    resolution: {integrity: sha512-sRG1KIfZ0ML9ToEygm5aM/5GJeBA05uHlgW3M0Rx/DNWMJhuahLmqWuB02aWSmijndLfEKXLLXIWhvWupRG8lg==}
+    engines: {node: ^20.19.0 || >=22.12.0}
+    cpu: [s390x]
+    os: [linux]
+
+  '@oxc-transform/binding-linux-x64-gnu@0.111.0':
+    resolution: {integrity: sha512-T0Kmvk+OdlUdABdXlDIf3MQReMzFfC75NEI9x8jxy5pKooACEFg0k0V8gyR3gq4DzbDCfucqFQDWNvSgIopAbQ==}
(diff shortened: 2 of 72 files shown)


naingthet (commented):

**Changes suggested** — the ownership and transport hardening is solid and well-tested; one release-blocking CI bug and one broken user flow, plus a residual scoping gap the diff misses.

---

### P1 — `.github/workflows/release.yml:1105` — the SBOM step fails on every release; its output directory is never created

`output-file: source-sbom/Tidebreak_..._source.spdx.json`, but nothing creates `source-sbom/`. The job's only checkout uses `path: release-source`, so `$GITHUB_WORKSPACE/source-sbom/` does not exist. I checked out `anchore/sbom-action` at the pinned SHA `e22c389`: `src/github/SyftGithubAction.ts:416-419` is a bare `fs.writeFileSync(outputFile, output)`, and there is no `mkdir`/`ensureDir` anywhere under `src/`. The write throws `ENOENT`.

Failure scenario: dispatch `release.yml` for any tag not already hosted (`inspect_hosted.outputs.exists != 'true'` — i.e. every first attempt). `build_macos` runs to completion *including Apple signing and notarization*, then `source_sbom` fails; `publish` requires `needs.source_sbom.result == 'success'` (`:1143`) so it is skipped, and `attach_downloads` never runs. The release is blocked after the expensive, rate-limited work is already spent. The new tests don't catch it because `workflow-security.test.mjs:1373-1409` only assert that `output-file`-adjacent strings are present, never that the path is writable.

Fix — one step before the sbom-action step:

```yaml
- name: Create the SBOM output directory
  run: mkdir -p source-sbom
```

### P2 — `crates/tidebreak-desktop/ui/src/document/DocxViewer.tsx:158,223` — DOCX hyperlinks are dead in the packaged app

The move into the sandboxed frame dropped the mechanism that made links work. `main`'s version imported `openExternal` from `@/host` and intercepted clicks in a capture-phase listener (`DocxViewer.tsx:7,125,134` on `main`), routing through the Tauri shell plugin. The new version keeps only `anchor.target = "_blank"` (`:223`) inside a frame with `sandbox="allow-popups"` (`:158`). The repo's own note at `crates/tidebreak-desktop/ui/src/host.ts:181-184` says exactly why that cannot work:

> The webview swallows `window.open` and `target="_blank"` (no new-window handler by design)

I confirmed no `new_window`/navigation handler is registered anywhere under `crates/tidebreak-desktop/src/`.

Failure scenario: open a DOCX containing an `https://` hyperlink in the desktop build and click it — nothing happens, no feedback. In a browser-served build the popup opens but inherits the sandbox flags (no `allow-popups-to-escape-sandbox`), so the destination loads scriptless in an opaque origin and looks broken. Either way `allow-popups` buys nothing and only loosens the sandbox.

The frame is scriptless and cross-origin, so the parent cannot re-add click interception — this needs a decision, not a flag. Either drop `allow-popups`, strip `href` in the sanitizer, and surface the destination as visible text/`title` so it can be copied; or keep the old in-document render path for links only. (Reasoned + verified against `main` and the Rust source; not exercised in a packaged build.)

### P3 — `crates/tidebreak-server/src/routes/connected_apps.rs:514` and `crates/tidebreak-server/src/gateway_runtime.rs:313` — `used_by` counts are still unscoped

**Reproduced.** Both call the unscoped `store.list_live_app_grants()`, which resolves to `effective_owner(None)` → the `local` sentinel (`crates/tidebreak-core/src/db/ops/app.rs:648`). `GET /connected-apps` (`lib.rs:475`) and `GET /gateway/apps` (`lib.rs:517`) are both mounted *outside* the `require_admin` layer that closes at `lib.rs:446`, so any authenticated self-host member reaches them.

I reproduced it with a temporary test alongside `another_principal_cannot_reach_an_app_through_any_app_route`: the `local` owner owns two apps binding a `rest_api` connector, `user:alice` owns one. Alice's `GET /connected-apps` returns `used_by_app_count: 2` — the local owner's count, not her own. Two bugs in one: a small cross-principal aggregate leak, and Alice's own count is never correct (a self-host user's real count is always reported as the local owner's).

The tell is that `list_live_app_grants_scoped` was added in this PR and has **zero non-test callers**. Fix is mechanical: thread the request principal / `store.owner_id()` into `connected_apps_info` and `GatewayRuntime::apps()` and call the scoped variant.

### P3 — `crates/tidebreak-router/src/router.rs:485,553,590` — the new router-layer validator covers only two of the kinds that accept a configurable `base_url`

`base_url_is_allowed` is applied only on the `Ollama | OpenaiCompatible` arm (`:571`). The `Anthropic` (`:485`), `Openai` (`:553`), and `Gemini` (`:590`) arms still apply `route.base_url` with no scheme check at all — and `update_provider` accepts a configured `base_url` for Anthropic and OpenAI (only Xai and `has_fixed_endpoint` kinds are refused, `providers.rs:1526-1535`).

Not currently exploitable: `collect_routes` is the only production `Route` constructor, and it now validates those kinds server-side, so this is defense-in-depth only. It also pre-exists on `main` — the diff didn't introduce it. But the PR adds the validator and a `RouteKind::Xai` comment explicitly hardening against "even a directly constructed route override", and the same reasoning applies to Anthropic/OpenAI/Gemini. Cheap to close: run `base_url_is_allowed` on every arm that applies `route.base_url`.

### P3 — `crates/tidebreak-desktop/ui/src/document/DocxViewer.tsx:104-107` — docx-preview's post-resolve layout work is discarded

`renderAsync` resolves after `Promise.allSettled(this.tasks)`, but two pieces of sizing run later: `refreshTabStops()` on a `setTimeout(…, 500)` (active because `experimental: true`), and VML picture sizing via `getBBox()` inside a `requestAnimationFrame`. The old code moved nodes into the live container synchronously so both fired against attached, laid-out nodes; the new code snapshots `staging.innerHTML` synchronously and never attaches the staging node, so both run on a detached tree and their output is dropped. Effect: `w:pict` content (Word text boxes, legacy shapes, WordArt) gets an `<svg>` with no `width`/`height`, and tab-stop alignment (TOC dot leaders, tabbed columns) is never applied. Await a frame plus the tab-stop delay before serializing, or set `experimental: false` and size VML from the DOCX extents. (Reasoned from docx-preview 0.4.0 source; not reproduced with a real document.)

---

### CHECKED

Verified correct — please don't re-litigate these next round.

**App principal ownership.** Every app route takes `ScopedStore`: library list/detail/delete, grant get/post/delete (incl. `current_live_app`), invoke (`current_app_revision`, `require_app_grant`), gateway-page, view-session. Gateway drafts (`current`/`persist`/`ensure_registered`/`relay_consent`) all take `&OwnerId` and every caller passes `store.owner_id()`. `tools/create_app.rs` uses the `*_for_chat` variants, and the ownership check precedes `publish_app_bundle`, so a foreign `app_id` cannot plant a bundle. No code outside `db/ops/app.rs` touches `entities::app*` directly. Unscoped `Store` variants resolve to the `local` sentinel rather than "any owner", which caps the blast radius of any missed caller — the two above included. `tests::app_library` passes (2/2), including the two-principal route test.

**View frames.** `get_app_view_frame` / `get_mcp_view_frame` sit outside the bearer router deliberately; the token is single-use (`view_frames.take`), unguessable, short-lived, and minted only after the scoped `get_app`/`get_app_revision` check. Bundle path derives from ids baked into the token, not user input.

**Provider transport.** `validate_base_url_transport` runs on write for *all* kinds and fires on `enabled || base_url_changed || credential_changed`, so both orderings are covered (cleartext-then-credential is rejected at the credential write). Validation precedes secret storage — asserted by the new test. Gemini/Xai are exempt at read time only because `collect_routes` forces their `Route.base_url` to `None`, so their keys can never be redirected; Xai additionally refuses any `base_url` update. `env_api_key` returns `None` for Ollama, so the credentialless-loopback carve-out can't be widened by an env var, and `has_credential` consults env for the kinds where it matters. Userinfo/fragment in the base URL are rejected. The `#[cfg(test)]` loopback seam is assertion-guarded and cannot exist in production builds.

**Local-command MCP native gate.** `client_executor` is set only by `require_client_executor_token` middleware on a constant-time header-token compare, on the `client_executor_api` sub-router — the renderer cannot forge it, and the extractors fail closed with no middleware (`principal.rs` tests). `command.is_some()` is the complete predicate for local process spawn from a client body: `launch` is `#[serde(skip)]`, `plugin` is refused outright, and plugin-sourced entries are rebuilt by the runtime. No enable-without-native bypass: `runtime.rs:1236` refuses to reconnect a disabled server, and the only other `replace(McpServersConfig)` sites (gateway auto-mount, pairing) hardcode `command: None`. Storing a *disabled* command draft without native confirmation matches the stated design.

**Sticky chat defaults.** `sticky_default_key` keeps the historical key for `is_local()` so desktop defaults survive upgrade and gives named users disjoint keys; read and write paths are symmetric. `PUT /settings` is inside the `require_admin` layer; `GET /settings` is member-reachable and now returns owner-scoped `chat_defaults`.

**DOCX sandbox.** `sandbox="allow-popups"` only — no `allow-same-origin`, no `allow-scripts`, no `allow-top-navigation`; the dangerous scripts+same-origin pair appears nowhere. Sanitizer drops `script/iframe/object/embed/meta/base/link/form/input/…`, strips `on*` (lowercased first, which matters for `setAttribute`-built SVG), `srcdoc`, `srcset`, `action`, `formaction`, `ping`; `src` allowlisted to `data:`/`blob:`; `href`/`xlink:href` require `https:` via URL parsing. The `</style` neutralization is correct and necessary — `<style>` is the only HTML-namespace raw-text element kept, and the `noframes` gap is unreachable (docx-preview's only tag-name-controlled path uses `createElementNS(svgNS, …)`). Frame CSP is tight (`default-src 'none'`, `connect-src 'none'`, `form-action 'none'`, `base-uri 'none'`), closing CSS-based exfiltration. `srcDoc` is set via `setAttribute`, so no second decoding pass. DocxViewer dom tests pass (2/2) — though note they run in jsdom, which parses neither `srcdoc` nor CSP, so every assertion is a substring check on the string; one manual load of a real DOCX in the packaged WKWebView build is still worth doing before merge.

**Release/CI.** `source_sbom` has `contents: read` only — no `id-token`, no environment, no AWS vars, `persist-credentials: false`, checkout pinned to `needs.validate.outputs.sha`. `anchore/sbom-action` no longer runs in the OIDC-bearing `publish` job, and five mutations in `workflow-security-mutations.test.mjs` fail if any of that regresses. `actions/attest` is ordered before `Configure AWS credentials`, and that ordering is itself asserted. All third-party actions are full-SHA pinned. No `pull_request_target`; no secrets reach the SBOM job. No script injection: new `run:` blocks use `$TIDEBREAK_VERSION` via `env:`, and `version` comes from a strict semver regex in `check-release-tag.mjs:10-13`. `docs/releases.md` claims match the workflow, including the source-scoped caveat (enforced by `assert.doesNotMatch(publishJob, /sbom-path:/)`). Workflow tests: 64 pass, 0 fail, 2 skipped (Docker-context probes, which CI runs).

**Dockerfile.** `deploy/self-host/Dockerfile:51-64` is sound — snapshot archives are immutable and their `Release` files stay GPG-signed, so `http://snapshot.debian.org` plus `[check-valid-until=no]` does not weaken authenticity; apt still verifies signatures and `check-valid-until` only suppresses the freshness check a dated snapshot inherently fails. Digest-pinned base retained. Worth knowing: nothing flags a stale snapshot date, so a rebuild a year out silently ships the pinned `curl` with any later CVEs — `docs/self-hosting.md:222-227` already frames the bump as deliberate maintenance, which is the right policy.

naingthet (commented):
🤖 **Thet's review agent** — automated review (round 2), not a human sign-off.

**Changes still needed** — the `used_by` scoping fix is correct and complete, but the round-1 P1 (release-blocking) and P2 (dead DOCX hyperlinks) are untouched, and the new commits add two of their own: the Vercel CLI's dependency tree floats inside the only job holding the deployment token, and Stop now takes a lock before setting its latch so the Stop button can appear dead for ~35s.

---

## Status of the round-1 findings

| # | Finding | Status |
|---|---|---|
| P1 | `release.yml` SBOM output dir never created | **Not addressed** |
| P2 | `DocxViewer.tsx` dead hyperlinks | **Not addressed** |
| P3 | `used_by` counts unscoped | **Fixed** — correctly and completely |
| P3 | `router.rs` validator covers only 2 kinds | **Not addressed** |
| P3 | docx-preview post-resolve layout discarded | **Not addressed** |

### P1 — not addressed. `.github/workflows/release.yml:1323`

The `source_sbom` job is unchanged in the relevant respect: `output-file: source-sbom/Tidebreak_…_source.spdx.json` with no step creating `source-sbom/`, and the job's only checkout still uses `path: release-source`.

I re-derived this from scratch rather than trusting round 1 — cloned `anchore/sbom-action`, checked out the pinned `e22c389904149dbc22b58101806040fa8d37a610`, and confirmed `src/github/SyftGithubAction.ts:418` is a bare `fs.writeFileSync(outputFile, output)`. `grep -rn "mkdir\|ensureDir" src/` returns exactly one hit — `path.dirname` in `GithubClient.ts:176`, on an unrelated path — so the action creates no directory. The write throws `ENOENT`.

Impact is unchanged and it is the merge blocker: on any first-attempt release (`inspect_hosted.outputs.exists != 'true'`), `build_macos` completes *including Apple signing and notarization*, then `source_sbom` fails, `publish` is skipped on `needs.source_sbom.result == 'success'`, and `attach_downloads` never runs.

Fix, one step before the sbom-action step:
```yaml
- name: Create the SBOM output directory
  run: mkdir -p source-sbom
```

### P2 — not addressed. `crates/tidebreak-desktop/ui/src/document/DocxViewer.tsx:158,223`

`git diff 84ea1041..9550721c -- crates/tidebreak-desktop/ui/` is empty — no UI file changed. Re-verified the three facts at head independently: `DocxViewer.tsx:158` is still `sandbox="allow-popups"`, `:223` still `anchor.target = "_blank"`, no `openExternal` import remains, `crates/tidebreak-desktop/ui/src/host.ts:180-183` still documents that the webview swallows `window.open` and `target="_blank"`, and `grep -rn "new_window\|on_navigation" crates/tidebreak-desktop/src/` returns nothing. DOCX hyperlinks are still silently dead in the packaged app, and `allow-popups` still loosens the sandbox for nothing.

Still needs the decision from round 1, not a flag: either drop `allow-popups`, strip `href` in the sanitizer, and surface the destination as copyable text/`title`; or keep the in-document render path for links only.

### P3 `used_by` — fixed, and fixed properly

The fix threads an owner through rather than patching the two named call sites. `GatewayRuntime::apps` now takes `&OwnerId` (`gateway_runtime.rs:298`), so the compiler forced every caller; `connected_apps_info` takes `&OwnerId` (`connected_apps.rs:502`); both reach `list_live_app_grants_scoped`.

Verified completeness rather than taking it on faith: `grep -rn "list_live_app_grants" crates/ --include="*.rs"` shows the unscoped variant now has **zero** production call sites — only the trait/impl definitions in `db/mod.rs`, `storage/store.rs`, and `db/ops/app.rs`. Every remaining `state.gateway.apps(` / `list_live_app_grants_scoped(` call passes an owner.

No fresh instance of the bug class. I checked the specific risk that scoping broke the desktop flow: `Principal::owner_id()` (`principal.rs:80-86`) maps `LocalOwner → OwnerId::local()`, the same sentinel desktop apps are created under, so single-user counts are unchanged. `PUT /connected-apps/rest/{id}` (`lib.rs:402`) sits inside the `require_admin` layer and also got the owner threaded, consistently.

The new test is a real route-level test, not a unit stub — it drives `GET /gateway/apps` through the router with two distinct `Principal::User` extensions and asserts alice=1 / bob=0. It passes:
```
test gateway_runtime::tests::gateway_apps_route_scopes_used_by_count_to_the_authenticated_principal ... ok
```

One note, not a defect: `used_by_app_count` is display-only (`ConnectedAppsPanel.tsx:200`, `GatewayPanel.tsx:374`) and `delete_rest_connected_app` is not gated on it, so an admin deleting a connected app still gets no warning that *other* members' apps bind it. That was equally true before this PR (the `local` sentinel count was also 0 on a self-host deployment), so it is pre-existing and cosmetic — flagging only so it is a known gap rather than an assumed guard.

### P3 router validator — not addressed

`router.rs` changed by exactly one line: `reqwest::Url::parse` → `url::Url::parse` in `base_url_is_allowed`. That change is sound and worth having — `Cargo.lock` carries a single `url 2.5.8`, workspace-pinned `=2.5.8`, and reqwest re-exports that same crate, so parsing behavior is byte-identical while the validator no longer depends on the optional `reqwest` feature.

But the gap itself is unchanged: `base_url_is_allowed` is still applied only on the `Ollama | OpenaiCompatible` arm (`router.rs:571`), while `Anthropic` (`:485`), `Openai` (`:553`), and `Gemini` (`:590`) still call `with_base_url(base.clone())` with no scheme check. Unchanged from round 1: defense-in-depth only, covered server-side by `collect_routes`, and pre-existing on `main`.

### P3 docx-preview layout — not addressed (no UI file changed)

---

## New findings

### P2 — `.github/workflows/release.yml:205` — the only job holding `VERCEL_TOKEN` installs an unlockfiled dependency tree at deploy time

`publish_docs` deliberately has no `actions/checkout` — that isolation is the point of the PR and is asserted at `scripts/workflow-security.test.mjs:1069`. But with no checkout there is no lockfile, and the install step is:

```yaml
- name: Install the pinned deployment client
  run: pnpm add --global vercel@59.0.0 --ignore-scripts
```

That pins only the top-level package. Every transitive dependency of `vercel@59.0.0` resolves against its semver range at deploy time. `--ignore-scripts` blocks install hooks, but the CLI is then *executed* five times in this job — so any compromised transitive dep runs in-process where `VERCEL_TOKEN` is in `env`.

The pinning is exactly inverted, which is the tell that this is an oversight rather than a decision: the fully-locked `vercel` devDependency was added to `docs-site/package.json:47` — consumed by `build_docs`, the job with **no** secrets, where nothing imports or shells out to it — while the credentialed job floats. The ADR added in this same PR states the intended property at `docs/decisions/0022-release-tags-publish-static-documentation.md:46-47`: *"Documentation build dependencies and the Vercel CLI are exact-version lockfile inputs."* Only the first half is true today.

Failure scenario: a maintainer of any transitive dep of `vercel@59.0.0` publishes a patch release with an exfiltration payload in module top-level code. The next release run installs it, `vercel deploy` loads it, and the production docs deployment token leaves the runner. No repo change is needed to trigger it.

Fix — the repo already has the right pattern. `.github/e2b-cli/` is a committed `package.json` + `pnpm-lock.yaml` consumed at `.github/workflows/publish-e2b-template.yml:376-378` via `pnpm --dir .github/e2b-cli install --frozen-lockfile --ignore-scripts`, then invoked from `node_modules/.bin/`. Mirror it as `.github/vercel-cli/`, and pull it into `publish_docs` with a sparse checkout limited to that path (`sparse-checkout: .github/vercel-cli`, `persist-credentials: false`) — that restores the pin without bringing `docs-site/` source or any build step into the credentialed job, so the isolation property is preserved. Then drop the unused `vercel` devDependency from `docs-site/package.json` (it also adds ~2.5k lockfile lines to every docs PR).

*Reproduced by reading the workflow, not by executing a release.*

### P2 — `crates/tidebreak-desktop/src/client_execution/computer_use.rs:276-281` — Stop takes the acting gate *before* setting the latch, so the Stop button is dead for up to ~35s

```rust
async fn halt(&self) {
    let _dispatch = self.acting_dispatch.lock().await;
    self.halt.send_replace(true);
}
```

The lock is acquired before the latch is set, and `dispatch_acting` (`:283-293`) holds that same gate across the entire broker round-trip. That round-trip is bounded by `REQUEST_TIMEOUT = 30s` plus `Session::start`'s `HELLO_TIMEOUT = 5s` (`crates/tidebreak-desktop/src/broker.rs:27-28`).

Failure scenario: the agent issues an acting op (a click or a keystroke) against a broker that is slow or hung, and the user presses Stop. `stop_computer_use_control` (`:481-488`) awaits `halt()` and only then calls `emit_state`, so for up to ~35 seconds `is_halted()` stays `false`, the Tauri command does not return, the UI never repaints to the stopped state, and every `wait_for_halt()` waiter (`:1163`, `:1300`) stays parked. The user sees a Stop button that did nothing while host actions continue — on the one control whose entire purpose is to be believed immediately.

Fix — set the latch first, then drain. Both properties survive:
```rust
async fn halt(&self) {
    self.halt.send_replace(true);                    // visible immediately
    let _drain = self.acting_dispatch.lock().await;  // return once in-flight work finished
}
```
Linearization is unaffected: `dispatch_acting` reads `is_halted()` under the same gate, and the `watch` channel's internal lock supplies the happens-before, so any dispatch that still observes `false` is by definition one that began before Stop landed.

Worth noting for the future even though it is not exploitable today: tokio's `Mutex` is FIFO-fair, and the computer-use executor is currently a single serial task, so no second op can slip in behind the queued `halt()`. If acting dispatches ever become concurrent (a second executor task, or per-chat parallelism), `halt()` would queue *behind* every already-waiting dispatch and each would then run after the user pressed Stop — silently breaking the contract documented at `:193-197`.

*Reasoned from the source and the constants; not reproduced against a hung broker.*

### P3 — `computer_use.rs:1228-1230` — the post-consent re-issue lost its halt check for non-acting ops

This PR removed an **unconditional** gate. On `main` (`computer_use.rs:1103-1106`) the re-issue was preceded by:

```rust
if cu.is_halted() {
    revoke_once_grant(state, decision, capability, bundle_id.as_deref(), call).await;
    return stopped_resolution();
}
```

At head that is replaced by `if acting { cu.dispatch_acting(…) } else { state.broker.operation(envelope).await }` — and the `else` branch has no halt check at all.

Failure scenario: the agent calls `computer_capture_screen` (or `computer_read_app_content`) against an app with no grant. The broker returns `Denied`, the native consent prompt opens, the user clicks **Allow** and then immediately presses **Stop** — landing after the `is_halted()` at `:1166` but before the re-issue. On `main` this resolved `stopped_by_user` and no screenshot was taken; at head the capture proceeds. Screen contents are the privacy-relevant payload here.

Mitigating, and the reason this is P3 rather than P2: the fresh-dispatch path was already `acting`-gated on `main` (`:912`), so a *first-time* capture after Stop always proceeded. The change makes the two paths consistent rather than opening something wholly new — but it is a narrowing of a stop gate in a hardening PR, and it is not called out in the commit message. If acting-only is the intended contract, say so in the comment at `:1003-1005`; otherwise restore the check in the `else` branch.

*Verified by diffing against `main`, not reproduced at runtime.*

### P3 — `.github/workflows/release.yml:285`, `:325` — the security headers are generated by an untested script and only presence-checked

`docs-site/scripts/package-vercel-output.mjs:34` is now the sole definition of the docs CSP, COOP, CORP, Permissions-Policy, Referrer-Policy, and `X-Frame-Options`. No test covers that script (it is referenced only by `docs-site/package.json:11`). The release smoke test value-checks COOP, CORP, `nosniff`, and `X-Frame-Options: DENY`, but CSP is checked by name only —

```
grep -Fqi 'content-security-policy:' "$RUNNER_TEMP/docs-headers"
```

— and `Permissions-Policy` / `Referrer-Policy` are not checked at all. The production-alias verification at `:325` is weaker still: name-only CSP plus `X-Frame-Options`.

Failure scenario: change line 34 to drop `frame-ancestors 'none'` and `object-src 'none'`, or to `default-src *`. All 70 workflow tests pass, both smoke gates pass, and the weakened policy ships to production unnoticed.

Fix: add unit coverage asserting the exact directive set `package-vercel-output.mjs` emits, and value-match at least `frame-ancestors 'none'` and `object-src 'none'` in the two smoke greps.

*Reproduced by reading; the CSP string itself is sound as written.*

### P3 — `.github/workflows/release.yml:156-159`, `:173-179` — the digest manifest is a retention record, not a verification control

The manifest is computed over `docs-site/out` **before** `package:vercel` runs, uploaded with 90-day retention, and then never downloaded or verified — `publish_docs` fetches only the `-prebuilt` artifact and deploys it unchecked. Two consequences: nothing validates the deployed bytes, and `.vercel/output/config.json` — which carries the entire header policy above — is generated after the manifest and is covered by no digest at all. Retention is also asymmetric (prebuilt 1 day, manifest 90), so after 24h you cannot retro-verify what shipped.

In-run artifacts are not attacker-writable, so this is defense-in-depth — but ADR 0022:39's *"retains a digest manifest"* should not be read as a check. Fix: compute the manifest over `.vercel/output` after `package:vercel`, and `sha256sum -c` it in `publish_docs` before deploying.

### P3 — `.github/workflows/release.yml:236,253,265,301,311` — `VERCEL_TOKEN` is passed on argv as well as in `env`

Every `vercel` invocation sets `env: VERCEL_TOKEN` *and* passes `--token "$VERCEL_TOKEN"`. The CLI reads the env var, so the flag is redundant, and argv is world-readable via `/proc` to anything else running in the job — which, per the P2 above, includes a floated third-party dependency tree. GitHub masks the value in logs, so this is not a log-leak issue. Fix: drop the five `--token` flags.

### Not a finding, recorded so it is not re-raised

`build_docs`'s checkout (`release.yml:129`) omits `persist-credentials: false`, leaving the run's `GITHUB_TOKEN` in `.git/config` while `next build`/`eslint`/`tsc` execute. I checked the base branch before raising it: `release.yml` has 10 checkouts, 9 omit the flag, and the base commit had **zero** occurrences — the one that sets it (`:1316`, `source_sbom`) was *added by this PR*. So this is a pre-existing repo-wide convention that this PR improved rather than regressed, and the token is `contents: read`. Worth adopting in the new job for consistency; not a defect in this PR.

---

## CHECKED

- **`used_by` scoping is complete**, by grep over every caller, not just the two routes named in round 1 — zero production callers of the unscoped variant remain. Desktop `LocalOwner` still maps to the `local` sentinel, so no legitimate flow broke.
- **Server-side enforcement backs the MCP disclosure.** `/native/mcp/servers` is mounted on the `client_executor_api` sub-router (`lib.rs:236-238`) under the `require_client_executor_token` route layer (`:322-323`), so the new disclosure is not a UI-only gate. The renderer route hard-refuses `command.is_some() && enabled` with `native_confirmation_required` (`routes/mcp_gateway.rs:53-63`), and the new test `enabled_local_mcp_commands_require_the_native_executor_surface` asserts both halves. The desktop preview's gate condition (`enabled && command is a string`) matches the server's exactly, so no server is skipped by the preview and re-enabling a previously-disabled command server re-enters the native path.
- **The disclosure is truthful and leaks nothing.** `ambient_environment: "cleared"` matches the unconditional `command.env_clear()` at `mcp_config/types.rs:359`; `env_from` names, stored `env` names and `cwd` match `build_command`. Only env *names* plus a set/preserve effect are shown — parent env values never appear, asserted by `stored_secret_sources_and_preservation_are_visible_without_values`. `url`/`bearer_token_env`/`gateway_endpoint` are mutually exclusive with `command` and `plugin` is refused, so nothing execution-relevant is undisclosed. No preview/apply TOCTOU: the previewed `config` value is the one serialized into the PUT.
- **`native_command_token` hardening is a genuine improvement** — control/format/line-separator characters now reject rather than being replaced with U+FFFD, closing the gap where the dialog showed a sanitized string while the unsanitized config was saved.
- **Stop linearization is otherwise sound.** All three broker-acting paths go through the gate (fresh `:1050`, post-consent re-issue `:1219`, confirmation redemption `:1324`), and the stopped post-consent path still calls `revoke_once_grant` so a one-shot grant cannot survive a Stop. No lock-order or self-deadlock bug: `dispatch_acting` releases the gate before `dispatch_confirmation`/`dispatch_consent` re-acquire it, and both native prompts await *outside* the gate — it is never held across a user-interaction await. `ComputerUseState` is per-process desktop state, so "global stop" carries no cross-user concern, and confirmations are redeemed by single-use `confirmation_id`, so one run's confirmation cannot satisfy another's action.
- **No injection surface added.** The diff adds no `Command::new`, no shell, no string-built command line; `argv` is passed as a `Vec<String>` to `command.args()`.
- **No secrets in logs from the new Rust code.** `git diff 84ea1041..9550721c -- '*.rs'` adds zero `tracing::*` / `println!` / `eprintln!` lines.
- **No script injection in the new workflow jobs.** No `${{ }}` appears inside any `run:` block in lines 117-332; `RELEASE_TAG`, `RELEASE_SHA`, `DEPLOYMENT_URL`, `DEPLOYMENT_ID` all reach the shell via `env:`. `RELEASE_TAG` is additionally constrained by the semver regex in `check-release-tag.mjs`.
- **Action pinning holds.** Every `uses:` in `release.yml` is a 40-hex SHA; the new jobs introduce no new action. Triggers remain `release` + `workflow_dispatch` — no `pull_request_target`, no `workflow_run`.
- **Credential isolation is genuine, not nominal.** `build_docs` declares `contents: read`, no `environment:`, and references no `secrets.*`; `publish_docs` has no checkout and runs no repository code — its workspace holds only the downloaded output and a `jq`-generated `project.json`. The Next build and the deployment token never coexist in one job. That is the core property of these commits and it holds.
- **`package-vercel-output.mjs` copies nothing sensitive.** All paths derive from `import.meta.url`, no user input, no traversal surface; it copies only `docs-site/out`. No tracked `public/`, no `productionBrowserSourceMaps`, no `.env`, no `.git`. Only `static/` and `config.json` are emitted — no `functions/`, so the deployment has no server runtime. Root `.gitignore` correctly gains `.vercel/`.
- **`GET /connected-apps` exposes no credential values to members** — the projection is credential *status* only (`Present`/`Missing`/`None`) plus catalog counts and `base_url`; never definitions, documents, or secret values.
- **`url` crate swap is behavior-neutral** — single `url 2.5.8` in `Cargo.lock`, workspace-pinned exact, and reqwest re-exports the same crate.
- **Tests.** `cargo test -p tidebreak-server` → 876 passed, 0 failed. `cargo test -p tidebreak-desktop` → 149 passed (including the three new stop tests). `cargo test -p tidebreak-router` → 175 + 5 passed. `node --test scripts/workflow-security*.test.mjs` → 70 pass, 0 fail. One flake noted and dismissed: `tests::configuration::api_key_put_configures_it_and_delete_reverts` failed once in a parallel run and passed on re-run and in isolation; it races on the process-global env guard at `configuration.rs:3116`, which pre-exists outside this PR's diff. Running `--test-threads=1` fails several unrelated timing tests, so the suite is parallel-by-design.
- **Known non-coverage, unchanged by this PR**, flagged only so it is not mistaken for a new gap: `COMPUTER_RETURN_TO_TIDEBREAK_TOOL` and `COMPUTER_WAIT_TOOL` bypass the halt latch entirely, and read-only ops are never halted — consistent with the contract documented at `computer_use.rs:1003-1005`.

naingthet (commented):
> 🤖 **Thet's review agent** — automated review (round 4, release & supply-chain lens), not a human sign-off.

A concurrent app-runtime lens covers the desktop UI, Rust services, and tenant isolation; this pass deliberately does not touch them.

**Verdict — changes requested.** The two headline items from round 3 are genuinely fixed and I verified both by mutation testing. But the docs pipeline this PR introduces has a release-blocking defect that no existing test catches: `vercel curl` does not accept `--token`, so the `publish_docs` smoke gate fails on every run. Separately, the PR #2080 test hole is **not** closed — the 9 new test lines are entirely about the SBOM.

---

## Status of open items in this lens

**P1 SBOM ENOENT — ✅ genuinely fixed.** Verified, not assumed:
- `release.yml:1346-1347` `mkdir -p source-sbom` sits in the same `source_sbom` job as the generator at `:1350`, and I confirmed the job declares no `defaults:` and no `working-directory:` (the only `defaults:` blocks in the file are at `:999` and `:1125`, other jobs), so both steps run in `$GITHUB_WORKSPACE`.
- Path matches exactly, `${{ needs.validate.outputs.version }}` interpolation included, across generator (`:1354`), checksum (`:1363`), and upload (`:1372`, `if-no-files-found: error`).
- Downstream chain intact: `publish` downloads into `dist` (`:1463-1465`), `:1471` re-verifies `sha256sum --check --strict` before the attestation subjects are built at `:1472-1476`.
- Mutation-tested three ways against `scripts/workflow-security.test.mjs` — removing the mkdir step, pointing it at a wrong directory name, and moving it *after* the generator each take the suite from 33/0 to 32/1. The guard is real, not a string-presence tautology.

**P2 global `pnpm add --global vercel@59.0.0` — ✅ fixed.** Replaced by a sparse checkout of `.github/vercel-cli` (`:198-199`), `pnpm install --frozen-lockfile --ignore-scripts` (`:210`) against a committed 3037-line `pnpm-lock.yaml` pinning `vercel@59.0.0`, invoked as the local binary. Transitive deps no longer float in the credentialed job, and `--ignore-scripts` blocks install-time execution. Good fix.
> **Disputing the ADR citation only:** this repository has no ADR directory — `find . -ipath '*adr*' -name '*.md'` returns nothing — so "contradicts ADR 0022:46-47" is unverifiable here and I can neither confirm nor apply it. The item is fixed on its own merits regardless.

**P3 token on argv as well as env — ❌ not addressed, and it escalates to P1 for `vercel curl`.** See NEW-1.

**P3 CSP checked by name only — ❌ not addressed.** `:293` and `:333` are still `grep -Fqi 'content-security-policy:'`. See NEW-3.

**P3 header-generation script untested — ❌ not addressed.** `docs-site/scripts/package-vercel-output.mjs` is referenced only by `docs-site/package.json:11`; a repo-wide grep finds zero tests touching it.

**P3 digest manifest never verified / does not cover `config.json` — ❌ not addressed, and narrower than described.** See NEW-2.

**PR #2080 test hole — ❌ NOT closed.** The ~9 new lines in `scripts/workflow-security.test.mjs:1499-1507` guard the SBOM mkdir; none touch the README/DMG assertion. See NEW-4.

---

## New findings

### NEW-1 · P1 · `.github/workflows/release.yml:258-261, 270-273` — `vercel curl` cannot accept `--token`; the docs smoke gate fails on every release

**Reproduced, not reasoned.** `vercel curl` does not use the normal flag parser. It uses `parseCurlLikeArgs`, whose known-flag sets are `VC_STRING_FLAGS {--deployment, --protection-bypass}`, `VC_IGNORED_STRING_FLAGS {--scope, --team}`, `VC_BOOLEAN_FLAGS {--yes, --help, --trace, --json}` (`vercel@59.0.0`, `dist/commands-bulk.js:20202-20204`). `--token` is in none of them, so it falls through to `result.toolFlags.push(arg)` and is handed to the real `curl` binary.

I extracted `parseCurlLikeArgs` from the installed `vercel@59.0.0` bundle and ran it on this workflow's exact argv shape:

```
target    = /docs/
toolFlags = ["--token","<value>","--fail","--silent","--show-error","--output",...]
```

The token value and its flag both land in curl's argv. `curl` then rejects it:

```
curl: option --token: is unknown        (exit 2)
```

`commands-bulk.js:21102-21104` resolves the vercel exit code straight from curl's, so the step exits non-zero under the runner's default `bash -e`.

**Failure scenario:** every release. `Create an unaliased production deployment` (`:230`) succeeds and creates a real production-target deployment. The very first `request /docs/` at `:269` then dies before any assertion runs — so none of the CSP/COOP/CORP/nosniff/frame-options checks at `:293-297` ever execute. `Promote the verified deployment` (`:302`) never runs, docs never go live, and an orphaned unaliased production deployment is left behind each attempt. The gate does not merely fail to catch regressions; it cannot pass.

This is new surface: `origin/main`'s `release.yml` contains zero occurrences of `vercel` or `publish_docs`, so this is not a pre-existing regression.

**Fix:** the CLI reads `VERCEL_TOKEN` from the environment, which is already set on these steps. Drop the `--token` line from both `vercel curl` invocations (`:261` and `:273`) and rely on the env var. `--token` on `deploy` (`:243`), `promote` (`:309`), and `inspect` (`:319`) is fine — those use the normal parser — but dropping it there too removes the secret from argv uniformly.

**Secondary:** because the value reaches argv, it appears in `output_manager.debug("Executing: curl …")` (`:21088`) and in the spawned process's command line. GitHub's automatic masking covers `secrets.VERCEL_TOKEN` in logs, so I'd treat the disclosure as low on its own — the functional break is the reason this is P1.

**Also worth a test:** `scripts/workflow-security.test.mjs:1086` asserts `vercel curl` is invoked, but nothing asserts anything about how it is authenticated. A regression test that rejects `--token` inside the `vercel curl` invocations would have caught this.

### NEW-2 · P2 · `.github/workflows/release.yml:156-159, 173-178` — the docs digest manifest hashes a different tree than the one that gets deployed, and is never verified

The manifest is `sha256sum` over `docs-site/out` (`:156-159`), computed *before* `pnpm --dir docs-site package:vercel` runs at `:160`. What is actually transferred to the credentialed `publish_docs` job is `.vercel/output` (`:166-170`) — a different tree, and one that contains `config.json`, the file carrying the entire CSP / COOP / CORP / Permissions-Policy / X-Frame-Options set (`docs-site/scripts/package-vercel-output.mjs:29-63`). That file is generated after the digest and is covered by no digest at all.

On top of that, the manifest artifact is uploaded with 90-day retention and then never downloaded: the only `download-artifact` in `publish_docs` is the `-prebuilt` one at `:214`, and a grep for the manifest name finds exactly one hit, the upload at `:176`. So it is a write-only artifact — it enforces nothing.

**Failure scenario:** any corruption or tampering of the prebuilt artifact between jobs — including of the security headers in `config.json` — reaches the credentialed deploy with no integrity check, and the retained manifest could not detect it even forensically, because it never covered those bytes.

**Fix:** compute the digest over `.vercel/output` *after* `package:vercel`, so `config.json` is included, and add a verification step in `publish_docs` that downloads the manifest artifact and runs `sha256sum --check --strict` against the downloaded tree before the token is used — mirroring what `publish` already does correctly for the SBOM at `:1471`.

### NEW-3 · P3 · `.github/workflows/release.yml:293, 333` — the CSP assertion checks only that the header exists

Both smoke gates use `grep -Fqi 'content-security-policy:'`, matching the header *name*. Their neighbours check values (`x-frame-options: DENY`, `cross-origin-opener-policy: same-origin`), so CSP is the single header whose content is unverified — and the longest, most regression-prone one.

**Failure scenario:** `package-vercel-output.mjs` regresses to `default-src *` or drops `object-src 'none'`; both gates stay green and ship it.

**Fix:** assert the load-bearing directives, e.g. add `grep -Eqi "content-security-policy:.*default-src 'self'"` plus checks for `object-src 'none'` and `frame-ancestors 'none'`. Better still, since `config.json` is generated from a constant, add a unit test over `package-vercel-output.mjs`'s route config and let the smoke gate confirm the deployed headers match it.

### NEW-4 · P3 · `scripts/workflow-security.test.mjs:1568-1570` — the PR #2080 hole is still open; the guard is a literal tautology

```js
if (/Tidebreak-macos-universal\.dmg/.test(attachJob)) {
  assert.match(attachJob, /Tidebreak-macos-universal\.dmg/);
}
```

The assertion is the loop condition, so it can never fail. Combined with the alternation at `:1577` (`Tidebreak-macos-(?:universal|apple-silicon)\.dmg`), the README link and the uploaded asset are never cross-checked.

**Mutation-tested:** I pointed the README at `Tidebreak-macos-universal.dmg` while removing that asset from `attach_downloads`. Result: **33 pass, 0 fail.** The hole reproduces exactly as it did before this round.

**Failure scenario:** the README's permanent one-click `/releases/latest/download/` link 404s for every user, with a green suite.

**Fix (verified — fails under the mutation, passes clean at 33/0):**

```js
assert.match(attachJob, /Tidebreak-macos-universal\.dmg/);
assert.match(attachJob, /Tidebreak-macos-apple-silicon\.dmg/);

const macDownloadLink = readFileSync(repositoryFile("README.md"), "utf8")
  .match(/releases\/latest\/download\/(Tidebreak-macos-[\w-]+\.dmg)/)?.[1];
assert.ok(macDownloadLink, "the README must publish a macOS download link");
assert.ok(
  attachJob.includes(`downloads/${macDownloadLink}`),
  `attach_downloads must upload ${macDownloadLink}`,
);
```

---

## CHECKED

- `actionlint .github/workflows/release.yml` — clean, exit 0.
- `node --test scripts/workflow-security.test.mjs` at `defd6285` — 33/0 pass.
- Mutation tests run: SBOM mkdir removed / wrong dir / reordered after generator (all correctly fail); README-vs-attach mismatch (incorrectly passes); proposed NEW-4 fix (correctly fails under mutation, 33/0 clean). Worktree restored to a clean tree afterward.
- `anchore/sbom-action` and the whole SBOM→checksum→upload→download→attest chain traced end to end.
- `vercel@59.0.0` installed and inspected directly; `parseCurlLikeArgs` extracted and executed on this workflow's argv; `curl --token` rejection confirmed live.
- **Non-finding, checked and cleared:** `vercel promote --yes` only rebuilds via `promoteByCreation` when `deployment.target !== "production"` (`commands-bulk.js:58997-59009`). This workflow deploys with `--prod` (`:232`), so the target *is* production and `--skip-domain` only defers aliasing — the promoted id matches the id the alias check at `:322` compares against. No defect.
- **Non-finding, per prior rounds:** `persist-credentials` on checkouts — not raised.
- Base-branch comparison: `origin/main`'s `release.yml` has no `vercel`/`publish_docs` content, so NEW-1/NEW-2/NEW-3 are introduced by this PR, not inherited.
- Action pinning: every `uses:` in `release.yml` is pinned to a 40-hex SHA with a version comment; `permissions:` blocks are per-job and least-privilege (`source_sbom` is `contents: read` with no `environment:` and no secrets, and the test at `:1494-1498` enforces that).

naingthet (commented):
🤖 **Thet's review agent** — automated review (round 4, app-runtime lens), not a human sign-off.

A concurrent release/supply-chain lens covers `.github/workflows/**`, action pinning, secrets and SBOM/attestation; this pass deliberately does not look at CI.

**Verdict: the round-3 findings in this lens are genuinely fixed, not merely touched.** Two P3 hardening notes below, neither a merge blocker. All 12 `computer_use` Rust tests and all 27 `ui/src/document` tests pass at `defd6285`.

---

## Status of open items

**P2 `DocxViewer.tsx` — dead DOCX hyperlinks — FIXED, and fixed the safe way.**
The fix does *not* route document URLs to `openExternal`. It defangs instead: the anchor keeps no `href` at all, and the validated URL is surfaced as visible text plus a `title` ("External link (copy this address): …"). The sandbox went *stricter* (`allow-popups` → `sandbox=""`, `:171`). So the specific risk I was asked to hunt — a new outbound native capability driven by untrusted document content — was never created. I reproduced the sanitizer's behaviour against a hostile fixture: `javascript:`, `file:`, `data:` and `https://user:pw@…` hrefs are all stripped with no text and no title emitted; `#` bookmarks keep working; and because `safeExternalHref` returns `parsed.href`, `URL` normalization percent-encodes `"`, `<` and `>` before the value reaches either the text node or the `title`, so there is no attribute- or HTML-injection path out of the new surface. Coverage is attribute-driven (`href` / `xlink:href` over `querySelectorAll("*")`), so headers, footers, footnotes, endnotes and VML shapes are all covered, not just body text. Confirmed reproduced, not reasoned.

**P2 `computer_use.rs:277-282` — `halt()` ordering — FIXED, guarantee preserved.**
`halt()` now does `send_replace(true)` first, then acquires `acting_dispatch`. Linearization still holds: `dispatch_acting` takes the gate *then* checks `is_halted` (`:290-294`), so any dispatch starting after the latch publishes fails closed and releases the gate; anything already in flight drains before `halt()` returns. `snapshot()`/`emit_state` read the latch without the gate, so Stop is visible immediately instead of after a ~35s broker round-trip. The new test `stop_sets_the_latch_before_waiting_for_an_in_flight_action` asserts both halves (latch visible within 100ms *and* `halt` still pending). The race moved nowhere.

**P3 `computer_use.rs` post-consent re-issue — FIXED (round-3 finding confirmed accurate).**
I checked against `origin/main`: `main:1103` has an unconditional `if cu.is_halted()` immediately after `broker.control(grant)`; at `3166eb1f` that check was gone, leaving only the `acting`-gated `dispatch_acting` path, so a non-acting op (`capture_screen`) in the re-issue window skipped Stop. `defd6285:1208-1211` restores it, with `revoke_once_grant` before returning. Parity with `main` restored. (For the record, the entry-point `if acting && cu.is_halted()` at `:1027` is identical on `main` — that one was never a regression.)

**P3 `DocxViewer.tsx:104-107` tab-stop/VML sizing on a detached tree — FIXED.**
Two changes, both correct. `experimental: false` (`:20`) is load-bearing: in docx-preview 0.4.0 that flag is the sole gate on `refreshTabStops()`, which fires on a **500ms `setTimeout`** after `renderAsync` resolves and then does live layout reads plus a `document.body` temp-div probe — i.e. it would fire long after `stagingHost.remove()`. Separately, `renderVmlElement` schedules a `requestAnimationFrame` `getBBox()` resize that is *not* experimental-gated, which the new offscreen closed shadow root (`:95`) plus the awaited `nextAnimationFrame()` (`:107`) correctly accommodates. I also verified the shadow root is safe: docx-preview does no document-scoped `querySelector`/`getElementById`, never writes to `document.head`, and routes all `<style>` output through the passed `styleContainer`, so the encapsulation claim holds.

**P3 `router.rs` URL validator — NOT ADDRESSED, and I'm content to leave it.** Unchanged in this commit; pre-exists on `main`, server-side enforcement covers it, client-side is defense-in-depth only.

---

## New findings

### P3 — `DocxViewer.tsx:226` — the sanitizer hands its validated URL to itself through a DOM attribute it never strips first

This is the bug class the fix was closing, reappearing one layer down: trusting in-band data that untrusted input could also write. `safeExternalHref`'s output is stashed on the element as `data-tidebreak-external-href` in the first loop, and the second loop (`:233`) reads it back and treats it as pre-validated. Nothing strips a pre-existing `data-tidebreak-external-href` before the first loop runs, and that loop only touches `on*` / `src` / `href` / `xlink:href`.

Reproduced: an anchor arriving with `data-tidebreak-external-href="javascript:…"` and no `href` renders as `seeded (javascript:…)` with `title="External link (copy this address): javascript:…"`. Impact is bounded to spoofing, not execution — there is no `href`, the frame is `sandbox=""` with `script-src 'none'`, so the string is inert; the harm is that the viewer's own "copy this address" chrome vouches for a string the validator never saw. **And I could not find a live path to it:** docx-preview 0.4.0 emits zero `data-*` attributes (its VML attribute passthrough is a fixed whitelist of `width/height/cx/cy/rx/ry/fill/x1/y1/x2/y2/stroke*`), so today this is unreachable. Filing it as hardening against a future library bump, not as a live bug.

Fix — drop the DOM channel entirely: collect into a `WeakMap<Element, string>` in the first loop and read it in the second. If you'd rather keep the attribute, add a strip pass at the top of `sanitizeRenderedDocument` — that also cleans up the related cosmetic leak where the marker survives into the frame on non-anchor elements, since the second loop only queries `a`.

### P3 — `DocxViewer.tsx:107,258` — the unconditional `requestAnimationFrame` wait strands the preview whenever rAF is paused

`nextAnimationFrame()` is awaited on every render, but Chromium suspends rAF entirely for hidden or occluded windows. Failure scenario: open a DOCX, then switch to another app (or minimize) before the parse finishes — `renderAsync` resolves, the rAF never fires, so `.then` never runs and the viewer sits on the spinner. It recovers on refocus (rAF resumes and the pending callback runs), so this is an annoyance rather than a hang, but there is no timeout and no retry. Reproduced with rAF stubbed to never fire: no iframe, spinner still mounted, staging host still attached after 250ms; unmount cleans up correctly, so there is no DOM leak.

Fix: race it, e.g. `await Promise.race([nextAnimationFrame(), new Promise((r) => setTimeout(r, 100))])`. `setTimeout` is throttled in background windows but still fires, unlike rAF. The rAF is only needed for VML `getBBox` sizing, so a timeout fallback degrades to slightly-off shape bounds rather than a stuck preview.

---

## Workbook-viewer reuse

**Disputing the premise here.** The DOCX hyperlink bug is *not* repeated in the workbook viewer merged as #2078. `NativeSpreadsheetViewer.tsx` parses OOXML in a WASM worker and draws a virtualized **canvas** — it never turns file content into DOM, has no `innerHTML`/`dangerouslySetInnerHTML`, no `srcdoc`, and no hyperlink code path at all (workbook hyperlinks simply aren't rendered as links). So there is nothing there for this fix to be shared with, and no dead-link regression to match.

That said, on the factoring question as asked: `sanitizeRenderedDocument` (`:187`) and `safeExternalHref` (`:264`) are module-private — `DocxViewer.tsx` exports only `DOCX_RENDER_OPTIONS` and the default component. If a future viewer *does* render untrusted file content to DOM, it would need those exported or it will grow a second copy. Worth a small follow-up to lift them into a shared `document/sanitize.ts` while the logic is fresh; not a blocker for this PR.

---

## CHECKED

`git diff 3166eb1f8776..defd6285f420` in full · `halt`/`dispatch_acting`/`is_halted`/`wait_for_halt` linearization · all `is_halted()` call sites diffed three ways (`main` vs `3166eb1f` vs `defd6285`) · `cargo test -p tidebreak-desktop client_execution::computer_use` (12/12) · `vitest run src/document/` (27/27, 9 files) · hostile-DOCX sanitizer fixture covering `javascript:`/`file:`/`data:`/credentialed-https/quote-and-tag-bearing https URLs, pre-seeded marker attributes, and marker-on-non-anchor · rAF-suspended render path · docx-preview 0.4.0 dist audited for `document.head`/`document.body` writes, `data-*` emission, `experimental` gating, and timers · closed-shadow-root boundary vs. every document-scoped lookup in `ui/src` · workbook viewer (`NativeSpreadsheetViewer.tsx`, `SpreadsheetViewer.tsx`, `readOnlyWorkbookProjection.ts`) for a parallel sanitizer/hyperlink surface. Scratch tests reverted; worktree removed.

Not reviewed by this pass (other lens): `.github/workflows/**`, `scripts/workflow-security.test.mjs`, `docs-site/**` canonical-URL changes.

naingthet (commented):

Both round-4 findings are genuinely fixed and verified by mutation-testing the new tests against the parent commit; one unreachable hardening nit remains, non-blocking. Incremental review of `defd6285..e109b3ec`, **app-runtime lens only** (`DocxViewer.tsx` / `DocxViewer.dom.test.tsx`); the release/supply-chain slice is another reviewer's.

*(Note: these files live at `crates/tidebreak-desktop/ui/src/document/`, not `packages/ui/src/document/`.)*

### Both open P3s: fixed, and the fixes are real

I reverted `DocxViewer.tsx` to its `defd6285` version while keeping the new tests, and both new tests fail — `expected '<!doctype html>…' not to contain 'javascript:alert'` and `expected null not to be null`. At head both pass. The tests genuinely pin the behaviour rather than merely accompanying it.

I specifically hunted the two adjacent-bug-class risks these fixes could have introduced. **Neither is present:**

- **WeakMap node identity holds.** Population (`:188`, `:229`) and read (`:236`) are in the same synchronous function with no `await` and no node-replacing mutation between them — the first loop only calls `removeAttribute`. docx-preview's node re-creation all happens before `renderAsync` resolves, so there is no window. More importantly the design **fails closed, not open**: `element.removeAttribute(name)` runs unconditionally after the `WeakMap.set`, so a lost entry would lose the *display* of the URL, never surface an unvalidated one. Verified with a hostile fixture — a spoofed marker plus a legitimate `https` href on the same anchor renders the real URL and never `evil.example.com`; a spoofed marker alone yields no `title` and no appended URL text.
- **The race loser is harmless.** `Promise.race` (`:266`) does not cancel the rAF, but the rAF callback only calls `resolve()` on an already-settled promise, and the continuation is gated by `if (cancelled) return` (`:108`). Reproduced: captured the rAF callback, let the 100ms timeout win, unmounted, then fired the callback late — no thrown error, no DOM write, no stranded staging host. Also confirmed a stranded-then-superseded render resolves to the second document with no leakage of the first.

### P3 — `DocxViewer.tsx:208` — attribute strip compares case-insensitively but removes case-sensitively (unreachable today; hardening only)

The guard lowercases (`const name = attribute.name.toLowerCase()`, `:198`) but the removal passes the original name to `removeAttribute`, which itself lowercases for HTML-namespace elements. An attribute whose literal name is case-preserved therefore matches the guard and survives the removal.

Reproduced in jsdom: after running the strip loop over an anchor carrying case-preserved `ONCLICK` and `DATA-TIDEBREAK-EXTERNAL-HREF`, `outerHTML` is unchanged — both attributes are still there. This applies to every name in the list (`on*`, `srcdoc`, `srcset`, `action`, `formaction`, `ping`), not just the new marker.

**Reachability is nil, which is why this is P3 and not higher.** The HTML parser lowercases attribute names, so parsed content can't carry one; the only way to create one on an HTML element is `setAttributeNS(null, …)`, and docx-preview 0.4.0 makes zero `setAttributeNS` calls. Its single dynamic `setAttribute(k, v)` site (`renderVmlChildElement`) targets a `createSvgElement` result — an SVG-namespace element, where `removeAttribute` *is* case-exact, so the strip works correctly there. And for the marker specifically it still fails closed: the second loop reads the `WeakMap`, never the attribute, so a surviving marker is inert chrome in a `sandbox=""` frame with `script-src 'none'` rather than a spoofing vector.

Fix is one line — remove the node itself instead of by name, which is case-exact in both namespaces:

```
element.removeAttributeNode(attribute);
```

Verified: with `removeAttributeNode`, the same fixture strips to `<a></a>`.

Filing this for consistency with round 4, which fixed the marker round-trip on the same "unreachable, but the sanitizer shouldn't have the gap" standard. Not a blocker.

### CHECKED

`git diff defd6285f420..e109b3ecd3fa -- crates/tidebreak-desktop/ui/src/document/` (17 lines source, 27 test) · `vitest run src/document/` at head — **28 passed / 28, 9 files** · `tsc --noEmit` — **clean, exit 0** · mutation test: parent `DocxViewer.tsx` against head tests — **2 failed / 1 passed**, confirming both new tests are load-bearing · late-rAF-after-unmount and stranded-then-superseded render paths reproduced · WeakMap fail-closed fixture (spoofed marker alone, spoofed marker + legitimate href, marker on a non-anchor — the round-4 cosmetic leak is now fixed by the global strip) · jsdom attribute-case mechanism isolated, including `removeAttributeNode` as the fix and the SVG-namespace contrast · docx-preview 0.4.0 dist audited for `setAttributeNS` and dynamic `setAttribute` sites to establish reachability. Scratch tests deleted; worktree removed.

Round-4's settled ground (defanged anchors, `sandbox=""`, `URL` normalization, attribute-driven coverage) was not re-derived.

naingthet (commented):

**Verdict: no blocking findings.** All five open items from round 4 are genuinely fixed and I reproduced each fix. Three P3 coverage-completeness gaps remain, none of which should hold the merge.

This was an **incremental review of `defd6285..e109b3ec`**, **release / supply-chain lens only** (`.github/workflows/release.yml`, `.github/workflows/staging-publish.yml`, `README.md`, `docs-site/**`, `scripts/workflow-security*.test.mjs`). The `packages/ui` DocxViewer slice was reviewed separately and is out of scope here. Prior rounds' CHECKED sections were treated as settled.

---

## Round-4 items — all verified fixed

**1. P1 `vercel curl --token` — FIXED, verified by execution.** I installed the pinned `vercel@59.0.0` and instrumented `parseCurlLikeArgs` to print the argv it forwards to the real curl binary. Round-4 argv produced `TOOLFLAGS=["--token","<redacted>","--fail","--silent","--output",...]` — `--token` is in neither `VC_STRING_FLAGS` (`--deployment`, `--protection-bypass`), `VC_IGNORED_STRING_FLAGS` (`--scope`, `--team`), nor `VC_BOOLEAN_FLAGS`, so it falls through to `toolFlags`. Head argv produces `TOOLFLAGS=["--fail","--silent","--output",...]` — clean. I also confirmed the CLI reads `VERCEL_TOKEN` from the environment globally (a malformed env value produced `You defined "--token", but its contents are invalid`), so dropping the flag does not lose authentication. Not just "the flag is gone" — the new argv parses correctly.

**2. P2 digest covers the wrong tree — FIXED.** The digest is now computed inside `.vercel/output` *after* `package:vercel`, so `config.json` and every security header it carries are hashed; the manifest is uploaded (`release.yml:178`), downloaded (`:224`), and checked with `sha256sum --check --strict` before the deployment is created (`:229`). Ordering is enforced by a new assertion, and I reproduced that it bites: moving the digest back before `package:vercel` → **32 pass / 1 fail**; deleting the verify step → **32 pass / 1 fail**.

**3. P3 CSP checked by name only — FIXED.** Both the staged smoke gate and the production-alias gate now value-check `default-src 'self'`, `object-src 'none'`, and `frame-ancestors 'none'`. The test counts each directive exactly twice, so dropping either call site fails.

**4. P3 tautology in `workflow-security.test.mjs` — FIXED.** The `if (/universal/.test(attachJob)) assert.match(attachJob, /universal/)` self-satisfying block is gone, replaced by an unconditional assertion.

**5. #2080 test hole — CLOSED, reproduced.** I re-ran the exact round-4 mutation: README points at `Tidebreak-macos-universal.dmg` (which it now does at head) while `attach_downloads` no longer produces that asset (case arm and `universal_dmg` retargeted to `Tidebreak-macos-arm64.dmg`, leaving **0** occurrences of the universal name in the workflow). Result: **`workflow-security.test.mjs` 32 pass / 1 fail**, **`workflow-security-mutations.test.mjs` 42 pass / 1 fail**. Round 4 saw 33/0 on this same mutation. The hole is genuinely closed.

---

## New findings

### P3 — `docs-site/scripts/vercel-output-config.test.mjs:19` — `Permissions-Policy` and `Referrer-Policy` are guarded by nothing, at build time or deploy time

The new test asserts CSP, COOP, CORP, `nosniff`, and `X-Frame-Options`, but not the two remaining headers. The release smoke gate does not check them either — `grep -rni "permissions-policy\|referrer-policy" .github/workflows scripts` returns no hits outside the config module itself.

**Reproduced.** I deleted lines 10 and 11 of `vercel-output-config.mjs` (`Permissions-Policy: camera=(), geolocation=(), microphone=(), payment=(), usb=()` and `Referrer-Policy: strict-origin-when-cross-origin`) and the entire suite stayed green: workflow-security **33/0**, mutations **43/0**, docs config **2/0**. The docs site would ship with default referrer behaviour and no permissions policy, and nothing anywhere would report it.

**Fix:** add to the test at `:19`
```js
assert.equal(
  docsRoute.headers['Permissions-Policy'],
  'camera=(), geolocation=(), microphone=(), payment=(), usb=()',
);
assert.equal(docsRoute.headers['Referrer-Policy'], 'strict-origin-when-cross-origin');
```

### P3 — `docs-site/scripts/package-vercel-output.mjs:32` — the new test covers the exported constant, not the `config.json` that actually ships

Extracting `VERCEL_OUTPUT_CONFIG` makes the *value* testable, but nothing asserts that the packager writes that value. The script that produces the shipped artifact is still untested — the round-4 "generated by an untested script" concern is narrowed, not closed.

**Reproduced.** I changed line 32 to serialize `{ version: 3, routes: [{ handle: 'filesystem' }] }` instead of `VERCEL_OUTPUT_CONFIG` — i.e. a docs deployment with *zero* security headers. Every pre-deployment gate stayed green: workflow-security **33/0**, mutations **43/0**, docs config **2/0**. It would be caught only by the live-header smoke test, which runs after a production deployment has already been created — the same late-failure shape as the round-4 P1.

**Fix:** have the test run the packager against a temp dir and assert on the emitted file rather than (or in addition to) the constant — e.g. spawn `node scripts/package-vercel-output.mjs` with the output root pointed at `fs.mkdtempSync(...)`, then `JSON.parse` the written `config.json` and run the existing assertions over that object.

### P3 — `.github/workflows/release.yml:178` — the manifest *producer* is unasserted while the *consumer* is mandatory

`publish_docs:224` downloads artifact `tidebreak-docs-<tag>-manifest` unconditionally, and `download-artifact` hard-fails when the artifact is absent. The test asserts the download step's artifact name but never asserts that `build_docs` uploads it.

**Reproduced.** I deleted the `Retain the static-content digest manifest` step (`:178`–`:183`) and the suite stayed green: workflow-security **33/0**, mutations **43/0**. A future edit that drops or renames that upload passes CI and then breaks every release at the download step. Failure is loud and fails closed, hence P3 rather than higher.

**Fix:** in the `release documentation is built from the validated tag` test, add
```js
assert.match(build, /name: tidebreak-docs-\$\{\{ needs\.validate\.outputs\.tag \}\}-manifest/);
```
and a matching entry in `workflow-security-mutations.test.mjs` that removes the upload step.

---

## CHECKED

- `actionlint` over the full workflow tree: **exit 0, no diagnostics.**
- Baseline at `e109b3ec`: `scripts/workflow-security.test.mjs` **33 pass / 0 fail**; `scripts/workflow-security-mutations.test.mjs` **43 pass / 0 fail** (41 mutations); `docs-site/scripts/vercel-output-config.test.mjs` **2 pass / 0 fail**.
- Mutations that correctly go red (each run individually, tree restored between): removing `pnpm --dir docs-site test:vercel-output` (32/1, 40/3); removing the `Verify the transferred documentation` step (32/1, 40/3); moving the digest before `package:vercel` (32/1, 42/1); re-adding `--token` to the deploy call (covered by the new `doesNotMatch`); removing the `object-src` grep; removing `persist-credentials: false` (32/1, 42/1); removing `timeout-minutes: 90` (32/1, 42/1); weakening CSP `object-src` and COOP in the shared config (docs config 1/1 each); the #2080 README/asset mutation (32/1, 42/1).
- Mutations that do **not** go red — the three P3s above (M6 header deletion, M7 packager divergence, M9 manifest-upload deletion).
- Manifest artifact plumbing traced end to end: upload path `${{ runner.temp }}/tidebreak-docs-<tag>.sha256` lands at artifact root, downloads to `$RUNNER_TEMP/docs-manifest/`, and matches the path the verify step checks. `find . -type f` inside `.vercel/output` covers `config.json`; `include-hidden-files: true` is set on the prebuilt upload.
- Step ordering in `publish_docs` confirmed: verify → deploy → smoke → promote → verify alias, with the new ordering assertion enforcing it.
- Not filed, noted only: `sha256sum --check --strict` detects modified and missing files but not *added* ones. Since `config.json` is covered, routes and headers cannot be altered this way, and both artifacts originate in the same workflow run — low value, but a `find`-count or manifest-line-count assertion would close it if you want completeness.
- `origin/main` checked before attributing anything to this PR; none of the three findings exist there (the code is new in this delta).
- Severity is **trending down**: round 4 carried a P1 and a P2; round 5 carries neither, and the three remaining P3s are all test-coverage completeness rather than defects in shipped behaviour.

PR #2076, opened 2026-08-14
feat(desktop): package universal macOS builds

naingthet:
## Summary

- build production, staging, and cache-warm macOS jobs with Tauri’s universal-apple-darwin target
- compile the host broker for arm64 and x86_64, combine it with lipo, and verify both slices in the signed app bundle
- publish one macos/universal artifact set while advertising its signed updater archive under both darwin-aarch64 and darwin-x86_64
- preserve the existing Apple Silicon GitHub download filename as a byte-identical compatibility alias while introducing Tidebreak-macos-universal.dmg
- update release documentation and workflow policy coverage for the new packaging contract

## Compatibility

Existing Apple Silicon installations continue to receive updates through darwin-aarch64. Intel installations use darwin-x86_64; both keys resolve to the same universal updater archive. The existing permanent README download URL remains valid through the compatibility asset alias.

Previously published arm64-only immutable releases cannot be resumed under the new platform descriptor, and fail closed during hosted-release validation rather than being republished with a different artifact shape.

## Safety and risk

The signing jobs still restore only credential-free compiler outputs and discard all restored linked products before rebuilding with signing material. Release and staging verification now reject the bundle unless both the main executable and host-broker sidecar contain arm64 and x86_64 slices.

The main operational risk is increased macOS compilation time because both Rust targets are built sequentially. The release and cache jobs retain their existing 60-minute timeouts and use a single cache archive containing both slices.

## Validation

- cargo tauri build --target universal-apple-darwin --no-bundle --ci
- cargo tauri build --target universal-apple-darwin --bundles app --no-sign --ci
- verified the bundled app executable reports x86_64 arm64 with lipo -archs
- verified the bundled host broker reports x86_64 arm64 with lipo -archs
- actionlint on all changed workflows
- manifest and workflow-security test suites: 62 tests passed
- Node syntax checks for the sidecar and manifest scripts
- git diff --check

### DIFF
--- a/.github/workflows/cache-macos.yml
+++ b/.github/workflows/cache-macos.yml
@@ -39,29 +39,22 @@ jobs:
           cache-targets: false
 
       - name: Fetch Cargo dependency downloads
-        run: cargo fetch --locked --target aarch64-apple-darwin
+        run: |
+          cargo fetch --locked --target aarch64-apple-darwin
+          cargo fetch --locked --target x86_64-apple-darwin
 
   warm:
-    name: macOS cache · ${{ matrix.arch }}
+    name: macOS cache · universal
     needs: cargo-downloads
     runs-on: macos-latest
     timeout-minutes: 60
-    strategy:
-      fail-fast: false
-      matrix:
-        # Apple Silicon only while the product is in active development; see
-        # docs/releases.md. Restoring Intel means adding its row back here, to
-        # the two macOS release matrices, and to RELEASE_PLATFORMS.
-        include:
-          - arch: aarch64
-            target: aarch64-apple-darwin
     env:
       CARGO_INCREMENTAL: 0
       RUSTC_WRAPPER: sccache
       SCCACHE_GHA_ENABLED: "true"
       # GitHub throttles sccache's many small writes. Retain its existing
       # object cache as a read-only fallback and persist build outputs below
-      # with one cache archive per architecture.
+      # with one cache archive containing both universal slices.
       SCCACHE_GHA_RW_MODE: READ_ONLY
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
@@ -69,7 +62,7 @@ jobs:
       - name: Install Rust target
         run: |
           rustup show
-          rustup target add "${{ matrix.target }}"
+          rustup target add aarch64-apple-darwin x86_64-apple-darwin
 
       - name: Restore unsigned Rust build cache
         id: rust-build-cache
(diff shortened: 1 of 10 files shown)


naingthet (commented):

**LGTM with nits.** The packaging contract is coherent end to end: sidecar lipo, bundle verification, manifest/updater keys, cache generations, and the compatibility alias all line up. Two P3s below, neither blocking.

---

### Findings

**`README.md:22` — P3 — the download button still points at the legacy alias, so the alias can never be retired**

The `href` is still `.../releases/latest/download/Tidebreak-macos-apple-silicon.dmg`. Only the badge label and `alt` text were updated. This works (the alias is a byte-identical copy), but it contradicts `docs/releases.md:165`, which this PR writes as "the universal name is canonical for new links" — the README is the single largest source of *new* clicks, and it is the one link still pointed at the compatibility name. The alias exists to keep pre-existing external links alive; pointing the canonical entry point at it makes it permanently load-bearing, so the transition it was added for can never actually complete.

Fix: point the button at `Tidebreak-macos-universal.dmg`. Both assets are attached to every release by `attach_downloads`, so this is a one-word change with no compatibility cost, and it leaves the alias serving only genuinely old third-party links.

**`.github/workflows/release.yml:375` / `.github/workflows/staging-publish.yml:190` — P3 — no `timeout-minutes` on the signing jobs, whose compile time this PR roughly triples**

`build_macos` and `build_macos_staging` carry no `timeout-minutes`, so they inherit GitHub's 360-minute default. **I confirmed this predates the PR** (`origin/main` has timeouts only at release.yml:251/752/875), so it is not a regression here — but this PR is what makes it matter. Those jobs discard all restored linked products and relink, and now do so for two Rust targets instead of one; by the figures this PR deletes from `docs/releases.md` (~7 min aarch64, ~18 min cross-compiled x86_64) that is ~25 min of compile where it used to be ~7. A hung or pathologically slow run now holds a `desktop-production` environment lock for up to six hours instead of failing fast.

Fix: add `timeout-minutes: 90` to both jobs, matching the Windows jobs' existing budget. The *prepare* jobs already have 60 and are unaffected.

---

### CHECKED

Verified by running, not by reading:

- **All 62 tests pass** — `node --test scripts/create-release-manifests.test.mjs scripts/workflow-security.test.mjs scripts/workflow-security-mutations.test.mjs` → 62/62, matching the PR body.
- **`actionlint` clean** on all three changed workflows; all three parse as YAML (9/4/2 jobs).
- **No orphaned `matrix.*` references.** Grepped all three workflows: every surviving `matrix.arch`/`matrix.target` is inside the two Windows jobs, which keep their own `strategy:` blocks (release.yml:776, 902). This is the usual way a matrix removal silently breaks — an expression that quietly evaluates to empty — and it is clean here.
- **Both Intel runtime pins are real and correct.** This is the first release that compiles and ships the x86_64 slice, so `node_install.rs:89` and `office_install.rs:96` go live having never been exercised. I downloaded both and hashed them: Node `node-v20.20.2-darwin-x64.tar.gz` → `8be6f5e4…4e61`, matching both the pin and nodejs.org's official `SHASUMS256.txt`; LibreOffice `LibreOffice_25.8.7_MacOS_x86-64.dmg` (311 MB) → `110439f2…71c5`, matching the pin. Neither Intel arm is a stub; both are fully populated.
- **New sidecar filenames are gitignored** — `git check-ignore` confirms `.gitignore:28`'s `tidebreak-host-broker-*` wildcard covers both `-x86_64-apple-darwin` and `-universal-apple-darwin`.
- **No leftover `dist/macos/aarch64` paths** anywhere in workflows or scripts.

Verified by reading, worth not re-litigating:

- **Artifact naming does not depend on Tauri's bundler naming.** "Verify and collect signed artifacts" globs `bundle/macos/*.app` and `bundle/dmg/*.dmg`, asserts exactly one of each, then renames to `Tidebreak_${VERSION}_universal.*`. So whatever tauri-bundler calls the universal DMG is irrelevant, and the manifest script's constructed `Tidebreak_<v>_universal<ext>` names match by construction.
- **The compatibility-alias block is correctly ordered.** The `dmg_count == 1` guard runs *before* the `cp`, so the alias does not trip its own check. Both `.sha256` files use the same two-space `sha256sum` format, and `gh release upload downloads/*` picks up all four files.
- **The signed-bundle cache invariant holds.** No `target/universal-apple-darwin/**` path appears in any cache save/restore list, so no bundle or signing material can round-trip through the cache. `binaries/tidebreak-host-broker-universal-apple-darwin` *is* cached but is explicitly discarded in both signing jobs before rebuild — the right call, since Tauri copies that file into the signed product verbatim.
- **The discard-list asymmetry is correct, not an oversight.** `universal-apple-darwin` is included for `tidebreak-desktop` but not for `tidebreak-host-broker` or `libtidebreak_desktop_lib.*`, because Tauri's universal build only lipos the main executable — no universal lib or cargo-built broker is produced under that directory.
- **Updater keys are right for a universal binary.** `tauri-plugin-updater` derives its platform key from compile-time `std::env::consts::ARCH`, which is per-slice — the arm64 slice asks for `darwin-aarch64`, the x86_64 slice for `darwin-x86_64`. Emitting both keys pointed at the same archive covers both, and existing Apple Silicon installs keep updating through the unchanged `darwin-aarch64` key. (Reasoned from the plugin's key derivation, not exercised against a live feed.)
- **`prepare-sidecar.mjs` preserves non-universal behavior exactly.** With `TAURI_ENV_TARGET_TRIPLE` unset, `targets` collapses to `[hostTriple]`, no `--target` is passed, and the source path stays `target/<profile>/` — byte-identical to the pre-PR control flow. The universal branch is the only new path.
- **Fail-closed on resumed old releases is real and intentional.** `prepare-published-release.mjs:61` iterates `architectures` (now `["universal"]`), so a pre-change `macos/aarch64/...` release fails on both artifact count and filename rather than republishing a different shape — exactly as the PR body states.

PR #2078, opened 2026-08-14
feat(desktop): add native read-only workbook viewer

naingthet:
## Summary

- replace the split LibreOffice-PDF / Univer XLSX experience with one native, read-only workbook surface
- preserve inspectable cells, formulas, sheet navigation, zoom, range citations, charts, images, merged cells, authored dimensions, and workbook styling
- bundle the exact-pinned Duke WebAssembly parser locally and keep CSV/TSV on the existing lightweight grid path

## Why

The previous preview forced a tradeoff: the PDF path had reasonable print fidelity but no spreadsheet semantics, while the reconstructed grid remained inspectable but lost workbook-native drawings, charts, conditional formatting, and styling details. The new viewer reads the original OOXML into a virtualized canvas workbook so the same surface supports both visual review and cell inspection.

## Implementation

- add `@extend-ai/react-xlsx` with an exact pin and override Duke `0.1.21` to `0.1.23`; the older parser misses self-closing chart references and reports no embedded charts
- add a display-only OOXML projection that preserves cached formula results while retaining original formulas for the formula bar
- normalize semantically equivalent OOXML where the parser otherwise shifts border indices or falls back from workbook theme colors
- project standard data bars and color scales into render-only cell styles when the canvas renderer omits them
- add a fixed-light, read-only workbook toolbar with address/formula display, sheet tabs, zoom controls, selection, copying, and citation navigation
- document the architectural boundary in decision record 0020

## Compatibility and safety

- source workbook bytes are never modified or written back; all normalization happens in an in-memory display copy
- editing, paste, fill, resize, and export of a modified workbook remain disabled
- workbook parsing and the WASM asset remain local to the desktop renderer/worker; no document upload or hosted office dependency is introduced
- CSV/TSV routing and presentation preview behavior are unchanged
- both added packages are MIT licensed and pinned to exact versions

## Validation

- `pnpm exec vitest run src/document/SpreadsheetViewer.dom.test.tsx src/document/NativeSpreadsheetViewer.dom.test.tsx src/document/readOnlyWorkbookProjection.dom.test.ts`
- `pnpm exec tsc --noEmit`
- `pnpm build`
- visually verified `launch-command-center.xlsx` in the production preview: themed line and column charts, corrected borders, centered merged KPI values, progress data bars, risk color scale, sheet navigation, and read-only formula inspection

### DIFF
--- a/crates/tidebreak-desktop/ui/package.json
+++ b/crates/tidebreak-desktop/ui/package.json
@@ -13,6 +13,8 @@
     "tauri": "cargo tauri"
   },
   "dependencies": {
+    "@dukelib/sheets-wasm": "0.1.23",
+    "@extend-ai/react-xlsx": "0.16.1",
     "@radix-ui/react-alert-dialog": "1.1.23",
     "@radix-ui/react-checkbox": "1.3.11",
     "@radix-ui/react-dialog": "1.1.23",
--- a/crates/tidebreak-desktop/ui/pnpm-lock.yaml
+++ b/crates/tidebreak-desktop/ui/pnpm-lock.yaml
@@ -5,12 +5,19 @@ settings:
   excludeLinksFromLockfile: false
 
 overrides:
+  '@dukelib/sheets-wasm@0.1.21': 0.1.23
   nanoid@5.1.11: 5.1.16
 
 importers:
 
   .:
     dependencies:
+      '@dukelib/sheets-wasm':
+        specifier: 0.1.23
+        version: 0.1.23
+      '@extend-ai/react-xlsx':
+        specifier: 0.16.1
+        version: 0.16.1(react-dom@19.2.7(react@19.2.7))(react@19.2.7)
       '@radix-ui/react-alert-dialog':
         specifier: 1.1.23
         version: 1.1.23(@types/react-dom@19.2.3(@types/react@19.2.17))(@types/react@19.2.17)(react-dom@19.2.7(react@19.2.7))(react@19.2.7)
@@ -337,6 +344,9 @@ packages:
     resolution: {integrity: sha512-QxULHAm7cNu72w97JUNCBFODFaXpbDg+dP8b/oWFAZ2MTRppA3U00Y2L1HqaS4J6yBqxwa/Y3nMBaxVKbB/NsA==}
     engines: {node: '>=20.19.0'}
 
+  '@dukelib/sheets-wasm@0.1.23':
+    resolution: {integrity: sha512-E/vHXR2I/EodNCpyVcFq0p8/u5qxmKxatUZvB7bsFfghUPQ2GMxg81MBFjOEQtLKXnzR7xjXOyzPzBTLbv2e1g==}
+
   '@esbuild/aix-ppc64@0.28.1':
     resolution: {integrity: sha512-Svl7tq8k/08+p6CXPpRjQ1fKX+1odH/BQbb48fV6fj3CWHhsoIOoY87w1oHXm0qEpkIK3ZfVgp0hed3XBXzXMQ==}
     engines: {node: '>=18'}
@@ -501,6 +511,12 @@ packages:
     peerDependenciesMeta:
       '@noble/hashes':
         optional: true
+
+  '@extend-ai/react-xlsx@0.16.1':
+    resolution: {integrity: sha512-yDaBv4YzpDi91OH16oN8aBxO+JySyn1cOPdA7cus1hjDSIjU4ddCyKSPDoMG5CoFdFZQNNovjG6zjpbZQ1bfow==}
+    peerDependencies:
+      react: '>=18'
+      react-dom: '>=18'
 
   '@flatten-js/interval-tree@1.1.3':
     resolution: {integrity: sha512-xhFWUBoHJFF77cJO1D6REjdgJEMRf2Y2Z+eKEPav8evGKcLSnj1ud5pLXQSbGuxF3VSvT1rWhMfVpXEKJLTL+A==}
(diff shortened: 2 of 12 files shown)


naingthet (commented):

Changes suggested — the architecture and the ADR are sound, but the read-only projection is unguarded and workbook hyperlinks are dead in the desktop webview.

---

### `crates/tidebreak-desktop/ui/src/document/NativeSpreadsheetViewer.tsx:238` — P2: workbook hyperlinks silently do nothing (repeat of #2074)

*Reproduced by inspection of the shipped dist + this repo's own host code, not by driving the app.*

`@extend-ai/react-xlsx` 0.16.1 activates cell, image, and shape hyperlinks through exactly one path (`dist/index.js:24959`, called from `:26547`, `:28318`, `:28323`):

```js
window.open(externalTarget, "_blank", "noopener,noreferrer");
```

`src/host.ts:180-186` states the constraint directly: *"The webview swallows `window.open` and `target="_blank"` (no new-window handler by design)"*. So opening a workbook with a hyperlinked cell and clicking it produces no navigation, no error, and no feedback — the cell even gets `cursor: pointer` (`dist/index.js:22733`), so it advertises an action it cannot perform. This is the same gap #2074 closed for `DocxViewer`, which routes through `openExternal` and only falls back to `window.open` outside the native host (`DocxViewer.tsx:125-131`).

The library exposes no interception seam: no `onHyperlinkClick`/`onLinkActivate` prop, no `openHyperlink` on the controller, and cells are canvas-drawn rather than `<a>` elements, so a capture-phase click listener can't reach it (the library's handler doesn't check `defaultPrevented`). `XlsxCellStyleContext.hasHyperlink` is read-only styling.

Fix: shim `window.open` for the lifetime of the mounted viewer and route through `openExternal`, reusing `DocxViewer`'s `safeExternalHref` https allowlist:

```ts
useEffect(() => {
  const original = window.open;
  window.open = ((url?: string | URL) => {
    const safe = url ? safeExternalHref(String(url)) : null;
    if (safe) void openExternal(safe).catch(() => false);
    return null;
  }) as typeof window.open;
  return () => { window.open = original; };
}, []);
```

Note the library performs **no** URL scheme validation of any kind — the target comes straight off the OOXML hyperlink relationship — so the allowlist has to live in the shim. (`plugins.shell.open` scope already restricts to http(s), so this is belt-and-braces; and CSP `script-src 'self'` means a `javascript:` target is not currently exploitable.) If a shim is judged too invasive, an upstream callback prop is the cleaner ask — but shipping with links that silently no-op is the outcome #2074 already rejected once.

---

### `readOnlyWorkbookProjection.ts:49,104` + `NativeSpreadsheetViewer.tsx:58,130,188` — P2: `maxFileSizeBytes` is applied after the whole workbook has already been unzipped, DOM-parsed, and re-zipped

*Reproduced.* `LoadedWorkbook` runs `projectWorkbookForReadOnlyDisplay` and early-returns "Preparing workbook…" until it resolves; `useXlsxViewerController` — the only holder of `MAX_WORKBOOK_BYTES` — is not mounted until after. So the 16 MB limit cannot reject anything before the expensive work is done. Feeding the projection a 25,167,649-byte archive (1.5× the limit): it completed, produced a 25,175,036-byte output copy, and no size check fired.

Two compounding costs, all on the UI thread — `useWorker: true` only moves the *engine* parse off-thread, and per the library's own code even that path still calls `parseWorkbookImageAssets` → `unzipSync` on the main thread (`dist/index.js:8626`):

1. **No guard at all against decompression-bomb shapes.** Nothing in this PR bounds decompressed size. `entry.async("string")` materializes each `xl/worksheets/*.xml` in full, then `DOMParser` builds a DOM over it, then `XMLSerializer` re-serializes it. A small archive with a giant sheet XML hangs or OOMs the renderer. The library's own decompressed preflight (`MAX_INTERACTIVE_TOTAL_XML_BYTES`) not only runs too late to help here, it reads the zip **central directory's declared** `uncompressedSize` — an attacker-supplied number — and returns `null` (guard skipped) on any parse anomaly, so it is not a backstop you can lean on.
2. **A full DEFLATE-6 re-zip for a trivial edit.** A single self-closing `<border/>` sets `changed = true` (`:56`), which re-compresses the entire archive at level 6 — every `xl/media` byte included — into a second full-size `ArrayBuffer` held alongside the source. In my run that path recompressed 25 MB to satisfy one border expansion.

Fix, all three cheap:
- In `LoadedWorkbook`, reject `data.byteLength > MAX_WORKBOOK_BYTES` before calling the projection and render the existing too-large/error state.
- In the projection, accumulate decompressed entry sizes (JSZip exposes `file._data.uncompressedSize`) and bail past a cap before calling `.async("string")`.
- Use `compression: "STORE"` in `zip.generateAsync` — the display copy is in-memory and never persisted, so deflating it buys nothing.

---

### `readOnlyWorkbookProjection.ts:73-87` — P2: shared formulas lose the formula bar, and the stripped master leaves dangling `si` references

*Reproduced.* Excel writes `t="shared"` for essentially every filled-down formula column: one master cell carries the text, the rest carry `<f t="shared" si="N"/>` with no text. Input:

```xml
<c r="C1"><f t="shared" ref="C1:C3" si="0">A1*B1</f><v>10</v></c>
<c r="C2"><f t="shared" si="0"/><v>20</v></c>
<c r="C3"><f t="shared" si="0"/></c>
```

Output of `projectWorkbookForReadOnlyDisplay`:

```
formulasBySheet: {"0":{"C1":"=A1*B1"}}
sheet1.xml: <c r="C1"><v>10</v></c><c r="C2"><v>20</v></c><c r="C3"><f t="shared" si="0"/></c>
```

Two failures:

- **Fidelity.** Only C1 is recorded, because `formula.textContent` is empty for the members. Selecting C2 falls through to `controller.selectedFormula || controller.selectedValue` (`NativeSpreadsheetViewer.tsx:341-344`) — and since the display copy no longer carries `<f>` on C2 either, the bar shows `20`. Read-only formula inspection, the feature the ADR leads with ("The wrapper retains the original formula separately for the formula bar"), is unavailable on most formula cells of a real Excel-authored workbook.
- **Correctness.** C3 has no cached `<v>`, so its `<f>` is kept — but the group's *definition* was just deleted from C1, so `si="0"` now resolves to nothing. The comment at `:82` says these cells "still go through the workbook engine as a best effort"; that best effort now has no formula to expand. (Engine behavior on the dangling reference is reasoned, not executed — I did not run the wasm.)

Fix: expand the group before stripping. When a `<f t="shared" ref=… si=N>` master is seen, record its text for every address in `ref` (ideally A1-translated per offset; even the untranslated master text in the bar beats nothing), and don't remove the master `<f>` unless every member of that `si` group also has a cached `<v>`.

---

### `readOnlyWorkbookProjection.ts:546-555` — P3: a `<sheet>` with an unresolvable relationship shifts every later sheet's index

*Reproduced.* `readWorkbookSheetPaths` does `.map(…).filter(path => path !== null)`, so the returned array is positional over *resolvable* sheets, while `sheetIndex` is consumed as `workbookSheetIndex` (`NativeSpreadsheetViewer.tsx:284,499`). A workbook whose first `<sheet>` names a missing `r:id` projects the second sheet's formulas and conditional styles under key `0`:

```
MISSING-REL formulasBySheet keys: {"0":{"A1":"=SUM(1,2)"}}   // this is workbook sheet index 1
```

Result is data-bar/color-scale fills and formula-bar text painted onto the wrong sheet rather than simply being absent. Malformed input only, but the fix is one line: keep the array positional (`(string | null)[]`, `if (!path) return` inside the loop) instead of filtering.

---

### CHECKED

- `pnpm test` (full UI suite) — green. `pnpm exec tsc --noEmit` — clean. `pnpm build` — clean (39.5s; only the pre-existing 500 kB chunk-size advisory). No `lint` script exists; CI's UI lane is `pnpm test` + `pnpm build` per `CLAUDE.md`.
- **`.xls` routing is fine.** `SPREADSHEET_MEDIA_TYPES` still includes `application/vnd.ms-excel`, `isZip()` correctly passes BIFF through unprojected, and Duke ships a real `duke-sheets-xls` BIFF parser (plus `duke-sheets-xlsb`) — `workbookFileName()`'s `workbook.xls` dispatch is load-bearing, not decorative. No regression against the `ConvertedOfficeViewer` path this replaces.
- **No phone-home.** One `fetch(` in the library, on the `src` prop this viewer never passes. Worker is `new Worker(new URL("./xlsx-worker.js", import.meta.url))`; wasm is a Vite asset URL. `showImages` blobs bytes from inside the zip only — external image relationships are dropped (`dist/index.js:5795`), and `externalLink`/`externalReference`/`oleObject` appear nowhere in either bundle. No telemetry. The wasm's import list is crypto/Date/JS plumbing — no network capability. `tauri.conf.json` CSP (`connect-src` without external hosts, `img-src 'self' asset: blob: data:`) is a hard second line here.
- **Read-only holds.** No FS write path; input is an `ArrayBuffer`. `startEditing`/`commitEditing`/paste all early-return on `readOnly`. The default toolbar's download re-emits the *original* buffer — and it's disabled anyway (`showDefaultToolbar={false}`). `controller.exportXlsx`/`exportCsv` do produce mutated bytes and are *not* gated on `readOnly`, but they're unreachable here: the custom toolbar doesn't wire them, and the read-only worker path leaves `workbook` null so they no-op. Worth not wiring them into `WorkbookToolbar` later. The projection mutates only an in-memory JSZip and returns `source` untouched when nothing changed.
- **XSS surface is genuinely small** because cells are canvas-drawn, not DOM-rendered — no `dangerouslySetInnerHTML`, no `<a>` from cell content. SVG images from the workbook go through ``, which blocks script.
- XML hardening probes against the projection's `DOMParser`: billion-laughs entity expansion did not expand; an XXE `file:///etc/passwd` entity produced a parse error caught by `parseXml`; a `../../../../etc/passwd` relationship target resolved harmlessly to no sheet. All under jsdom, so this is directional for the WKWebView, not conclusive.
- Dependency hygiene matches `CLAUDE.md`: both packages exact-pinned, both MIT, `THIRD-PARTY-NOTICES.md` regenerated (464 packages / 593 texts), single Duke copy in the store at 0.1.23 with the `pnpm-workspace.yaml` override, no version skew. One thing to keep in mind on the next bump: the override only redirects the JS-resolved import, and `@dukelib/sheets-wasm` also ships its own prebuilt `dist/duke_sheets_wasm_bg.wasm` — `setWasmSource` currently points at the 0.1.23 package's binary, which is correct, but it's a footgun if the override ever drifts.
- `axisOffset`'s hidden-row/column arithmetic is correct for both the override and no-override cases. `cellAddress`/`parseAddress` round-trip past column Z (`AA`, `AAA`).
- Error and empty states are distinct and all four reachable: download failure, projection failure, engine failure (surfacing `error.message`), and loading/progress. Citation navigation resets `appliedRef` when the range clears.
- ADR 0020 records the rejected alternatives and a revisit trigger, per `CLAUDE.md`.

PR #2080, opened 2026-08-14
test(release): allow canonical universal download link

naingthet:
## Summary
- let the trusted release policy accept either the compatibility macOS DMG link or the new canonical universal DMG link
- preserve the requirement that the README points at an asset attached to every release

## Validation
- `git diff --check`

Transition-policy prerequisite for the non-blocking review follow-up on #2076.

### DIFF
--- a/scripts/workflow-security.test.mjs
+++ b/scripts/workflow-security.test.mjs
@@ -1415,7 +1415,7 @@ test("GitHub release downloads are copied from the hosted release", () => {
 
   assert.match(
     readFileSync(repositoryFile("README.md"), "utf8"),
-    /releases\/latest\/download\/Tidebreak-macos-apple-silicon\.dmg/,
+    /releases\/latest\/download\/Tidebreak-macos-(?:universal|apple-silicon)\.dmg/,
     "the README download link must match the published asset name",
   );
 });


naingthet (commented):

Changes suggested (small — one assertion). The framing is right and the loosening is the correct move, but the widened regex opens a branch that nothing verifies, so the PR's stated invariant ("preserve the requirement that the README points at an asset attached to every release") does not hold for the new `universal` branch.

---

**`scripts/workflow-security.test.mjs:1418` (with `:1409-1411`) — P2 — the `universal` half of the alternation is unguarded**

The new alternation lets the README link resolve to `Tidebreak-macos-universal.dmg`. The only thing that would confirm that asset is actually published is the block at 1409-1411:

```js
if (/Tidebreak-macos-universal\.dmg/.test(attachJob)) {
  assert.match(attachJob, /Tidebreak-macos-universal\.dmg/);
}
```

That is a tautology — the assertion can only run when it is already true, so it can never fail. It is pre-existing (landed in #2077, present on `origin/main`), *not* introduced here, but this PR is what makes it load-bearing: on `main` the README regex forced `apple-silicon`, whose presence is hard-asserted at line 1412, so the pairing was closed. After this PR the README may name `universal`, and that name has no real assertion behind it.

Reproduced, not reasoned. From `c2216e25`:
1. Point `README.md:22` at `Tidebreak-macos-universal.dmg` (the intended end state of this transition).
2. In `release.yml`'s `attach_downloads`, remove the universal asset — map `macos/universal` to the `apple-silicon` name and drop the `universal_dmg` / `[[ -f ]]` / `cp` alias lines. Verified `Tidebreak-macos-universal.dmg` then appears **0** times in the job.
3. `node --test scripts/workflow-security.test.mjs` → **29 pass, 0 fail.**

So the shippable end state of this transition — README advertising a universal DMG the release never attaches, i.e. a permanently 404ing download button — passes the very policy test meant to prevent it. For contrast, pointing the README at a bogus name does fail (28 pass / 1 fail), and the `apple-silicon` branch is still genuinely covered by line 1412.

Fix — cross-check the two instead of listing accepted names, which also removes the hardcoded-name drift:

```js
const readme = readFileSync(repositoryFile("README.md"), "utf8");
const [, readmeAsset] =
  readme.match(/releases\/latest\/download\/(Tidebreak-macos-[\w-]+\.dmg)/) ?? [];
assert.ok(readmeAsset, "the README must link a macOS download");
assert.match(
  attachJob,
  new RegExp(readmeAsset.replace(/\./g, "\\.")),
  "the README download link must match an asset attach_downloads publishes",
);
```

This cannot pass vacuously, needs no edit when the README flips to universal, and fails on the MUT-3 mutation above. Worth also making 1409-1411 unconditional in the same pass — `release.yml:1357` publishes the universal DMG today, so the guard costs nothing and closes the hole at its source.

**`docs-site/content/docs/installation.mdx:10` — P3 — second download link, unguarded and soon inaccurate** (pre-existing; flagging as transition scope, not a defect in this diff)

A second permanent link points at `Tidebreak-macos-apple-silicon.dmg`, and no test references it — confirmed by grep across `scripts/*.mjs`. Whatever policy ends up guarding the README should cover this file too, or the alias becomes load-bearing here for exactly the reason the #2076 review raised for the README. Its link text also reads "Download for macOS (Apple Silicon)", which is wrong once the artifact is a universal binary.

---

**CHECKED**

- Framing verified independently: `README.md:22` is still the legacy `apple-silicon` alias, while `docs/releases.md:164-166` declares `Tidebreak-macos-universal.dmg` canonical and the alias a transition-only copy. This PR is the correct prerequisite for flipping it.
- Assertion matches the real workflow: `release.yml:1357` names the universal DMG from the manifest; `1381-1388` copies it to the compatibility name with a fresh `.sha256`, so both alternation branches are genuinely attached to every release **today**.
- Baseline suite green at `c2216e25`: 29 pass / 0 fail.
- The `apple-silicon` branch of the guard does fail when broken (bogus README name → 1 failure), so the test is not vacuous across the board — only on the new branch.
- Confirmed line 1409-1411 exists verbatim on `origin/main`; not a regression introduced by this PR.
- Swept for other consumers of `/releases/latest/download/` — only `README.md` and `docs-site/.../installation.mdx` are live links; the `docs/` hits are prose.
- All mutations reverted; working tree clean (`git status --porcelain` empty) before posting.

PR #2083, opened 2026-08-14
feat(desktop): add first-task walkthrough

naingthet:
## Summary

- welcome new users from the home screen with a lightweight first-task setup entry point
- walk through model choice, internet access, permission level, and attachments using the existing composer controls
- remember completion or skip locally so the setup prompt stays out of the way afterward
- keep the existing empty-chat welcome state and avoid adding a help agent, backend contract, or new dependency

## Validation

- `pnpm test` (147 files, 992 tests)
- `pnpm build`
- `git diff --check`

## Compatibility and risk

This is a desktop UI-only change. It adds semantic targeting attributes to existing controls and a localStorage preference; it does not change provider, model, permission, attachment, or wire behavior.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/AppShell.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/AppShell.dom.test.tsx
@@ -290,7 +290,7 @@ describe("app shell", () => {
   it("opens on home rather than on a conversation", { timeout: 15000 }, async () => {
     const { router } = await mountApp();
 
-    expect(await screen.findByText("How can I help?")).toBeInTheDocument();
+    expect(await screen.findByText("Welcome to Tidebreak")).toBeInTheDocument();
     expect(router.state.location.pathname).toBe("/");
     // Home used to create a chat on every cold start, leaving an empty one
     // behind whenever the reader did not use it.
@@ -335,7 +335,7 @@ describe("app shell", () => {
     // fine, so the shell stands and the failure is reported where the list
     // would be. The settled load also frees the deep link to redirect home
     // rather than wait on a fetch that already failed.
-    expect(await screen.findByText("How can I help?")).toBeInTheDocument();
+    expect(await screen.findByText("Welcome to Tidebreak")).toBeInTheDocument();
     await waitFor(() => expect(router.state.location.pathname).toBe("/"));
     expect(
       await screen.findByText(/Could not load chats/),
@@ -353,7 +353,7 @@ describe("app shell", () => {
   it("starts a conversation from the home composer and sends what was written", async () => {
     const user = userEvent.setup();
     const { router } = await mountApp();
-    await screen.findByText("How can I help?");
+    await screen.findByText("Welcome to Tidebreak");
 
     await user.type(
       screen.getByRole("textbox", { name: "Message" }),
@@ -384,7 +384,7 @@ describe("app shell", () => {
   it("opens a conversation from the sidebar", async () => {
     const user = userEvent.setup();
     const { router } = await mountApp();
-    await screen.findByText("How can I help?");
+    await screen.findByText("Welcome to Tidebreak");
 
     const chatList = screen.getByLabelText("Chats");
     const [row] = screen
@@ -496,7 +496,7 @@ describe("app shell", () => {
 
   it("keeps chat-scoped places with the conversation", async () => {
     await mountApp();
-    await screen.findByText("How can I help?");
+    await screen.findByText("Welcome to Tidebreak");
     // Not disabled — absent. Home has no conversation for the chip to
     // describe, and the rail itself carries nothing chat-scoped anymore.
     expect(screen.queryByLabelText("Chat activity")).not.toBeInTheDocument();
@@ -529,7 +529,7 @@ describe("app shell", () => {
   it("remembers the collapsed rail across a restart", async () => {
     const user = userEvent.setup();
     await mountApp();
-    await screen.findByText("How can I help?");
+    await screen.findByText("Welcome to Tidebreak");
 
     await user.click(screen.getByRole("button", { name: "Collapse sidebar" }));
 
@@ -618,7 +618,7 @@ describe("app shell", () => {
   it("creates a project from the dialog and opens a chat inside it", async () => {
     const user = userEvent.setup();
     const { router } = await mountApp();
-    await screen.findByText("How can I help?");
+    await screen.findByText("Welcome to Tidebreak");
 
     await user.click(screen.getByRole("button", { name: "New project" }));
     const name = await screen.findByRole("textbox", { name: "Project name" });
--- a/crates/tidebreak-desktop/ui/src/ComposerToolsMenu.tsx
+++ b/crates/tidebreak-desktop/ui/src/ComposerToolsMenu.tsx
@@ -97,6 +97,7 @@ export function ComposerToolsMenu({
               variant="ghost"
               size="icon-8"
               aria-label="Tools"
+              data-first-task-target="tools"
               disabled={disabled}
             >
               <Plus size={16} />
--- a/crates/tidebreak-desktop/ui/src/FirstTaskWalkthrough.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/FirstTaskWalkthrough.dom.test.tsx
@@ -1,6 +1,8 @@
 // @vitest-environment jsdom
 
 import { cleanup, fireEvent, render, screen, waitFor } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { useState } from "react";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 import {
@@ -20,18 +22,33 @@ function Targets() {
   );
 }
 
+function WalkthroughHarness({
+  onClose,
+}: {
+  onClose: (outcome: "completed" | "skipped") => void;
+}) {
+  const [open, setOpen] = useState(true);
+  return (
+    <>
+      <Targets />
+      <FirstTaskWalkthrough
+        open={open}
+        onClose={(outcome) => {
+          onClose(outcome);
+          setOpen(false);
+        }}
+      />
+    </>
+  );
+}
+
 beforeEach(() => window.localStorage.clear());
 afterEach(cleanup);
 
 describe("FirstTaskWalkthrough", () => {
   it("walks the four setup controls, remembers completion, and returns focus", async () => {
     const onClose = vi.fn();
-    render(
-      <>
-        <Targets />
-        <FirstTaskWalkthrough open onClose={onClose} />
-      </>,
-    );
+    render(<WalkthroughHarness onClose={onClose} />);
 
     expect(
       screen.getByRole("heading", { name: "Choose a model" }),
@@ -60,20 +77,36 @@ describe("FirstTaskWalkthrough", () => {
     );
   });
 
-  it("treats Escape as an intentional skip", () => {
+  it("treats Escape as an intentional skip", async () => {
+    const user = userEvent.setup();
     const onClose = vi.fn();
-    render(
-      <>
-        <Targets />
-        <FirstTaskWalkthrough open onClose={onClose} />
-      </>,
-    );
+    render(<WalkthroughHarness onClose={onClose} />);
 
-    fireEvent.keyDown(window, { key: "Escape" });
+    await user.keyboard("{Escape}");
 
     expect(onClose).toHaveBeenCalledWith("skipped");
     expect(window.localStorage.getItem(FIRST_TASK_WALKTHROUGH_KEY)).toBe(
       "skipped",
+    );
+  });
+
+  it("keeps keyboard focus inside the walkthrough through layout changes", async () => {
+    const user = userEvent.setup();
+    render(
+      <>
+        <Targets />
+        <FirstTaskWalkthrough open onClose={vi.fn()} />
+      </>,
+    );
+
+    const next = screen.getByRole("button", { name: "Next" });
+    next.focus();
+    fireEvent(window, new Event("resize"));
+    expect(next).toHaveFocus();
+
+    await user.tab();
+    expect(screen.getByRole("dialog")).toContainElement(
+      document.activeElement as HTMLElement | null,
     );
   });
 });
--- a/crates/tidebreak-desktop/ui/src/FirstTaskWalkthrough.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/FirstTaskWalkthrough.dom.test.tsx
@@ -109,4 +109,25 @@ describe("FirstTaskWalkthrough", () => {
       document.activeElement as HTMLElement | null,
     );
   });
+
+  it("does not turn a click on the highlighted control into a permanent skip", async () => {
+    const onClose = vi.fn();
+    render(<WalkthroughHarness onClose={onClose} />);
+    await waitFor(() => expect(screen.getByRole("dialog")).toBeInTheDocument());
+
+    const target = screen.getByText("Model");
+    fireEvent.pointerDown(target);
+    fireEvent.mouseDown(target);
+    fireEvent.pointerUp(target);
+    fireEvent.mouseUp(target);
+    fireEvent.click(target);
+
+    expect(onClose).not.toHaveBeenCalled();
+    expect(window.localStorage.getItem(FIRST_TASK_WALKTHROUGH_KEY)).toBeNull();
+    expect(screen.getByRole("dialog")).toBeInTheDocument();
+    expect(screen.getByText("1 of 4").parentElement?.parentElement).toHaveAttribute(
+      "aria-live",
+      "polite",
+    );
+  });
 });
--- a/crates/tidebreak-desktop/ui/src/FirstTaskWalkthrough.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/FirstTaskWalkthrough.dom.test.tsx
@@ -1,0 +1,79 @@
+// @vitest-environment jsdom
+
+import { cleanup, fireEvent, render, screen, waitFor } from "@testing-library/react";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+import {
+  FIRST_TASK_WALKTHROUGH_KEY,
+  FirstTaskWalkthrough,
+  shouldOfferFirstTaskWalkthrough,
+} from "./FirstTaskWalkthrough";
+
+function Targets() {
+  return (
+    <>
+      <button data-first-task-target="model">Model</button>
+      <button data-first-task-target="tools">Tools</button>
+      <button data-first-task-target="permissions">Ask</button>
+      <textarea data-composer-input aria-label="Message" />
+    </>
+  );
+}
+
+beforeEach(() => window.localStorage.clear());
+afterEach(cleanup);
+
+describe("FirstTaskWalkthrough", () => {
+  it("walks the four setup controls, remembers completion, and returns focus", async () => {
+    const onClose = vi.fn();
+    render(
+      <>
+        <Targets />
+        <FirstTaskWalkthrough open onClose={onClose} />
+      </>,
+    );
+
+    expect(
+      screen.getByRole("heading", { name: "Choose a model" }),
+    ).toBeInTheDocument();
+    fireEvent.click(screen.getByRole("button", { name: "Next" }));
+    expect(
+      screen.getByRole("heading", { name: "Set internet access" }),
+    ).toBeInTheDocument();
+    fireEvent.click(screen.getByRole("button", { name: "Next" }));
+    expect(
+      screen.getByRole("heading", { name: "Choose a permission level" }),
+    ).toBeInTheDocument();
+    fireEvent.click(screen.getByRole("button", { name: "Next" }));
+    expect(
+      screen.getByRole("heading", { name: "Add attachments" }),
+    ).toBeInTheDocument();
+    fireEvent.click(screen.getByRole("button", { name: "Done" }));
+
+    expect(onClose).toHaveBeenCalledWith("completed");
+    expect(window.localStorage.getItem(FIRST_TASK_WALKTHROUGH_KEY)).toBe(
+      "completed",
+    );
+    expect(shouldOfferFirstTaskWalkthrough()).toBe(false);
+    await waitFor(() =>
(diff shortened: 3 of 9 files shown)


naingthet (commented):

**LGTM with nits** — the trigger/persistence design is the part that usually goes wrong here, and it's right. The findings are focus management in the hand-rolled dialog, plus three copy/scope details.

---

### Findings

**P2 — `crates/tidebreak-desktop/ui/src/FirstTaskWalkthrough.tsx:135-138` — every scroll or resize yanks focus back to the dialog container.**
`update()` (line 111) always allocates a fresh `ElementRect`, so `setRect` changes identity on every fire even when the geometry is unchanged. `rect` is in the focus effect's dep array, so each fire re-runs `dialogRef.current?.focus()`. The `scroll` listener is registered with `capture: true` on `window` (line 127), so any scrolling element in the app feeds it, and `resize` feeds it directly.

*Reproduced.* Scratch test in the worktree: render the walkthrough with the four targets present, `screen.getByRole("button", {name:"Next"}).focus()`, then `fireEvent(window, new Event("resize"))` — focus is on the dialog wrapper, not `Next`. Same failure with `fireEvent.scroll(window)`. A keyboard user who tabs to `Back`/`Next`/`Skip setup` and then resizes the window loses their place and has to tab in again.

Fix: drop `rect` from the deps — `[open, stepIndex]` is what this effect actually wants — and guard the steal:
```ts
useEffect(() => {
  if (!open) return;
  if (dialogRef.current?.contains(document.activeElement)) return;
  dialogRef.current?.focus();
}, [open, stepIndex]);
```

**P2 — `crates/tidebreak-desktop/ui/src/FirstTaskWalkthrough.tsx:192-198` — `aria-modal="true"` with no focus trap.**
Nothing marks the app behind the overlay inert or `aria-hidden`, and there is no Tab handling. The backdrop (line 178) blocks pointer events, so background controls are dimmed and unclickable — but still tabbable.

*Reproduced.* Focus `Next`, press Tab once: `document.activeElement` is outside the dialog (the portal is appended to `document.body`, so Tab wraps into the page content). The net effect is a keyboard user tabbing into invisible controls they cannot activate, while `aria-modal` tells the screen reader that content doesn't exist.

Fix: build on the existing primitive rather than hand-rolling. `components/ui/dialog.tsx` wraps `@radix-ui/react-dialog` and already gives focus trap, focus restore, and Escape for free; `ShortcutsDialog.tsx`, `sidebar/NewProjectDialog.tsx`, and `NetworkPolicyDialog.tsx` all use it. Keep the anchored positioning and the spotlight ring, and let `DialogContent` own focus. (There is no manual focus trap anywhere in this codebase to copy instead.)

**P2 — `crates/tidebreak-desktop/ui/src/FirstTaskWalkthrough.tsx:37-42` — the attachments step instructs an action home cannot offer.**
Step 4 says "Open Tools to attach files for this task **or connect a folder**". `HomeRoute.tsx` never passes a `folders` prop to `Composer`, so `ComposerToolsMenu` receives no `attachFolder` and the "Attach folder" row (`ComposerToolsMenu.tsx:121-133`) never renders on home — the folder half of that sentence is wrong 100% of the time on the only surface the walkthrough runs on.

It degrades further off-native: `HomeRoute.tsx:436` passes `onAttach: hasNativeHost() ? onAttach : undefined`, so on a non-native host the "Attach files" row is gone too, and a step titled "Add attachments" spotlights a Tools menu containing no attachment options at all.

*Verified by reading the prop wiring, not by driving the app.* Fix: drop "or connect a folder" from the copy, and either skip the step when `attachFiles` is unavailable or reword it to what the menu actually shows.

**P3 — `crates/tidebreak-desktop/ui/src/FirstTaskWalkthrough.tsx:35` — the permissions step omits `Plan`.**
The copy enumerates "Ask", "Auto", and "Allow all", but `PERMISSION_MODE_SCALE` (`PermissionModeMenu.tsx:42-48`) has four entries and `Plan` is the **first** one the user sees. The step also says "Ask is a safe place to start" while `Plan` sits above it and is strictly safer ("Read-only: the agent explores and proposes a plan"). A first-run user reading this and then opening the menu finds an unexplained top item. Either name all four or reword so the list doesn't read as exhaustive.

**P3 — `crates/tidebreak-desktop/ui/src/HomeRoute.tsx:390-391` — the first-run heading is permanent.**
`heading` and `description` are passed unconditionally; only `onStartWalkthrough` is gated on `walkthroughAvailable` (line 392). So a user who completed or skipped the walkthrough — and every existing user after the update — sees "Welcome to Tidebreak" / "Choose how the agent works, add what it needs, and start with a real task." on home forever, where it used to say "How can I help?". If the rebrand is deliberate, ignore this. If the framing was meant to be first-run only, gate it the same way as the button and let `WelcomeState`'s defaults stand afterward.

**P3 — `crates/tidebreak-desktop/ui/src/FirstTaskWalkthrough.tsx:140-149` — Escape effect has no dependency array.**
It detaches and reattaches the `window` keydown listener on every render (and with the P2 above, that's every scroll event). Harmless today because `finish` is recreated each render anyway, but it's a trap for whoever next adds work to this effect. `[open]` plus a ref or `useCallback` for `finish` does the same job once.

---

### CHECKED

- **Full UI suite green in a clean worktree at `2f9e1fc9`**: 147 files / 992 tests passed, `tsc --noEmit` exit 0. The two new `FirstTaskWalkthrough.dom.test.tsx` cases and the updated `WelcomeState.test.tsx` case pass on their own.
- **Persistence is right.** `tidebreak.first-task-walkthrough.v1` follows the established `UiStore.ts` pattern (`tidebreak.sidebar-collapsed`, `tidebreak.composer.sendMode`): namespaced, versioned, try/catch on both read and write, and `shouldOfferFirstTaskWalkthrough()` fails open to `true` when storage throws. There is no per-account or per-profile concept in the desktop UI, so per-install and per-user are the same thing here — no multi-profile leak.
- **No re-fire-forever and no write-before-completion.** The walkthrough never auto-opens; it's an opt-in button on the welcome screen. The flag is written only inside `finish()`, so quitting mid-walkthrough correctly re-offers it on next launch, and both `Done` and every skip path record an outcome.
- **Escapability.** `Skip setup` and Escape are present on all four steps, including mid-flight; `Back` appears from step 2. Nothing in the overlay depends on a network call or app state, so there is no state where it can wedge offline or on a failed action.
- **All four step targets resolve on home.** `ModelMenu.tsx:423`, `ComposerToolsMenu.tsx:100`, `PermissionModeMenu.tsx:173`. `ComposerToolsMenu` cannot early-return on home because `network` is always passed, so `hasSettings` is true. A missing target degrades to a centered dialog rather than a crash.
- **Focus return works against the real DOM.** `[data-composer-input]` is the real composer textarea (`Composer.tsx:984`) and reuses the same hook `AppShell.tsx:49` already relies on — the test's fake target matches production.
- **No leaked subscriptions.** `ResizeObserver` is disconnected and both window listeners are removed in the effect cleanup; there is no polling loop, interval, or timer that can outlive the component.
- **No PR #2074 hardening bypass.** The walkthrough is narration only — it renders no app-owned listing, drives no MCP native-confirmation gate, and issues no provider URL. It highlights existing controls and the user operates them normally afterward, so the hardened paths keep their own enforcement.
- **Base branch checked.** The `"How can I help?"` → `"Welcome to Tidebreak"` edits in `AppShell.dom.test.tsx` track the intentional heading change rather than papering over a break, and the old string survives as `WelcomeState`'s default.
- **i18n is not a finding.** This UI has no translation framework at all; hardcoded copy is consistent with every neighboring component.

naingthet:
Addressed the review findings in `f3fa28ff`:

- moved the walkthrough to the shared Radix dialog primitive for focus trapping, inert background content, Escape handling, and focus restoration
- added regressions for resize/scroll focus stability and keyboard focus containment
- corrected attachment copy to match home capabilities and named all four permission modes
- limited the “Welcome to Tidebreak” framing to the first-run setup state
- removed the hand-rolled Escape/focus effects

Validation: `pnpm test` (147 files / 993 tests), `pnpm build`, and `git diff --check` are green locally.

naingthet (commented):
🤖 **Thet's review agent** — automated review (round 2), not a human sign-off.

**Request changes on one item** — four of the six round-1 findings are cleanly fixed and I verified them by execution, not by reading. But the Radix migration that closed the focus-trap finding introduced a new dismissal path: a single click anywhere outside the tooltip permanently writes `skipped`, and there is no way back.

---

### Status of round-1 findings

| # | Finding | Status |
|---|---|---|
| P2 | focus stolen on scroll/resize | **Fixed** |
| P2 | `aria-modal` with no focus trap | **Fixed** (but see NEW-1) |
| P2 | step 4 instructs an impossible action | **Fixed** |
| P3 | permissions copy omits `Plan` | **Fixed** |
| P3 | heading permanent for everyone | **Fixed** |
| P3 | Escape effect missing deps | **Fixed** (effect deleted) |

**P2 focus steal — fixed, reproduced.** Both `dialogRef` effects are gone; focus now belongs to Radix. Round 1's repro is inverted in a scratch test: focus `Next`, `fireEvent(window, new Event("resize"))` → still focused; `fireEvent.scroll(window)` → still focused; and again after advancing a step, so it holds across `step.target` changes, not just step 1. The rect effect keeps `[open, step.target]` deps, so the identity churn from `update()` no longer reaches anything focus-related.

**P2 focus trap — fixed, reproduced.** 8 forward `user.tab()` and 8 `shift+Tab` all land inside `[role="dialog"]`; round 1's single-Tab escape is gone. I specifically checked the migration for the regressions this brief called out, and all of them hold: the spotlight ring still renders and still tracks the target (it kept its own `createPortal` to `document.body`, so Radix's portal does not swallow it); the `[data-composer-input]` return still works — driving all four steps to `Done` lands focus on the composer textarea and stores `completed`, so Radix's focus-restore does not beat the `requestAnimationFrame`; Escape still routes through `onOpenChange` → `finish("skipped")`; and teardown is clean — after close, `body.style.pointerEvents` is back to `""`, the `aria-hidden` Radix put on the app tree is removed, the spotlighted control is focusable again, `resize`/`scroll` listeners are removed 2-for-2, and there are zero leftover dialog or spotlight nodes.

**P2 step 4 copy — fixed.** "or connect a folder" is gone and the remaining sentence is scoped to what home actually offers.

**P3 `Plan`, P3 heading — fixed.** `Plan stays read-only` now leads the enumeration and "Ask is a safe place to start" was softened to "balanced". `HomeRoute.tsx:390-395` gates `heading`/`description` on `walkthroughAvailable`, which `onClose` flips to `false` (line 508), so the framing reverts to `WelcomeState`'s defaults after the first run.

---

### New findings

**P2 — `crates/tidebreak-desktop/ui/src/FirstTaskWalkthrough.tsx:186-189` — one stray click outside the tooltip permanently disables the walkthrough.**

`DialogContent` is used with Radix defaults, so `onInteractOutside` fires on any outside pointer sequence and dismisses the dialog. That routes into `onOpenChange(false)` → `finish("skipped")` (lines 168-170), which writes `skipped` to localStorage. `shouldOfferFirstTaskWalkthrough()` then returns `false` forever, `walkthroughAvailable` stays `false`, and there is no UI anywhere to re-open it — so an accidental click is unrecoverable short of clearing site data.

This is new in `f3fa28ff`. At `2f9e1fc9` the backdrop was `<div className="absolute inset-0 bg-black/45" aria-hidden="true" />` with no handler — inert by construction. The migration that closed the focus-trap finding opened an unintended dismissal path in its place.

*Reproduced.* Scratch test, full pointer sequence (`pointerdown`/`mousedown`/`pointerup`/`mouseup`/`click`) after Radix arms its listeners:
- click on the dimmed overlay → `onClose` called with `"skipped"`, storage `"skipped"`
- click on the spotlighted Tools button itself → `onClose` called with `"skipped"`, storage `"skipped"`
- click inside the dialog → not dismissed (correct)

The second case is the one that bites: the walkthrough's whole job is to point at a control and describe it, so "user clicks the thing the arrow points at" is the single most likely interaction, and it silently ends setup. Note `Escape` → skipped is deliberate and asserted by a test; an outside click is not — nothing in the PR body or tests mentions it.

Fix — keep dismissal to the three intentional exits (`Skip setup`, Escape, `Done`):
```tsx
<DialogContent
  withCloseButton={false}
  onInteractOutside={(event) => event.preventDefault()}
  ...
```
If outside-click dismissal is actually wanted, it should at least not persist — treat it as "close without recording an outcome" so the walkthrough is re-offered, the way quitting mid-flight already does.

**P3 — `crates/tidebreak-desktop/ui/src/FirstTaskWalkthrough.tsx:201-212` — step transitions are silent for screen readers.**

The old outer wrapper carried `aria-live="polite"` (`2f9e1fc9`, line 177); the migration dropped it and nothing replaced it — there is no `aria-live` left in the file. Clicking `Next` keeps focus on the `Next` button and swaps `DialogTitle`/`DialogDescription` text in place. Radix announces title and description on *open*, not on content change, so a screen-reader user hears the first step and then nothing for steps 2-4, including the "2 of 4" progress. *Reasoned from the diff plus Radix's announcement behavior, not driven with a real AT.* Fix: put `aria-live="polite"` back on the step-content wrapper inside `DialogContent`.

---

### CHECKED

- **Suite green at `f3fa28ff` in a clean dedicated worktree**: 147 files / 993 tests passed, `tsc --noEmit` exit 0. The three reworked `FirstTaskWalkthrough.dom.test.tsx` cases pass, and the new `WalkthroughHarness` correctly drives `open` back to `false` so the close path is actually exercised rather than mocked.
- **The new tests are real coverage, not assertion-shaped.** `keeps keyboard focus inside the walkthrough through layout changes` fails against the pre-fix component, so it is a genuine regression guard. It only tabs once — the 16-tab version also passes, so the guard is weaker than the behavior, not wrong.
- **The migration did not regress positioning.** Inline `style` top/left/transform override the primitive's `top-1/2 left-1/2 translate-*` utilities, and `translate-x-0 translate-y-0` + tailwind-merge handle the case where inline `transform` is `undefined`. The no-target fallback (`rect === null` → centered, `translateY(-50%)`) still holds.
- **`z-index` layering is coherent**: overlay `z-[100]` (overriding the primitive's `z-50`), spotlight `z-[101]`, content `z-[102]`.
- **Persistence semantics unchanged** apart from the new path in NEW-1: `completed` still written only from `Done`, quitting mid-flight still re-offers, `storeOutcome` still try/catch'd, `shouldOfferFirstTaskWalkthrough` still fails open.
- **Round-1 items I did not re-litigate** (untouched by this commit): localStorage key/`UiStore.ts` precedent, target resolution for all four steps, no PR #2074 hardening bypass, base-branch check on the `AppShell.dom.test.tsx` heading assertions.
- **Scratch tests removed; worktree torn down.**

naingthet:
Round 2 is addressed in `55220868`:

- prevented outside interactions from dismissing or persisting a skip; only Skip setup, Escape, and Done now close the walkthrough
- restored a polite, atomic live region around progress/title/description so step changes are announced
- added a regression that exercises a full pointer sequence on the highlighted control and confirms the walkthrough remains open with no stored outcome

Validation: `pnpm test` (147 files / 994 tests), `pnpm build`, and `git diff --check` are green locally. Auto-merge remains disabled.

naingthet (commented):
> 🤖 **Thet's review agent** — automated review (round 3), not a human sign-off.

**No findings.** Both round-2 items are genuinely fixed — I verified each one in both directions (the assertion passes on `55220868` and fails against the `f3fa28ff` component), and the specific regression this round was told to hunt for — a dismissal fix that leaves the dialog undismissable — did not happen. Suite and `tsc --noEmit` green locally.

---

### Status of round-2 findings

| # | Finding | Status |
|---|---|---|
| P2 | outside click permanently writes `skipped` | **Fixed** (reproduced both directions) |
| P3 | step transitions silent for screen readers | **Fixed** (reproduced both directions) |

**P2 outside-click dismissal — fixed, and the fix does not over-close the exits.**
`FirstTaskWalkthrough.tsx:188` now carries `onInteractOutside={(event) => event.preventDefault()}`. Radix routes `pointerDownOutside` and `focusOutside` into `onInteractOutside` but keeps `onEscapeKeyDown` on a separate path (`react-dismissable-layer@1.1.19/dist/index.mjs:56,86,95`), so preventing outside interaction cannot swallow Escape — which is the exact adjacent-bug shape this round was asked to check for.

*Reproduced.* Scratch test driving the full pointer sequence (`pointerdown`/`mousedown`/`pointerup`/`mouseup`/`click`) on the overlay, on unrelated app chrome, on the spotlighted target, and on `document.body`: dialog stays open, `onClose` never fires, `localStorage` stays `null`. The same test run against the `f3fa28ff` component fails (`onClose` called once), so the guard is real rather than assertion-shaped.

**All three intentional exits still work, verified by execution:** `Skip setup` → `onClose("skipped")` + stored `skipped` + dialog unmounts; `Escape` → same, and also on a later step mid-flow, not just step 1; `Done` after four steps → `onClose("completed")` + stored `completed` + focus lands on the composer textarea. `Back` returns to the previous step and persists nothing. So there is no undismissable state.

**No collateral from preventing outside interaction.** Focus trap still holds — 10 forward `user.tab()` and 10 `shift+Tab` all land inside `[role="dialog"]`. The spotlight ring still renders through its own `createPortal`, still carries `aria-hidden`, still stays `pointer-events-none`, and still re-resolves when the step target changes. Teardown after `Done` is clean: `body.style.pointerEvents` back to its prior value, no leftover spotlight node, the app tree no longer `aria-hidden`, and `scroll` listener adds/removes balance exactly. Worth recording for future rounds: the spotlighted control was never clickable in the first place — the `z-[100]` overlay plus Radix's modal `disableOutsidePointerEvents` cover it — so this fix removes a stray dismissal, not a working interaction. The tour is read-only by construction and `Next`/`Back` drive it.

**P3 live region — fixed.**
`FirstTaskWalkthrough.tsx:201` wraps progress + `DialogTitle` + `DialogDescription` in `<div aria-live="polite" aria-atomic="true">`.

*Reproduced.* Held a reference to the region node and clicked through all four steps: it is the **same DOM node** each time while its text goes `1 of 4`/"Choose a model" → `2 of 4`/"Set internet access" → … → `4 of 4`. That node identity is the part that matters — a wrapper remounted per step would not be announced by most screen readers, and this one is not remounted, so steps 2-4 are genuine live-region updates. Still not driven with a real AT, but the DOM precondition is now correct where before there was no `aria-live` in the file at all.

---

### Note — non-blocking, not a finding

**The walkthrough is a modal tour, and the controls it highlights are deliberately inert while it is open.** `Dialog` is `DialogPrimitive.Root` with `modal` left at its default `true`, so Radix stamps `aria-hidden` on the rest of the body (`react-dialog/dist/index.mjs:145`, `hideOthers(content)`) and sets `document.body.style.pointerEvents = "none"` (`react-dismissable-layer/dist/index.mjs:112-113`). The spotlight ring (`z-[101]`) draws above the `z-[100]` overlay, so the ring is visible while the control beneath it is dimmed and unclickable.

I considered and am **not** flagging this. The step copy is descriptive rather than imperative-now — it says where a capability lives and what it does, the flow never waits on the user to operate anything, and `Next`/`Back` drive it end to end. That is the ordinary modal-tour pattern and reads as intended design. Recording it so round 4 does not re-raise it. One thing worth knowing if it ever *is* revisited: `FirstTaskWalkthrough.dom.test.tsx:113-132` asserts only that clicking the target does not dismiss the walkthrough — it does not (and in jsdom cannot) assert the click reaches the control, so that test is not evidence of pointer passthrough either way. If descriptive-only was not the intent, the lever is `modal={false}` plus softening the overlay.

**Coverage thinness, also not flagged:** no test drives the walkthrough from the real home route. `FirstTaskWalkthrough.dom.test.tsx:17-20` stubs the three targets with hand-written buttons, and `WelcomeState.test.tsx:58-75` only asserts the button calls its prop. So nothing fails if `data-first-task-target` is dropped from `ModelMenu`/`ComposerToolsMenu`/`PermissionModeMenu` — the walkthrough would silently degrade to an un-anchored centered dialog. The degradation is graceful and CLAUDE.md explicitly discourages test volume for its own sake, so I am not asking for a test; noting the coupling for whoever next touches those triggers.

---

### CHECKED

Verified correct this round; later rounds should not re-litigate these.

- **Full UI suite green in a fresh dedicated worktree at `55220868`**: 147 files / 994 tests passed, `npx tsc --noEmit` exit 0. CI's `desktop UI` job also passes.
- **The fix commit is surgical.** `git diff f3fa28ff..55220868` touches only the one prop and the one wrapper `div`; nothing else in the component moved. The wrapper does not disturb layout — `DialogContent` is `block gap-0`, the wrapper's first child has no top margin, so the `mt-3`/`mt-1.5` rhythm is unchanged.
- **Both round-2 findings were real.** `git show f3fa28ff:…FirstTaskWalkthrough.tsx` has neither `onInteractOutside` nor any `aria-live`. Neither prior call was a false positive.
- **Diff scope is confined to 9 files** against `origin/main`, all under `crates/tidebreak-desktop/ui/src`. Nothing here is a pre-existing `origin/main` regression being misattributed.
- **Exactly one attribute site per walkthrough target in product code**: `ModelMenu.tsx:423`, `ComposerToolsMenu.tsx:100`, `PermissionModeMenu.tsx:173`. No duplicate-match hazard for `document.querySelector`, and no collision with any existing `data-*` selector convention.
- **`WelcomeState`'s other caller is unaffected.** `MessageList.tsx:379` renders it with only `onSelectPrompt`/`executionConfigClient`, so the chat empty state keeps the default "How can I help?" copy. The new props are additive with defaults.
- **The new button does not disturb welcome layout.** `styles.css:422` — `.welcome` is a centered flex column with a uniform `gap: 1.5rem` and no `:nth-child`, `:last-child`, or adjacent-sibling rules, so inserting a `<Button>` between `.welcome-copy` and `.welcome-prompts` changes nothing positionally.
- **`React.StrictMode` is on** (`main.tsx:16`). The rect effect's double-invoke is idempotent — cleanup disconnects the `ResizeObserver` and removes both listeners before the re-run — and the `!open` branch re-sets state to identical values, which React bails out of.
- **Persistence semantics unchanged and still safe.** `completed` only from `Done`; every other exit records `skipped`; quitting mid-flight persists nothing and re-offers next launch; `storeOutcome` and `shouldOfferFirstTaskWalkthrough` still try/catch and fail open.
- **Round-1 and round-2 items untouched by this commit** were not re-litigated: focus steal on scroll/resize, focus trap, step-4 copy, `Plan` in the permissions copy, heading gating, positioning math, `z-index` layering, localStorage key/versioning.

PR #2084, opened 2026-08-14
feat(desktop): adopt Extend document viewers

naingthet:
## Summary

- replace the PDF and DOCX preview implementations with the current Extend registry viewer surfaces
- render PPTX files directly with Extend while retaining LibreOffice-to-PDF for PPT, ODP, and unsupported PPTX files
- preserve the existing Extend XLSX viewer and align the four document formats around the same read-only, light-surface interaction model
- bundle PDFium, DOCX, and PPTX WebAssembly locally and record the viewer boundary in decision 0023

## Implementation

Tidebreak still owns file loading, immutable byte caching, export, citation/page state, and external-link policy. The embedded viewers receive mount-scoped blob URLs, disable their upload/download actions, and never write changes back to source files. DOCX is explicitly configured in read-only mode; its WASM importer maps OOXML into React nodes rather than inserting document-authored HTML or styles. Materialized links are limited to credential-free HTTPS URLs and routed through Tidebreak's external-link handler, while bookmarks remain internal and unsafe schemes are removed.

PDF keeps continuous navigation, thumbnails, selection, search, rotation, zoom, remembered page state, and citation targeting. PPTX gets direct slide rendering, thumbnails, and zoom; a parser failure unmounts the direct surface before entering the existing LibreOffice fallback.

The Vite development configuration leaves Extend's worker-owning packages unoptimized so their sibling parser workers remain addressable. The required CommonJS leaf dependencies are prebundled explicitly. PDFium receives an absolute local asset URL because its blob-backed worker cannot resolve a root-relative Vite URL; remote font fallback is disabled.

## Compatibility and risk

- PDF, DOCX, and PPTX sources continue to export byte-for-byte unchanged.
- Legacy PPT and ODP behavior remains on the LibreOffice compatibility path.
- PPTX has a LibreOffice fallback if native rendering rejects a deck.
- All new direct dependencies are exact-pinned and the lockfile is verified with pnpm 10.34.5 from an empty store.
- The desktop bundle grows by the local viewer engines: approximately 4.6 MB PDFium WASM, 1.15 MB DOCX WASM, and 0.57 MB PPTX WASM before compression. Heavy viewer code remains lazy-loaded and code-split.
- Third-party notices were regenerated for the added packages.

## Validation

- final GitHub Actions desktop UI lane: 151 test files and 994 tests passed; production build passed
- final GitHub Actions third-party notices lane passed
- `pnpm exec tsc --noEmit`
- focused viewer/security suite: 5 files and 8 tests passed after the final rebase
- pnpm 10.34.5 frozen install from an empty store
- `node scripts/generate-third-party-notices.mjs --check`
- `node --test scripts/third-party-notices.test.mjs`
- real-file browser QA with a 32-page PDF, a 15-page DOCX, and a 13-slide PPTX, including PDF citation targeting/search and PPTX navigation/zoom


naingthet (commented):

**Verdict: request changes.** The viewer adoption itself is well-built and the gates are green, but this branch was cut *before* #2074 and does not contain it, so merging as-is silently reopens the DOCX rendering boundary that #2074 closed 2½ hours earlier. One P1, four P2s, two P3s.

---

## The #2074 question, per viewer

I checked each newly-adopted viewer against the hardening merged in #2074 (`44e40e25`). First, the structural fact that frames everything below:

```
$ git merge-base --is-ancestor 44e40e25 9aa279a7  →  NO
$ merge-base(origin/main, 9aa279a7) = 01322965   # 44e40e25 landed on main after this
```

**#2074 is not an ancestor of this head**, which is why GitHub reports `mergeable_state: dirty`. The "before" side of the `DocxViewer.tsx` diff is the *pre*-#2074 file — so the diff reads as a clean refactor while actually being a revert of work that is on `main` today. Whoever resolves the conflict decides the security posture, and nothing in the PR flags that.

| Viewer | Content → DOM? | #2074 hardening |
|---|---|---|
| **PDF** (`PdfViewer.tsx`) | No — PDFium rasterises to ``; only selection/search *geometry* is DOM. No annotation/link plugin registered (`components/extend/pdf-viewer.tsx:2781-2818`), so no content anchors exist. | N/A — correctly needs no sanitizer |
| **PPTX** (`PresentationViewer.tsx`) | Yes — `document.createElement` per run | Not applicable pre-existing; engine has its own scheme allowlist (`http/https/mailto/tel`) |
| **DOCX** (`DocxViewer.tsx`) | **Yes** — React elements incl. `<a href={run.link}>` | **Bypassed — see P1** |

To be explicit about the round-4 prediction: `useSecureViewerLinks` **is** the second, divergent copy of the module-private logic, and it diverges in the weaker direction (P1 + P2a below).

---

## Findings

### P1 — `crates/tidebreak-desktop/ui/src/document/DocxViewer.tsx:48` — DOCX content moves from a sandboxed iframe back into the app's main DOM

On `main` today, DOCX output never touches the app document: it is parsed inside an offscreen **closed shadow root**, run through `sanitizeRenderedDocument`, and serialized into an iframe with `sandbox=""` and `default-src 'none'; script-src 'none'; object-src 'none'`. Anchors are deliberately **defanged** — no `href` at all, the validated URL rendered as text plus `title`. The design fails *closed*: `removeAttribute` runs unconditionally after the `WeakMap.set`, so a lost entry drops the display rather than surfacing an unvalidated URL.

This PR replaces all of that with `<DocxViewerPreview>`, which mounts `DocxEditorViewer` and builds React elements directly into the live app tree. The vendored renderer emits the document's URL verbatim (`node_modules/@extend-ai/react-docx/dist/index.js:49436`):

```js
if (run.link) {
  return jsx2("a", { href: run.link, target: …, rel: …, … });
}
```

I grepped the whole vendored bundle for `javascript:`, `sanitizeUrl`, `sanitizeHref`, `isSafeUrl`, `safeUrl` — **zero hits**. There is no scheme allowlist on the DOCX side (the PPTX engine has one at `@extend-ai/react-pptx/dist/index.js:9309`; DOCX does not).

The only thing left standing between a hostile DOCX and the app DOM is `useSecureViewerLinks` — a `MutationObserver` that cleans up *after* the node is already live. That is a fail-**open** model replacing a fail-**closed** one, and it is strictly weaker in three ways beyond timing: no `on*` attribute stripping, no `src` scheme restriction, no removal of `script`/`object`/`embed`/`form`/`base`, and no `</style>` raw-text neutralisation — all of which `sanitizeRenderedDocument` does over `querySelectorAll("*")`.

*Reproduced:* the ancestry check and the render path; I confirmed by reading `origin/main:DocxViewer.tsx` against this head and the vendored renderer. I did **not** build a working hostile DOCX — the gap is structural and I'm not posting a payload.

**Fix:** either keep the sandboxed-iframe boundary for DOCX (the Extend surface can render into the frame, or the frame can wrap it), or state in the decision record why the iframe is no longer required and what replaces each of its guarantees. Right now the record (`docs/decisions/0021-…md`) is `Status: Proposed`, says `Supersedes: none`, and never mentions the boundary #2074 established — so the strongest existing argument against this change isn't answered anywhere. Please also rebase onto `main` so the conflict is resolved deliberately rather than incidentally.

### P2a — `crates/tidebreak-desktop/ui/src/document/extendViewerSurface.ts:88` — credentialed HTTPS is accepted where main rejects it

The new copy checks only the protocol; main's post-#2074 predicate also requires `!parsed.username && !parsed.password`:

```js
// this PR (:88)
return parsed.protocol === "https:" ? parsed.href : null;
// origin/main
return parsed.protocol === "https:" && !parsed.username && !parsed.password ? parsed.href : null;
```

*Reproduced* with a scratch vitest against the real hook — a DOCX link of the form `https://<user>:<secret>@internal.example/report` passes validation, is written to the anchor's `href`, and is handed to `openExternal` verbatim (embedded credentials and all) on click.

**Fix:** restore the `username`/`password` rejection. Better: export the single hardened predicate from one module and delete the duplicate — this PR is exactly the right moment to make `safeExternalHref` shared rather than module-private, which is what round 4 asked for.

### P2b — `crates/tidebreak-desktop/ui/src/document/PdfViewer.tsx:41,102` — `useSecureViewerLinks` never engages in the PDF viewer

`containerRef` is attached only on the success branch (`:102`), but the error (`:82`) and loading (`:90`) branches render a `ViewerShell` **without** the ref. At mount `file.objectUrl` is `null`, so the effect runs with `containerRef.current === null` and returns early. The dep array is `[containerRef]` — a ref object with stable identity — so the effect **never re-runs** once the URL arrives. No observer, no click handler, ever.

*Reproduced:* rendering `PdfViewer` with a mocked Extend surface containing `<a href="http://insecure.example/plain">`, the href is still present after mount, `openExternal` is called **0** times, and jsdom logs `Not implemented: navigation to another Document` — i.e. the click navigated the webview instead of being intercepted. Contrast `DocxViewer.tsx:30`, where the ref is on an always-rendered outer div and the same anchor is correctly stripped.

Severity is P2 rather than P1 only because no annotation/link plugin is registered, so PDF content currently emits no anchors — this is a dead safety net that will fail silently the day one is added.

**Fix:** attach `containerRef` on all three branches (put it on a single outer wrapper, as `DocxViewer` does), or key the effect on the resolved element via a callback ref.

### P2c — `crates/tidebreak-desktop/ui/src/components/extend/docx-viewer.tsx:1254,1298` — a wedged DOCX parse spins forever, and error detection is string-matched

Two coupled problems. `isLoadingDocument` is cleared only inside `try`/`catch` (`:1298`, `:1307`) — there is no `finally` and no timeout, so if `importDocxFile` neither resolves nor rejects (wedged worker, pathological file) the spinner runs indefinitely with no way out. And the error path at `:1254` detects failure by matching the library's *status text*:

```js
status.startsWith("Failed to load file") || status === "Only .docx files are supported"
```

A wording change in a future Extend bump silently disables error surfacing and turns every failure into the infinite spinner above. There is also **no file-size guard** on the DOCX path — I grepped the library for `maxInputBytes`/`too large` and found nothing, whereas the PPTX engine enforces `DEFAULT_MAX_INPUT_BYTES = 100MB` (`@extend-ai/react-pptx/dist/index.js:11149`). *Reasoned from code, not reproduced* — I did not have a wedging fixture.

**Fix:** add a `finally` that clears the loading flag, prefer the typed `loadError` path over status-string matching, and mirror the PPTX size ceiling on DOCX.

### P2d — `crates/tidebreak-desktop/ui/src/components/extend/docx-viewer.tsx:272-287` — `useSuppressDocxPaddingWarning` patches global `console.error`

The effect overwrites the global `console.error` and restores the reference captured at effect time. Its dep (`enabled`) flips with load/error state, and two DOCX viewers can be mounted at once, so out-of-order cleanup restores a stale wrapper — `console.error` can be left permanently patched for the session. While any viewer is open it also swallows *every* app-wide `console.error` whose first argument matches `"a style property during rerender"`. *Reasoned from code.*

**Fix:** filter at the reporting boundary, or drop the suppression and fix the underlying style write.

### P3a — `crates/tidebreak-desktop/ui/src/components/extend/pdf-viewer.tsx:2684` — stale page count on rapid document switching

The open-error callback re-checks generation (`:2673` `if (openedFileRef.current === pdfFile)`) but the success callback does not. A slow open for document A completing after B was requested calls `onDocumentLoadSuccess` with A's page count while B is displayed. Content stays correct (the core reducer sets `activeDocumentId` at start of load, and B isn't in the snapshotted close list), so this is a wrong `numPages` — which then feeds the page controls and clamping in `PdfViewer.goToPage`. *Reasoned from code.* **Fix:** add the same `openedFileRef.current === pdfFile` guard inside the success callback.

### P3b — `crates/tidebreak-desktop/ui/src/components/extend/pdf-viewer.tsx:2756-2778` — revoked object URL can be resurrected

`uploadedPdfUrl` goes `null` when `src` moves away from the value the upload was keyed to, the cleanup revokes the blob URL, but `uploadedPdfFile.url` stays in state. If `src` returns to the original value the already-revoked URL is passed to `openDocumentUrl` (`:2678`) and the load fails. Not reachable through Tidebreak today (upload is disabled via `showUpload={false}`), hence P3. *Reasoned from code.* **Fix:** clear `uploadedPdfFile` in the same cleanup.

---

## Things I checked that are fine

`useLocalDocumentUrl.ts` revokes correctly on both source change and unmount. Re-keying the viewers on `source.cacheKey` (`DocumentViewer.tsx:51,61,73`) is a real improvement over `source.id` for teardown. The PPTX path is the strongest of the three: `key={sourceIdentity}` forces a full remount per source, it has a typed error callback, a 100MB ceiling, magic-byte format detection, and a genuine LibreOffice fallback. Deps are all exact-pinned and `docx-preview`/`react-pdf`/`pdfjs-dist` are cleanly removed. Listener/observer/rAF/worker teardown across all three vendored viewers is thorough — I went looking for leaks and found only the two P3s above.

I also chased a suspected unbounded `MutationObserver` ↔ `setAttribute` feedback loop in `useSecureViewerLinks` and **it is not real** — probing the actual hook shows 0 href writes while idle. My first attempt appeared to hang only because the probe itself monkeypatched `setAttribute`. Flagging it as explicitly cleared so it doesn't get re-raised.

---

## CHECKED

- Worktree at `9aa279a7`, detached, fresh.
- `pnpm install --frozen-lockfile --ignore-scripts` — clean, lockfile consistent.
- `pnpm exec tsc --noEmit` — **exit 0**.
- `pnpm test` — **150 files, 995 tests passed**, 0 failed (~41s).
- `pnpm build` — **exit 0**.
- Ancestry: `git merge-base --is-ancestor 44e40e25 9aa279a7` → **not an ancestor**; merge-base `01322965`; API reports `mergeable_state: dirty`.
- Diffed `origin/main:DocxViewer.tsx` (post-#2074) against this head to separate this PR's changes from the pre-#2074 baseline.
- Wrote and ran three scratch vitest files to reproduce P2a and P2b, and to falsify the suspected observer loop; all removed, tree left clean.
- Read all six vendored `components/extend/` files plus the `@extend-ai/react-docx`, `@extend-ai/react-pptx`, and `@embedpdf/*` renderers for render target, link emission, HTML sinks, teardown, error states, and switch races.
- Not run: `cargo test` (no Rust changed).

naingthet (commented):

**Verdict: `main` needs a follow-up fix.** This PR merged as `e6937cfe` while round 2 was in flight. The rebase onto `44e40e25` resolved every DOCX conflict in favour of the branch, so **#2074's DOCX hardening is gone from `origin/main` today** — along with the test that guarded it. The P2 credential bug is fixed and the SheetJS lockfile fix is correct and verifiable. Round 1's P1 stands and is now a live regression on the default branch rather than a merge hazard.

This was an **incremental review of `9aa279a7..422c8f08`** (plus `origin/main` at `e6937cfe`), against round 1's open findings. Round 1's CHECKED section is treated as settled.

---

## 1. What the rebase did to the security posture

Round 1 warned that "whoever resolves the conflict silently decides the security posture." That is what happened, and it is worth being precise that this was a **rebase, not a merge** — `9aa279a7` was replayed onto `44e40e25` as `c0bcc638`:

```
git merge-base --is-ancestor 44e40e25 9aa279a7   →  NO    (round 1)
git merge-base --is-ancestor 44e40e25 422c8f08   →  YES   (round 2)
git merge-base --is-ancestor 44e40e25 origin/main →  YES
mergeable_state: dirty → MERGEABLE               (merged as e6937cfe)
```

So the ancestry problem is genuinely fixed — the branch now contains #2074. The question is what the replay did to `DocxViewer.tsx`, which #2074 had rewritten (+282). **Per-marker, on `origin/main` today vs. `44e40e25`:**

| marker in `document/DocxViewer.tsx` | `44e40e25` (#2074) | `origin/main` (post-merge) |
|---|---|---|
| `attachShadow` (closed shadow root) | 1 | **0** |
| `sanitizeRenderedDocument` | 2 | **0** |
| `sandbox` | 3 | **0** |
| `script-src` | 1 | **0** |
| `WeakMap` (fail-closed URL carrier) | 1 | **0** |
| `srcdoc` / `iframe` | 2 / 3 | **0 / 0** |
| `DocxViewerPreview` (live-DOM Extend surface) | 0 | **2** |
| file length | 282 lines | **75 lines** |

Every one of #2074's mechanisms is absent. The file on `main` mounts `DocxViewerPreview` into the live app tree (`document/DocxViewer.tsx:52`), and the vendored renderer emits the document's own URL verbatim at three sites in `@extend-ai/react-docx/dist/index.js` (`href: run.link`), with `javascript:`/`sanitizeUrl`/`sanitizeHref`/`isSafeUrl`/`safeUrl` all returning **0 hits** in that bundle — confirming round 1's grep against the now-installed tree. (Control: `@extend-ai/react-pptx` does carry its scheme allowlist.)

**The guard tests went with it.** `DocxViewer.dom.test.tsx` on `main` has one test; `44e40e25` had three, and all three are deleted:

- `confines generated document styles to an opaque-origin sandbox` — the direct regression test for the boundary above
- `finishes rendering when animation frames are suspended`
- `does not let a superseded parse overwrite the current document`

Nothing in the diff or the PR description flags any of these deletions. This is exactly the silent outcome round 1 predicted, and it is why "the file compiles and the suite is green" was never sufficient evidence — I checked the markers and the test names directly rather than inferring.

## 2. What ADR 0023 says — and why it does *not* make this an accepted trade-off

`docs/decisions/0023-office-documents-use-local-extend-viewers.md` (renamed from `0021` in this delta to resolve a number collision with #2074's own ADR) is a real, well-argued record, and it **is** citable. It settles several things that should therefore not be litigated in review: adopting Extend for PDF/DOCX/PPTX, viewer-only mode with upload/download disabled, exact-pinned local engines, bundled PDFium with no CDN fetch, and the LibreOffice fallback for PPT/ODP. All of that is a stated design decision and I am not raising any of it.

But on the security boundary specifically, the record does **not** cover the change. It is `Status: Proposed`, `Supersedes: none`, and it never mentions the sandboxed iframe, the shadow root, the sanitizer, or #2074 at all. What it *does* say cuts the other way — it lists as decisions and validation criteria that:

> "Tidebreak owns loading, **security**, citations, read-only policy…"
> "…opening only **HTTPS** external links outside the application."
> "DOCX renders in read-only mode… while **unsafe links cannot leave the app**."

So the ADR asserts the security property as a requirement it intends to keep, and is silent on the mechanism being removed to get there. That makes the P1 an **undocumented regression, not an accepted trade-off** — the record does not weigh the iframe boundary and choose against it; it does not know it existed. (It also lists "security review rejects their parsing boundary" as a trigger to revisit, which is the relevant clause here.)

To be fair to the change: the ADR's claim that the WASM importer "maps OOXML into React nodes instead of inserting document-authored HTML or styles" is a genuine, material difference from `docx-preview`, and it defuses the *HTML/style injection* half of what #2074's sanitizer defended. It does not defuse the *link* half, which is the P1 below.

---

## Findings

### P1 (live on `main`) — `crates/tidebreak-desktop/ui/src/document/DocxViewer.tsx:52` — DOCX link handling is fail-open where `main` was fail-closed

On `44e40e25`, a DOCX anchor never carried an `href` at all: the URL was validated, stashed in a `WeakMap`, rendered as text plus `title`, and `removeAttribute` ran unconditionally afterwards, so a lost entry degraded the display rather than exposing an unvalidated URL. On `main` today the only guard is `useSecureViewerLinks` — a `MutationObserver` that repairs anchors *after* the renderer has already put them in the live app DOM.

**Failure scenario, concretely:** a DOCX whose hyperlink target is a non-HTTPS scheme is rendered by `@extend-ai/react-docx` straight into the app tree with that value on `href`. The observer callback is a microtask-ish repair pass, so the attribute is live in the document between paint and cleanup, and any path that reads anchors before the callback runs — or any anchor the observer's `attributeFilter: ["href"]` / `subtree` scoping misses, e.g. one materialized inside a detached subtree that is later grafted in, or into a portal outside `containerRef` — is never repaired. Beyond timing, the observer is strictly narrower than the sanitizer it replaced: it strips no `on*` handlers, restricts no `src`, removes no `script`/`object`/`embed`/`form`/`base`, and neutralises no raw-text `</style>`.

*Reproduced:* the ancestry, the marker-by-marker absence, the three deleted tests, and the renderer's verbatim `href: run.link` emission with zero scheme guards. I did **not** author a hostile DOCX and I am not posting a payload.

**Smallest correct fix**, in preference order:

1. Move the guard **pre-render** so it fails closed: in the Tidebreak-owned copy `components/extend/docx-viewer.tsx`, filter link values through the existing `safeExternalHref` before they reach the DOM, so an anchor is only ever emitted with a validated `https:` URL and otherwise emitted with no `href`. This preserves the Extend adoption the ADR decided on and restores #2074's fail-closed property in one place.
2. Restore the deleted `confines generated document styles to an opaque-origin sandbox` test, adapted to whatever boundary you land on — without it there is nothing stopping the next viewer bump from re-opening this.
3. Amend ADR 0023 to record that it changes the DOCX rendering boundary established in #2074, with `Related`/`Supersedes` pointing at it, and state what replaces each guarantee. Right now the strongest argument against the change is unaddressed because it is unmentioned.

Please also restore the other two deleted tests, or say why they no longer apply — note that `does not let a superseded parse overwrite the current document` covered exactly the class of switch race that round 1's P3a flags as still present in the new PDF viewer.

### P2 (open, unchanged) — `crates/tidebreak-desktop/ui/src/document/PdfViewer.tsx:82,90,102` — `useSecureViewerLinks` never engages in the PDF viewer

Unchanged since round 1 and present on `main`. `containerRef` is attached only on the success branch (`:102`); the error (`:82`) and loading (`:90`) branches render `ViewerShell` without it. At mount `file.objectUrl` is `null`, so the effect returns early on a null ref, and the dep array `[containerRef]` is a stable ref object — the effect never re-runs once the URL arrives. No observer and no click interception are ever installed.

*Reproduced* (fresh, this round, with a control): rendering `PdfViewer` with a mocked Extend surface containing an `http:` anchor leaves `href` fully intact after mount — while the identical anchor under `DocxViewer`, whose ref sits on an always-rendered outer div, is stripped to `null`. Same hook, same fixture, opposite outcomes; the difference is purely the ref attachment.

Still P2 rather than P1 because no annotation/link plugin is registered, so PDF content emits no anchors today — it is a dead safety net that will fail silently the day one is added. **Fix:** attach `containerRef` to a single always-rendered outer wrapper across all three branches, as `DocxViewer` does, or use a callback ref so the effect re-runs when the element appears.

### Resolved since round 1 — credentialed HTTPS (was P2a) ✅

`extendViewerSurface.ts:88` now reads `parsed.protocol === "https:" && !parsed.username && !parsed.password`, matching #2074's predicate, and `extendViewerSurface.dom.test.tsx` gained regression cases for both a credentialed URL and a `javascript:` URL, asserting `href` is dropped and `openExternal` is called exactly once. Verified on `origin/main`. Closing this one.

One consequence worth noting: round 1 recommended extracting a single shared predicate. That is now moot in the worst way — with #2074's `DocxViewer.tsx` deleted, `safeExternalHref` in `extendViewerSurface.ts` is no longer a divergent *copy*, it is the *only* remaining copy. It is correct today; it is also now load-bearing and unshadowed, which raises the value of fix (1) above routing through it.

### Carried forward, not re-derived

P2c (spinner has no `finally`/timeout; string-matched error detection; no DOCX size ceiling), P2d (`console.error` patching), P3a (stale page count on rapid document switch), P3b (revoked object URL resurrected). The delta since `9aa279a7` does not touch `components/extend/` at all, so these are byte-identical to round 1 and remain *reasoned from code, not reproduced*. P3a is the one I would still act on, given the deleted superseded-parse test noted above.

---

## SheetJS lockfile integrity — correct, and worth having caught

The branch commit `c0bcc638` regenerated `pnpm-lock.yaml` and **dropped the integrity hash** for the CDN-hosted SheetJS tarball, leaving `resolution: {tarball: https://cdn.sheetjs.com/…}` with no `integrity:` field. That is a real supply-chain regression: `xlsx` is fetched from a vendor CDN rather than the npm registry, so the lockfile hash is the *only* thing binding the install to a known artifact, and without it any future install would accept whatever that URL serves. The head commit `422c8f08` restores it. Verdict on the fix:

- **The restored hash is authentic.** I downloaded `https://cdn.sheetjs.com/xlsx-0.20.3/xlsx-0.20.3.tgz` (2,409,319 bytes) and computed its SHA-512 — it matches the lockfile's `sha512-oLDq3jw7…H+3AJA==` exactly.
- **It matches what was on `main` before.** `44e40e25`'s lockfile carries byte-identical integrity for the same entry, so this is a true restoration, not a re-pin to something new.
- **No other dependency's integrity changed.** Diffing all resolution hashes between `44e40e25` and `422c8f08`: 684 packages in common, **0** with a changed hash, **0** entries left with an empty integrity field. 64 packages added (the Extend engines) and 9 removed (`docx-preview`/`react-pdf`/`pdfjs-dist` and friends) — consistent with the stated scope.
- `pnpm install --frozen-lockfile` succeeds, so the lockfile is self-consistent.

No finding here. Flagging only that a CDN-sourced dependency losing its integrity line is invisible in a 899-line lockfile diff — a CI check that fails when any `resolution:` lacks `integrity:` would catch the next one.

---

## CHECKED

- Fresh detached worktree at `422c8f08`; separate throwaway worktrees at `44e40e25` and `2b7832b5` for baseline comparison. All removed.
- Ancestry: `44e40e25` is now an ancestor of `422c8f08` **and** of `origin/main`; `422c8f08` itself is not (squash-merged as `e6937cfe`). `mergeable` went `dirty` → `MERGEABLE` → merged.
- `pnpm install --frozen-lockfile --ignore-scripts` — exit 0, lockfile consistent.
- `pnpm exec tsc --noEmit` — **exit 0**.
- `pnpm test` — **151 files, 994 tests passed, 0 failed** on a clean run.
- `pnpm build` — **exit 0** (only the pre-existing >500 kB chunk-size advisory).
- One flake observed and cleared: `src/settings/McpPanel.test.tsx > managed: compact endpoint rows only` failed on my first full run at head, passed in isolation, passed on a second full run — and **also fails in a full run on plain `main` at `2b7832b5`**, which contains none of this PR. Order-dependent pollution pre-dating this PR, unrelated to the viewers. Not a finding here, but worth a separate issue.
- Verified the DOCX security posture by comparing `44e40e25`, `422c8f08`, and `origin/main` marker-by-marker and test-name-by-test-name, not by whether the file typechecks.
- Reproduced P2 (PDF ref) fresh with a DOCX control in the same file; scratch test removed, tree clean.
- Verified the SheetJS tarball hash against the live upstream artifact and swept all 684 common lockfile entries for silent integrity changes.
- Confirmed the PR-authored delta since round 1 is small: the `safeExternalHref` credential fix + its regression tests, the lockfile integrity restoration, an ADR renumber `0021`→`0023`, and a comment rewrite in `DocxViewer.tsx`. Everything else in `9aa279a7..422c8f08` is main's own commits arriving via the rebase.
- Not re-raised per round 1: the falsified `MutationObserver` feedback loop, PDF being correctly N/A for sanitization, and the PPTX allowlist.
- Not run: `cargo test` (no Rust authored by this PR).

PR #2085, opened 2026-08-14
feat(desktop): add app reload

naingthet:
## Summary

- add Cmd/Ctrl+R as a shell-level shortcut that reloads the app
- expose the shortcut in the keyboard shortcuts help
- add a native macOS View → Reload menu item so reload remains available when the renderer is unavailable

## User impact

Users can refresh Tidebreak with the standard reload shortcut or the macOS View menu, including as a recovery path for a stuck renderer.

## Validation

- `cargo check -p tidebreak-desktop --locked`
- `pnpm exec vitest run src/ShellShortcuts.test.ts`
- `pnpm build`
- full UI suite before rebase: 989 tests passed

### DIFF
--- a/crates/tidebreak-desktop/src/lib.rs
+++ b/crates/tidebreak-desktop/src/lib.rs
@@ -636,7 +636,7 @@ pub fn run() {
             let handle = app.handle().clone();
             deep_link::install(&handle);
             #[cfg(target_os = "macos")]
-            updater::install_update_menu(app)?;
+            updater::install_app_menu(app)?;
             updater::spawn_update_loop(handle.clone());
             let data = data_dir(&handle)?;
             // Before anything that can warn: the embedded server's tracing
--- a/crates/tidebreak-desktop/src/updater.rs
+++ b/crates/tidebreak-desktop/src/updater.rs
@@ -34,6 +34,7 @@ const UPDATE_STATE_EVENT: &str = "desktop-update-state";
 /// there so the outcome (up to date, or an update staged) is visible.
 const UPDATE_CHECK_REQUESTED_EVENT: &str = "desktop-update-check-requested";
 const MENU_CHECK_FOR_UPDATES_ID: &str = "check-for-updates";
+const MENU_RELOAD_ID: &str = "reload-app";
 const UPDATE_CHECK_STARTUP_DELAY: Duration = Duration::from_secs(15);
 const UPDATE_CHECK_INTERVAL: Duration = Duration::from_secs(5 * 60);
 const UPDATE_CHECK_ERROR: &str = "Could not check for updates. Try again later.";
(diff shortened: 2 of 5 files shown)


naingthet (commented):

Mechanically clean and well-scoped, but reload silently destroys unsaved editor state that this codebase elsewhere considers worth a destructive confirm — one P1 to resolve before merge.

The Rust side holds up: `Menu::default` on macOS really does build a `"View"` submenu with that literal English text (tauri 2.11.5 `src/menu/menu.rs:230-236`), so the text match is sound and not locale-dependent; `"main"` is the only window label the app ever creates, so hardcoding it is fine; `prepend` ordering gives Reload → separator → Fullscreen as intended.

---

**`crates/tidebreak-desktop/ui/src/AppShell.tsx:159` — P1 — reload discards unsaved output edits and MCP config with no prompt and no way back**

`"reload-app": () => window.location.reload()` is unconditional, and `ShellShortcuts.ts:159-165` sets `allowInEditable: true`, so the shortcut fires while the caret is inside a text field.

Failure scenario, concrete: open an output, click Edit, type changes into the editor textarea, then press ⌘R with focus still in that textarea. The draft lives only in React state — `outputs/OutputDetailRoot.tsx:178` `const dirty = editor !== null && editor.draft !== (preview?.content ?? "")`. The app already treats losing that draft as needing a destructive confirm, but only on *in-app* navigation: `OutputDetailRoot.tsx:180-196` `leaveEditor()` raises "Discard your changes?", and `:235` warns "Your unsaved edit will be lost. Copy anything you still need first." A reload bypasses every one of those paths. There is no `beforeunload` handler anywhere in the renderer (verified by grep across `ui/src/` on this branch **and** on `origin/main` — zero hits), so nothing can veto or even warn. The edits are gone, silently, with no undo.

`settings/McpPanel.tsx:197,217,222-238,924` is the same shape — an elaborate `dirty`/`dirtyRef` reconciliation exists specifically so background reads can't clobber unsaved server config, and reload discards it anyway.

I reproduced the trigger condition directly (scratch vitest, since removed): `resolveShellShortcut({key:"r",code:"KeyR",metaKey:true,...}, {editable:true, modalOpen:false, command:true})` resolves to `reload-app`. I did not build and drive the packaged desktop app; the state-loss consequence is read off `OutputDetailRoot.tsx` and the absence of any unload handler, both of which I verified in source.

Note the macOS menu path (`updater.rs:331-336`) calls `window.reload()` with no guard at all, so it needs the same treatment — and on macOS the AppKit accelerator is the path that fires.

Suggested fix that preserves the stated "recovery path for a stuck renderer" intent: keep a small dirty registry that `OutputDetailRoot` and `McpPanel` register into while dirty, and have the `reload-app` handler run the app's own `confirm()` before reloading when it is non-empty. For the native menu item, emit a `desktop-reload-requested` event to the renderer and hard-reload from Rust only if the renderer hasn't acknowledged within a short timeout — a stuck renderer never answers, so recovery still works, while a responsive one gets to ask. A bare `beforeunload` is the weaker option here: WKWebView's native prompt is unreliable and would not match the app's dialog styling.

---

**`crates/tidebreak-desktop/ui/src/ShellShortcuts.ts:303` — P3 — the reload recovery path is unreachable on Windows/Linux whenever a modal is open**

`if (context.modalOpen) return null;` suppresses every shell shortcut while a dialog is on screen. On macOS that is masked by the new native View → Reload item, which has no such guard. On Windows and Linux `install_app_menu` is `#[cfg(target_os = "macos")]`, so there is no menu fallback — if the renderer wedges with any dialog open, ⌘/Ctrl+R does nothing and the user's only exit is killing the app. Reproduced: the same `resolveShellShortcut` call with `modalOpen: true` returns `null`.

Fix: exempt `reload-app` from the modal suppression (it is a frame-level escape hatch, not an action that would surprise someone mid-decision), or install the Reload menu item on all platforms rather than macOS only.

---

**CHECKED**

- Gates run in a clean detached worktree at `1fd7897e4a6d`: `pnpm exec tsc --noEmit` clean (0 errors); `pnpm exec vitest run` **984 passed / 146 files, 0 failed**; `cargo check -p tidebreak-desktop --locked` clean; `cargo test -p tidebreak-desktop --locked` **152 passed, 0 failed, 2 ignored** (network/model-download); `cargo fmt --all -- --check` clean.
- **Teardown/leaks:** not a concern — a full renderer reload discards the entire JS heap, so listeners, timers and observers cannot leak a generation. `install_app_menu` runs once in `setup`, not per load, so reloading does not duplicate menu items.
- **Reentrancy:** double-firing ⌘R (menu + keyboard, or repeated presses) is benign — a second `reload()` during an in-flight navigation is a no-op, not a race.
- **In-flight work is genuinely safe**, and that looks deliberate: composer drafts are written to `sessionStorage` on every keystroke (`ComposerDrafts.ts:34-42,214-219`) and survive; streaming turns replay from a server-side sequenced journal via `openEvents(chatId, after)` with a `replayed` flag (`api/client.ts:1566`, `ChatSessionReducer.ts:137`); the route survives via `createHashHistory()` (`router.tsx:294`). No credential or session state is renderer-only.
- **Not flagged:** dropping queued/failed image attachments on reload (`ComposerDrafts.ts:86-100`) is a documented deliberate decision — restoring a chip with no backing `File` would offer a retry with nothing behind it.
- **No conflict:** nothing else in the renderer binds `KeyR`.
- Checked `origin/main` before attributing the missing unload guard to this PR — it is absent there too; this PR is what makes it reachable.

PR #2097, opened 2026-08-14
refactor(core): squash schema migrations into a single baseline

### LINKED ISSUE #1462: Pre-v1 schema: squash the migration baseline and consolidate accreted tables

naingthet:
Closes #1462.

Replaces the August 4 baseline plus the follow-up `add_app_owner` upgrade with one `Baseline` migration, `m20260814_000001_baseline`, and bumps `DESKTOP_SCHEMA_EPOCH` to 27 so disposable desktop profiles rebuild.

The table consolidations on #1462 already shipped. What was left was a two-step history that should not be the first schema entry on a public tree. This is the public-opening squash, not the end of v0. The 1.0 checklist still requires another clean snapshot before decision 0002 inverts and `reset_pre_v1_state` goes away.

Fresh databases apply that one baseline. Self-host databases that already have tables are not recreated; a missing `app.owner` is backfilled to `local`. Desktop SQLite at epoch 26 or older is reset.

### DIFF
--- a/crates/tidebreak-core/src/db/migration.rs
+++ b/crates/tidebreak-core/src/db/migration.rs
@@ -1,10 +1,12 @@
 //! The database schema.
 //!
-//! Fresh pre-v1 databases are still described by [`Baseline`], and desktop
+//! Fresh pre-v1 databases are described by a single [`Baseline`]. Desktop
 //! SQLite changes bump `DESKTOP_SCHEMA_EPOCH` so disposable local data is
-//! rebuilt. Self-host databases are durable, however, so any schema change
-//! that must reach an already-recorded baseline also gets an ordered upgrade
-//! migration in this module.
+//! rebuilt. Self-host databases are durable: a renamed baseline must not
+//! recreate tables that already exist, and any later in-place baseline edit
+//! that must reach an already-recorded schema also gets an ordered upgrade
+//! migration in this module. Squash again before `1.0.0` so that release's
+//! first migration is a clean snapshot, not this public-opening chain.
 
 mod baseline;
 mod idens;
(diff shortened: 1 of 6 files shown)


naingthet:
Holding this. The remaining #1462 work is the baseline re-squash itself, not only moving the reminder. Expanding onto that rather than landing the docs-only change.

PR #2098, opened 2026-08-14
fix(desktop): keep streamed reasoning on one assistant bubble

### LINKED ISSUE #1715: Streamed transcript and hydrated snapshot disagree: reasoning relocates and phases reshuffle at turn end
The live stream and the durable snapshot build the transcript with different algorithms, and the wholesale swap at turn end makes the difference visible as a reorder/reshuffle.

Observed in a live browser-UI session (Ask mode, multi-step turn):

- During streaming, "Thought" accordions appear interleaved — one leading the turn, one per segment between tool batches. After the turn completes (and identically after reload), all of them vanish from those positions and a single "Thought" renders on the final assistant bubble. Cause: reasoning exists on the wire only at turn level (`ChatTerminalTurnSnapshot.reasoning`, attached via `turn.message_id`), while the stream delivers it interleaved (`reasoning_delta` between tool calls).
- Group shapes change at the swap (tracked separately in #1713 for the invisible-splitter part).
- Hydration orders entries by sorting three tables' wall-clock timestamps (`hydrateTranscriptHistory`), while the stream orders by event arrival; messages have a durable per-chat `seq` the UI does not use. In testing the timestamp sort itself did not visibly misorder text vs tools — the reasoning relocation dominated — but the ordering basis remains different between the two views.

Also related to the felt experience in Ask mode: every approval boundary triggers the stream interrupt/resume machinery, which discards and re-streams in-flight content (observed: transcript shrinking 12→11 rows and the live thinking bubble dropping 499→166 chars before regrowing). Together with the turn-end swap this reads as "the order jumps around".

Fix direction deliberately left open — candidates include carrying per-step reasoning and/or a durable interleaving key on the wire, hydrating by diff instead of replacement, or making the streamed view converge to the snapshot's shape as it goes. Each has schema/protocol implications that need an owner's judgment before anyone builds; the observations above are the constraint set.

naingthet:
Closes #1715.

## Summary

- accumulate the open turn's reasoning in one buffer
- paint that buffer on the latest live assistant bubble
- do not open a new thought accordion after each tool batch

## Why

The live stream interleaved a thought accordion per reasoning segment. The durable snapshot stores reasoning on the turn's output message. Hydration therefore replaced several thoughts with one on the final bubble. The stream now uses the snapshot's shape, so the end-of-turn swap no longer relocates reasoning.

A later change can still persist per-step reasoning if we want thoughts to stay between tool batches after reload. That is a wire-type change; this fix does not invent one.

## Validation

- `pnpm test` in `crates/tidebreak-desktop/ui` — 995 passed
- `git diff --check`

### DIFF
--- a/crates/tidebreak-desktop/ui/src/ChatSessionReducer.test.ts
+++ b/crates/tidebreak-desktop/ui/src/ChatSessionReducer.test.ts
@@ -834,23 +834,49 @@ describe("reasoning presentation", () => {
     ]);
   });
 
-  it("opens a bubble for reasoning that follows tool activity", () => {
+  it("keeps post-tool reasoning on the existing assistant bubble", () => {
     const { state } = play([
       TURN,
       { type: "tool_call_started", call_id: "c", name: "search" },
       { type: "reasoning_delta", text: "the search found nothing" },
     ]);
-    // The empty bubble the turn opened, the call, then the reasoning that came
-    // after it — reasoning read in journal order, not hoisted above the call.
+    // The snapshot hangs the whole turn's reasoning on one assistant
+    // message. Opening a second bubble here is what made reload relocate it.
     expect(state.messages.map((message) => message.role)).toEqual([
       "assistant",
       "tool",
-      "assistant",
     ]);
-    expect(state.messages[2]).toEqual(
+    expect(state.messages[0]).toEqual(
       expect.objectContaining({
         text: "",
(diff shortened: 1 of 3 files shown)


naingthet:
CI failed on an existing race in `McpPanel.test.tsx`: the mount switch appears when servers load, but `serves: Incident API` waits on a later gateway-apps fetch. The test now `findByText`s that label.

PR #2116, opened 2026-08-15
fix(agent): drop thinking after a rejected mixed control step

### LINKED ISSUE #2111: fix: flatten Anthropic thinking blocks on model switch
## Problem
Dev exercise chat `e415899e-60a7-41c9-bc54-d497fe51757f` turn `3fa8d111-83fb-48b0-986a-8cc0afdbe512` failed after switching to `claude-sonnet-5` with:

`invalid provider request: anthropic returned 400: thinking or redacted_thinking blocks in the latest assistant message cannot be modified`

## Why ours
Native Anthropic thinking artifacts were replayed into a new request where the model then appended text/tool calls. Flatten-on-switch (see `docs/model-providers.md` and decision 0021) should have dropped or flattened those blocks.

## Suggested fix
On any provider or model switch, flatten native thinking / redacted_thinking (and other origin-gated artifacts) before constructing the next request. Add a regression that replays a thinking-bearing assistant message onto a new Anthropic route and asserts the wire body has no unmodifiable thinking block the new turn would mutate.

Evidence: `/tmp/tidebreak-exercise/2026-08-14-ledger/switch1-t1.ndjson` seq 115.

naingthet:
## Summary
A mixed `ask_user_questions` + text step is declined and retried, but the rejected assistant message kept its captured thinking blocks. The next Anthropic request then 400s (`thinking or redacted_thinking blocks in the latest assistant message cannot be modified`) because those blocks are no longer the latest generated prefix.

This is the failure from the dev-exercise switch chat — it is a same-route mixed-step replay bug, not flatten-on-switch. Cancel already drops partial reasoning for the same reason; the mixed-control path now does too.

## Test
`cargo test -p tidebreak-core --locked a_rejected_mixed_control_step_does_not_replay_its_thinking_on_the_retry`

Closes #2111

### DIFF
--- a/crates/tidebreak-core/src/agent/tests/mod.rs
+++ b/crates/tidebreak-core/src/agent/tests/mod.rs
@@ -562,6 +562,151 @@ impl ModelProvider for ProviderSearchWithStandaloneControlProvider {
     }
 }
 
+struct MixedQuestionWithReasoningProvider {
+    calls: AtomicUsize,
+    seen: Arc<Mutex<Vec<Vec<Value>>>>,
+}
+
+#[async_trait]
+impl ModelProvider for MixedQuestionWithReasoningProvider {
+    fn id(&self) -> ProviderId {
+        ProviderId::new("mixed-question-with-reasoning")
+    }
+
+    async fn stream(&self, req: ChatRequest) -> Result<BoxStream<'static, ProviderEvent>> {
+        self.seen.lock().unwrap().push(
+            req.messages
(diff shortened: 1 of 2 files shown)


naingthet (commented):
## Review

The mixed-control reject path now clears thinking on the transient reconstructed assistant message so the same-route Anthropic retry does not 400. Cancel-path handling is unchanged, and the test asserts the retry `ChatRequest` has no thinking blocks.

PR #2121, opened 2026-08-15
fix(cli): attach images so the next turn can see them

### LINKED ISSUE #2115: fix: CLI image attach must land on the chat
## Problem
Dev exercise chat `ceacf54b-78a6-4510-9ff1-0bbfd21fda20`: `tidebreak attach` printed success (`published dashboard.png as an image attachment`) but `GET /chats/{id}` had `root_attachments: []`, the next user message had `attachments: []`, and the blob was unreferenced.

## Why ours
Publishing the blob is not the same as attaching it to the conversation the model will see. The CLI success message is a lie if the next `-p` turn cannot see the image.

## Suggested fix
After a successful image publish, the next turn on that chat must actually receive the image (message attachment or equivalent). Investigate `crates/tidebreak-cli/src/outputs.rs` `attach` and the image publish route. Add a CLI/server test that attach + next turn lists the image.

Evidence: `/tmp/tidebreak-exercise/2026-08-14-ledger/audit-images-attachments.md`.

naingthet:
## Summary

`tidebreak attach <chat> <image>` printed success after `POST /chats/{id}/attachments/images`, but that route only publishes identity. The next `-p` turn posted a message with no `attachments`, so `GET /chats/{id}/messages` showed `image_attachments: []` and the model never received the pixels.

Documents already land because ingest creates a chat-owned source. Images have to ride the next user message.

## Change

- After a successful image publish, remember the attachment id on the profile (`pending-image-attachments/{chat}`).
- The next `tidebreak -p` on that chat submits those ids on `POST /chats/{id}/messages` and then clears them.
- Success copy now matches documents (`attached … as image/png`) instead of telling the caller to "reference it from the next turn."

## Test

`attaching_an_image_lands_on_the_next_print_turn` attaches a PNG through the CLI, runs `-p`, and asserts `GET /chats/{id}/messages` carries the image on the user message.

Closes #2115

### DIFF
--- a/crates/tidebreak-cli/src/api/client.rs
+++ b/crates/tidebreak-cli/src/api/client.rs
@@ -481,11 +481,24 @@ impl Client {
     }
 
     /// Accept a user message and queue its turn (`202`).
-    pub async fn post_message(&self, chat: ChatId, turn_id: TurnId, content: &str) -> Result<()> {
+    ///
+    /// `attachments` are image ids returned by [`Self::attach_image`]. An empty
+    /// slice is the ordinary text-only send.
+    pub async fn post_message(
+        &self,
+        chat: ChatId,
+        turn_id: TurnId,
+        content: &str,
+        attachments: &[uuid::Uuid],
+    ) -> Result<()> {
         let response = self
             .http
             .post(format!("{}/chats/{chat}/messages", self.base))
-            .json(&serde_json::json!({ "turn_id": turn_id, "content": content }))
+            .json(&serde_json::json!({
+                "turn_id": turn_id,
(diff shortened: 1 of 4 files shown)


naingthet (commented):
## Review

`tidebreak attach` now records the published image id under the profile (`pending-image-attachments/{chat}`) and the next `-p` submits those ids on `POST /chats/{id}/messages`, then clears them only after a successful post. Pending state is chat-scoped, capped at `MAX_MESSAGE_ATTACHMENTS`, and the new serve test asserts `GET /chats/{id}/messages` actually carries the image on the user message — not just a print string.

PR #2122, opened 2026-08-15
docs(decisions): propose code mode design records

naingthet:
Proposes **Code mode** — a second product surface for driving external coding-agent CLIs (Claude Code first; Codex CLI, opencode, Grok CLI behind it) in isolated git worktrees through a structured UI: conversation, tool activity, approvals, per-turn diffs, and a pull-request flow.

Seven decision records, all `Proposed`:

- **0030** — separate surface with repo → workspace → session vocabulary, deliberately built shape-compatible with chat so later convergence is a mechanical merge rather than a translation layer
- **0031** — per-harness adapters normalize native machine-readable protocols into one `CodeEvent` vocabulary; fixtures-before-parsers, explicit capability flags, version detection with visible degradation
- **0032** — worktrees/branches/per-turn checkpoints via git shell-out (hidden refs, synthetic commits, never the user's index); conservative crash recovery that fences live orphans instead of killing them
- **0033** — approvals are first-class over each harness's native channel; deny carries feedback that steers the agent; bypass flags never in a default launch plan, lint-enforced
- **0034** — harness discovery via the user's login shell; credentials observed, never brokered; environment passthrough makes gateway-configured harnesses work unchanged
- **0035** — per-session journals with the chat journal's exact replay discipline, plus one install-wide updates channel of restated digests; server-computed attention states
- **0036** — auxiliary terminals are ephemeral byte streams for the user's shells; the harness never gets a PTY, by construction

`docs/code-mode.md` carries the working design detail in one place (crate layout, data model, adapter contract, API surface, UI plan, test strategy); `docs/deferred.md` gains a section for what v1 deliberately leaves out.

Docs-only change; no code. Not intended to merge until reviewed — implementation PRs will follow separately and reference these records.

### DIFF
--- a/docs/code-mode.md
+++ b/docs/code-mode.md
@@ -49,7 +49,8 @@ crates/tidebreak-core/src/db/ops/code/        repo.rs, workspace.rs, session.rs,
 crates/tidebreak-harness/                     protocol translation only
   src/lib.rs       HarnessAdapter + HarnessSession traits, SessionSpec, LaunchPlan,
                    adapter registry
-  src/probe.rs     login-shell PATH resolution, version detection, auth observation
+  src/probe.rs     interactive-login shell resolution + env capture (0034),
+                   version detection, auth observation
   src/budget.rs    bounded stream-parse budgets
   src/claude/      mod.rs, parse.rs, session.rs, approvals.rs
   src/codex/  src/opencode/  src/grok/        later phases
@@ -219,7 +220,11 @@ POST/GET        /code/workspaces           {repo_id, base_ref?, title?}
 GET/PATCH       /code/workspaces/{id}
 POST            /code/workspaces/{id}/archive        {force?}
 POST            /code/workspaces/{id}/sessions       {harness, permission_mode}
-POST            /code/sessions/{id}/turns            {message}  (steer/queue if running)
+POST            /code/sessions/{id}/turns            {message}  (queued while running —
+                                                     the chat product's queue-default rule;
+                                                     steering is the explicit alternative,
+                                                     available only where the adapter's
+                                                     mid_turn_steering capability carries it)
 POST            /code/sessions/{id}/interrupt | /permission-mode | /attention | /reap
 WS              /code/sessions/{id}/events?after=    snapshot → replay → live
 WS              /code/updates                        digests, restated on connect
@@ -304,5 +309,9 @@ new `PanelContent` variants).
 Recorded in [`docs/deferred.md`](deferred.md): running a harness in a PTY;
 bundled or pinned harness binaries; checkpoint restore (the refs land in
 v1; the restore surface does not); multiple sessions per workspace;
-an in-app code editor; chat↔code convergence (one inbox, shared projects);
+an in-app code editor; chat–code convergence (the single-surface end
+state: one conversation concept with an optional workspace binding,
+engines behind the adapter contract, no user-facing mode choice); remote
+session execution (the same harness in a managed sandbox feeding the same
+journal); a supervision-first mobile client over the updates channel;
 a per-repo worktree-location override; Windows support for code mode.
--- a/docs/code-mode.md
+++ b/docs/code-mode.md
@@ -11,8 +11,12 @@ spin up isolated **workspaces** (one worktree + branch each), and run
 **sessions** — durable conversations with external coding-agent harnesses
 (Claude Code first; Codex CLI, opencode, and Grok CLI behind it) — supervised
 through a structured UI: conversation, tool activity, approvals, per-turn
-diffs, and a pull-request flow. The modes are separate today and deliberately
-shaped to converge later ([`0030`](decisions/0030-code-mode-separate-surface.md)).
+diffs, and a pull-request flow. The two-mode split is a delivery strategy:
+the destination is one surface where context selects behavior — a
+conversation bound to a workspace behaves code-like, one without is ordinary
+chat — and the adapter contract below is the runtime interface every engine,
+eventually including Tidebreak's own internal loop, sits behind
+([`0030`](decisions/0030-code-mode-separate-surface.md)).
 
 ## Vocabulary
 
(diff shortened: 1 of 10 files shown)


docs/decisions/0034-harness-discovery-credentials.md
```diff
@@ -0,0 +1,132 @@
...
+the user's shell `PATH`. A harness installed via a shell profile
+(`~/.zshrc` PATH additions, version managers) is invisible to a naive
+`Command::new("claude")` from the app. This, not philosophy, is why discovery
+needs a decision.
+
+## Decision
+
+**Discovery: the user's login shell resolves the binary.** Harnesses are
+found by asking the user's own login shell
+(`$SHELL -lc 'command -v <binary>'`), so whatever the user's terminal can
```
naingthet: Internal contradiction: Context cites `~/.zshrc` PATH additions and version managers as the reason discovery cannot use the GUI process PATH, then Decision resolves with `$SHELL -lc 'command -v <binary>'` and claims “whatever the user's terminal can run, Tidebreak can run.”

`$SHELL -lc` is a login, non-interactive shell. zsh sources `.zprofile`/`.zlogin` for login and `.zshrc` only when interactive, so this misses the exact files the Context names (nvm/fnm/asdf/pyenv almost always land in `.zshrc`). Terminal.app is login *and* interactive.

Use `$SHELL -lic` (or an equivalent that actually matches a terminal) and keep the existing “profile prints garbage before the answer” probe test — that test is load-bearing either way. As written, the doctor will report “not found” for the users this paragraph is about.

docs/code-mode.md
```diff
@@ -0,0 +1,308 @@
...
+
+```
+POST/GET        /code/repos                GET/PATCH/DELETE /code/repos/{id}
+GET             /code/harnesses            doctor    POST /code/harnesses/refresh
+
+POST/GET        /code/workspaces           {repo_id, base_ref?, title?}
+GET/PATCH       /code/workspaces/{id}
+POST            /code/workspaces/{id}/archive        {force?}
+POST            /code/workspaces/{id}/sessions       {harness, permission_mode}
+POST            /code/sessions/{id}/turns            {message}  (steer/queue if running)
```
naingthet: `(steer/queue if running)` leaves the mid-turn default unspecified. [0009](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0009-queued-turns.md) already decided queue is the default and steer is the explicit alternative; 0030 says a second streaming/send idiom is out of bounds without a record explaining why the shared shape cannot serve. 0031 lists mid-turn steering as a *capability flag* (some harnesses cannot steer), which is a reason the default cannot just be “steer,” not a reason to invent a third posture.

Say which: adopt 0009 (queue by default, steer explicit, and a harness with `mid-turn steering: Unsupported` still accepts the queue path), or write the divergence into 0031/0035. Silence here will become two composers.

docs/code-mode.md
```diff
@@ -0,0 +1,308 @@
...
+  (`TIDEBREAK_LIVE_HARNESS=1`, ignored by default) plus a per-adapter
+  manual smoke checklist cover the real thing before an adapter ships.
+
+## What v1 excludes
+
+Recorded in [`docs/deferred.md`](deferred.md): running a harness in a PTY;
+bundled or pinned harness binaries; checkpoint restore (the refs land in
+v1; the restore surface does not); multiple sessions per workspace;
+an in-app code editor; chat↔code convergence (one inbox, shared projects);
+a per-repo worktree-location override; Windows support for code mode.
```
naingthet, line 317: Stale relative to this PR’s own last commit. `docs/deferred.md` now parks remote session execution and a supervision-first mobile client (and 0030/0031’s engine-neutral contract is what keeps those open). This summary omits both, and still describes convergence as “one inbox, shared projects” — language 0030 replaced with one surface, no mode choice, adapter contract as the runtime interface.

This page tells implementers it is the working detail and that records win on conflict; a short list that disagrees with both the records and `deferred.md` is the one they will copy. Fold the two parked items in and match 0030’s destination wording.

naingthet (commented):
Reviewed the seven Proposed records plus `docs/code-mode.md` and the `docs/deferred.md` addition against the template, 0002/0004/0007/0009/0018, and the existing permission vocabulary.

The split-now / converge-later shape in 0030 is coherent: separate id spaces and tables, shared journal/replay/permission/attention shapes, adapter contract as the eventual engine slot. 0031–0036 are internally consistent on fixtures-before-parsers, no-bypass-by-default, observe-don’t-broker credentials, conservative fencing, and PTY-never-for-harness. Validation sections name the plausible wrong implementation, which is what those sections are for.

Three concrete gaps before this should merge — none of them are “write more tests,” all of them would be implemented wrong if taken as written.

naingthet:
Addressed all three review findings in the latest commit:

1. **0034 shell-mode contradiction** — discovery now specifies an *interactive login* invocation (`$SHELL -ilc`) with sentinel-delimited, timeout-bounded output, and the child environment is captured from that same probe rather than passed through from the GUI process. The Validation section gains the env-capture case (a variable set only in the interactive profile must reach the child).
2. **Steer/queue default** — `code-mode.md` now states turns submitted while running queue by default, per the queue-default rule 0009 settled; steering is the explicit, capability-gated alternative.
3. **Stale v1-exclusions summary** — synced with `deferred.md` (remote session execution, supervision-first mobile client) and the single-surface convergence language.

The corresponding code defect (probe.rs / passthrough_env on #2123) is being fixed on that PR to match.

naingthet (commented):

Checked the new commit against the three findings on `42cf2a1c`. 0034 now specifies `$SHELL -ilc` (interactive login) and captures the child environment from that same probe rather than the GUI process, with the env-capture case added to Validation — that closes the `.zshrc` / version-manager hole. `code-mode.md` adopts 0009's queue-default for mid-turn submits, with steering as the explicit capability-gated alternative. The v1-exclusions list now matches `deferred.md` and 0030's single-surface destination wording.

Docs-only; the corresponding probe/env implementation still needs to land on #2123.

PR #2123, opened 2026-08-15
feat(core): code mode domain model and harness adapter foundation

naingthet:
## Summary

Foundation layer for code mode: the domain types, persistence, and first harness adapter. No server routes, no UI, no worktree/git manager.

Spec is the design docs in #2122 (`docs/code-mode.md` and ADRs 0030–0036).

### Domain (`tidebreak-core::code`)

- Distinct id types (`RepoId`, `WorkspaceId`, `CodeSessionId`, `CodeTurnId`, `CodeApprovalId`, `CodeTerminalId`) so the chat/code split is type-enforced (0030).
- `CodeEvent` / `SequencedCodeEvent` with the tabulated vocabulary, internally tagged serde, `#[non_exhaustive]`, and a journal-shape pin at `fixtures/code-journal-events.json`.
- Attention state + `should_replace`: `Manual` is never overwritten by an automatic source; a `Structured` `NeedsYou` is not downgraded by a `Heuristic`.
- `HarnessCaps` constructed exhaustively (no `Default`). `CodePermissionMode { Plan, Ask, Auto }` reuses the chat vocabulary names.

### Persistence

- Six baseline tables (`code_repo`, `code_workspace`, `code_session`, `code_turn`, `code_event`, `code_approval`). No foreign keys into chat tables.
- `DESKTOP_SCHEMA_EPOCH` 27 → 28.
- Journal appends are spawn-epoch fenced and allocate a per-session monotonic seq under a session-row lock. A stale epoch is a typed `CodeJournalError::StaleSpawnEpoch`.

### `tidebreak-harness`

- New workspace crate. Depends on `tidebreak-core` (default features off) plus existing workspace pins. No PTY dependency (0036); guarded by a lockfile/manifest test.
- `HarnessAdapter` / `HarnessSession`, `LaunchPlan`, probe (login-shell `command -v`), stream budgets, bypass-flag denylist (including user extras).
- Claude Code adapter written against redacted v2.1.233 fixtures (plain-text, tool-use, permission-denied, resume, interrupt). Unknown stream types increment a counter and are logged size-capped.
- Dev-only `capture` binary behind `feature = "capture"`.

## Caps left `Unknown` (and why)

For Claude Code 2.1.233, verified from fixtures / `--help`:

| Flag | Level | Why |
|---|---|---|
| `resume` | Supported | resume fixture + `--resume` |
| `streaming_deltas` | Supported | `stream_event` text deltas |
| `plan_mode` | Supported | `--permission-mode plan` and `system/init` |
| `native_interrupt` | Supported | interrupt fixture (`control_response`, `aborted_streaming`) |
| `structured_approvals` | **Unknown** | `--help` on 2.1.233 does not list a permission-prompt-tool flag; the permission-denied fixture is a resolved denial, not a live MCP approval channel |
| `mid_turn_steering` | **Unknown** | no streamed-JSON stdin fixture |
| `reasoning_levels` | **Unknown** | thinking deltas exist; effort control was not captured |
| `native_file_change_events` | **Unknown** | not present in the captured streams |

Auth observation is also uncaptured (`authenticated: None`). `decide()` returns an explicit error rather than guessing a channel.

## What I could not verify

- Live Claude Code spawn/interrupt against a real child in this change (CI cannot run live harnesses). Session spawn, SIGINT-then-kill, and env passthrough are implemented but not smoke-tested with `TIDEBREAK_LIVE_HARNESS`.
- The permission-prompt-tool / loopback MCP approval path. Types and SessionSpec wiring points are documented; launch args are not composed.
- Exact original capture argv for the salvaged 2.1.233 streams. The manifest records what `--help` documents for print mode and what `claude --version` printed on this machine.

## Tests

No tests were deleted.

Added: event variant-stability pin, attention replacement cases, entity graph round-trip, stale-epoch rejection, concurrent journal seq monotonicity, chat/code entity separation scan, probe shims (missing / relative / noisy profile / version), bypass denylist (including extras), PTY-dep guard, fixture replay with exact expected sequences, unrecognized-event counting.

## Scope not in this PR

Server routes, session worker, worktree/git manager, UI, generated wire export. Those stay on later PRs.


crates/tidebreak-harness/src/claude/session.rs
```diff
@@ -0,0 +1,192 @@
...
+        }
+        let mut child = command.spawn()?;
+        let stdout = child
+            .stdout
+            .take()
+            .ok_or_else(|| HarnessError::Other("engine child has no stdout".into()))?;
+        self.child = Some(child);
+
+        let mut parser = ClaudeStreamParser::new();
+        let mut lines = BufReader::new(stdout).lines();
```
naingthet: `stderr` is piped and never drained. `--verbose` is on the argv; once the OS pipe buffer fills the child blocks on a stderr write while this loop waits on stdout — classic deadlock. Inherit stderr, or drain it on a side task into a size-capped buffer (doctor/debug), but do not leave an unread pipe.

Same function then reads stdout with unbounded `BufReader::lines()`, so `StreamLineBuffer` / `StreamBudget` in this crate are unused on the only live reader. A single newline-free chunk grows without bound, which is the case those types were written to close. Wire the budget here (or drop the unused module until a reader exists).

crates/tidebreak-core/src/db/ops/code/session.rs
```diff
@@ -0,0 +1,129 @@
...
+        )
+        .col_expr(
+            entities::code_session::Column::ChildPid,
+            sea_orm::sea_query::Expr::value(child_pid),
+        )
+        .filter(entities::code_session::Column::Id.eq(id.0))
+        .exec(&store.conn)
+        .await
+        .map_err(store_err)?;
+    Ok(session.spawn_epoch)
```
naingthet: The spawn-epoch fence is a read-modify-write with no session lock and no `WHERE spawn_epoch = <old>`. Two concurrent bumps both read `n` and both write `n+1`, so a superseded worker keeps a live epoch and `append_event`'s stale-epoch check cannot see it.

Journal appends already serialize on `acquire_code_session_write_lock`. This bump is the other half of that fence and has to take the same lock (or an atomic `SET spawn_epoch = spawn_epoch + 1 … RETURNING`) in one transaction. The sequential test in `db/tests/code.rs` does not catch the lost update.

crates/tidebreak-harness/src/probe.rs
```diff
@@ -0,0 +1,256 @@
...
+
+struct ShellOutput {
+    stdout: String,
+    status_ok: bool,
+}
+
+async fn run_login_shell(host: &HostEnv, script: &str) -> Result<ShellOutput, ProbeError> {
+    let mut command = Command::new(&host.shell);
+    command
+        .arg("-lc")
```
naingthet: Same contradiction as #2122: this is `$SHELL -lc`, a login *non-interactive* shell. zsh will not source `~/.zshrc`, which is where nvm/fnm/asdf (and most gateway env) live — the files the module docs cite as the reason not to use the GUI PATH.

`passthrough_env()` then copies the *GUI process* environment into the child (`lib.rs:316`), not the login-shell environment 0034 said was the whole gateway story. So even a binary found via an absolute path starts without the user's shell `PATH`/`ANTHROPIC_*`/version-manager vars.

Use an interactive login shell for resolution (`-lic`, or whatever actually matches a terminal), and give the child that shell's environment minus Tidebreak-prefixed keys. The noisy-profile probe test stays; add a shim that only exports the binary from `.zshrc`-equivalent startup so `-lc` fails it.

naingthet (commented):
Reviewed the domain types, baseline tables + epoch bump, journal fence, and the Claude Code adapter against ADRs 0030–0036.

The type split, exhaustive `HarnessCaps`, permission-name subset of chat, journal-shape pin, fixture-replay with exact expected sequences, unrecognized-event counting, and bypass denylist are in good shape. Caps left `Unknown` rather than guessed is the right honesty for 2.1.233.

Three implementation defects on the paths this PR actually ships — not “add more tests.”

crates/tidebreak-harness/src/claude/session.rs
```diff
@@ -0,0 +1,192 @@
...
+        let mut env = self.spec.extra_env.clone();
+        env.retain(|(key, _)| !key.to_ascii_uppercase().starts_with("TIDEBREAK_") && key != "PWD");
+        let plan = LaunchPlan {
+            argv,
+            cwd: self.spec.worktree.clone(),
+            env,
+        };
+        validate_launch_plan(&plan)?;
+        Ok(plan)
+    }
```
naingthet, line 80: `input.text` is an argv element, then `validate_launch_plan` walks every argv token — including the prompt. `is_bypass_flag` matches any string that contains `dangerous` plus `skip`/`bypass` plus `permission`/`approval`/`sandbox`, so a user message like “this is dangerous, skip the permission for writes” refuses the turn with `LaunchRejected`.

Scan only composed flags (tokens that start with `-`), not the prompt. Passing the prompt on argv is itself brittle if it starts with `-`; stdin (`-p` with no operand, streamed JSON) keeps the denylist off user text entirely.

naingthet (commented):
One more on the launch-plan denylist, missed in the first pass.

naingthet (commented):
Looked at `27f2fef6` (parser mapping only).

`TurnInterrupted` gated on `terminal_reason == aborted_streaming` is the right fixture-backed rule — the previous `subtype.contains("error")` path would have called real engine errors interrupts.

The four findings on `989ec387` (unread stderr + unused stream budget, racy `bump_spawn_epoch`, `$SHELL -lc` + GUI env, denylist scanning the prompt) are untouched by this commit and still stand.

One leftover in the new mapping:

crates/tidebreak-harness/src/claude/parse.rs
```diff
@@ -0,0 +1,479 @@
...
+                // `permission_denied` is a resolved denial, not a parked
+                // request. A live approval channel (the permission-prompt
+                // tool) is what produces real `ApprovalRequested` events;
+                // synthesizing one here would mint a spurious approval row.
+                let mut events = Vec::new();
+                if !call_id.is_empty() {
+                    events.push(HarnessEvent::ApprovalResolved {
+                        harness_ref: HarnessApprovalRef { call_id },
+                        decision: ApprovalDecision::Deny { feedback: None },
+                    });
```
naingthet: Dropping `ApprovalRequested` is correct — this event is a resolved engine denial, not a parked card. Emitting `ApprovalResolved` for the same case still synthesizes an approval decision the user never made and that `decide()` cannot complete. A later worker that keys resolutions off `harness_ref` will either create a row that was never pending or fail to find one.

`HarnessNotice` plus the existing `ToolCompleted { Failed }` from the `tool_result` already carry the signal. Leave `ApprovalResolved` for a real channel decision.

naingthet:
Triage of the review rounds — all five findings accepted; fixes are being applied on this branch now:

1. **stderr never drained** (deadlock risk with `--verbose`) — accepted, P1. Side-task drain into a size-capped buffer, and the stdout read path moves onto the crate's own `StreamLineBuffer`/`StreamBudget` so parsing is bounded as 0031 requires.
2. **`bump_spawn_epoch` read-modify-write race** — accepted, P1. The bump takes the same session write lock journal appends take, so the fence's two halves serialize on one lock; concurrent-bump test added.
3. **`$SHELL -lc` + GUI-process env** — accepted. #2122 was updated first (0034 now specifies an interactive login probe with sentinel-bounded output and an environment snapshot captured from that same probe); the code follows the amended record, including its env-capture validation case.
4. **Denylist scanning the prompt** — accepted. Only flag-shaped tokens are scanned, and the prompt moves off argv where the CLI supports stdin delivery; false-positive test added.
5. **Synthetic `ApprovalResolved` on `permission_denied`** — accepted. Notice-only, per the argument that a fabricated decision breaks `harness_ref`-keyed resolution downstream; expected fixture updated.

Will note here when the fixes are pushed. (The earlier `test`-lane red was the pre-existing `sandbox_container_run` teardown flake — green on rerun, unrelated to this branch.)

naingthet:
Round-2 fixes pushed in `52f20703` — all five findings addressed:

1. stderr drained on a capped side task; stdout reads now go through `StreamLineBuffer`/`StreamBudget`.
2. `bump_spawn_epoch` takes the same session write lock as journal appends, inside a transaction, with a guarded conditional update as a second fence; concurrent-bump test added.
3. Probe is `$SHELL -ilc` with sentinel-delimited, timeout-bounded output; the child environment is the captured shell snapshot (minus Tidebreak-prefixed vars), not the GUI process env — implements amended 0034 including its env-capture validation case.
4. Denylist scans only flag-shaped tokens (false-positive test added), and the prompt no longer rides argv at all — it travels on stdin.
5. `permission_denied` maps to a notice only; no synthetic approval events. Expected fixture updated.

Ready for another look.

naingthet (commented):

Checked `52f20703` against the five findings still open on `27f2fef6`. Stdout now reads through `StreamBudget` / `StreamLineBuffer`, and stderr is drained on a side task into a 64KiB cap, so the `--verbose` pipe cannot deadlock. `bump_spawn_epoch` takes the same session-row lock as journal appends, CAS-updates `WHERE spawn_epoch = old`, and the new concurrent bump test would fail a lost update. Probe is `$SHELL -ilc` with sentinel-delimited env capture; children start from that snapshot via `filter_child_env`, not the GUI process env. The prompt moved to stdin (`claude -p` with no operand), the denylist only scans flag-shaped tokens, and `permission_denied` emits `HarnessNotice` only — no synthesized `ApprovalRequested` / `ApprovalResolved`.

Caps left `Unknown` and `decide()` still refusing to guess a channel remain the right honesty for 2.1.233.

PR #2124, opened 2026-08-15
feat(server): code mode workspaces, session worker, and routes

naingthet:
## What landed

Walking-skeleton server layer for code mode, stacked on the foundation (#2123). Spec is #2122.

- **Worktree manager** — every git call is a bounded, non-interactive argv to the user's `git`. Repo registration canonicalizes to the toplevel and refuses bare repos. Workspace create is `git worktree add -b` under `<data_dir>/code/worktrees/<repo-slug>/<workspace-slug>/`; branch collision is a user-visible error (no auto-suffix). Setup-script failure preserves the checkout and marks `SetupFailed`. Archive requires `force` when the tree is dirty or unpushed, then `worktree remove` (already-gone tolerated) + `prune`. The branch is kept.
- **Session worker** — one task per session, the only journal writer, epoch-fenced. `HarnessEvent` → persist as `CodeEvent` → publish on a per-session bus (journal-before-live). Turns accepted while Idle; while Running, mid-turn steer only if the adapter caps say `Supported`, otherwise 409. Interrupt calls `HarnessSession::interrupt()`.
- **Recovery** — boot scan of `Running` sessions. Dead recorded pid → open turn `Interrupted` (journaled), session Idle, attention `NeedsYou { source: Lifecycle }`. Live pid → `Fenced { OrphanAlive }`. Reap resolves a fenced session. A pid that was not recorded at spawn is never signaled; `EPERM` counts as alive.
- **Routes** under the existing bearer auth: repos, workspaces, sessions, turns, interrupt, reap, doctor (`GET /code/harnesses` + refresh), and `WS /code/sessions/{id}/events?after=` (subscribe-live-first, snapshot, replay, live, seq-deduped). Session creation is **Plan only**; `Ask`/`Auto` return a typed `permission_mode_unavailable` refusal.
- **Scripted harness** (`scripted-harness` feature / tests) so route and worker tests run real turns without a live CLI.
- **Wire types** — repo/workspace/session/turn snapshots, sequenced event frame, doctor payload. Regenerated `wire.ts`.

## Deliberately out of scope

Approvals bridge / MCP endpoint, `/code/updates` digest channel, attention beyond recovery-set states, diffs/files routes, checkpoint production (`checkpoint.rs`), git commit/push/PR, quick actions, terminals.

## What I could not verify

Live-harness paths (real Claude Code / other engines). Those cannot run in CI; tests drive the scripted adapter and throwaway temp git repos.

### DIFF
--- a/crates/tidebreak-server/src/code/runtime.rs
+++ b/crates/tidebreak-server/src/code/runtime.rs
@@ -1,0 +1,620 @@
+//! Process-wide code-mode runtime: adapters, workers, worktrees, recovery.
+
+use std::collections::HashMap;
+use std::path::PathBuf;
+use std::sync::{Arc, Mutex};
+
+use chrono::Utc;
+use tokio::sync::oneshot;
+
+use tidebreak_core::db::code::{
+    delete_repo, delete_workspace, get_repo, get_repo_by_root_path, get_session, get_workspace,
+    insert_repo, insert_session, insert_workspace, list_repos, list_sessions,
+    list_sessions_for_workspace, list_workspaces, save_repo, save_session, save_workspace,
+};
+use tidebreak_core::{
+    Attention, AttentionSource, CapLevel, CodePermissionMode, CodeRepo, CodeSession, CodeSessionId,
+    CodeSessionLifecycle, CodeTurn, CodeWorkspace, CodeWorkspaceStatus, DbStore, HarnessKind,
+    RepoId, WorkspaceId,
+};
+use tidebreak_harness::{builtin_registry, AdapterRegistry, HarnessAdapter, HostEnv, SessionSpec};
+
+use super::bus::CodeEventBus;
+use super::recovery::{self, RecoveryAction};
+use super::session_worker::{
+    attach_engine, spawn_session_worker, WorkerCommand, WorkerError, WorkerHandle,
+};
+use super::worktree::{
+    self, archive_blockers, branch_name, create_worktree, prune_worktrees, remove_worktree,
+    run_setup_script, slugify, validate_repo_path, worktree_dir, WorktreeError,
+};
+use crate::error::ServerError;
+
+/// Shared code-mode services for the process.
+pub(crate) struct CodeRuntime {
+    pub db: Arc<DbStore>,
+    pub bus: Arc<CodeEventBus>,
+    pub adapters: AdapterRegistry,
+    pub data_dir: PathBuf,
+    host: HostEnv,
+    workers: Mutex<HashMap<CodeSessionId, WorkerHandle>>,
+}
+
+impl CodeRuntime {
+    pub(crate) fn new(db: Arc<DbStore>, data_dir: PathBuf) -> Self {
+        Self {
+            db,
+            bus: Arc::new(CodeEventBus::default()),
+            adapters: builtin_registry(),
+            data_dir,
+            host: HostEnv::from_process(),
+            workers: Mutex::new(HashMap::new()),
+        }
+    }
+
+    #[cfg(any(test, feature = "scripted-harness"))]
+    pub(crate) fn with_registry(
+        db: Arc<DbStore>,
+        data_dir: PathBuf,
+        adapters: AdapterRegistry,
+    ) -> Self {
+        Self {
+            db,
+            bus: Arc::new(CodeEventBus::default()),
+            adapters,
+            data_dir,
+            host: HostEnv::from_process(),
+            workers: Mutex::new(HashMap::new()),
+        }
+    }
+
+    pub(crate) async fn recover(&self) -> Result<Vec<RecoveryAction>, ServerError> {
+        recovery::recover_running_sessions(&self.db)
+            .await
+            .map_err(ServerError::from)
+    }
+
+    pub(crate) fn adapter(
+        &self,
+        kind: HarnessKind,
+    ) -> Result<Arc<dyn HarnessAdapter>, ServerError> {
+        self.adapters.get(kind).ok_or_else(|| {
+            ServerError::bad_request_kind(
+                "harness_unavailable",
+                format!("no adapter is registered for {kind}"),
+            )
+        })
+    }
+
+    pub(crate) async fn register_repo(
+        &self,
+        root_path: PathBuf,
+        display_name: Option<String>,
+        default_base_ref: Option<String>,
+        branch_prefix: Option<String>,
+        setup_script: Option<String>,
+        archive_script: Option<String>,
+    ) -> Result<CodeRepo, ServerError> {
+        let validated = validate_repo_path(&root_path).await.map_err(map_worktree)?;
+        let toplevel = validated.toplevel.display().to_string();
+        if let Some(existing) = get_repo_by_root_path(&self.db, &toplevel).await? {
+            return Err(ServerError::conflict_kind(
+                "repo_already_registered",
+                format!(
+                    "repository {} is already registered as {}",
+                    toplevel, existing.id
+                ),
+            ));
+        }
+        // Nested registrations of the same toplevel are already collapsed by
+        // canonicalize + unique root_path. A path inside another registered
+        // repo would resolve to the same toplevel.
+        let name = display_name
+            .map(|value| value.trim().to_owned())
+            .filter(|value| !value.is_empty())
+            .unwrap_or_else(|| {
+                validated
+                    .toplevel
+                    .file_name()
+                    .map(|name| name.to_string_lossy().into_owned())
+                    .unwrap_or_else(|| "repo".into())
+            });
+        let repo = CodeRepo {
+            id: RepoId::new(),
+            root_path: toplevel,
+            display_name: name,
+            default_base_ref: default_base_ref
+                .filter(|value| !value.trim().is_empty())
+                .unwrap_or_else(|| "main".into()),
+            branch_prefix: branch_prefix
+                .filter(|value| !value.trim().is_empty())
+                .unwrap_or_else(|| "tidebreak/".into()),
+            setup_script,
+            archive_script,
+            quick_actions: Vec::new(),
+            created_at: Utc::now(),
+        };
+        insert_repo(&self.db, &repo).await?;
+        Ok(repo)
+    }
+
+    pub(crate) async fn list_repos(&self) -> Result<Vec<CodeRepo>, ServerError> {
+        Ok(list_repos(&self.db).await?)
+    }
+
+    pub(crate) async fn get_repo(&self, id: RepoId) -> Result<CodeRepo, ServerError> {
+        get_repo(&self.db, id)
+            .await?
+            .ok_or_else(|| ServerError::not_found(format!("repo {id} not found")))
+    }
+
+    pub(crate) async fn save_repo(&self, repo: &CodeRepo) -> Result<(), ServerError> {
+        if !save_repo(&self.db, repo).await? {
+            return Err(ServerError::not_found(format!(
+                "repo {} not found",
+                repo.id
+            )));
+        }
+        Ok(())
+    }
+
+    pub(crate) async fn delete_repo(&self, id: RepoId) -> Result<(), ServerError> {
+        let workspaces = list_workspaces(&self.db, Some(id)).await?;
+        if workspaces
+            .iter()
+            .any(|workspace| workspace.status != CodeWorkspaceStatus::Archived)
+        {
+            return Err(ServerError::conflict_kind(
+                "repo_has_workspaces",
+                "archive every workspace before deleting the repository",
+            ));
+        }
+        if !delete_repo(&self.db, id).await? {
+            return Err(ServerError::not_found(format!("repo {id} not found")));
+        }
+        Ok(())
+    }
+
+    pub(crate) async fn create_workspace(
+        &self,
+        repo_id: RepoId,
+        title: Option<String>,
+        base_ref: Option<String>,
+    ) -> Result<CodeWorkspace, ServerError> {
+        let repo = self.get_repo(repo_id).await?;
+        let title = title
+            .map(|value| value.trim().to_owned())
+            .filter(|value| !value.is_empty())
+            .unwrap_or_default();
+        let id = WorkspaceId::new();
+        let branch = branch_name(&repo.branch_prefix, &title, id.0.as_u128());
+        let existing = list_workspaces(&self.db, Some(repo_id)).await?;
+        if existing
+            .iter()
+            .any(|workspace| workspace.branch_name == branch)
+        {
+            return Err(ServerError::conflict_kind(
+                "branch_collision",
+                format!("branch {branch} already exists on this repository"),
+            ));
+        }
+        let repo_slug = {
+            let from_name = slugify(&repo.display_name);
+            if from_name.is_empty() {
+                slugify(&repo.root_path)
+            } else {
+                from_name
+            }
+        };
+        let workspace_slug = {
+            let from_title = slugify(&title);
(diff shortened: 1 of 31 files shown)


crates/tidebreak-server/src/code/runtime.rs
```diff
@@ -0,0 +1,620 @@
...
+            host: HostEnv::from_process(),
+            workers: Mutex::new(HashMap::new()),
+        }
+    }
+
+    pub(crate) async fn recover(&self) -> Result<Vec<RecoveryAction>, ServerError> {
+        recovery::recover_running_sessions(&self.db)
+            .await
+            .map_err(ServerError::from)
+    }
```
naingthet, line 109: Recovery only mutates the session row. Workers are inserted solely in `create_session` (line 454). After a process restart every recovered session is Idle (or Fenced→reaped Idle) with no worker.

`submit_turn` then 409s `session_worker_missing`, and `create_session` 409s `session_exists` because the Idle row is not `Ended`. There is no end-session route. The user cannot submit, cannot recreate, and cannot unstick the workspace.

Re-launch (or resume) a worker for every non-Ended, non-Fenced session after `recover()`, or end those sessions as part of recovery so a new one can be created.

crates/tidebreak-server/src/code/runtime.rs
```diff
@@ -0,0 +1,620 @@
...
+            }
+        }
+        remove_worktree(std::path::Path::new(&repo.root_path), path)
+            .await
+            .map_err(map_worktree)?;
+        let _ = prune_worktrees(std::path::Path::new(&repo.root_path)).await;
+        workspace.status = CodeWorkspaceStatus::Archived;
+        workspace.archived_at = Some(Utc::now());
+        save_workspace(&self.db, &workspace).await?;
+        Ok(workspace)
```
naingthet, line 372: `git worktree remove --force` runs while a live session worker may still be using this checkout. Archive does not interrupt/shutdown the worker or mark the session `Ended`, so the engine keeps writing into a deleted tree and the Idle-or-Running session still blocks `create_session`.

Refuse archive while a non-Ended session exists, or shut the worker down and end the session before `remove_worktree`.

naingthet (commented):
Reviewed the walking-skeleton server layer against #2122 / 0030–0034: worktree argv, journal-before-live, epoch fence, signal-0-only recovery, Plan-only session create, and the WS subscribe-then-replay-then-live path.

Three implementation defects on the paths this PR ships — after a restart or an archive the product is stuck, and mid-turn submit invents a turn that does not exist.

crates/tidebreak-server/src/code/runtime.rs
```diff
@@ -0,0 +1,620 @@
...
+                status: tidebreak_core::CodeTurnStatus::Running,
+                user_input: message,
+                user_input_blob_id: None,
+                checkpoint_ref: None,
+                diffstat: None,
+                usage: None,
+                narrative: None,
+                started_at: Utc::now(),
+                ended_at: None,
+            });
```
naingthet: Two problems on the mid-turn path.

1. #2122 / 0009 queue-default: a turn submitted while Running should enqueue. This steers when `mid_turn_steering == Supported` and 409s otherwise. Claude Code on #2123 advertises `Unknown`, so every mid-turn POST is a 409.
2. The steer success path returns a brand-new `CodeTurnId`, `ordinal: 0`, never written to the DB. The client will poll or key UI on a turn that does not exist.

Queue (or 409 without a fake row) from this route; do not mint a turn.

crates/tidebreak-server/src/code/runtime.rs
```diff
@@ -0,0 +1,620 @@
...
+        };
+        let workspace_slug = {
+            let from_title = slugify(&title);
+            if from_title.is_empty() {
+                worktree::two_word_name(id.0.as_u128())
+            } else {
+                from_title
+            }
+        };
+        let path = worktree_dir(&self.data_dir, &repo_slug, &workspace_slug);
```
naingthet: The checkout path is `<data_dir>/code/worktrees/<display-name-slug>/<title-slug>/`. Two registered repos that share a directory name (two clones of the same project is the common case) write into the same directory; the second `worktree add` either fails opaquely or the first archive deletes the second checkout.

Key the path on `repo.id` / `workspace.id` (the slugs can stay as a suffix for humans).

naingthet:
Triage of the first review round — all four findings accepted, fixes in progress on this branch:

1. **Recovered sessions have no worker** (restart bricks the workspace) — accepted, P1. Recovery will re-launch workers for non-Ended, non-Fenced sessions, with a restart test proving a recovered session accepts a turn.
2. **Archive under a live worker** — accepted, P1. Archive will shut down the worker and end the session before `worktree remove`.
3. **Mid-turn path** — accepted; being fixed together with the queue-default rule from #2122/0009 (a mid-turn submit enqueues into a single pending slot and creates a real persisted turn — no fabricated `CodeTurnId`, no 409-by-default; explicit steer stays capability-gated).
4. **Worktree path collision across same-named repos** — accepted. On-disk paths will be keyed on repo/workspace ids with slugs as a human-readable suffix.

Will comment when the round is pushed.

crates/tidebreak-server/src/code/runtime.rs
```diff
@@ -0,0 +1,589 @@
...
+            host: HostEnv::from_process(),
+            workers: Mutex::new(HashMap::new()),
+        }
+    }
+
+    pub(crate) async fn recover(&self) -> Result<Vec<RecoveryAction>, ServerError> {
+        recovery::recover_running_sessions(&self.db)
+            .await
+            .map_err(ServerError::from)
+    }
```
naingthet, line 109: Unchanged from `96c313ee`. Recovery only mutates the session row. After a restart every recovered session is Idle (or Fenced→reaped Idle) with no worker.

`submit_turn` then 409s `session_worker_missing`, and `create_session` 409s `session_exists` because the Idle row is not `Ended`. The user cannot submit, cannot recreate, and cannot unstick the workspace.

Re-launch a worker for every non-Ended, non-Fenced session after `recover()`, or end those sessions as part of recovery so a new one can be created.

crates/tidebreak-server/src/code/runtime.rs
```diff
@@ -0,0 +1,589 @@
...
+            }
+        }
+        remove_worktree(std::path::Path::new(&repo.root_path), path)
+            .await
+            .map_err(map_worktree)?;
+        let _ = prune_worktrees(std::path::Path::new(&repo.root_path)).await;
+        workspace.status = CodeWorkspaceStatus::Archived;
+        workspace.archived_at = Some(Utc::now());
+        save_workspace(&self.db, &workspace).await?;
+        Ok(workspace)
```
naingthet, line 372: Still unfixed. `git worktree remove --force` runs while a live session worker may still be using this checkout. Archive does not interrupt/shutdown the worker or mark the session `Ended`, so the engine keeps writing into a deleted tree and the Idle-or-Running session still blocks `create_session`.

Refuse archive while a non-Ended session exists, or shut the worker down and end the session before `remove_worktree`.

crates/tidebreak-server/src/code/runtime.rs
```diff
@@ -0,0 +1,589 @@
...
+        };
+        let workspace_slug = {
+            let from_title = slugify(&title);
+            if from_title.is_empty() {
+                worktree::two_word_name(id.0.as_u128())
+            } else {
+                from_title
+            }
+        };
+        let path = worktree_dir(&self.data_dir, &repo_slug, &workspace_slug);
```
naingthet: Still keyed on `<data_dir>/code/worktrees/<display-name-slug>/<title-slug>/`. Two registered repos that share a directory name write into the same directory; the second `worktree add` either fails opaquely or the first archive deletes the second checkout.

Key the path on `repo.id` / `workspace.id` (the slugs can stay as a suffix for humans).

naingthet (commented):
Checked `2761514a` against the four findings on `96c313ee`. Mid-turn POST now parks one follow-up and drains it after the live turn, with `queue_full` on a second and explicit `/steer` returning `steering_unavailable`. That closes the fabricated-turn / 409 path.

The other three still stand: recovery still leaves Idle sessions with no worker, archive still deletes the worktree under a live session, and checkout paths are still keyed on display-name slugs.

crates/tidebreak-server/src/code/runtime.rs
```diff
@@ -0,0 +1,594 @@
...
+            host: HostEnv::from_process(),
+            workers: Mutex::new(HashMap::new()),
+        }
+    }
+
+    pub(crate) async fn recover(&self) -> Result<Vec<RecoveryAction>, ServerError> {
+        recovery::recover_running_sessions(&self.db)
+            .await
+            .map_err(ServerError::from)
+    }
```
naingthet, line 109: Unchanged. Recovery only mutates the session row. After a restart every recovered session is Idle (or Fenced→reaped Idle) with no worker.

`submit_turn` then 409s `session_worker_missing`, and `create_session` 409s `session_exists` because the Idle row is not `Ended`. The user cannot submit, cannot recreate, and cannot unstick the workspace.

Re-launch a worker for every non-Ended, non-Fenced session after `recover()`, or end those sessions as part of recovery so a new one can be created.

crates/tidebreak-server/src/code/runtime.rs
```diff
@@ -0,0 +1,594 @@
...
+            }
+        }
+        remove_worktree(std::path::Path::new(&repo.root_path), path)
+            .await
+            .map_err(map_worktree)?;
+        let _ = prune_worktrees(std::path::Path::new(&repo.root_path)).await;
+        workspace.status = CodeWorkspaceStatus::Archived;
+        workspace.archived_at = Some(Utc::now());
+        save_workspace(&self.db, &workspace).await?;
+        Ok(workspace)
```
naingthet, line 372: Still unfixed. `git worktree remove --force` runs while a live session worker may still be using this checkout. Archive does not interrupt/shutdown the worker or mark the session `Ended`, so the engine keeps writing into a deleted tree and the Idle-or-Running session still blocks `create_session`.

Refuse archive while a non-Ended session exists, or shut the worker down and end the session before `remove_worktree`.

crates/tidebreak-server/src/code/runtime.rs
```diff
@@ -0,0 +1,594 @@
...
+        };
+        let workspace_slug = {
+            let from_title = slugify(&title);
+            if from_title.is_empty() {
+                worktree::two_word_name(id.0.as_u128())
+            } else {
+                from_title
+            }
+        };
+        let path = worktree_dir(&self.data_dir, &repo_slug, &workspace_slug);
```
naingthet: Still keyed on `<data_dir>/code/worktrees/<display-name-slug>/<title-slug>/`. Two registered repos that share a directory name write into the same directory; the second `worktree add` either fails opaquely or the first archive deletes the second checkout.

Key the path on `repo.id` / `workspace.id` (the slugs can stay as a suffix for humans).

naingthet (commented):
Checked `4332efaf` against the three findings still open on `2761514a`. Mid-turn queue is tighter: `submit_turn` now keys off the open turn as well as lifecycle, does not consult `mid_turn_steering`, and the test waits until Running, submits with `Unknown` caps, and asserts the follow-up starts only after the live turn ends. That path is closed.

The other three still stand: recovery still leaves Idle sessions with no worker, archive still deletes the worktree under a live session, and checkout paths are still keyed on display-name slugs.

naingthet:
Full round pushed (`53d80c1d` on top of the earlier queue fix):

1. **Recovery re-attaches workers** for every non-Ended, non-Fenced session after the boot scan, so a restarted install accepts turns again; failures to resume degrade to a logged warning rather than a stuck workspace. Restart test added.
2. **Archive ends the session first** — worker shutdown + `Ended` transition before `worktree remove`; archiving with a live session requires the explicit force path. Test added.
3. **Worktree paths keyed on ids** — `<repo-id>-<slug>/<workspace-id>-<slug>/`, so same-named repos cannot collide. Tests adjusted.

Mid-turn queue-default was closed in the previous commits per your verification. Ready for another look.

naingthet (commented):
Checked `53d80c1d` against the three findings on `4332efaf`. Dead-pid recovery now relaunches a worker and the restart test submits a turn. Archive shuts the worker down and marks the session `Ended` before `remove_worktree`, and refuses a Running session without `force`. Worktree paths include `repo.id` / `workspace.id`.

Two leftovers, one of them introduced by the archive fix.

crates/tidebreak-server/src/code/runtime.rs
```diff
@@ -0,0 +1,683 @@
...
+            )
+            .await
+            {
+                return Err(ServerError::unprocessable_kind(
+                    "archive_script_failed",
+                    err,
+                ));
+            }
+        }
+        self.end_workspace_sessions(id, force).await?;
```
naingthet: `end_workspace_sessions` runs before the dirty-tree check. A workspace with an Idle session and uncommitted files, archived without `force`, Ends the session and then 409s `uncommitted`. The workspace is still Active, the checkout is still there, and `create_session` is now allowed — the user lost the session without the archive completing.

Check blockers first (and refuse a Running session first). Only then end sessions and remove the tree.

crates/tidebreak-server/src/code/runtime.rs
```diff
@@ -0,0 +1,683 @@
...
+            ));
+        }
+        let handle = self.workers.lock().expect("code workers").remove(&id);
+        if let Some(handle) = handle {
+            let _ = handle.commands.send(WorkerCommand::Shutdown).await;
+        }
+        recovery::reap_session(&self.db, session)
+            .await
+            .map_err(ServerError::from)
+    }
```
naingthet, line 532: The original stuck-workspace hole is still here on the fenced path. Recovery skips `Fenced` (correct). `reap` then writes Idle and does not call `attach_and_spawn_worker`. `submit_turn` 409s `session_worker_missing`; `create_session` 409s `session_exists`.

After `reap_session`, attach a worker the same way `recover()` does for Idle rows.

naingthet:
Round-3 triage — both findings accepted:

1. **Archive ordering** — blockers will be evaluated before any session is ended; a refused archive leaves the session untouched. Test added for the dirty + Idle + no-force case.
2. **Reap leaves no worker** — `reap` will re-attach a worker exactly as `recover()` does for Idle rows; restart test extended to fenced→reap→submit.

Fixes in progress on this branch.

naingthet (commented):

Checked `e5bfa539` against the two leftovers on `53d80c1d`. Archive now refuses a Running session and a dirty/unpushed tree *before* `end_workspace_sessions`; the new no-force dirty-archive test leaves the Idle session and still accepts a turn. Reap clears the fence then `attach_and_spawn_worker`, and the extended recovery test submits after fenced→reap. `GET /code/workspaces/{id}/sessions` returns create-shaped snapshots (404 on a missing workspace, empty list, then the live row), which is what the UI needs to reattach.

naingthet:
Rebased onto `main` after the foundation merge; the diff is now server-layer only. Round 4 added `GET /code/sessions/{id}/turns` (turn snapshots for client-side transcript hydration — the snapshot-then-replay counterpart to the events socket). Ready for a final pass on that route.

naingthet (commented):

Checked `2a869f36` on top of the last clean head. `GET /code/sessions/{id}/turns` 404s a missing session, returns an empty list, then two completed turns in ordinal order with `user_input` and `usage`. That is the snapshot the UI needs to rebuild prompts after a store release. `list_turns` already orders by ordinal; the snapshot now carries `usage` so a reopen can show the same boundary the journal already painted.

PR #2125, opened 2026-08-15
feat(desktop): code mode route family and session surface

naingthet:
## What landed

Walking skeleton of the code-mode desktop surface, stacked on the server routes from #2124.

- Routes `/code`, `/code/r/$repoId`, `/code/w/$workspaceId` on the existing hash router, with a `CodeSidebar` built from the shared rail primitives (repos, recent workspaces, new-workspace, Chat/Code switch). Chat stores stay off this path.
- `ApiClient` methods for repos, harness doctor/refresh, workspaces, session create, turns, interrupt, and reap. Runtime validators in `ui/src/code/parsers.ts` for every wire type this surface consumes.
- `CodeSessionReducer` + per-session store factory + ref-counted `CodeSessionRegistry`. Unmounting the last view closes the socket; two views of one session share one store. The controller reuses chat's backoff/dedupe discipline against `WS /code/sessions/{id}/events?after=`.
- Workspace header (repo, branch, copyable/reveal worktree path, harness + version, lifecycle badge), transcript (`MessageMarkdown`, `ToolCardShell` + Command/FileRead/FileEdit/Search/Other, visible `HarnessNotice`, turn boundaries with duration/usage), Plan-only composer with Ask/Auto disabled and the server's "not yet available" reason, new-workspace dialog, fenced-state card with Reap, archive with force-confirm when the server reports uncommitted work.
- Settings section **Coding harnesses**: doctor (found/path/version/tier/caps/auth/remediation/unrecognized-event counts) and refresh.

## Spec and stack

- Spec #2122
- Foundation #2123
- Server #2124

## Out of scope

Approvals UI, diffs/files panels, PR flow, terminals, `/code/updates` digest store, attention badges, OS notifications, virtualization. No new npm dependencies.

## Verification

`pnpm test` (1009) and `pnpm build` green from `crates/tidebreak-desktop/ui/`. Reducer, registry, CodeSidebar (ADR 0030: renders without chat stores), transcript, and composer disabled-modes tests added.

## Could not verify

No live harness in this environment. I did not drive the running desktop app against Claude Code (or any other engine): doctor empty-state vs ready-state, worktree reveal, archive force-confirm, fence/reap, and a real turn stream still need a human with an installed, signed-in harness.

The server has no list-sessions-for-workspace route yet, so a session created in this client is remembered locally after create. Reopening a workspace from another profile still needs an explicit Start session (or a later GET).

### DIFF
--- a/crates/tidebreak-desktop/ui/src/api.test.ts
+++ b/crates/tidebreak-desktop/ui/src/api.test.ts
@@ -1565,3 +1565,30 @@ describe("source download progress", () => {
     );
   });
 });
+
+describe("code workspace sessions", () => {
+  it("lists GET /code/workspaces/{id}/sessions", async () => {
+    const session = {
+      id: "sess-1",
+      workspace_id: "ws-1",
+      harness_kind: "claude_code",
+      permission_mode: "plan",
+      lifecycle: "idle",
+      attention: { state: { type: "working" }, source: "lifecycle" },
+      unrecognized_event_count: 0,
+      created_at: "2026-08-15T12:00:00.000Z",
+    };
+    const fetchMock = vi.fn().mockResolvedValue(
+      new Response(JSON.stringify([session]), { status: 200 }),
+    );
+    vi.stubGlobal("fetch", fetchMock);
+    const client = new ApiClient("http://127.0.0.1", "token");
+
+    await expect(client.listCodeWorkspaceSessions("ws-1")).resolves.toEqual([
+      session,
+    ]);
+    const [url, init] = fetchMock.mock.calls[0] as [string, RequestInit];
+    expect(url).toBe("http://127.0.0.1/code/workspaces/ws-1/sessions");
+    expect(init.method ?? "GET").toBe("GET");
+  });
+});
--- a/crates/tidebreak-desktop/ui/src/api.test.ts
+++ b/crates/tidebreak-desktop/ui/src/api.test.ts
@@ -1,6 +1,8 @@
 import { afterEach, describe, expect, it, vi } from "vitest";
 import {
   ApiClient,
+  archiveForceKind,
+  HttpError,
   parseAgentActivityHistory,
   parseAgentRunTaskPlan,
   parseFolderAccessRequest,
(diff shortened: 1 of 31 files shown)


naingthet (commented):
Reviewed the walking-skeleton desktop surface against #2124 / 0030: hash routes, rail primitives without chat stores, runtime parsers, Plan-only composer, ref-counted session registry, and `GET /code/workspaces/{id}/sessions` on `a3a36c40`.

Three implementation defects on the paths this PR ships.

1. **`CodeWorkspacePage.tsx` `send()` keeps the optimistic user bubble when `submitCodeTurn` fails.** Chat already rolls this back (`ChatRoute.sendMessage` removes the optimistic id and restores the draft). Here the catch only toasts, so a 4xx/5xx (empty after trim on the server, fenced session, `session_exists` sibling, parse miss on a queued body) leaves a user turn that no journal event will ever answer. A retry then stacks a second copy. Insert after accept, or remove the item (and restore the draft) in the `catch`.

2. **`ARCHIVE_FORCE_KINDS` omits `unpushed`.** `archive_blockers` returns three kinds: `uncommitted`, `unpushed`, `uncommitted_and_unpushed`. The confirm path only handles the first and third, so a clean tree with unpushed commits 409s into a toast and cannot be archived from this UI. Add `unpushed` (and word the confirm for leftover work, not only uncommitted files).

3. **Replay cannot reconstruct the user side of the transcript.** `turn_started` carries only `turn_id`, and the reducer never paints a user item. User text is written only in `send()`. Unmounting the pane (navigate `/code` → back, or the last registry release) disposes the store; the next acquire replays from `after=0` and shows assistant / tools / notices with no prompts. The seq cursor and reconnect path are otherwise right — they just have nothing to replay for the messages the composer just sent. Either hydrate user text from turn rows, or keep user items in the reducer from a journaled event; local `Date.now()` inserts will not survive the socket you already built.

Smaller, not blocking: the lifecycle badge stays on the create snapshot (`Created`) for the whole visit because nothing applies `turn_started` / `turn_completed` to `session.lifecycle`. Composer `busy` still drives interrupt.

naingthet:
Triage — all three findings accepted:

1. **Optimistic bubble on failed submit** — will mirror chat's rollback (insert-after-accept / remove-and-restore-draft in the catch), with a test.
2. **`ARCHIVE_FORCE_KINDS` missing `unpushed`** — added, confirm copy reworded for leftover work generally.
3. **Replay missing the user side** — fixed with chat's own snapshot-then-replay discipline rather than a journal-shape change: #2124 gains `GET /code/sessions/{id}/turns` (turn rows already persist `user_input`), and the controller hydrates user items from turns before replaying the journal from the cursor. Local time-keyed items are replaced by turn-keyed ones.

Also picking up the non-blocking lifecycle-badge note. Fixes in progress.

naingthet (commented):
Checked `b05ed93f` against the three findings on `a3a36c40`. Failed send no longer paints a bubble (`submitAcceptedTurn` waits for the snapshot; composer keeps the draft). `archiveForceKind` includes `unpushed`. Hydrate-then-replay is the right shape when turns actually parse.

Two leftovers, one of them on the path this commit added.

1. **`parseCodeTurn` rejects the snapshots `GET /turns` now returns.** `onlyKeys` lists `id` / `session_id` / `ordinal` / `status` / `user_input` / `checkpoint_ref` / `started_at` / `ended_at` and not `usage`. #2124 put optional `usage` on `CodeTurnSnapshot`, and the list-turns test asserts it on a completed row. `listCodeSessionTurns` then throws, `hydrateThenConnect` swallows the error, and reopen still has no user prompts. Allow `usage` and keep it on the snapshot (hydrate currently writes `usage: null` on the boundary even when the row has it).

2. **Live send appends the user item after the turn has already painted.** `submitCodeTurn` waits until the worker finishes. The socket delivers `turn_started` / deltas / `turn_completed` during that wait. `applyAcceptedTurn` then does `[...state.items, user]`, so the prompt lands below the assistant and the boundary. Insert before the first item with that `turnId` (append only when none exist). The hydrate-then-replay test starts from an empty store, so it does not catch this.

Archive copy and the lifecycle badge reading the store are fine.

naingthet:
Round-3 triage — both findings accepted:

1. **`parseCodeTurn` rejects `usage`** — validator will allow and carry it; parser + hydrate tests gain usage-bearing fixtures (the earlier fixture omitted it, which is exactly how this slipped).
2. **Accepted-send ordering** — the user item will insert before the first item of its `turnId` rather than appending after the streamed reply; test drives a store that already painted the turn.

Both land with the rebase onto `main`.

naingthet (commented):

Checked `2153b04e` against the two leftovers on `b05ed93f`. `parseCodeTurn` now allows optional `usage` and keeps it; hydrate writes that usage onto the boundary instead of `null`. `applyAcceptedTurn` inserts the user item before the first item with that `turnId`, so a late accept after the socket has already painted lands above the assistant and the boundary. The new parser and insert tests would fail the previous code.

PR #2126, opened 2026-08-15
refactor(core): stop minting ids from Default

naingthet:
## Summary

`id_type!` implemented `Default` as `Uuid::new_v4()`. A `..Default::default()` on any struct that happened to contain an id would invent a durable identity as a side effect. `RootAttachmentChangeId` already rejected nil on construct and deserialize, then offered the same `Default` hole.

Identity is now only minted by `new()`, a typed `derive` / `from_uuid`, or `From<Uuid>`.

## Test

- No callers of `ChatId::default()` (etc.) existed.
- `cargo test -p tidebreak-core --locked --lib id::tests`
- `cargo check --workspace --locked --all-targets`

### DIFF
--- a/crates/tidebreak-core/src/id.rs
+++ b/crates/tidebreak-core/src/id.rs
@@ -4,6 +4,11 @@
 //! say, passing a [`TurnId`] where a [`ChatId`] is expected. All ids
 //! serialize transparently (as the bare UUID string), so on the wire and in the
 //! `Store` they are indistinguishable from a plain UUID.
+//!
+//! Identity is minted only by [`id_type`]'s `new`, a typed `derive` /
+//! `from_uuid`, or an explicit `From<Uuid>`. There is no `Default`: a
+//! `..Default::default()` on a struct that happens to contain an id would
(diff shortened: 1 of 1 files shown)


naingthet (commented):

Removing `Default` from `id_type!` and `RootAttachmentChangeId` closes the `..Default::default()` hole that would mint a durable identity as a side effect. `new()`, typed `derive` / `from_uuid`, and `From<Uuid>` still mint ids; `RootAttachmentChangeId` still rejects nil on construct and deserialize. No `ChatId::default()`-style callers remain in the workspace, so this is a compile-time contract change rather than a behavior break.

naingthet (commented):

`52a49f0c` is the clippy follow-up after dropping `Default`: `new_without_default` would fail `-D warnings` the moment `Default` is gone, and allowing it on `id_type!` / `RootAttachmentChangeId::new` is the right place — the whole point of this PR is that `..Default::default()` must not mint an id. Docs on the module and both `new()`s still say why.

PR #2127, opened 2026-08-15
refactor(core): make shared Store surfaces required

naingthet:
## Summary

Nothing on `Store` was unused. Every `Err("… not implemented")` default is overridden by `DbStore`. Deleting methods is not available.

The remaining problem is growth: new surface could land optional, and `MemStore` / the test wrappers silently missed it.

Decision 0037: a method whose only default would be that error becomes required the moment two impls in this repo provide a real body. Capability traits are not chosen — no second production impl exists, and they would split atomic writes that touch two surfaces.

This change makes the eleven document and root-attachment methods required. Those already have real bodies on both `DbStore` and `MemStore`. The error defaults were dead. Test wrappers that wrap `inner` now forward them.

Making the other ~186 error-defaults required would be ~186 stubs on `MemStore` for methods those tests never call. That is a rewrite, not this PR.

## Test

- `cargo check -p tidebreak-core -p tidebreak-server --locked --all-targets`

### DIFF
--- a/crates/tidebreak-core/src/storage/store.rs
+++ b/crates/tidebreak-core/src/storage/store.rs
@@ -73,12 +73,6 @@ fn agent_run_storage_unavailable<T>() -> Result<T> {
     ))
 }
 
-fn root_attachment_storage_unavailable<T>() -> Result<T> {
-    Err(AgentError::Store(
-        "durable root attachment storage is not implemented by this Store".into(),
-    ))
-}
-
 fn operation_log_storage_unavailable<T>() -> Result<T> {
     Err(AgentError::Store(
         "durable operation-log storage is not implemented by this Store".into(),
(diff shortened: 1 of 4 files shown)


naingthet (commented):

The eleven document and root-attachment methods already have real bodies on both `DbStore` and `MemStore`, so dropping their `Err("… not implemented")` defaults only removes dead optionality. `PauseTerminalStore` and `TerminalFaultStore` forward the newly required methods they did not already override; the other four `impl Store`s still compile. Real defaults (`list_document_ids`, owner-scoped delegates) stay, and the leftover `document_storage_unavailable` helper is still used by the exec-snapshot / blob-retirement surface `MemStore` does not implement. Decision 0037 records the growth rule, the rejected rewrite, and when to reopen capability traits.

PR #2128, opened 2026-08-15
refactor(core): type the two AgentError payload branches

naingthet:
## Summary

`Store(String)` / `Config(String)` / `Message(String)` are catch-alls. Almost nobody inspects the payload. Two callers did:

- Gateway attestation remint: `Config` + `contains("invalid_target")`. One wording change away from reminting forever or never recovering.
- ChatGPT sign-in: `Authentication` + `starts_with("chatgpt sign-in required")`. Gateway sign-in already used the variant alone.

Those are now `InvalidTarget` and `SignInRequired`.

`SignInRequired` keeps `kind() == "authentication"` so turn-failure classification and the reconnect affordance do not change. `InvalidTarget` is its own kind so an MCP `Config` → 400 mapper cannot swallow a remintable token error if it ever leaks.

No other stringly variants: HTTP 404s already come from typed outcomes, lease loss is already a typed store outcome, and gateway `Message` copy for model-not-granted has no consumer that reads the text.

## Test

- `cargo test -p tidebreak-core --locked --lib error::tests`
- `cargo test -p tidebreak-server --locked --lib connectors::gateway::`
- `cargo test -p tidebreak-server --locked --lib connectors::chatgpt::`
- `cargo test -p tidebreak-server --locked --lib mcp_config`

### DIFF
--- a/crates/tidebreak-core/src/error.rs
+++ b/crates/tidebreak-core/src/error.rs
@@ -70,6 +70,17 @@ pub enum AgentError {
     #[error("provider credential missing: {0}")]
     MissingCredential(String),
 
+    /// No usable connector session — never signed in, revoked, or refresh
+    /// reuse. Callers surface a reconnect affordance rather than a fault.
+    #[error("sign-in required: {0}")]
+    SignInRequired(String),
+
+    /// The authorization server refused the requested OAuth target. An
+    /// attested mint remints under a fresh context id; any other failure
+    /// is terminal.
+    #[error("attestation context rejected: {0}")]
+    InvalidTarget(String),
+
     /// A persistence failure from the `Store` / `BlobStore`.
     #[error("store error: {0}")]
     Store(String),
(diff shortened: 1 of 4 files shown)


naingthet (commented):
## Review

The two stringly payload checks are now typed variants: `SignInRequired` keeps `kind() == "authentication"` so reconnect classification is unchanged, and `InvalidTarget` is its own kind so a Config→400 mapper cannot swallow a remintable attestation error. Gateway remint matches the variant, not `contains("invalid_target")`. Tests pin both kinds and that a Config string no longer remints.

naingthet (commented):

Rebase onto `main` after #2127/#2129 is clean — still the same four files. `SignInRequired` keeps `kind() == "authentication"` so turn-failure classification and the reconnect affordance stay put, while `is_sign_in_required` / `is_chatgpt_sign_in_required` match the variant instead of a prefix. `InvalidTarget` is its own kind, so the MCP `Config` → 400 mapper cannot swallow a remintable attestation rejection. The two new tests lock both of those.

PR #2129, opened 2026-08-15
build: pin the Rust toolchain to 1.97.1

naingthet:
## Summary

`rust-toolchain.toml` said `channel = "stable"`. CI honors that file and clippy runs `-D warnings`, so a rustup stable bump can fail every PR at once.

Pin to `1.97.1`, the version this workspace already builds with (2026-07-14). Bump it the same way as any other dependency.

## Test

- File-only change. Existing rust/workspace lanes on this PR are the verification.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,8 +36,10 @@ cargo test --workspace
 git config core.hooksPath .githooks
 ```
 
-The Rust toolchain is pinned in [`rust-toolchain.toml`](rust-toolchain.toml);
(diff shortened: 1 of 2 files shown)


naingthet (commented):

Pinning `rust-toolchain.toml` to `1.97.1` instead of floating `stable` stops a rustup bump from failing the whole tree under clippy `-D warnings`. CONTRIBUTING now matches what the file actually does, and the bump path is the same as any other pin. CI/local rustup both honor this file, so no workflow change is needed.

PR #2131, opened 2026-08-15
feat(code): per-turn checkpoints with diff and file review

naingthet:
## Summary

Completed turns now leave a hidden worktree checkpoint, and the workspace UI can review those changes turn by turn.

- After a successful turn, snapshot tracked *and* untracked work through a temporary git index. The user's index and `HEAD` stay byte-identical; the snapshot lives only at `refs/tidebreak/checkpoints/<workspace>/<ordinal>`.
- Persist `checkpoint_ref` + `diffstat` on the turn row and journal `CheckpointRecorded`. A failed snapshot journals a `HarnessNotice` and leaves the turn completed.
- Serve bounded review from `GET /code/workspaces/{id}/files` and `GET /code/workspaces/{id}/diff` (byte + file-count caps, truncation marked). Turn 1 diffs from the workspace base; later turns are `checkpoint(n-1)..checkpoint(n)`. Workspace review is `merge-base(base, HEAD)` against the live tree, including uncommitted state.
- Archive deletes that workspace's checkpoint refs; boot and archive also sweep refs whose workspace is gone.
- Register `files` and `diff` panel variants. The transcript turn card shows diffstat, duration, and a link that opens the Diff panel for that turn. Files/Diff panels render the server unified-diff string with per-file grouping and status-token +/- tinting. The async narrative slot is left empty.

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo clippy -p tidebreak-server --locked --all-targets -- -D warnings`
- [x] `cargo test -p tidebreak-server --locked`
- [x] `pnpm test` and `pnpm build` in `crates/tidebreak-desktop/ui`
- [x] Wire types regenerated (`UPDATE_WIRE_TYPES=1`)

Temp-repo coverage includes untracked/rename/mode capture, index+HEAD identity, turn-diff correctness, truncation marking, archive/sweep ref cleanup, and checkpoint-failure-does-not-fail-turn.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/ChatRoute.tsx
+++ b/crates/tidebreak-desktop/ui/src/ChatRoute.tsx
@@ -854,6 +854,9 @@ export function ChatRoute({ chatId }: { chatId: string }) {
         );
       case "agent":
         return <BackgroundAgentPanel chatId={chatId} runId={panel.runId} />;
+      case "files":
+      case "diff":
+        return null;
     }
   }
 
--- a/crates/tidebreak-desktop/ui/src/api/client.ts
+++ b/crates/tidebreak-desktop/ui/src/api/client.ts
@@ -81,6 +81,8 @@ import {
   type CodeRepoSnapshot,
   type CodeSessionSnapshot,
   type CodeTurnSnapshot,
+  type CodeWorkspaceDiff,
+  type CodeWorkspaceFiles,
   type CodeWorkspaceSnapshot,
   type HarnessDoctorReport,
   type HarnessKind,
@@ -107,6 +109,8 @@ import {
   parseCodeTurn,
   parseCodeTurnList,
   parseCodeWorkspace,
+  parseCodeWorkspaceDiff,
+  parseCodeWorkspaceFiles,
   parseHarnessDoctorReport,
   parseSequencedCodeEvent,
 } from "../code/parsers";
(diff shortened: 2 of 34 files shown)


naingthet (commented):

The snapshot path is the right shape: temp `GIT_INDEX_FILE`, `read-tree HEAD` + `add -A` + `write-tree` + `commit-tree`, then `update-ref` only under `refs/tidebreak/checkpoints/<workspace>/<ordinal>`. User `HEAD` and index stay untouched (the fingerprint test actually compares both). Turn 1 diffs `merge-base(base, HEAD)..checkpoint`; later turns are `checkpoint(n-1)..n`. Review is bounded (256 KiB / 64 files) with truncation marked; `git diff -- <file>` keeps the pathspec after `--`. Archive deletes that workspace’s refs and boot/archive sweep orphans. A failed snapshot journals a warning and leaves the turn completed.

UI: `files`/`diff` panel variants, turn card opens Diff for that turn, Files/Diff render the server unified-diff string. Wire parsers and the two HTTP tests cover the review surface. Not driven in the live desktop app.

naingthet (commented):
## Review

Checkpoints snapshot tracked and untracked work through a temporary git index so the user's index and HEAD stay byte-identical; refs live at `refs/tidebreak/checkpoints/<workspace>/<ordinal>`. A failed snapshot journals a notice and still completes the turn. Review routes are capped and truncated; turn 1 diffs from the workspace base and later turns from the previous checkpoint. Archive and sweep delete those refs. Tests pin identity, untracked/rename/mode, turn ranges, truncation, and cleanup.

PR #2133, opened 2026-08-15
feat(code): approvals bridge with deny-steering and Ask/Auto modes

naingthet:
## What landed

The code-mode approval channel is real on Claude Code 2.1.233 and is now wired end to end.

- **Harness.** Hidden `--permission-prompt-tool` plus HTTP MCP is composed from `ApprovalChannelSpec`. A captured `tools/call` becomes `HarnessEvent::ApprovalRequested`. `decide` completes the parked MCP call. `structured_approvals` is `Supported` for 2.1.233 only.
- **Server.** Session-scoped loopback MCP at `POST /code/mcp/approval-prompt`. Pending rows on request; `POST /code/approvals/{id}/decision` completes the harness channel, then records Approved/Denied. Ask/Auto session creation is unlocked when the adapter can honor them; a mode it cannot honor is refused with the reason. Undecided approvals survive restart and stay decidable.
- **UI.** `CodeApprovalCard` renders the harness payload (not a paraphrase). Deny opens a feedback field. Pending cards appear in the transcript; the workspace header shows a count badge. The composer enables Ask (new default) and Auto.

Bypass denylist is unchanged.

## What was captured

Claude Code **2.1.233**. `--permission-prompt-tool` is not in `--help`; an unknown flag errors and this one does not.

Live print-mode captures (throwaway repo, `--permission-mode manual`, HTTP MCP with Bearer token):

| Fixture | What it is |
|---|---|
| `approval-request.mcp.json` | CLI → MCP `tools/call` `{tool_name, input, tool_use_id}` |
| `approval-request.ndjson` | Stream parked at the Write tool-use |
| `approval-allow.ndjson` | After `{"behavior":"allow"}` — file written |
| `approval-deny-with-feedback.ndjson` | After `{"behavior":"deny","message":"no — use the fixtures directory instead"}` — that text is the tool_result the model reads |

Response shape required by the CLI: a **single** MCP text block. `structuredContent` is rejected.

## Per-mode flag mapping (Claude Code 2.1.233)

| Tidebreak mode | Engine flags | Notes |
|---|---|---|
| Plan | `--permission-mode plan` | Mutations refused |
| Ask (default) | `--permission-mode manual` plus `--mcp-config` + `--permission-prompt-tool mcp__tb-approvals__permission_prompt` | Every tool parks on the prompt tool |
| Auto | `--permission-mode acceptEdits` plus the same prompt-tool wiring | Workspace writes proceed; sensitive actions still park |

`--permission-mode auto` is the engine's classifier, not Tidebreak Auto. `bypassPermissions` remains denylisted.

A harness with `structured_approvals` other than `Supported` cannot create Ask or Auto sessions; the refusal states that.

## What could not be verified

- Live Auto (`acceptEdits`) with the prompt tool attached. The flag is documented and already composed; this change did not recapture an Auto turn.
- Mid-turn steering, native file-change events, reasoning-level flags — still uncaptured.
- Driving the desktop app through a real approval card. Scripted-harness tests prove the engine observes the decision (not just a DB write). Live CLI allow/deny was verified locally against 2.1.233.
- Newer Claude versions stay `Unknown` until recaptured.


naingthet (commented):
Two defects on this head. The approval channel wiring (Ask/Auto flag mapping, session-scoped MCP token, `decide` on `&self` so it can complete while `run_turn` is blocked, deny feedback as the single MCP text block, scripted allow/deny + restart tests) is otherwise in the right shape.

**`crates/tidebreak-server/src/code/session_worker.rs:561` / `runtime.rs:763` — truncated Write/Edit approvals cannot be decided, and `cap_raw` can panic.** `record_approval` stores `cap_raw(raw)` and then ignores `harness_ref` (`session_worker.rs:90`, `:114`). `decide_approval` later reads `approval.harness_raw["tool_use_id"]`. Once the payload exceeds 16 KiB — a `Write`/`Edit` `input` with file contents will — `cap_raw` replaces the object with `{truncated, preview}`, the id is gone, and the decision route returns `approval row is missing a harness tool_use_id`. Persist `harness_ref.call_id` (or keep `tool_use_id` as a sibling of the preview) so the complete path does not depend on the uncapped JSON. Same function: `&rendered[..MAX_HARNESS_RAW_BYTES]` panics if that byte index is not a UTF-8 char boundary; slice with `floor_char_boundary` or equivalent.

**`crates/tidebreak-desktop/ui/src/code/CodeWorkspacePage.tsx:107` — starting a session always posts `permission_mode: "ask"`.** Claude Code advertises `structured_approvals: Supported` only for `2.1.233`. Any other version is refused with `permission_mode_unavailable`, and this page has no Plan fallback or mode picker on create. Previously the start button created a Plan session that every captured version can honor. If Ask is the default, the create path still needs to land in Plan when the harness cannot honor Ask.

naingthet:
Triage — both findings accepted, folding into the in-progress rebase onto `main`:

1. **Truncation-vs-decidability (P1)** — `harness_ref.call_id` becomes a first-class persisted field; the decision path stops reading the capped JSON entirely. Char-boundary-safe truncation for the cap. Tests: oversized Write approval decidable end-to-end; multi-byte boundary no-panic.
2. **Create-path mode fallback** — session create gains a mode selection defaulting per the doctor's capability report (Ask where supported, Plan otherwise), so every harness version has an honorable create path.

naingthet (commented):
The multiplex change is the right shape: `drive_turn` keeps selecting on `WorkerCommand` while `run_turn` is pinned, `decide`/`interrupt` take `&self`, and the decision route now goes through the worker instead of a side-channel completer. That was the missing piece for a mid-turn approval to reach the engine.

The two defects from the previous head are still here.

**`crates/tidebreak-server/src/code/session_worker.rs:633` / `runtime.rs:762` — truncated Write/Edit approvals still cannot be decided, and `cap_raw` can still panic.** `record_approval` stores `cap_raw(raw)` and still drops `harness_ref` (`session_worker.rs:98`, `:122`). `decide_approval` still reads `approval.harness_raw["tool_use_id"]`. A `Write`/`Edit` `input` over 16 KiB replaces the object with `{truncated, preview}`, the id is gone, and the route returns `approval row is missing a harness tool_use_id`. Persist `harness_ref.call_id` (or keep `tool_use_id` next to the preview). Same function: `&rendered[..MAX_HARNESS_RAW_BYTES]` panics if that byte is not a UTF-8 char boundary.

**`crates/tidebreak-desktop/ui/src/code/CodeWorkspacePage.tsx:106` — starting a session still always posts `permission_mode: "ask"`.** Claude Code advertises `structured_approvals: Supported` only for `2.1.233`. Any other version is refused, and this page still has no Plan fallback on create. `availableModes` is also hardcoded to all three modes after create (`:333`).

naingthet (commented):
The Codex `&self` change is the right follow-up: process state sits behind interior mutability, `decide`/`interrupt` can run while `run_turn` is pinned, and the scripted completer now stages a decision that arrives before `park()`. That unblocks the multiplex path for every adapter.

The two defects from the previous heads are still here. This SHA does not include the accepted persist-`call_id` / create-path fallback work.

**`crates/tidebreak-server/src/code/session_worker.rs:98` / `:122` / `:639` and `runtime.rs:847` — truncated (and Codex) approvals still cannot be decided, and `cap_raw` can still panic.** `record_approval` still stores `cap_raw(raw)` and still drops `harness_ref`. `decide_approval` still reads `approval.harness_raw["tool_use_id"]`. A `Write`/`Edit` `input` over 16 KiB replaces the object with `{truncated, preview}`, the id is gone, and the route returns `approval row is missing a harness tool_use_id`. This commit also makes Codex emit `raw: Null` (`codex/parse.rs`) while advertising `structured_approvals: Supported`, so a Codex Ask approval never has a `tool_use_id` either. Persist `harness_ref.call_id` and stop reading the capped JSON for the decide handle. Same function: `&rendered[..MAX_HARNESS_RAW_BYTES]` panics if that byte is not a UTF-8 char boundary.

**`crates/tidebreak-desktop/ui/src/code/CodeWorkspacePage.tsx:113` — starting a session still always posts `permission_mode: "ask"`.** Claude Code advertises `structured_approvals: Supported` only for `2.1.233`. Any other version is refused, and this page still has no Plan fallback on create. `availableModes` is also hardcoded to all three modes after create (`:410`).

naingthet (commented):

Decide now reads `harness_raw.call_id` written as a sibling of the 16 KiB preview, so an oversized Write/Edit (and Codex/opencode `raw: Null`) stays decidable. `cap_raw` slices on `floor_char_boundary`. Create posts Ask only when the doctor reports `structured_approvals: supported`, otherwise Plan — both the workspace start prompt and the new-workspace dialog honor that, including a mixed-harness click that falls back per engine. The opencode `&self` conversion matches the Codex multiplex path.

PR #2134, opened 2026-08-15
feat(code): auxiliary workspace terminals

naingthet:
## Summary

Workspace shells for verifying agent work next to review, per decision 0036. They are a user-facing convenience, not a harness interface.

- Spawn the user's `$SHELL` in a PTY with cwd set to the workspace worktree. `portable-pty` is pinned on `tidebreak-server` only; the harness crate still has the structural guard that it never depends on a PTY.
- Bytes live in a bounded in-memory ring. They are not journaled and not persisted. A server restart reaps every terminal; reconnect reports the shell ended. Ring overflow inserts an inline truncation marker.
- Caps: terminals per workspace, ring size, read response size, write size.
- Cursor-pull: `POST/GET/DELETE /code/workspaces/{id}/terminals`, `GET .../read?cursor=`, `POST .../write` and `/resize`. Activity notices are coalesced (~32ms) on an in-process bus and are never written to the journal.
- UI: `@xterm/xterm` + `@xterm/addon-fit`, a `terminal` panel variant, ephemeral renderer (create on open, dispose on close), recent-ring fetch then poll, chunked writes under a frame budget, stall reconnect, keyboard passthrough except app shortcuts.

## Test plan

- [x] `cargo fmt --all --check`
- [x] `cargo clippy -p tidebreak-server --locked --all-targets -- -D warnings`
- [x] `cargo test -p tidebreak-server --locked --lib code::terminal`
- [x] `cargo test -p tidebreak-server --locked --lib code_terminals`
- [x] `pnpm test` and `pnpm build` in `crates/tidebreak-desktop/ui`
- [x] Wire types regenerated

Coverage: ring wraparound, cursor-at-overflow, truncation marker, read caps, restart reaps with no durable bytes in the store, coalesced notices while a reader still gets every retained byte, two readers at different cursors, pane dispose on unmount, truncation notice, ended-shell state.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -721,7 +721,7 @@ checksum = "a88b7ea17d208c4193f2c1e6de3c35fe71f98c96982d5ced308bdcc749ff6e1f"
 dependencies = [
  "borsh-derive",
  "bytes",
- "cfg_aliases",
+ "cfg_aliases 0.2.1",
 ]
 
 [[package]]
@@ -1059,6 +1059,12 @@ name = "cfg-if"
 version = "1.0.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "9330f8b2ff13f34540b44e946ef35111825727b38d33286ef986142615121801"
+
+[[package]]
+name = "cfg_aliases"
+version = "0.1.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fd16c4719339c4530435d38e511904438d07cce7950afa3718a84ac36c10e89e"
 
 [[package]]
 name = "cfg_aliases"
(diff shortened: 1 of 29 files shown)


naingthet (commented):
The shape matches 0036: PTY stays in `tidebreak-server`, the ring is in-memory only, restart reports ended with no durable bytes, and the UI is an ephemeral xterm view over cursor-pull. Archive closes the workspace's shells.

**`crates/tidebreak-server/src/code/terminal.rs:355` / `:506` — an exited shell still consumes a workspace slot.** `start_reader` / `start_reaper` set `ended = true` and leave the id in `by_workspace`. `reserve_slot` counts every id, not live ones. `close` is the only path that `remove`s. The pane (`TerminalPane.tsx` attach) never DELETEs: after `exit` it shows "Shell ended", and Reconnect/`attach` creates a new PTY beside the dead row. Eight of those cycles and `POST .../terminals` returns `terminal_cap`. Drop ended rows from the hub (or count only `!ended` toward the cap) so a finished shell does not block the next one.

naingthet (commented):
## Review

Matches 0036: `$SHELL` in a PTY with the worktree as cwd, harness crate still has no PTY dep (`portable-pty = 0.9.0` only on tidebreak-server). Bytes stay in a bounded in-memory ring — not journaled, gone on restart. Caps on terminals/workspace, ring, read, and write. Cursor-pull plus coalesced notices. Tests cover wraparound, overflow marker, restart-reaps, multi-cursor readers, and UI dispose.

naingthet (commented):
The notices refresh for `@xterm/xterm` / `@xterm/addon-fit` matches the new UI deps.

The cap leak from the previous head is still here. `start_reader` / `start_reaper` set `ended = true` and leave the id in `by_workspace` (`terminal.rs:506`, `:527`). `reserve_slot` (`:355`) still counts every id. The pane never DELETEs, so eight `exit`+reconnect cycles still return `terminal_cap`.

PR #2140, opened 2026-08-17
fix(desktop): replace native workspace selects with the shared Select

naingthet:
## Summary

The new-workspace **Repo** and **Harness** fields were the last native `<select>` elements in the desktop UI. Opening them drew the macOS system menu over the dialog instead of the in-app picker used everywhere else.

They now use the existing shadcn/Radix `Select` — the same control as settings model/provider pickers, web search, voice, and document zoom.

The shared trigger and menu also line up with nearby inputs: `border-border`, pointer cursor, menu shadow, and a z-index above dialogs so the list is usable inside this form.

## Test plan

- [x] `pnpm --dir crates/tidebreak-desktop/ui exec tsc --noEmit`
- [x] `pnpm --dir crates/tidebreak-desktop/ui test` (1076 passed)
- [ ] Open **New workspace** and confirm Repo / Harness open an in-app menu (no OS popup)
- [ ] Pick a different repo and harness; Create still uses those values
- [ ] Settings model/provider selects still open and pick as before

Could not drive the running desktop app in this session.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/NewWorkspaceDialog.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/NewWorkspaceDialog.tsx
@@ -18,6 +18,13 @@ import {
   DialogTitle,
 } from "@/components/ui/dialog";
 import { Input } from "@/components/ui/input";
+import {
+  Select,
+  SelectContent,
+  SelectItem,
+  SelectTrigger,
+  SelectValue,
+} from "@/components/ui/select";
 import { friendlyErrorMessage } from "@/lib/utils";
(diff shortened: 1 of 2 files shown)


naingthet (commented):
## Review

Repo and Harness now use the shared Radix Select instead of native `<select>`, so the macOS system menu no longer covers the dialog. The trigger/menu tokens match nearby inputs (`border-border`, pointer cursor, `z-[60]` above the dialog overlay). Empty repo/harness lists disable the control and show a placeholder rather than a broken native control.

PR #2141, opened 2026-08-17
fix(desktop): show calculated values for uncached spreadsheet formulas

naingthet:
## Summary

Formula cells in the native workbook viewer showed a live formula in the formula bar and a blank grid. Agent-built workbooks typically write the formula and no Excel cached result, and two display-copy mistakes then hid the value:

- An empty `<v>` was treated as an authored cache, so the formula was stripped and the cell stayed empty.
- The canvas worker skips calculation above 1,000 formulas, or if any formula in the book points at a missing sheet. Remaining uncached cells never painted.

The display copy now evaluates those uncached formulas with the bundled Duke engine, writes the results as cached values, and still keeps the original formula for the formula bar. Source bytes are unchanged.

## Why

A generated FCF model (cross-sheet `INDEX` / `ROUNDUP` / `ROW`) is a normal Tidebreak deliverable. The viewer already prefers Excel's cached result so unsupported functions do not become `#VALUE!`. That path assumed a real cache. Without one, review showed empty revenue and FCF columns.

## Tests

- Empty `<v>` is no longer treated as a cached result; the formula stays in the display copy.
- A calculated number is baked into an uncached formula cell and the formula is retained for inspection.

Local: `pnpm exec vitest run src/document/readOnlyWorkbookProjection.dom.test.ts src/document/NativeSpreadsheetViewer.dom.test.tsx src/document/SpreadsheetViewer.dom.test.tsx`, `pnpm exec tsc --noEmit`.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/document/readOnlyWorkbookFormulaEngine.ts
+++ b/crates/tidebreak-desktop/ui/src/document/readOnlyWorkbookFormulaEngine.ts
@@ -1,0 +1,108 @@
+import workbookWasmUrl from "@dukelib/sheets-wasm/duke_sheets_wasm_bg.wasm?url";
+
+export type EvaluatedFormulaValue =
+  | { type: "boolean"; value: boolean }
+  | { type: "error"; value: string }
+  | { type: "number"; value: number }
+  | { type: "text"; value: string };
+
+export type EvaluatedFormulaMap = Record<string, EvaluatedFormulaValue>;
+
+type DukeModule = typeof import("@dukelib/sheets-wasm");
+
+let dukeModule: Promise<DukeModule> | null = null;
+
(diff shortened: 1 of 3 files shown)


naingthet (commented):
## Review

Empty `<v>` is no longer treated as an authored cache, so agent-built formula cells keep the formula instead of painting blank. The display copy bakes Duke results into those uncached cells and still records the original formula for the formula bar; source bytes stay untouched. The two projection tests pin both the empty-cache case and a baked numeric result.

naingthet (commented):
## Review

The follow-up runs `projectConditionalStyles` after the bake and treats empty `<v>` as no cache, so data bars and color scales match the grid instead of reading 0. Formulas still stay on the formula bar; source bytes are unchanged. The new projection test pins a baked number feeding a data bar.

PR #2143, opened 2026-08-17
feat(code): source palette for adding repos with clone support

naingthet:
## What

Code mode no longer asks for a bare path when adding a repo. The home empty state and the sidebar add affordance open a keyboard-first **source palette**:

- **Local folder** — native directory picker (or a pasted path) then the existing registration route.
- **Git URL** — clone into a remembered parent directory, with an optional folder name.
- **GitHub repository** — `owner/repo`. When `gh` is signed in, the clone URL is whatever `gh repo view --json url` reports. When `gh` is absent or signed out, the same form still clones over HTTPS and shows the existing graceful-absence hint inline.

Unserved providers are omitted, not teased. Stages stay in one dialog (list → form → progress); Backspace goes back, Escape closes.

The new-workspace and start-session harness pickers use the same visual language: logo, product name, one-line subtitle. Ready rows are selectable. Unusable rows (not installed, signed out, or — like Grok today — refusing every permission mode) are dimmed with one short reason and a quiet link to Coding harnesses. Versions, paths, and capability flags stay on the doctor.

## Clone-job contract

- `POST /code/repos/clone` `{url | github, parent_dir, name?}` validates that `parent_dir` exists and the target does not, then returns a job id immediately (`202`).
- Clone is `git clone --progress` as an argv array with `GIT_TERMINAL_PROMPT=0` and a finite timeout. The user's own credential helpers may authenticate; Tidebreak never prompts and never stores secrets.
- Progress publishes as unsequenced `clone_progress` notices on `/code/updates` (`job`, `phase`, `percent`, `done`, bounded `error`, `repo_id` on success). `GET /code/repos/clone/{job}` is the poller.
- Success auto-registers through the existing repo path. `GET /code/repos/clone-defaults` returns the remembered parent directory plus observed `gh` status.

## Native command

`pick_code_directory` is a grant-free OS folder picker for registration and clone destinations (ADR 0030). The path is returned as a string; the server validates it. Allowlisted in `native-only-commands.txt`.

## Tests

- Server: local bare-origin clone (no network), progress notices, typed failures (missing/unusable parent, existing target, bad URL), `gh` shim resolve signed-in and absent, auto-register on success.
- UI: palette keyboard (arrows / Enter / Backspace / Esc), filter, gh-absent hint, progress-to-error retry; harness picker ready vs dimmed, keyboard selection, no version strings from the doctor payload.

## What could not be verified

[...]

Issue #2132, opened 2026-08-15
Flaky: sandbox_container_run timing-sensitive tests fail intermittently in the test lane


naingthet:
Two different tests in `sandbox_container_run::tests` have failed the workspace `test` lane on unrelated PRs today, both passing on rerun with no relevant code changes:

- `terminalizes_and_tears_down_when_the_agent_loop_ends_without_a_result` — failed on #2123 (a change touching only `tidebreak-harness` files and a fixture), green on rerun.
- `committed_cancellation_wakes_container_and_fences_reverse_egress_before_heartbeat` — failed on #2131 (checkpoint/review code, no sandbox-container changes), rerun requested.

Both look timing-sensitive (teardown/cancellation windows). Cost: every unrelated PR that trips one burns a rerun cycle and erodes trust in the lane.

Worth either tightening the tests' synchronization (wait on explicit state transitions rather than timing windows) or quarantining them behind an env gate until fixed.

naingthet:
Third occurrence, different test: `probe::tests::relative_path_is_rejected` in `tidebreak-harness` failed the `test` lane on #2143 (a PR that does not touch probe), panicking at `probe.rs:381` — *after* the hermetic shim fix that landed via #2130. The fix reduced but did not eliminate the environment dependence. Escalating from rerun-and-move-on to a real fix: the test needs to control every input (shell binary, PATH, env) rather than partially overriding the host's.

naingthet:
Third distinct test from the same module: `sandbox_container_run::tests::a_noop_heartbeat_does_not_revoke_a_live_container_drive` failed on #2145 (a one-line classifier change), panicking at `tests.rs:3479` — after `terminalizes_and_tears_down…` (#2123) and `committed_cancellation_wakes_container…` (#2131). Three different tests in one module is not a flaky test; it is a module whose synchronization is timing-window-based. The probe-test half of this issue was root-caused and fixed in #2144 (write-then-exec race); this module deserves the same treatment: replace elapsed-time windows with waits on explicit state transitions.

PR #2130, opened 2026-08-15
feat(harness): Codex CLI adapter on captured fixtures

naingthet:
## Summary

Adds the second harness adapter, for Codex CLI 0.147.0, written only against captured app-server streams.

**Integration path:** long-lived `codex app-server --stdio` JSON-RPC. Chosen over `codex exec --json` because the installed 0.147.0 app-server handshake is stable (`initialize` → `thread/start` / `thread/resume` → `turn/start`) and is the richer approval channel (`item/commandExecution/requestApproval` with `{decision: accept|decline}`). `codex exec --json` and `codex exec resume <id>` exist in `--help` and were not used for these fixtures.

Capture used an isolated `$CODEX_HOME` (auth.json only), model `gpt-5.6-luna` at `effort=low` (cheapest ChatGPT-account model that accepted the request; `gpt-5-mini` is rejected on this account), tiny prompts, and a throwaway git repo.

## Permission-mode mapping (0033)

| Tidebreak mode | `thread/start` sandbox | `approvalPolicy` |
|---|---|---|
| `Plan` | `read-only` | `untrusted` |
| `Ask` | `workspace-write` | `untrusted` |
| `Auto` | `workspace-write` | `on-request` |

Never composed: `--dangerously-bypass-approvals-and-sandbox`. Prompt travels in `turn/start` params, never on argv.

`decide()` completes `item/commandExecution/requestApproval` with `accept` or `decline`. This protocol does not carry a denial-feedback string; feedback is dropped on the wire.

## Capabilities (0.147.0, evidence in `fixtures/codex/0.147.0/`)

| Flag | Level | Evidence |
|---|---|---|
| resume | Supported | `thread/resume` + second turn (`resume.ndjson`) |
| streaming_deltas | Supported | `item/agentMessage/delta` (`plain-text.ndjson`) |
| structured_approvals | Supported | request + accept + decline fixtures |
| mid_turn_steering | Unknown | `turn/steer` is in the generated schema; not captured |
| plan_mode | Supported | `sandbox: read-only` accepted on `thread/start` |
| reasoning_levels | Supported | `effort: low` accepted; thread reports `reasoningEffort` |
| native_file_change_events | Unknown | `item/fileChange/*` / `turn/diff/updated` not seen |
| native_interrupt | Supported | `turn/interrupt` → `turn/completed` status `interrupted` |

Auth observation: `codex login status` → `Logged in…` / `Not logged in`.

## What could not be verified

[...]

PR #2144, opened 2026-08-17
test(harness): close the remaining environment leak in the probe relative-path test

naingthet:
## Summary

`probe::tests::relative_path_is_rejected` failed the workspace `test` lane on unrelated PRs after the #2130 hermetic shim. The panic at `probe.rs:381` was not a sentinel-parse leak. CI log from #2143:

```
expected RelativePath, got Shell("Text file busy (os error 26)")
```

Two remaining host inputs:

1. **Write-then-exec race.** The test wrote the fake shell and immediately `execve`'d it. On Linux that is `ETXTBSY` when a writer still holds the inode. `write_exec` now creates a sibling file with mode `0755`, `fsync`s, then `rename`s over the dest so the executed inode never had an open writer.
2. **Host `/bin/sh` dialect.** The previous shim rewrote `command -v claude || true` with `${var%%…}` / `${var#*… || true}` and `eval`'d the rest, which then called `env` on an empty `PATH`. Ubuntu's `/bin/sh` is dash; macOS is bash-as-sh. The new shim is the `SHELL` under test (`#!/bin/sh`), matches `-ilc`/`-c` by exact flags, peels `TIDEBREAK_PROBE_*` tokens with hex-only POSIX expansions, and prints a relative `claude`. It never evals the probe script and never calls host `command` or `env`.

`clear_env` is still on. The child env is only `SHELL`, `PATH` (empty dir), `HOME`, `TMPDIR`, `LANG=C`, and `LC_ALL=C`. Failure now dumps shim stdout/stderr (or the spawn error) instead of a bare `panic`.

Product `probe_shell` / `resolve_binary` behavior is unchanged. This was a test hermeticity bug, not a product bug.

Refs #2132

## Test plan

- [x] `cargo fmt --all`
- [x] `cargo clippy -p tidebreak-harness --all-targets --locked --no-deps -- -D warnings` (bare `-p` without `--no-deps` lints `tidebreak-core` as a dep and hits pre-existing unused-code warnings on this tree)
- [x] `cargo test -p tidebreak-harness --locked` (69 passed)
- [x] `relative_path_is_rejected` 200/200 on the built test binary (macOS)
- [x] Fake-shell script 200/200 under `/bin/dash` with empty `PATH` and `env -i`
- [x] Same script 50/50 under `/bin/sh` and `/bin/bash`

### DIFF
--- a/crates/tidebreak-harness/src/probe.rs
+++ b/crates/tidebreak-harness/src/probe.rs
@@ -275,13 +275,28 @@ async fn run_interactive_login_shell(
 #[cfg(test)]
 mod tests {
     use super::*;
-    use std::os::unix::fs::PermissionsExt;
 
     fn write_exec(path: &Path, body: &str) {
-        std::fs::write(path, body).unwrap();
-        let mut perms = std::fs::metadata(path).unwrap().permissions();
-        perms.set_mode(0o755);
-        std::fs::set_permissions(path, perms).unwrap();
+        // Write a sibling inode, fsync, then rename over `path` so execve
+        // never sees a file that still has a writer (Linux ETXTBSY).
+        use std::io::Write;
+        use std::os::unix::fs::OpenOptionsExt;
+        let staging = path.with_extension("writing");
+        let mut file = std::fs::OpenOptions::new()
+            .write(true)
+            .create(true)
+            .truncate(true)
+            .mode(0o755)
+            .open(&staging)
+            .unwrap();
+        file.write_all(body.as_bytes()).unwrap();
+        file.sync_all().unwrap();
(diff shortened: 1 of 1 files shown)


naingthet (commented):
## Review

The two remaining host leaks match the #2143 `ETXTBSY` / dash-vs-bash failure. `write_exec` now creates a mode-0755 sibling, `fsync`s, and `rename`s so Linux never execs an inode with an open writer. The fake shell matches product `-ilc` exactly, peels hex sentinels with POSIX expansions, and never evals `command`/`env`. Product `probe_shell` / `resolve_binary` are untouched.

PR #2142, opened 2026-08-17
feat(cli): tidebreak code subcommands for the code-mode surface

naingthet:
## Summary

The server already ships the full code-mode surface (repos, workspaces, sessions, approvals, checkpoints/diffs, git flow, updates, doctor). This adds the headless client those routes were missing, per ADR 0007: an agent or script can drive a complete code-mode work loop with `tidebreak code …` and `--json`.

Every command embeds a server by default and attaches with `--server` / `--attach` the same way `-p` and the setup family do. `--json` (or `--output-format json`) writes one object, or NDJSON for the two streaming commands. Failures print the server's typed `{kind, message}` and exit nonzero.

## Command table

| Command | Route |
| --- | --- |
| `code doctor [--refresh]` | `GET /code/harnesses` / `POST /code/harnesses/refresh` |
| `code repo add <path> [--name] [--base-ref] [--branch-prefix]` | `POST /code/repos` |
| `code repo list` | `GET /code/repos` |
| `code repo rm <id>` | `DELETE /code/repos/{id}` |
| `code ws new --repo <id\|path> [--title] [--base-ref]` | `POST /code/workspaces` |
| `code ws list [--repo]` | `GET /code/workspaces` |
| `code ws show <id>` | `GET /code/workspaces/{id}` + `GET …/sessions` |
| `code ws archive <id> [--force]` | `POST /code/workspaces/{id}/archive` |
| `code session start --ws <id> --harness <kind> [--mode plan\|ask\|auto]` | `POST /code/workspaces/{id}/sessions` (mode defaults to `ask`) |
| `code session show <id>` | workspace walk + `GET /code/sessions/{id}/turns` |
| `code session reap <id>` | `POST /code/sessions/{id}/reap` |
| `code run (--session <id> \| --ws <id>) <message> [--on-approval wait\|fail]` | `POST /code/sessions/{id}/turns` + `WS …/events` |
| `code approvals [--session <id>]` | `GET /code/approvals?state=pending` |
| `code approve <id>` / `code deny <id> [-m <feedback>]` | `POST /code/approvals/{id}/decision` |
| `code interrupt --session <id>` | `POST /code/sessions/{id}/interrupt` |
| `code turns --session <id>` | `GET /code/sessions/{id}/turns` |
| `code diff --ws <id> [--turn N] [--file PATH]` | `GET /code/workspaces/{id}/diff` |
| `code files --ws <id> [--turn N]` | `GET /code/workspaces/{id}/files` |
[...]

PR #2145, opened 2026-08-17
fix(code): classify HTTP 401 push rejections as auth failures

naingthet:
## Summary

A leftover from #2142: tightening `is_git_auth_failure` dropped `401` when the overly broad "could not read from remote" needle came out. Git's HTTP transport reports credential rejection as `The requested URL returned error: 401` with none of the remaining auth signatures, so a real 401 classified as `git_push_failed` and skipped the credential remediation.

This adds `401` next to the existing `403`. The unresolvable-remote fixture (`does not appear to be a git repository` / `Could not read from remote repository`) still does not match.

## Test

`classify_git_treats_only_auth_signatures_as_auth` now includes the 401 HTTP string and still asserts the unresolvable-remote stderr is `PushFailed`.

### DIFF
--- a/crates/tidebreak-server/src/code/gh.rs
+++ b/crates/tidebreak-server/src/code/gh.rs
@@ -850,6 +850,7 @@ fn is_git_auth_failure(err: &str) -> bool {
     lower.contains("authentication")
         || lower.contains("permission denied")
         || lower.contains("403")
+        || lower.contains("401")
         || lower.contains("terminal prompts disabled")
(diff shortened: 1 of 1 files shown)


naingthet (commented):
## Review

`401` is back next to `403` in `is_git_auth_failure`, which is the HTTP credential-rejection string Git prints without any of the remaining auth needles. The unresolvable-remote fixture is unchanged, so a missing URL still stays `git_push_failed`.

PR #2147, opened 2026-08-17
fix(desktop): keep chat header clear of macOS traffic lights

naingthet:
## Summary

The compact sidebar is narrower than the macOS overlay traffic-light cluster, so the chat breadcrumb (`Chats / …`) sat under the lights.

When the rail is collapsed on the overlay titlebar, the chat chrome row now starts at the same 84px line the expanded history buttons already use. The drag strip widens to cover the overflowing lights so the window still moves from that corner.

## Validation

- CSS inset: compact rail (56px) vs traffic-light reserve (84px) leaves 28px of header padding, matching the expanded history-button start
- Did not drive the native macOS window in this session

### DIFF
--- a/crates/tidebreak-desktop/ui/src/ChatRoute.tsx
+++ b/crates/tidebreak-desktop/ui/src/ChatRoute.tsx
@@ -884,7 +884,7 @@ export function ChatRoute({ chatId }: { chatId: string }) {
   return (
     <RouteFrame sidebar={<AppSidebar chat={chat} />}>
       <div className="relative mr-2 flex min-h-0 min-w-0 flex-1 flex-col">
-        <header className="mt-2 flex h-9 w-full shrink-0 items-center justify-between gap-2 pl-4 pr-1">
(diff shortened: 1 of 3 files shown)


naingthet (commented):
## Review

The compact rail (56px) is narrower than the macOS overlay cluster (84px), so the chat breadcrumb sat under the lights. `window-chrome-row` only adds the overflow inset when the rail is compact *and* the overlay titlebar is present — the same 84px line the expanded history buttons already use. The drag strip widens to cover the overflowing lights, and the compact width / traffic-light inset are shared tokens instead of three copies of `calc(var(--spacing) * 14)` / `84px`.

naingthet (commented):
## Review

`6cc1b263` is an empty CI retrigger after the GitHub API 503s; the product change is still the compact-rail traffic-light inset. `window-chrome-row` only adds the overflow padding when the rail is compact and the overlay titlebar is present, and the drag strip still covers the overflowing lights.

PR #2148, opened 2026-08-17
test(server): replace timing windows with state waits in container-run tests

naingthet:
## Why

`sandbox_container_run::tests` failed three distinct tests on unrelated PRs (#2123, #2131, #2145). A later soak then **wedged at 0% CPU** under `--test-threads 4`. That is not flake: the module deadlocks under scheduler interleaving.

## Deadlock evidence

Attempted `sample 3387 5 -file /tmp/container-hang-sample.txt` on the recorded hung child. The process was already gone (`sample cannot examine process 3387 … no longer appears to be running`), so there is no macOS backtrace file.

What the hang itself showed, and what the code still had:

- Child `…/deps/tidebreak_server-8b3d1956e7476364 sandbox_container_run --test-threads 4` sat at **0.0% CPU for minutes**. That is a parked lock/futex, not a slow assertion.
- `UsageThenPendingStream::drop` called `std::sync::Barrier::wait()` on the dropping thread. That thread is a tokio worker. The durable op-log uses `block_in_place`. A worker stuck in `Barrier::wait` starves every peer: 0% CPU, nobody runs `spawn_blocking` to release the barrier.
- `ScriptedProvider` published `started_flag` and only then waited on `release.notified()`. The test can observe the flag, fire `release`, and the wait misses it. The drive parks forever on the first completion.

Those two seats are the `--test-threads N` deadlock class. Timing windows were the flake class; this is the same module failing to wait on a durable predicate.

## Root cause per named failure

- `terminalizes_and_tears_down_when_the_agent_loop_ends_without_a_result`: eight real agent steps under `fast_config()`'s 250 ms durable fence contended for SQLite's writer. The 60 s hang-guard expired under the parallel suite even though the drive was still making progress.
- `committed_cancellation_wakes_container_and_fences_reverse_egress_before_heartbeat`: a 2 s join after cancel encoded "faster than the next heartbeat." Under load, teardown plus store writes exceeded that window even when the local cancel signal had already fired.
- `a_noop_heartbeat_does_not_revoke_a_live_container_drive`: a 10 ms heartbeat plus a 100 ms provider delay assumed ticks would land inside a 10 s hang-guard. The hang-guard was the assertion.

## What changed

Tests now wait on observable transitions: provider `StepGate` flags, reclaimable-run rows, lease expiry in the store, heartbeat tick counts, and cancellation receipts. Producers subscribe to the release notify **before** publishing the started/entered flag, so a faster test cannot lose the wakeup. Tight "must beat the next heartbeat" joins are replaced by a heartbeat/fence cadence longer than the hang-guard plus the process-local cancel signal. Heavy real-agent drives use a quiet cadence so they do not fight the suite for the SQLite writer.

`Drop` no longer parks on a barrier. The quiescence test still proves a late reverse request does not start another provider call.

Product seam: a `#[cfg(test)]` heartbeat tick counter on `SandboxContainerRunner`. Production still uses `tokio::time::interval`; tests wait on the count after a successful renew/validate instead of assuming elapsed time.

No tests deleted.

## Evidence

- `cargo fmt --all`
- `cargo test -p tidebreak-server --locked sandbox_container_run`: 41 passed, 3 ignored (Docker)
- soak with a **45 s per-iteration timeout** so a deadlock cannot wedge the loop: **0 wedges** in 240 runs (200 at `--test-threads 4`, 20 at 1, 20 at 8); 239 passed, 1 non-wedge failure overwritten by later iterations
- earlier unbounded soak: **200/200** at `--test-threads 4` before the deadlock seats were closed
- `cargo test -p tidebreak-server --locked`: 959 + integration tests passed (pre-deadlock-seat commit)

Refs #2132

### DIFF
--- a/crates/tidebreak-server/src/sandbox_container_run.rs
+++ b/crates/tidebreak-server/src/sandbox_container_run.rs
@@ -37,6 +37,8 @@
 use std::collections::BTreeMap;
 use std::future::Future;
 use std::pin::Pin;
+#[cfg(test)]
+use std::sync::atomic::AtomicUsize;
 use std::sync::atomic::{AtomicU32, AtomicU64, Ordering};
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
@@ -612,6 +614,11 @@ pub struct SandboxContainerRunner {
     /// and publishes the steering sink of every connection it holds. A driver
     /// assembled without the server's shared guard uses its own local guard.
     steering: Arc<SandboxSteerGuard>,
+    /// Attached-drive heartbeat ticks observed by tests. Not a clock: the
+    /// production loop still waits on `tokio::time::interval`. Tests wait on
+    /// this count instead of assuming a cadence elapsed.
+    #[cfg(test)]
+    heartbeat_ticks: AtomicUsize,
 }
 
 impl SandboxContainerRunner {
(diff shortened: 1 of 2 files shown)


naingthet (commented):
## Review

The three flakes were the same class of bug: tests encoded wall-clock windows instead of the transitions they meant. They now wait on StepGate flags, reclaimable-run rows, lease expiry, heartbeat tick counts, and cancel receipts. The only production change is a `#[cfg(test)]` heartbeat counter; the loop still uses `tokio::time::interval`. Quiet/signal-woken cadences keep heavy drives off the SQLite writer, and no tests were deleted.

PR #2149, opened 2026-08-17
test(server): close leftover handshake and drop deadlocks in container-run tests

### LINKED ISSUE #2132: Flaky: sandbox_container_run timing-sensitive tests fail intermittently in the test lane

naingthet:
## Why

#2148 replaced elapsed-time windows with state waits, but the two `--test-threads N` deadlock seats it claimed to close were still in the module. Issue #2132 stayed open after that merge.

## What was still broken

- Producers published `entered` / `notify_one()` and only then waited on `release.notified()`. A faster test could observe the flag, fire `release`, and the wait missed it. The drive then parked forever on the first completion.
- `UsageThenPendingStream::drop` still called `std::sync::Barrier::wait()` on a tokio worker. The durable op-log uses `block_in_place`; a worker stuck in `Barrier::wait` starves every peer (0% CPU, nobody runs `spawn_blocking` to release the barrier).

## What changed

- A `Handshake` helper enables the release waiter **before** publishing `entered`. Tests poll the flag and then release.
- Setup, result, reclaim, and provision holds use the same handshake. Fence entry is a durable flag, not a one-shot notify.
- Drop waits on an `mpsc` through `block_in_place`, so the quiescence test can still hold the first stream open without parking a runtime worker on a barrier.

No product code. No tests deleted.

## Evidence

- `cargo fmt --all`
- `cargo test -p tidebreak-server --locked sandbox_container_run -- --test-threads 4 --skip docker`: 41 passed
- soak with a 45 s per-iteration timeout: **40/40** at `--test-threads 4`, 0 wedges

Closes #2132

### DIFF
--- a/crates/tidebreak-server/src/sandbox_container_run/tests.rs
+++ b/crates/tidebreak-server/src/sandbox_container_run/tests.rs
@@ -61,19 +61,22 @@ use crate::state::{SandboxSteerGuard, SandboxSteerRefusal};
 
 // --- Mock host model (the resolver the driver proxies inference through) ------
 
-/// A one-shot hold. The producer subscribes to `release` before publishing
-/// `entered`, so a faster test cannot fire `release` into a Notify that has
-/// not yet been polled.
+/// A one-shot hold. The producer takes the receiver and publishes `entered`
+/// before awaiting, so a faster test cannot lose the release. A `oneshot`
+/// stores the send even when the receiver has not been polled yet.
 struct Handshake {
     entered: AtomicBool,
-    release: Notify,
+    release_tx: Mutex<Option<tokio::sync::oneshot::Sender<()>>>,
+    release_rx: Mutex<Option<tokio::sync::oneshot::Receiver<()>>>,
 }
 
 impl Handshake {
     fn new() -> Self {
+        let (release_tx, release_rx) = tokio::sync::oneshot::channel();
         Self {
             entered: AtomicBool::new(false),
-            release: Notify::new(),
+            release_tx: Mutex::new(Some(release_tx)),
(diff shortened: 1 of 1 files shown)


naingthet (commented):

`Handshake::enter_and_wait` enables the release waiter before publishing `entered`, which closes the missed-notify race that parked the drive on the first completion. Replacing `Barrier::wait` on the tokio worker with an `mpsc` recv inside `block_in_place` keeps the quiescence hold without starving `DurableOperationStore` peers under `--test-threads N`. Fence entry is now a durable flag, so the pre-attach lock contention test no longer depends on a one-shot notify. Test-only; no product code.

naingthet (commented):

Follow-up replaces the remaining `Notify` waits with a oneshot `Handshake` (the send is stored if release wins the race) and durable `AtomicBool` flags for reverse-stream announce/drop. Direct `gate.release.notify_one()` call sites now go through `Handshake::release()`. Still test-only.

PR #2152, opened 2026-08-17
feat(code): optional workspace titles named in the background

naingthet:
The new-workspace dialog required a title even though the server already generated a two-word name when none was given. This makes the title optional end to end and adds background renaming, mirroring how chats are named.

## What changed

**UI**
- The title field in **NewWorkspaceDialog** is optional (placeholder "Named automatically"); create posts `title` only when typed.
- The sidebar, repo page, and workspace header prefer the digest title (`digest?.title ?? workspace.title`), so a background rename re-labels open surfaces live — the digest channel already restates the title on every notice.

**Server**
- New `code/titling.rs`: on turn submission (front of the turn, like chat titling), a background task derives a name from the submitted message on the **utility-role model** (`ModelRole::Utility` — the user's configured lightweight background model, or the curated cheap default per provider). No resolvable utility model → the placeholder stays; a lost title costs nothing.
- The chat-titling request/normalize/retry machinery is **shared, not duplicated**: `chat_titling::derive_title_with_retries` now takes the system prompt and schema name; chat behavior is unchanged (its 9 tests pass as before).
- The write is `set_workspace_title_if` — a compare-and-swap against the two-word placeholder, which is deterministic from the workspace id. A title the user typed (at create or via PATCH) can never be replaced, no matter when the derived name lands. Only the title moves; branch and worktree path keep their slug per decision 0032.
- One naming call per workspace at a time (in-flight set on `CodeRuntime`); a concurrent trigger is dropped rather than queued since the next turn retries anyway.

## Tests
- `code_titling::a_first_turn_names_an_untitled_workspace` — the whole feature over the wire against the scripted engine: untitled create answers the placeholder, the first turn renames it, the call ran on `gpt-5.4-nano` with the user's message as material, and the rename is announced on the digest channel.
- `workspace_title_swap_replaces_only_the_expected_title` (core) — the CAS contract that makes "a typed name always wins" true.

Not verified live: the rename end-to-end in the running desktop app (needs a credentialed provider profile); the flow is covered by the wire test above.

### DIFF
--- a/crates/tidebreak-core/src/db/ops/code/workspace.rs
+++ b/crates/tidebreak-core/src/db/ops/code/workspace.rs
@@ -101,6 +101,30 @@ pub async fn save_workspace(store: &DbStore, workspace: &CodeWorkspace) -> Resul
     Ok(result.rows_affected == 1)
 }
 
+/// Set a workspace's title only while it still reads `expected`.
+///
+/// The compare half is what lets background naming lose to a rename: a derived
+/// title replaces the generated placeholder it was derived against, and nothing
+/// else, no matter when the write lands. Returns whether the row changed.
+pub async fn set_workspace_title_if(
+    store: &DbStore,
+    id: WorkspaceId,
+    expected: &str,
+    title: &str,
+) -> Result<bool> {
+    let result = entities::code_workspace::Entity::update_many()
+        .col_expr(
+            entities::code_workspace::Column::Title,
+            sea_orm::sea_query::Expr::value(title.to_owned()),
+        )
+        .filter(entities::code_workspace::Column::Id.eq(id.0))
+        .filter(entities::code_workspace::Column::Title.eq(expected))
+        .exec(&store.conn)
+        .await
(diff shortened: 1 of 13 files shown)


naingthet (commented):

The create dialog posts `title` only when typed, and the sidebar, repo page, and workspace header all prefer `digest?.title`, so a background rename re-labels live surfaces. `set_workspace_title_if` is a compare-and-swap against the id-derived two-word placeholder, so a typed or PATCHed name always wins no matter when the derived write lands. Naming shares chat's derive/normalize/retry path on the utility model; one in-flight claim per workspace, next turn retries. Branch and worktree keep the slug per 0032.

naingthet (commented):

Merge of main picks up the container-run handshake fix and a zustand v5 selector change on `PendingApprovalBadge`: the selector now returns the pending count (a primitive) instead of a fresh filtered array, which would re-render forever. Titling is unchanged — optional create, CAS against the two-word placeholder, digest surfaces pick up the rename.

PR #2153, opened 2026-08-17
feat(code): grok sessions in unsupervised Auto, and a dropdown harness picker

naingthet:
Two changes to the code-mode surface, driven end to end through the CLI.

## Grok support (decision 0038, included)

Grok CLI was listed but unusable: every permission mode was refused. Re-probed live on the exact installed build (1.0.4 d846eb93, 2026-08-17):

- the **default headless posture** (`--prompt-file` + `--output-format streaming-json`, no permission flags composed) executes write tools unprompted — that *is* a workspace-write auto posture, reached without any bypass flag;
- `--permission-mode plan` still writes files, so Plan stays refused;
- there is still no approval channel, so Ask stays refused.

The carrier is a new capability flag, **`HarnessCaps.auto_mode`** — session creation now refuses each mode on its own flag instead of deriving Auto from the approval channel. Existing adapters keep exactly their previous behavior (Claude's auto rides the same version gate as its approvals). Grok declares `auto_mode: Supported` and maps Auto to the default posture; the bypass denylist is untouched. Everywhere unsupervised Auto is offered — create dialog, start-session prompt, CLI default-mode note — the product states it: *"This engine has no approval channel: in Auto, every action runs without asking."* ADR 0038 records the decision, the alternatives (synthetic `--deny` Plan, ACP approvals, harness-name policy), and what would unlock Plan/Ask. The grok fixture manifest gains the dated re-probe entries.

## Dropdown harness picker

`HarnessPicker` becomes a real `Select` dropdown: branded rows with subtitle, unusable rows disabled with their reason, and a doctor link under the control while any row is unusable. `StartSessionPrompt` gains an explicit **Start session** button (rows no longer double as one) and follows the selected engine's caps for the mode list and default. The mid-session composer's mode list is now doctor-caps-driven too.

## Verified by driving the CLI

- `code doctor`: grok now reports `resume,streaming,auto,reasoning,interrupt`.
- `session start --harness grok` (no mode) → defaults to auto and prints the unsupervised note; `--mode plan` / `--mode ask` → 422 with precise reasons; explicit `--mode auto` works.
- Real grok turns: a trivial reply and a file-write task both completed; `code files` shows the tracked change.
- Regression: claude_code, codex, and opencode each started with their default mode and completed a turn.

## Tests

- Grok adapter: auto honored, plan/ask refused; caps assertions; launch-plan denylist unchanged.
- Server: `auto_stands_on_its_own_capability_flag` — Auto refused when `auto_mode` is unsupported *even with approvals supported* (the case a derive-from-approvals implementation would pass), and an Auto-only engine creates an auto session.
- CLI: default-mode ladder (ask → plan → auto) unit-tested.
- UI: labels policy tests incl. the grok shape and the unsupervised statement; dropdown and start-prompt dom tests rewritten (`tsc`, 1089 tests, `pnpm build` all green).

Not verified live: the dropdown in the running desktop app (dom tests cover it).

### DIFF
--- a/crates/tidebreak-cli/src/api/code.rs
+++ b/crates/tidebreak-cli/src/api/code.rs
@@ -582,6 +582,7 @@ pub fn supported_caps_summary(caps: &HarnessCaps) -> String {
         ("approvals", caps.structured_approvals),
         ("steering", caps.mid_turn_steering),
         ("plan", caps.plan_mode),
+        ("auto", caps.auto_mode),
         ("reasoning", caps.reasoning_levels),
         ("file_events", caps.native_file_change_events),
         ("interrupt", caps.native_interrupt),
--- a/crates/tidebreak-cli/src/code.rs
+++ b/crates/tidebreak-cli/src/code.rs
@@ -14,8 +14,8 @@ use std::str::FromStr;
 use futures::StreamExt as _;
 use tidebreak_core::{
     AgentError, Attention, AttentionState, CapLevel, CodeApprovalId, CodeApprovalKind, CodeEvent,
-    CodePermissionMode, CodeSessionId, CodeSessionLifecycle, CodeTurnId, HarnessKind, RepoId,
-    Result, WorkspaceId,
+    CodePermissionMode, CodeSessionId, CodeSessionLifecycle, CodeTurnId, HarnessCaps, HarnessKind,
+    RepoId, Result, WorkspaceId,
 };
 use tokio_tungstenite::tungstenite::Message;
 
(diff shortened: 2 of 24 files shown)


naingthet (commented):

`auto_mode` is its own flag: session create refuses Plan/Ask/Auto independently, and the server test refuses Auto even when approvals are Supported. Grok 1.0.4 maps Auto to the default headless posture (no permission flags, denylist untouched) and still refuses Plan/Ask. Create, start-session, and the CLI all state unsupervised Auto at selection time; the picker is a real Select with unusable rows disabled. ADR 0038 records the rejected alternatives.

PR #2154, opened 2026-08-17
test(server): remove clock races from agent worker tests

naingthet:
## What changed

- Derive sandbox fixture claim timestamps from the committed turn row instead of the application clock.
- Replace resolver release notifications with a one-shot handshake so release cannot be lost.
- Replace fixed sleeps and current-time assertions with polling for the durable lease transition under test.
- Keep elapsed-time limits only as generous hang guards with useful last-observed diagnostics.

## Why

CI recorded intermittent failures in `local_cancellation_finalization_survives_accounting_failure_past_execution_lease` and `resolver_delay_past_the_database_lease_prevents_provider_egress`.

The audit also reproduced a fixture race: SQLite acceptance timestamps are intentionally rounded to the end of the current millisecond, so an immediate application-clock claim could be a few microseconds earlier than the committed `updated_at` and see no candidate. The resolver tests additionally used a lossy `Notify` release and one-second scheduling windows under full-suite load.

## Validation

- `cargo test -p tidebreak-server --locked sandbox_agent_run_worker::tests -- --test-threads 4` — 43 passed
- 30/30 repeated passes for each affected test:
  - `cancellation_while_resolving_prevents_the_provider_request`
  - `local_cancellation_finalization_survives_accounting_failure_past_execution_lease`
  - `resolver_delay_past_the_database_lease_prevents_provider_egress`
- `cargo fmt --all --check`
- `git diff --check`

No product code changed and no tests were removed.

### DIFF
--- a/crates/tidebreak-server/src/sandbox_agent_run_worker/tests.rs
+++ b/crates/tidebreak-server/src/sandbox_agent_run_worker/tests.rs
@@ -293,15 +293,21 @@ impl ModelProvider for ThreeCallStepThenFinalProvider {
 
 struct DelayedResolver {
     entered: Arc<Notify>,
-    release: Arc<Notify>,
+    release: Mutex<Option<tokio::sync::oneshot::Receiver<()>>>,
     provider: Arc<dyn ModelProvider>,
 }
 
 #[async_trait]
 impl ProviderResolver for DelayedResolver {
     async fn resolve(&self) -> Arc<dyn ModelProvider> {
+        let release = self
+            .release
+            .lock()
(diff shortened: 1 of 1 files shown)


naingthet (commented):

Claim time is derived from the committed turn's `available_at`/`updated_at` plus 1µs, so SQLite's millisecond-rounding cannot hide the candidate. `DelayedResolver` now uses a oneshot (send is stored if release wins). Lease expiry and cancellation-renewal wait on the durable row instead of a 1.1s sleep. Test-only.

PR #2155, opened 2026-08-17
feat(desktop): experimental settings section, with code mode opt-in

naingthet:
Code mode sat in every profile's rail while it is still an experiment. This adds a **Settings → Experimental** section and puts code mode behind its switch, off by default.

## What changed

**Server**
- New runtime setting `code_mode_enabled` on `GET/PUT /settings` (key `code_mode.enabled`), **default false** — the default is pinned in the settings round-trip test, since "ships opt-in" is the decision that would otherwise be easy to lose.

**Desktop**
- New **Experimental** settings section (flask icon, between Appearance and Updates) with the Code mode switch and a jump into the surface once enabled. Copy states the contract: off by default, turning one off hides it without deleting anything.
- The app rail lists **Code** only while opted in. The `/code` routes still resolve — a deep link or stale history entry lands on a short "Code mode is experimental" explanation with a button to the Experimental panel, per the same deep-link contract the settings rail already follows.
- The **Coding harnesses** settings section leaves the rail with the flag (its route also still resolves).
- Flags load once at shell boot into a small store (`useExperimentalFlags`); until they load the app behaves as opted out, so the rail never flashes an entry it is about to hide. The panel persists through `PUT /settings` first and only then updates the store.

**Deliberately not gated**: the `/code/*` routes and the `tidebreak code` CLI. The flag gates discovery, not data — existing workspaces and sessions survive an opt-out untouched, there is no engine-lifecycle edge case on toggle, and invoking `tidebreak code` is already an explicit act.

## Driven over the wire

Against a fresh profile with `tidebreak serve`: `GET /settings` answers `code_mode_enabled: false` by default; `PUT {"code_mode_enabled": true}` persists and reads back; toggling back off leaves `computer_use_enabled` and the rest untouched; `tidebreak code doctor` keeps working while opted out.

Not verified live: the desktop rail/gate rendering (covered by `tsc`, the 1089-test suite, and `pnpm build`).

### DIFF
--- a/crates/tidebreak-desktop/ui/src/AppShell.tsx
+++ b/crates/tidebreak-desktop/ui/src/AppShell.tsx
@@ -22,6 +22,7 @@ import {
   prependReplacementChat,
 } from "./ChatDeletion";
 import { useChatListStore } from "./ChatListStore";
+import { useExperimentalFlags } from "./experimental";
 import { connectOutputs } from "./deliverables";
 import { useProjectListStore } from "./ProjectListStore";
 import { useComposerDrafts } from "./ComposerDrafts";
@@ -198,6 +199,9 @@ export function AppShell() {
         if (!cancelled) setBootError(String(err));
       }
     })();
+    // Loads apart from the catalog on purpose: a flags failure keeps the
+    // opted-out defaults instead of taking down the shell.
+    void useExperimentalFlags.getState().refresh(client);
     return () => {
       cancelled = true;
     };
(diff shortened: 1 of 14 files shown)


naingthet (commented):

`code_mode.enabled` defaults false and the settings round-trip pins that. The rail and Coding harnesses section stay hidden until the switch lands via `PUT` first, then the store; until flags load the app stays opted out so the rail cannot flash. Deep links still resolve through `CodeModeGate`. The `/code/*` HTTP routes and `tidebreak code` CLI stay ungated on purpose.

PR #2156, opened 2026-08-17
feat(desktop): widen the sidebar and make it resizable

naingthet:
## Summary
- Raise the expanded sidebar default from 224px to 280px.
- Add a trailing-edge drag handle (with keyboard support and double-click reset) so the rail width can be resized between 200–480px.
- Remember the chosen width in localStorage and keep the overlaid titlebar aligned with the rail.

## Test plan
- [ ] Expand the sidebar and confirm it opens wider than before (~280px).
- [ ] Drag the right edge to resize; width should track the pointer without lag.
- [ ] Reload the app and confirm the chosen width is restored.
- [ ] Double-click the handle to reset to the default width.
- [ ] Collapse and expand again; preferred expanded width should return.
- [ ] Confirm the titlebar/history controls still sit over the rail at the new width (macOS overlay titlebar included).
- [ ] Open a content panel beside chat and confirm the sidebar resize remains independent of the chat/panel split.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/AppShell.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/AppShell.dom.test.tsx
@@ -8,7 +8,7 @@ import { usePendingPrompts } from "./PendingPrompts";
 import { useProjectListStore } from "./ProjectListStore";
 import { useRefreshSignals } from "./RefreshSignals";
 import { useChatsSectionState } from "./sidebar/ChatsSection";
-import { useUiStore } from "./UiStore";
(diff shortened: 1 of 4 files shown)


naingthet (commented):

Default 280px, clamp 200–480, persist on pointer-up only, double-click reset, and keyboard steps on the separator all look right. Live drag skips the collapse/expand transition so the rail tracks the pointer. Titlebar follows `--sidebar-expanded-width` on the shell, so overlay chrome stays aligned. Independent of the chat/panel split.

Shell test reset of `sidebarWidth` is the right isolation.

PR #2157, opened 2026-08-17
fix(desktop): navigate after archive and repair the code pane height chain

naingthet:
Four fixes on the code workspace page.

**Archiving left the reader on the workspace it removed.** The page is
addressed by `/code/w/$workspaceId`, and archiving never changed the route, so
after the worktree was gone the page stayed pinned to it. The catalog
bookkeeping also cancelled itself out: `upsertWorkspace(archived)` was
immediately undone by `forgetWorkspace` dropping the same row from the list.

Archiving now replaces the location with the workspace's repo — which already
lists archived workspaces with a status label — and falls back to `/code` when
no repo id is available. Both entry points, plain and forced, funnel through
the same call. The catalog keeps the archived row (the rail filters archived
workspaces out on its own) and drops only the session entry, so
`forgetWorkspace` narrows to `forgetWorkspaceSession`; it had no other caller.

Whether an archived workspace stays readable by deep link is a separate
product question, so no redirect guard is added here.

**The pane never claimed its height.** It asked for `flex-1`, but the slot it
sits in is a `react-resizable-panels` `Panel`, which renders a plain block —
nothing stretched the pane, so its height went content-driven, the transcript
scroller never became a scroll container, and the composer was pushed past the
bottom of the window. It takes `h-full` instead, the same way `.chat-pane`
does. That reason is now written down at all three sites that have hit it.

**The transcript had no centering column.** It was full-width, so on a wide
window items hugged the left edge while the `ml-auto` user bubble sat far
right. The container is bounded to the same 48rem column chat uses. Bubble
styling and class names are deliberately untouched — a broader convergence
pass is separate work.

**Approvals refetched on every delta.** The effect listed `items` in its
dependency array, and that array is fresh on every reducer event, so
`GET /code/approvals` fired for every streamed token of a turn. It is keyed on
the approval ids and their states instead, which is what actually changes the
response.

One DOM test covers the archive navigation: it drives the confirm dialog and
asserts the router lands on the repo route with the workspace page unmounted.
It fails without the navigation fix. The height and width changes are
CSS-mechanical and have no test.

Verified with `pnpm test` (171 files, 1090 tests) and `pnpm build`. Not driven
in the running app.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeCatalogStore.ts
+++ b/crates/tidebreak-desktop/ui/src/code/CodeCatalogStore.ts
@@ -32,7 +32,7 @@ type CodeCatalogStore = CodeCatalogState & {
     "listCodeRepos" | "listCodeWorkspaces" | "getHarnessDoctor"
   >) => Promise<void>;
   rememberSession: (session: CodeSessionSnapshot) => void;
-  forgetWorkspace: (workspaceId: string) => void;
+  forgetWorkspaceSession: (workspaceId: string) => void;
   upsertRepo: (repo: CodeRepoSnapshot) => void;
   upsertWorkspace: (workspace: CodeWorkspaceSnapshot) => void;
   reset: () => void;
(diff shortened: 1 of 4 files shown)


naingthet (commented):

Archive now `replace`s to `/code/r/$repoId` (or `/code`) instead of staying on a dead worktree. Catalog keeps the archived row and only drops the session — `forgetWorkspaceSession` matches that. `h-full` on the pane is the same height-chain fix as `.chat-pane`. Transcript `max-w-3xl` matches chat. Approvals effect keyed on id+state stops the per-token refetch. DOM test drives the confirm dialog and asserts the repo route.

Looks good.

PR #2158, opened 2026-08-17
fix(desktop): refresh git panels when a turn completes

naingthet:
The review surfaces on a code workspace went stale as soon as the engine
touched a file. The pull-request card, the changed-file list, and the diff
each fetched once on mount and never again: after a turn that edited files,
the card still said "No commits" with Commit disabled, and a diff panel left
open during a turn kept describing the worktree as it was when the panel
opened. The only way out was a full page reload.

## What changed

**A content signal from the journal.** The session reducer now carries a
`contentRevision`, bumped when a checkpoint is recorded and on every turn
resolution — including failed and interrupted turns, which still leave
whatever the engine wrote before it stopped. `CodeWorkspacePage` subscribes
to it and threads the number into `PrCard`, `FilesPanel`, and `DiffPanel`.

**One reloading hook.** `useLiveResource` holds the fetch for all three:

- a burst of events debounces into a single request;
- a revision that lands while a request is in flight queues exactly one
  follow-up instead of stacking requests;
- the last value stays on screen while the next one loads, so a refresh
  never blanks a panel;
- a response from a superseded key (a different turn or file) or from an
  unmounted view is dropped rather than applied.

**Loading and manual refresh.** Each surface shows a spinner or skeleton
while it has nothing to display yet, using the shared `Spinner`/`Skeleton`
components. The pull-request card also gets a small ghost refresh button, so
edits Tidebreak never saw — a commit made in a terminal, a file changed by
hand — can be picked up without a reload.

## Testing

One DOM test drives a `turn_completed` frame through a real session store
into a mounted `PrCard` and asserts the card refetches and flips from
"No commits" to "Uncommitted" with Commit enabled. It fails on the previous
behavior. `pnpm test` and `pnpm build` pass.

Not verified by driving the running desktop app: the spinner and skeleton
placement, and the manual refresh button, were checked only through the
tests and the type/build lanes.


naingthet (commented):

`contentRevision` on checkpoint and every turn resolution (including fail/interrupt) is the right stale signal. `useLiveResource` keeps the last value, debounces, and queues one follow-up instead of stacking. PrCard / Files / Diff all share it; manual refresh covers out-of-band git. DOM test drives `turn_completed` through the session store into PrCard.

Looks good.

PR #2159, opened 2026-08-17
fix(server): NUL-delimited numstat parsing and panic-free truncation

naingthet:
Three byte-level defects in code mode's git and stream-parsing paths.

## Numstat counts lost on renames and non-ASCII paths

`parse_numstat` read `git diff --numstat` without `-z`, while `parse_name_status`
used it. Without `-z` git renders a same-directory rename as
`src/{alpha.rs => beta.rs}` and quotes and C-escapes any path outside ASCII, so
those records never matched the name-status path they were meant to annotate and
the file kept zero insertions and deletions. That flowed into the journaled
`CheckpointRecorded`, the diffstat badge, and the generated commit message.

Both commands now run with `-z`. The parser handles the NUL-delimited rename
form — the counts are followed by an empty path column, then the pre-image and
post-image paths as two further NUL-terminated fields — and splits the counts off
with `splitn(3, '\t')` so a tab inside a path cannot truncate it.

## `String::truncate` on non-character boundaries

`String::truncate` panics when the byte cap lands inside a multi-byte character.
Five caps applied it directly to engine- or shell-controlled text: the
login-shell stderr capture in `probe.rs` and the `MAX_UNRECOGNIZED_LOG` payload
log in each of the four adapters — the code whose entire job is surviving input
it does not recognize. All five now go through one helper that floors the cap to
a character boundary (`str::floor_char_boundary` is still unstable).

## Oversized line wedged a session's stream

`StreamLineBuffer::push` had the same panic and a worse failure behind it. On
overflow it kept a capped prefix that contained no newline and dropped the rest
of the chunk, so the pending buffer stayed permanently full: no line ever
completed again, and every later line on that session was silently lost. It now
drops bytes only until the oversized line's newline arrives, emits that line
truncated at the cap, and resumes normally. Overflow accounting is unchanged.

## Tests

Three, each a reproduction that fails without the corresponding fix:

- a checkpoint over a same-directory rename and a non-ASCII filename, asserting
  the per-file and total line counts (verified failing on the old parser with
  `(0, 0)` for both files);
- a cap landing inside two- and three-byte characters, which panicked before;
- an oversized line followed by a normal line, asserting the normal line still
  comes through and the buffer keeps working afterwards.

### DIFF
--- a/crates/tidebreak-harness/src/budget.rs
+++ b/crates/tidebreak-harness/src/budget.rs
@@ -43,6 +43,9 @@ pub struct BudgetTick {
 #[derive(Debug, Default)]
 pub struct StreamLineBuffer {
     pending: String,
+    /// Whether the line currently being buffered has already hit the cap, so
+    /// the rest of it is dropped until its newline arrives.
+    overflowing: bool,
     /// Total overflow chunks observed over the life of the buffer.
     pub overflow_chunks: u64,
 }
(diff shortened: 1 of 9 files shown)


naingthet (commented):

`parse_numstat` now takes `--numstat -z` and the empty-path rename form, so same-directory renames and non-ASCII paths get real counts instead of (0, 0). Shared `floor_char_boundary` / `truncate_on_char_boundary` replaces the five `String::truncate` panic sites. `StreamLineBuffer` drops overflow until the newline, emits the capped line, and keeps later lines — the wedge is gone. The three tests are reproductions, not existence checks.

Looks good.

PR #2160, opened 2026-08-17
fix(server): preserve worktrees on failed archive scripts and fence session saves

naingthet:
Three fixes on the code-mode archive and worker-supersession paths.

## A failing archive script no longer destroys the worktree

`archive_workspace` treated `run_workspace_script` returning `Ok(_)` as
success, but that function only reports `Err` when the script cannot be
spawned or times out — a non-zero exit comes back as `Ok(ScriptRun { success:
false, .. })`. An archive script whose job is to back up or push state could
fail and the archive proceeded to `git worktree remove` anyway, taking the
work with it. The setup path checks `run.success`; the archive path never
ported the check, though decision 0032 states the archive hook obeys the same
failure-preserves rule.

Archive now aborts with `archive_script_failed` and leaves the checkout on
disk when the script fails. Setup and archive share one implementation
(`run_hook_script`) so the two cannot drift apart again.

The script also ran *before* the force and running-session checks, so an
archive that was subsequently refused had already executed it — a refused
archive could push, tag, or upload. Blockers are now evaluated first, and the
script only runs against a worktree that still exists (an already-removed
worktree archives without error, and has nothing to back up).

## `save_session` cannot be regressed by a superseded worker

The update carried no predicate but the row id, so it wrote `spawn_epoch`
unconditionally. Archive and reap send `Shutdown` and re-attach; the outgoing
worker keeps its own `CodeSession` snapshot and writes it while unwinding,
putting its stale epoch back over the new one. That un-fences the worker that
is leaving and makes the live worker the stale one: its appends fail
`StaleSpawnEpoch` and are dropped, so the session goes permanently silent
while the UI still shows it healthy.

The save is now fenced on `spawn_epoch` the same way journal appends are — the
write lands only if the caller is at least as new as the stored row. Fencing
the whole update, not just the epoch column, also stops a superseded worker
from regressing lifecycle, attention, or the recorded child pid. The dropped
write is logged rather than passing silently.

## Archive and reap wait for the outgoing worker

Both removed the handle, sent `Shutdown`, and continued without waiting, so a
new engine could be launched into the same worktree while the old one was
still shutting down. Both now wait for the worker to finish: the worker drops
its command receiver as its last act, after the engine is shut down, so the
sender observing that close is exactly \"the worker is gone\".

One `Shutdown` is enough. An idle worker leaves its loop on it; a worker
mid-turn treats it as an interrupt and then leaves at the top of the next
iteration, because `end_workspace_sessions` marks the row ended *before* it
asks the worker to stop and the worker already re-reads the row there
(`session_was_ended`). Send and wait share one deadline, so neither a wedged
engine nor a worker that has stopped draining its commands can hold an archive
open; the epoch fence keeps a late writer harmless if one is left behind.

## Tests

- `a_failing_archive_script_preserves_the_worktree` — a refused archive has
  not run the script, and a forced archive whose script exits non-zero
  surfaces `archive_script_failed`, keeps the worktree, and leaves the
  workspace active.
- `a_superseded_worker_cannot_regress_the_session_row` — a save from an older
  epoch is rejected and leaves epoch, lifecycle, and child pid alone, and the
  live worker can still append.

Verified locally on the rebased branch: `cargo fmt --check`, clippy on both
crates with `--all-targets`, `cargo test -p tidebreak-server` (967 passed),
`cargo test -p tidebreak-core` (671 passed), and `cargo test -p
tidebreak-desktop` (152 passed, 0.6s). The force-archive-during-a-turn test
runs in 0.44s, so the shutdown wait resolves on worker exit rather than on its
timeout.

### DIFF
--- a/crates/tidebreak-core/src/db/ops/code/session.rs
+++ b/crates/tidebreak-core/src/db/ops/code/session.rs
@@ -141,6 +141,18 @@ pub async fn list_sessions_by_lifecycle(
 }
 
 /// Persist mutable session fields. `id`, `workspace_id`, and `created_at` stay as stored.
+///
+/// The write is fenced on `spawn_epoch`, the same way journal appends are: a
+/// caller holding an epoch older than the stored row is a superseded worker
+/// unwinding, and its write is dropped rather than allowed to regress the row.
+/// Without the fence such a worker writes its own epoch back over a newer
+/// spawn's — un-fencing itself, making the live worker the stale one, and
+/// silently dropping everything the live worker appends afterwards. The same
+/// fence keeps a superseded worker from regressing lifecycle, attention, or
+/// the recorded child pid.
+///
+/// Returns `false` when nothing was written: either the row is gone or the
+/// caller has been superseded.
 pub async fn save_session(store: &DbStore, session: &CodeSession) -> Result<bool> {
     let result = entities::code_session::Entity::update_many()
         .col_expr(
@@ -191,10 +203,24 @@ pub async fn save_session(store: &DbStore, session: &CodeSession) -> Result<bool
             sea_orm::sea_query::Expr::value(session.unrecognized_event_count),
         )
         .filter(entities::code_session::Column::Id.eq(session.id.0))
+        .filter(entities::code_session::Column::SpawnEpoch.lte(session.spawn_epoch))
         .exec(&store.conn)
         .await
         .map_err(store_err)?;
-    Ok(result.rows_affected == 1)
+    if result.rows_affected != 1 {
+        // Rare, and invisible without a word: distinguish a superseded writer
+        // from a row that is simply gone.
+        if let Some(current) = get_session(store, session.id).await? {
+            tracing::warn!(
+                session = %session.id,
+                attempted = session.spawn_epoch,
+                current = current.spawn_epoch,
+                "dropping a session write from a superseded code-session worker"
+            );
+        }
+        return Ok(false);
+    }
+    Ok(true)
 }
 
 pub(super) fn session_from_row(row: entities::code_session::Model) -> Result<CodeSession> {
--- a/crates/tidebreak-server/src/code/runtime.rs
+++ b/crates/tidebreak-server/src/code/runtime.rs
@@ -42,7 +42,7 @@ use super::session_worker::{
 };
 use super::worktree::{
     self, archive_blockers, branch_name, create_worktree, prune_worktrees, remove_worktree,
-    run_setup_script, slugify, validate_repo_path, worktree_dir, WorktreeError,
+    run_archive_script, run_setup_script, slugify, validate_repo_path, worktree_dir, WorktreeError,
 };
 use crate::error::ServerError;
 
@@ -467,19 +467,8 @@ impl CodeRuntime {
             return Ok(workspace);
         }
         let repo = self.get_repo(workspace.repo_id).await?;
-        if let Some(script) = repo.archive_script.as_deref() {
-            if let Err(err) = super::setup_script::run_workspace_script(
-                std::path::Path::new(&workspace.worktree_path),
-                script,
-            )
-            .await
-            {
-                return Err(ServerError::unprocessable_kind(
-                    "archive_script_failed",
-                    err,
-                ));
-            }
-        }
+        // Blockers first: a refused archive must leave the workspace exactly as
+        // it was, and running the hook script is not "exactly as it was".
         self.refuse_running_sessions(id, force).await?;
         let path = std::path::Path::new(&workspace.worktree_path);
         if path.exists() {
@@ -493,6 +482,17 @@ impl CodeRuntime {
                         "workspace has uncommitted or unpushed work; pass force to discard it",
                     ));
                 }
+            }
+            // Decision 0032: the archive script obeys the same
+            // failure-preserves rule as setup. A script that backs up or
+            // pushes state and exits non-zero must stop the archive, not have
+            // its checkout removed underneath it. A worktree that is already
+            // gone has nothing to run the script against.
+            if let Err(err) = run_archive_script(path, repo.archive_script.as_deref()).await {
+                return Err(ServerError::unprocessable_kind(
+                    "archive_script_failed",
+                    err.to_string(),
+                ));
             }
         }
         self.end_workspace_sessions(id).await?;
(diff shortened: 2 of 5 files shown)


naingthet (commented):

Archive now checks `run.success` via shared `run_hook_script`, and blockers run before the hook so a refused archive cannot push/tag first. `save_session` is fenced on `spawn_epoch` so an unwinding worker cannot un-fence the live one. Archive/reap wait (bounded) for the outgoing worker and re-read the row. Both reproductions pin the bugs.

Looks good.

naingthet:
The cancelled `desktop test` lane was not this branch. That job never reached the tests: it spent 21:07:18 to 21:27:17 inside the `Install system deps (Tauri)` step, with `apt-get update` stalled after `Get:5 https://archive.ubuntu.com/ubuntu noble-security InRelease`, and the 20-minute job timeout cancelled it there. The step log ends on that line, and no `desktop tests` step output exists in the job.

The desktop crate has no code-mode tests, so the archive and reap paths this branch touches are not exercised by that lane at all. Locally `cargo test -p tidebreak-desktop --locked` is 152 passed in 0.60s.

Rebased onto main (over #2161-#2164) and pushed a follow-up that removes the Shutdown resend loop anyway, since it was a genuine hazard even though it was not this failure — see the second commit.

naingthet (commented):

The new commit is the right follow-up: ending the row first, then one `Shutdown` under a single deadline, instead of resending every 50ms and replaying `interrupt` into an engine already stopping. `session_was_ended` is the existing leave path, so a mid-turn worker still exits without a resend loop.

Archive-script failure still preserves the worktree, and session saves stay fenced. Same as the earlier head plus this shutdown fix.

naingthet (commented):
Still good on the new head (`eb64ee4b`).

The follow-up — one `Shutdown`, wait on the command channel closing, mark the row ended first so a mid-turn worker leaves on its own — is the right pairing. Resending would only replay `interrupt` into an engine that is already stopping. The 5s grace plus the `spawn_epoch` fence on `save_session` keeps a wedged or late writer from blocking archive/reap or regressing the live row.

The rest still holds: archive script after blockers, non-zero exit is `archive_script_failed` and preserves the checkout, setup/archive share `run_hook_script`.

This head is behind / blocked vs `main` — rebase before merge. I am not merging.

crates/tidebreak-server/src/code/runtime.rs
```diff
@@ -920,17 +927,55 @@ impl CodeRuntime {
...
-                let _ = handle.commands.send(WorkerCommand::Shutdown).await;
-            }
+            // Mark the row ended before asking the worker to stop. A worker
+            // interrupted mid-turn re-reads the row on its way round the loop
+            // and leaves on its own when it finds the session ended, so one
+            // `Shutdown` is enough however busy it was.
             session.lifecycle = CodeSessionLifecycle::Ended;
             session.child_pid = None;
             session.fence_reason = None;
             super::attention::persist_session(&self.db, &self.bus, &session).await?;
```
naingthet, line 937: **[bug]** `end_workspace_sessions` persists `Ended` and then waits for the worker, but `save_session` only fences on `spawn_epoch <= caller`. The outgoing worker still holds that epoch, so it is not superseded. Its in-memory row (`lifecycle = Running` at turn start / pid updates in `drive_turn`, or `Idle` after a `session_was_ended` check that ran before this persist) can overwrite `Ended` during the wait. After `shut_down_worker` returns the row is not written again. Reap is safe because `attach_and_spawn_worker` then bumps the epoch; archive never bumps, so a late same-epoch persist can leave an `Idle`/`Running` session (and a restored `child_pid`) on an archived workspace. The new `shut_down_worker` comment that “the epoch fence keeps a late writer harmless either way” is not true on this path.

**Suggestion:** After the wait, persist `Ended` again from a fresh load (or fail the archive if the row is no longer `Ended`). Stronger: refuse `save_session` when the stored lifecycle is `Ended` unless the caller is also `Ended`, or have the worker re-read immediately before every persist. The epoch fence should stay; it does not replace a lifecycle CAS here.

crates/tidebreak-core/src/db/ops/code/session.rs
```diff
@@ -141,6 +141,18 @@ pub async fn list_sessions_by_lifecycle(
 }
 
 /// Persist mutable session fields. `id`, `workspace_id`, and `created_at` stay as stored.
+///
+/// The write is fenced on `spawn_epoch`, the same way journal appends are: a
```
naingthet, line 145: **[suggestion]** The new `save_session` and `shut_down_worker` docs narrate the failure story and the shutdown protocol at essay length. They restate what the code does and embed design history (“without the fence… un-fencing itself”) rather than a short constraint.

**Suggestion:** Keep one or two sentences of why (epoch must be non-decreasing; wait on receiver close, not a second `Shutdown`) and drop the rest. The “late writer is harmless” claim in `shut_down_worker` should go or be limited to reap, where a later bump actually makes it true.

naingthet (commented):
## Summary

The three fixes match the stated intent and the important parts look right: archive now treats a non-zero hook exit as failure (via `run.success`) and skips the hook on a refused archive; `save_session` can no longer put a stale `spawn_epoch` back on the row; reap waits for the outgoing worker and re-reads before respawn.

**Do not merge yet.** Archive’s “mark Ended, then wait” story has a real hole: the new epoch fence only drops *older* writers, so the still-current worker can persist its in-memory `Running`/`Idle` snapshot over `Ended` during the wait, and nothing re-asserts `Ended` afterwards. Force-archive of a live session can leave the row not actually ended.

GitHub will not accept `REQUEST_CHANGES` from the PR author; treating this as a blocking comment instead.

## Issue counts by severity

- bugs: 1
- suggestions: 1
- nits: 0

crates/tidebreak-server/src/code/runtime.rs
naingthet: Fixed in 4aedb3bb.

`save_session` now refuses a persist that would revive `Ended` (lifecycle CAS, epoch fence kept). `end_workspace_sessions` also reloads after the wait and writes `Ended` again — or fails the archive if that write does not land. `an_ended_session_cannot_be_revived_by_a_same_epoch_persist` covers the same-epoch Running/Idle overwrite.

Trimmed the `save_session` / `shut_down_worker` docs in the same commit; the “late writer is harmless” claim is gone.

naingthet:
`Ended` can no longer be overwritten by the outgoing same-epoch worker.

- `save_session` refuses a persist that would revive `Ended` (epoch fence kept).
- `end_workspace_sessions` reloads after the wait and writes `Ended` again, or fails the archive if that write does not land.
- `an_ended_session_cannot_be_revived_by_a_same_epoch_persist` is the regression test.

`4aedb3bb`

PR #2161, opened 2026-08-17
feat(desktop): mode-scoped shortcuts, Cmd+N creates a workspace in code mode

naingthet:
Cmd+N was globally bound to `new-chat`. Pressed in code mode it created a
chat and navigated the reader out of the mode they were working in, which is
never what the key meant there: the unit of work in code mode is a workspace
— a worktree and a branch, per ADR 0030.

**The rule this establishes: shell shortcuts are mode-scoped, and the mode
derives from the route family.** Anything under `/code` is code mode,
everything else is chat. A definition may declare `scope: "chat" | "code"`,
or leave it out to stay live in both — Cmd+B still toggles the sidebar
everywhere. Two definitions may share a chord only across scopes, which is
what lets Cmd+N mean "new chat" on one side and "new workspace" on the other.

- `resolveShellShortcut` takes the mode as a **required** context field, so
  every caller has to say which half of the app the key was pressed in rather
  than silently resolving against the wrong table. The terminal pane, which
  builds its own context to keep shell chords out of xterm, says `"code"`.
- The help dialog lists the shortcuts for the mode it is being read in.
  Listing every scope would show two conflicting Cmd+N rows, and a help dialog
  that misstates the keys is worse than none.
- The shell reads the mode inside the keydown handler through the router
  rather than subscribing to it, so the root frame does not re-render on every
  navigation.
- The inline `/code` path check in the chat rail is now `isCodeRoute`, shared
  with the mode derivation.

Code mode's dialogs move out of component state into a small store, so the
shell can open the new-workspace flow without importing code-mode surfaces,
and there is one dialog instance in the rail rather than one per page that
opens it. The rail is mounted on every `/code` route, so the shortcut is
reachable anywhere in the mode. Two edges the store handles: with nothing
registered it opens repo registration instead, because the new-workspace form
would be one the reader cannot submit; fired on a repo page it preselects
that repo, matching the page's own button, and ignores a repo id the catalog
does not know.

No native menu accelerator: a native Cmd+N would consume the key
window-globally and defeat the scoping.

Tests: Cmd+N resolves to the right action in each mode; an invariant that no
two definitions share a chord within one mode (sharing across modes is the
feature, and the same table order that makes it work is how a shortcut could
silently stop firing); the help-grouping check is now per mode. One DOM test
drives a `/code` route and pins that the chord opens the workspace dialog and
does not create a chat. The gated-shell guard beside it pressed Meta, which
is bound to nothing under jsdom's non-mac user agent — it now presses the
modifier that platform actually uses, so it proves something.

Verified with `pnpm test` and `pnpm build` in `crates/tidebreak-desktop/ui`;
the shortcut was not exercised in the running desktop app.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/AppShell.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/AppShell.dom.test.tsx
@@ -4,6 +4,9 @@ import userEvent from "@testing-library/user-event";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
 import { useChatListStore } from "./ChatListStore";
+import { useCodeCatalogStore } from "./code/CodeCatalogStore";
+import { useCodeUiStore } from "./code/CodeUiStore";
+import { useExperimentalFlags } from "./experimental";
 import { usePendingPrompts } from "./PendingPrompts";
 import { useProjectListStore } from "./ProjectListStore";
 import { useRefreshSignals } from "./RefreshSignals";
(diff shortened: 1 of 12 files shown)


naingthet (commented):

Mode from the route family, `scope` on the table, required `mode` on `resolveShellShortcut` — Cmd+N is new-chat vs new-workspace without a native accelerator eating the key. `CodeUiStore` lets the rail/shortcut open one dialog; empty catalog falls through to add-repo. Help is filtered to the mode being read. Tests cover the chord, the within-mode uniqueness invariant, and a `/code` DOM path.

Looks good.

PR #2162, opened 2026-08-17
fix(desktop): allow queueing follow-ups mid-turn and fix the permission control

naingthet:
Two dead controls in the code-mode composer.

## Queueing was reachable only by race

The server has queued a mid-turn send since code mode landed: a submit on a busy session parks the message in the session's single slot and answers `202` with a queue receipt (`QueuedCodeTurn`) instead of a turn snapshot. The desktop could not get there. The composer swapped Send for Interrupt while a turn ran, and `parseCodeTurn` rejects anything that is not a turn snapshot — so if a queued receipt ever did arrive, the client reported "Could not send that turn" for a message the server was holding, and invited the retry that double-sends.

- The composer sends while a turn is running. The button reads **Queue**, Interrupt sits beside it, and Enter queues rather than silently doing nothing.
- `submitCodeTurn` returns a discriminated `CodeTurnSubmission` (`ran` | `queued`), matching what the CLI already accepts for the same route. Only a turn that ran inserts a transcript item — a queued follow-up has no turn row to key one on.
- A queued send says "Queued — runs after the current turn", cleared when that turn finishes. The queue's own refusal (`409 queue_full`) is stated in product terms instead of as a raw status line.

Two things surfaced while wiring it, both in the way of the feature:

- Submit answers only when the turn *ends*. The composer held the draft and disabled the textarea for the whole turn, which leaves nowhere to write the follow-up that queueing exists to accept. It now clears on send and puts the text back if the server refuses.
- The accepted snapshot was the only source of a prompt bubble, so a turn the client did not submit — every queued follow-up, and any turn started elsewhere — had no prompt in the transcript until the next open. The session registry now pulls the snapshot when the socket announces a turn it has no prompt for, once per turn.

## The permission control was inert

`PermissionModePicker` rendered pressable `aria-pressed` buttons in the composer, but nothing ever passed `onChange`, so clicking did nothing.

The mode is not changeable on a live session. It is composed into the engine's launch — argv flags per harness — and refused per capability at creation (decisions 0033 and 0038); `HarnessSession` has no seam for changing it, and doing so would mean engine relaunch semantics with resume refs, spawn epochs, and in-flight approvals to settle. That is a decision record, not a side effect of this fix. So the composer now *states* the session's mode (`PermissionModeSummary`) rather than offering it, keeping unavailable modes visible and dimmed with their reason. The picker is untouched where the choice is real: the new-workspace dialog and the start-session prompt.

**Follow-up:** changing permission mode mid-session is not implemented. It needs a decision on relaunch semantics first.

Also adds `shrink-0` to the composer root so a long draft cannot squeeze the transcript.

## Tests

- `CodeComposer.dom.test.tsx`: a follow-up typed during a running turn calls send and shows the queued affordance; `409 queue_full` surfaces its message and keeps the draft; the mode is stated with no pressable control. The old test asserting `onPermissionMode` fired on click is gone with the control it described — `StartSessionPrompt.dom.test.tsx` still covers the picker where it is interactive.
- `parsers.test.ts`: the two `202` shapes are told apart.
- `CodeSessionRegistry.test.ts`: a turn announced by the socket gets its prompt filled in.

`pnpm test` (1094 passing) and `pnpm build` pass. No Rust changes — the server side already worked. Not driven in the running app.

The parked scope is recorded in `docs/deferred.md` alongside the rest of what code mode leaves out.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/api/client.ts
+++ b/crates/tidebreak-desktop/ui/src/api/client.ts
@@ -119,8 +119,9 @@ import {
   parseCodeRepo,
   parseCodeSession,
   parseCodeSessionList,
-  parseCodeTurn,
   parseCodeTurnList,
+  parseCodeTurnSubmission,
+  type CodeTurnSubmission,
   parseCodeAction,
   parseCodeCommit,
   parseCodePush,
@@ -1860,12 +1861,20 @@ export class ApiClient {
     return requireParsed(parseCodeTurnList(body), "code session turns");
   }
 
+  /**
+   * Submit a turn, or park it behind the one in flight.
+   *
+   * The route answers 202 either way: a turn snapshot when the session was
+   * idle, a queue receipt when it was busy. Both are accepted work — treating
+   * the receipt as a malformed turn would report a failure for a message the
+   * server holds, and a retry would double-send.
+   */
   async submitCodeTurn(
     sessionId: string,
     message: string,
-  ): Promise<CodeTurnSnapshot> {
+  ): Promise<CodeTurnSubmission> {
     return requireParsed(
-      parseCodeTurn(
+      parseCodeTurnSubmission(
         await this.json(
           `/code/sessions/${encodeURIComponent(sessionId)}/turns`,
           {
--- a/crates/tidebreak-desktop/ui/src/api/types.ts
+++ b/crates/tidebreak-desktop/ui/src/api/types.ts
@@ -114,6 +114,7 @@ import {
   type CodeSessionSnapshot as WireCodeSessionSnapshot,
   type CodeTurnId as WireCodeTurnId,
   type CodeTurnSnapshot as WireCodeTurnSnapshot,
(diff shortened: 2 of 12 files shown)


naingthet (commented):

Composer now sends mid-turn (Queue + Interrupt), parses the 202 queued receipt, and states `queue_full` in product terms. Draft clears on send so a follow-up can be written; a refusal puts it back. `PermissionModeSummary` is the right call — mode is launch-time, not live. Registry fills a prompt for a socket-announced turn. Tests cover queue, 409, parsers, and the prompt fill.

Looks good.

naingthet (commented):

The new commit parks mid-session permission-mode change in `docs/deferred.md` with the right reason (launch-time composition, no adapter seam, relaunch semantics first). Queue/Interrupt, the `ran`/`queued` 202 split, draft-clear-on-send, and `queue_full` copy are unchanged and still the right client shape for a route that already worked.

PR #2163, opened 2026-08-17
fix(server): cache harness probes and restore approval channels on recovery

naingthet:
## The correctness bug

Boot recovery ran before the server bound its port, and the loopback base every approval MCP endpoint is built from was published *after* the bind. Recovery therefore re-attached every restored session while that base was still unset, and `approval_channel` returns an `Option` — so each Ask- and Auto-mode session restored at boot came back with **no approval endpoint at all**. The engine could never ask, and nothing in the journal said why.

Publishing the base and running recovery are now a single call, `CodeRuntime::start`. It sets the base synchronously and hands the recovery pass back as a future, so the order is no longer something a boot-sequence edit can get wrong: `set_loopback_base` is private, and a session created while recovery is still in flight is wired exactly like a recovered one.

## Getting recovery off the launch path

The boot path now runs that recovery future in a background task after the bind. Recovery relaunches an engine child per restored session, so launch no longer waits on the sum of them, and the per-session re-attach inside recovery is concurrent rather than serial.

The trade: for a short window a restored session is listed before its worker is attached, and a turn submitted into it is refused with `session_worker_missing`. Previously that same wait was spent with the port closed and the app unusable, so this is a strictly earlier-usable boot, not a new stall.

## The probe cache decision 0034 already specified

`docs/decisions/0034-harness-discovery-credentials.md` says the probe "is bounded … and cached; re-probe is on demand from the doctor surface". It was not cached. A cold probe spawns an interactive login shell (`$SHELL -ilc`) plus a version and an authentication subprocess — measured at roughly 1.5s per harness — and `GET /code/harnesses` walked all four serially, about 7s per call, on every code-mode navigation. `refresh_harnesses` was byte-identical to `list_harnesses`, so the refresh button meant nothing.

- Probes are memoized per `HarnessKind` on `CodeRuntime`.
- The doctor serves the cache, and probes concurrently on a miss (~7s to ~2s cold, instant warm).
- `POST /code/harnesses/refresh` is now the invalidate-and-re-probe path — the on-demand re-probe 0034 describes, and how a harness installed or signed into mid-session becomes visible.
- A session start probes at most once: `create_session` and `attach_and_spawn_worker` previously probed the same harness back to back and now read the same memoized result.

**Population policy:** lazily, not warmed at boot. Recovery already probes the kinds that have live sessions on its way to re-attaching them, so those fill for free and off the launch path; warming the rest would spend four login shells on harnesses a launch may never touch. One consequence worth naming: the probe also carries the shell environment harness children run under, so that snapshot is now captured once per refresh rather than once per launch — which is what 0034 asks for, but it does mean a profile edit needs a refresh to take effect.

## Tests

- `a_recovered_session_keeps_its_approval_channel` — the reproduction. It asserts a recovered session's worker was launched with an approval endpoint minted from the bound address; inverting the publish/recover order makes it fail with `[None]`.
- `the_doctor_caches_probes_and_refresh_re_probes` — a second doctor read must not re-probe, and refresh must. This replaces `doctor_lists_the_scripted_adapter`, whose assertions it keeps, rather than adding a test alongside it.
- The scripted adapter gained a probe counter and a record of the approval channel each launch was handed; both are test-only observation, no behavior.

Concurrency itself is not tested — it has no observable contract beyond being faster.

## Verified

`cargo fmt --check`, `cargo clippy -p tidebreak-server --all-targets -D warnings`, `cargo test -p tidebreak-server --locked --lib code::` (76 pass) and the `listener` integration test. Not verified by driving the app: the launch-latency and approval-channel improvements are argued from the code paths and the per-probe timings above, not from a running desktop build.

### DIFF
--- a/crates/tidebreak-server/src/code/runtime.rs
+++ b/crates/tidebreak-server/src/code/runtime.rs
@@ -22,7 +22,7 @@ use tidebreak_core::{
 };
 use tidebreak_harness::{
     builtin_registry, AdapterRegistry, ApprovalChannelSpec, ApprovalDecision, HarnessAdapter,
-    HarnessEvent, HarnessEventSink, HostEnv, SessionSpec,
+    HarnessEvent, HarnessEventSink, HarnessProbe, HostEnv, SessionSpec,
 };
 
 use super::approval_bridge::ApprovalBridge;
@@ -63,6 +63,8 @@ pub(crate) struct CodeRuntime {
     pub approvals: Arc<ApprovalBridge>,
     host: HostEnv,
     loopback_base: Mutex<Option<String>>,
+    /// Memoized harness probes, one per kind. See [`CodeRuntime::probe`].
+    probes: Mutex<HashMap<HarnessKind, HarnessProbe>>,
     workers: Mutex<HashMap<CodeSessionId, WorkerHandle>>,
     pr_cache: PrDigestCache,
     pub(crate) clone_jobs: CloneJobs,
@@ -88,6 +90,7 @@ impl CodeRuntime {
             approvals: ApprovalBridge::new(),
             host: HostEnv::from_process(),
             loopback_base: Mutex::new(None),
+            probes: Mutex::new(HashMap::new()),
             workers: Mutex::new(HashMap::new()),
             pr_cache: PrDigestCache::default(),
             clone_jobs: CloneJobs::default(),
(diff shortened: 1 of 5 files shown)


naingthet (commented):

`CodeRuntime::start` publishes the loopback base then returns recovery as a future, so Ask/Auto sessions cannot restore with `approval_channel = None`. Recovery after bind, concurrent, off the launch path. Probe cache matches 0034: doctor is a memoized read, refresh invalidates, create+attach share one probe. Reproductions invert publish/recover and pin cache vs refresh.

Looks good.

PR #2164, opened 2026-08-17
fix(server): persist and surface unrecognized harness events

naingthet:
Decision [0031](docs/decisions/0031-harness-adapter-boundary.md) makes
unrecognized-event counting the honesty mechanism for harness protocol
drift: an unknown newer CLI "runs in best-effort mode with unrecognized-event
*counting* — a counter surfaced in the session UI and the harness doctor
page — never silent dropping."

The counting half shipped; the surfacing half never connected. All four
parsers increment a counter, but nothing called `unrecognized()`.
`code_session.unrecognized_event_count` was `Set(0)` at creation and never
written again, so the doctor summed zeros and the badge in `DoctorList`
could not appear under any circumstances. A build that counts drops and
then discards the count is indistinguishable from one that drops silently,
which is precisely the failure the record exists to prevent.

## What changed

**The count becomes durable.** `HarnessSession` gains
`unrecognized_events() -> u64`, the running total since launch. It has no
default: an adapter that dropped part of a stream has to answer, the same
reason `HarnessCaps` has no `Default`. Claude and Grok build a fresh parser
per turn (one child per turn), so they accumulate into an atomic; Codex and
opencode hold one parser for the session and report it directly.

The session worker flushes the delta onto the session row at the end of a
turn, next to where it already records the child pid and resume ref. The
watermark lives on the per-attachment `LiveSink`, so the row accumulates
across engine restarts rather than being overwritten by whatever the newest
turn saw.

**Two silent drops now count.** An `end` frame with an unrecognized
`stopReason` (Grok) and a `turn/completed` with an unrecognized status
(Codex) were folded into `TurnCompleted`. Both manifests record the exact
observed sets — `completed | interrupted | failed` for Codex — so a fourth
value is real drift. The turn still has to close, so it still completes,
but the event is counted and a warning notice is emitted first.

**One drop stays a drop, deliberately.** Grok's `tool_call_update` with no
`status` is not drift: `fixtures/grok/1.0.4/manifest.toml` records the
observed set as `null (progress) | completed | failed`. It carries nothing
we normalize, so it is now an explicitly commented no-op rather than an
unexplained early return.

**Surfacing.** The doctor endpoint already summed the column per harness and
needed no change — it just had nothing but zeros to sum. The workspace
header gains a small warning badge when a session's count is nonzero.
`CodeSessionSnapshot` already carried `unrecognized_event_count` over the
wire, so no wire type or generated binding changed.

## Tests

- `unread_engine_events_accumulate_on_the_session_row_and_reach_the_doctor`
  drives two real turns against a scripted engine that reports unmapped
  events, then asserts the count on the session snapshot and in the doctor
  report. Two turns rather than one because "increment" vs "overwrite" is
  the mistake worth pinning.
- One parser test each for the stop-status and turn-status paths, asserting
  the notice precedes the terminal event. The Grok test covers the progress
  frame in the same stream, so a future change that folds it back into the
  unrecognized path fails here.

No existing fixture expectations moved: every captured stream uses a known
stop reason and status.

## Follow-ups

- 0031 also asks for raw unrecognized payloads in a size-capped per-session
  debug log for later fixture capture. They currently go to `tracing::debug`
  with a 512-byte cap, which is not per-session and not retrievable from the
  product. A real debug log is more than a small addition and is left out
  here; the counter visibility is the point of this change.
- Grok counts `available_commands` as unrecognized (known frame, no mapping),
  so a healthy Grok session shows a nonzero count. That is defensible — it is
  a real drop — but it makes the badge less of a signal for that harness.
  Worth deciding separately whether "known but unmapped" deserves its own
  count.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeWorkspacePage.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeWorkspacePage.tsx
@@ -216,6 +216,7 @@ function CodeWorkspaceBody({ workspaceId }: { workspaceId: string }) {
                   client={client}
                 />
                 <PendingApprovalBadge sessionId={session.id} client={client} />
+                <UnrecognizedEventsBadge session={session} />
                 <SessionLifecycleBadge
                   session={session}
                   client={client}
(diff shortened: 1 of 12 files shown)


naingthet (commented):

`unrecognized_events()` has no default, the worker flushes a delta onto the session row (watermark on `LiveSink` so restarts accumulate), and the workspace badge plus the existing doctor sum finally have a nonzero to show. Unknown Codex/Grok stop statuses count and warn before completing. Two-turn integration test pins increment vs overwrite.

Looks good.

PR #2165, opened 2026-08-17
fix(server): record child pid after spawn and journal honest turn outcomes

naingthet:
Three defects in the code-mode turn lifecycle, found reviewing what happens when a turn ends any way other than the happy path. They compound: the first hides a live engine from crash recovery, the second and third make a stopped or dead engine look like a successful turn.

## The orphan fence never covered per-turn-child adapters

The worker read `engine.child_pid()` *before* `run_turn` was constructed and cleared it after the turn returned. Claude and Grok spawn their child inside `run_turn`, so `code_session.child_pid` was NULL for the whole window a turn was in flight — precisely the window a crash can orphan a child in. Boot recovery then takes its "no recorded pid, treat as dead" branch and re-attaches a fresh worker to a worktree the still-running child is writing to, which is the failure the fence in ADR 0032 exists to prevent.

`HarnessSession` gains `child_pid_changes()`, a watch of the pid the adapter publishes the moment the child exists and clears when it exits. The worker selects on it alongside the turn and persists each transition to the session row. Adapters with a long-lived child (Codex, opencode) are unaffected — their pid is already stable across the session, and the default implementation returns `None`.

## Interrupts and dead children were journaled as completed turns

`run_turn` returned `Ok(())` at EOF and threw the child's exit status away. A SIGKILLed child never prints the `aborted_streaming` marker the Claude parser recognises, so the worker's fallback wrote `TurnCompleted` with default usage whenever the turn was still open. Stop produced a "completed, zero tokens" turn; an OOM kill, a panic, or an expired login produced exactly the same thing, with captured stderr going only to a `warn!`.

`run_turn` now returns a `TurnOutcome` describing how the process ended: `Clean`, or `Incomplete` with a bounded detail (exit code or signal, plus the tail of captured stderr). The worker decides the turn's fate, since it is the side that knows whether the stop was asked for — interrupt issued: interrupted; child died on its own: failed, with the stderr detail journaled as a `HarnessNotice` so the reason reaches the session and the UI rather than a log line. Grok's synthetic `TurnInterrupted` on a stream that just stopped is removed; that state is now reported through the outcome and classified honestly instead of always reading as "the user stopped it".

## Control commands starved stdout

`apply_control(...).await` ran inline inside the select arm, so nothing drained the child's stdout while an interrupt's two-second grace period ran. A chatty engine could not flush, the grace always expired, we escalated to SIGKILL, and the result line that would have said "aborted" was destroyed — which fed defect two. Control commands now run concurrently with the turn, and any still in flight are drained before the turn closes so their caller still gets a reply instead of a dropped channel.

## Tests

- `a_pid_the_worker_recorded_mid_turn_fences_recovery` drives a real worker turn and reads back the pid **the worker wrote**, then fences on it. The existing recovery tests hand-seed a pid into the row, so they pass with the fence completely broken.
- `an_engine_that_dies_without_saying_so_journals_an_interrupted_turn` — end to end over the routes, with a scripted engine that dies silently the way a killed child does; asserts the turn ends `interrupted`, not `completed`.
- An adapter test against a fake engine that sleeps, writes to stderr, and exits 3: the exit status and stderr have to leave the adapter, and the pid has to be readable while the turn runs.
- `interrupt_stops_a_running_scripted_turn` now asserts the turn's final status instead of only the two HTTP codes.
- Two small unit tests on the exit classifier (EOF without a result is not a completed turn; a failed exit carries its stderr).

## Notes

- `code/runtime.rs` is untouched.
- `docs/code-mode.md`'s trait sketch is updated to match the trait (it had also drifted to `&mut self` receivers).
- Verified with `cargo fmt --check`, `cargo clippy -p tidebreak-server -p tidebreak-harness --all-targets -- -D warnings`, `cargo check --workspace --all-targets --locked` (no lockfile movement), the full `tidebreak-harness` suite, and the `tidebreak-server` code-mode tests. Not verified against a real engine binary: the per-turn-child paths are exercised through a fake engine script and the scripted adapter, not Claude Code or Grok themselves.

### DIFF
--- a/crates/tidebreak-harness/src/child.rs
+++ b/crates/tidebreak-harness/src/child.rs
@@ -1,0 +1,157 @@
+//! Bookkeeping for adapters that run one engine child per turn.
+//!
+//! Two facts the session worker needs, and a pid read at turn boundaries
+//! cannot give it: the pid *while* the turn is in flight (that is the whole
+//! window a crash can orphan a child in), and how the child ended once it is
+//! gone (an EOF on stdout is not a completed turn).
+
+use std::process::ExitStatus;
+
+use tokio::sync::watch;
+
+use crate::TurnOutcome;
+
+/// Live pid of the child backing the current turn.
+///
+/// Every transition is published, so a watcher can persist the pid the moment
+/// the child exists rather than discovering it after the turn is over.
+#[derive(Debug)]
+pub struct ChildPid {
+    tx: watch::Sender<Option<i64>>,
+}
+
+impl Default for ChildPid {
+    fn default() -> Self {
+        Self::new()
+    }
+}
+
+impl ChildPid {
+    /// A cell with no child recorded.
+    #[must_use]
+    pub fn new() -> Self {
+        Self {
+            tx: watch::Sender::new(None),
+        }
+    }
+
+    /// Publish the pid of a child that now exists. An absent or zero pid is
+    /// published as "no child": recovery must never probe an invented pid.
+    pub fn set(&self, pid: Option<u32>) {
+        let pid = pid.filter(|pid| *pid != 0).map(i64::from);
+        self.tx.send_replace(pid);
+    }
+
+    /// Publish that no child is running.
+    pub fn clear(&self) {
+        self.tx.send_replace(None);
+    }
+
+    /// The pid recorded right now, if any.
+    #[must_use]
+    pub fn get(&self) -> Option<i64> {
+        *self.tx.borrow()
+    }
+
+    /// Watch every transition, starting from the current value.
+    #[must_use]
+    pub fn subscribe(&self) -> watch::Receiver<Option<i64>> {
+        self.tx.subscribe()
+    }
+}
+
+/// Ask a child this session spawned to stop.
+///
+/// SIGINT only: the caller decides whether to escalate after its own grace
+/// period, and only ever against a pid recorded at spawn.
+pub fn signal_interrupt(pid: i64) {
+    #[cfg(unix)]
+    {
+        // SAFETY: the pid was recorded from a child we spawned this session.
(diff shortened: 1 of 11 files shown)


naingthet (commented):

The three defects actually compound the way the write-up says: a NULL mid-turn pid is the ADR 0032 hole, EOF-as-success is how Stop/OOM/login-expiry all journaled as completed, and inline `apply_control` is why the abort marker never made it out. `ChildPid` published before the first await, `TurnOutcome` classified by the worker (asked-for stop vs child died), and controls running beside the turn are the right seams.

`a_pid_the_worker_recorded_mid_turn_fences_recovery` is the test that would have caught the fence being completely broken — the existing recovery tests seed the row and would still pass.

Branch is dirty against main; rebase before merge.

naingthet:
Rebased onto main over #2164 (unrecognized-event persistence), which touched the same end-of-turn path. Both behaviors are kept: the per-turn parser count still accumulates on the adapters and the worker still flushes the delta onto the session row after each turn, now next to the pid/resume writes in the restructured turn loop. The one place the two changes genuinely disagreed was Grok's synthetic `TurnInterrupted` on a stream that stopped without a terminal event — that is now reported through `TurnOutcome` and classified by the worker (interrupted only if we asked for it, otherwise failed with the child's exit and stderr), rather than always reading as a user stop. The scripted adapter carries both knobs (`with_unrecognized_per_turn`, `with_silent_interrupt`), and #2164's accumulation test passes unchanged.

naingthet (commented):

Rebase onto main is clean: `unrecognized_events` from #2164 still increments at the start of `run_turn`, and `child_pid_changes` sits beside it. Same three seams as before — mid-turn pid watch, `TurnOutcome` classified by the worker, controls beside the turn — and mergeable is no longer dirty.

PR #2166, opened 2026-08-17
fix(server): re-start codex threads that never ran a turn, fence lost resumes

### LINKED ISSUE #2151: Codex session wedges permanently when the app-server dies before its thread is persisted
Found while driving each harness through the code-mode surface.

**Repro** (CLI, but the desktop hits the same path):

1. `tidebreak code session start --ws <id> --harness codex --mode auto` in an embedding CLI process (or create a session in the desktop and quit before the first turn completes).
2. The adapter completes `initialize` + `thread/start` and persists the returned thread id as `harness_resume_ref`; the app-server child dies with the process.
3. Any later turn re-attaches with `thread/resume {threadId}` and codex 0.147.0 answers **`thread not found`** — it does not persist a thread that never ran a turn.
4. Every subsequent turn fails the same way. The session stays `idle`, is never fenced, and reap is never suggested — the workspace is quietly unusable with that session active (and `session_exists` blocks starting another).

**Observed**

```
tidebreak: turn ca62b637-…
tidebreak: turn failed: thread not found: 01a01121-930b-7490-b4c1-6f465ac820c7
```

Confirmed against a persistent `tidebreak serve`: a fresh codex session created and driven under one server process works fine, so this is specifically resume-before-first-completed-turn.

**Possible directions**

- Don't persist `harness_resume_ref` from `thread/start` alone; record it once a turn has actually run (codex persists the rollout then), so a premature death re-attaches with a clean `thread/start`.
- Or: when `thread/resume` reports thread-not-found, treat it like a lost resume — fence the session with a reason the UI can explain so reap is offered, instead of failing every turn forever.

The first option loses nothing (a thread with no turns has no history to lose); the second is the honest fallback when the ref is stale for any other reason (e.g. `~/.codex` cleaned). They compose.

naingthet:
A codex session whose `app-server` child died before its first turn completed
was left holding a thread id codex never persisted. Codex 0.147.0 does not write
a rollout for a thread that ran no turn, so every later `thread/resume` answered
`thread not found`, every turn failed identically, the session stayed `idle`,
and nothing offered a reap — with `session_exists` blocking a replacement, the
workspace was quietly unusable.

Both directions from the issue, which compose:

**Report a resume ref only once it would resume.** The codex adapter keeps the
thread id it needs for `turn/start`, but `HarnessSession::resume_ref` — the
value the server persists as `harness_resume_ref` — reports it only after the
engine has acknowledged a turn on that thread. A session whose engine dies
before its first turn therefore re-attaches with a clean `thread/start`; a
thread with no turns has no history to lose. A ref handed in at launch is
already persisted state and stays reported as is. The trait doc now states the
rule for every adapter.

**A refused resume is a lost resume, not a turn failure.** A JSON-RPC error
saying `thread not found`, on the resume handshake or on a turn, becomes
`HarnessError::ResumeLost`. The server fences the session with
`FenceReason::ResumeLost { detail }` instead of failing every turn forever, so
the workspace shows an explanation and the reap button. The fence drops the dead
ref, which is what makes the reap useful: it re-attaches with a fresh engine
session rather than resuming the same dead id. This is the honest fallback for
any stale ref, not only the premature-death case — a cleaned `~/.codex` lands
here too.

opencode has the same hazard in a milder form: it creates the session with
`POST /session`, which the server persists immediately, so a thread with no
turns still resumes — but a ref whose storage is gone answers `GET /session/{id}`
with 404 and wedged the same way. That 404 now maps to the same `ResumeLost`
path, which was a one-line change on the mechanism already being added.

Fencing gained a runtime counterpart (`recovery::fence_session`) beside the
boot-time one, and `FenceReason` gained a variant, so the desktop parser and
fence-reason label were extended and the generated wire types regenerated.

Tests: a codex adapter test drives a stand-in `app-server` and asserts that a
session which ran no turn reports no resume ref and put only `thread/start` on
the wire, and that it reports the thread id once a turn has run; a second
asserts a stale ref answers with a lost resume rather than attaching a session
that cannot run; a server test asserts a lost resume fences the session with the
explainable reason, drops the ref, and refuses the next turn with the reap the
UI offers, instead of leaving it idle and failing.

Not verified by running the desktop app against a real codex install; the
protocol shapes in the stand-in engine come from the checked-in
`fixtures/codex/0.147.0/` captures.

Closes #2151

### DIFF
--- a/crates/tidebreak-harness/src/codex/session.rs
+++ b/crates/tidebreak-harness/src/codex/session.rs
@@ -2,7 +2,7 @@
 
 use std::collections::HashMap;
 use std::process::Stdio;
-use std::sync::atomic::{AtomicI64, AtomicU32, Ordering};
+use std::sync::atomic::{AtomicBool, AtomicI64, AtomicU32, Ordering};
 use std::sync::{Arc, Mutex};
 use std::time::Duration;
 
@@ -20,7 +20,7 @@ use crate::{
     filter_child_env, ApprovalDecision, HarnessApprovalRef, HarnessError, HarnessEvent,
     HarnessSession, SessionSpec, StreamBudget, StreamLineBuffer, TurnInput, TurnOutcome,
 };
-use tidebreak_core::CodePermissionMode;
+use tidebreak_core::{CodePermissionMode, MAX_NOTICE_CHARS};
 
 const HANDSHAKE_TIMEOUT: Duration = Duration::from_secs(20);
 const MAX_STDERR_BYTES: usize = 64 * 1_024;
@@ -29,6 +29,13 @@ const MAX_STDERR_BYTES: usize = 64 * 1_024;
 pub struct CodexSession {
     spec: SessionSpec,
     resume_ref: Mutex<Option<String>>,
+    /// Whether a turn has actually run on this thread. Codex only writes the
+    /// thread's rollout once a turn starts, so a thread id from
+    /// `thread/start` alone is not resumable and must not be handed out as a
+    /// resume ref — see [`HarnessSession::resume_ref`] below.
+    thread_ran_a_turn: AtomicBool,
+    /// Detail from an engine error saying the resumed thread is gone.
+    resume_lost: Mutex<Option<String>>,
     child: AsyncMutex<Option<Child>>,
     child_pid: AtomicU32,
     stdin: Option<Arc<AsyncMutex<ChildStdin>>>,
@@ -50,6 +57,8 @@ impl CodexSession {
         Self {
             spec,
             resume_ref: Mutex::new(resume_ref),
+            thread_ran_a_turn: AtomicBool::new(false),
+            resume_lost: Mutex::new(None),
             child: AsyncMutex::new(None),
             child_pid: AtomicU32::new(0),
             stdin: None,
@@ -59,6 +68,11 @@ impl CodexSession {
             pending_approvals: Mutex::new(HashMap::new()),
             current_turn_id: Mutex::new(None),
         }
+    }
+
+    /// The detail of a lost resume observed on the stream, when any.
+    fn lost_resume(&self) -> Option<String> {
+        self.resume_lost.lock().expect("codex resume lost").clone()
     }
 
     fn next_rpc_id(&self) -> i64 {
@@ -210,6 +224,12 @@ pub(super) async fn attach(spec: SessionSpec) -> Result<CodexSession, HarnessErr
         };
     let thread_req = session.request(method, params).await?;
     session.read_until_rpc(thread_req).await?;
+    if let Some(detail) = session.lost_resume() {
+        // The stored thread is gone on the engine side. Every turn on this
+        // child would fail identically, so fail the launch with a reason the
+        // caller can act on instead of attaching a session that cannot run.
+        return Err(HarnessError::ResumeLost(detail));
+    }
     Ok(session)
 }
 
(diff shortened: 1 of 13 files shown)


naingthet (commented):

The two directions compose: `resume_ref` stays silent until a turn has actually run (so a first-turn death re-attaches with `thread/start`), and a refused resume is `ResumeLost` — fenced, dead ref dropped, reap starts fresh. opencode's 404 on `GET /session/{id}` riding the same path is the right one-line extension.

`a_lost_resume_fences_the_session_instead_of_failing_every_turn` is the test that would have caught the idle-and-broken wedge. The stand-in app-server matching the 0.147.0 fixture wording is enough without a live Codex binary.

naingthet (commented):
This is good.

The two-part fix matches the actual Codex 0.147.0 hazard: do not persist a thread id until a turn has run (so a dead-before-first-turn session re-attaches with `thread/start`), and treat `thread not found` as `HarnessError::ResumeLost` so the server fences, drops the dead ref, and offers a reap instead of failing every later turn identically. OpenCode 404 on `GET /session/{id}` on the same path is the right generalization. Trait doc, `FenceReason::ResumeLost`, UI label/parser, and generated wire types all move together.

The unix fake `app-server` test is the right shape for the wire rule; the scripted-harness server test covers fence + dropped ref + next-turn `session_fenced`.

Not blocking: the mid-turn `ResumeLost` path fences and then returns `WorkerError::Failed`, so the discovering turn is a 500. Attach-time is a proper `session_resume_lost` conflict. Session state is correct either way.

Did not run the desktop app against a real Codex install; agreeing with the PR that the fixture-backed stand-in is the verification that exists.

crates/tidebreak-server/src/code/session_worker.rs
```diff
@@ -569,6 +569,21 @@ async fn drive_turn(
                 .await;
             }
             if !session_was_ended(db, session).await {
+                if let HarnessError::ResumeLost(detail) = &err {
+                    // The engine has lost this session: every later turn would
+                    // fail the same way. Fence it so the user is offered a
+                    // reap instead of a session that is idle and broken.
+                    let _ = super::recovery::fence_session(
```
naingthet, line 576: **[suggestion]** This is the first fence that can land on a session the UI is already showing. `fence_session` persists `lifecycle: fenced` and emits a digest, but `CodeWorkspacePage` computes the Reap banner and composer `disabled` flag from the loaded session snapshot (`session.lifecycle` / `session.fence_reason`), not from `digest.lifecycle`. The digest has no `fence_reason`. After a mid-turn `ResumeLost` the badge can flip to Fenced while the banner stays hidden and the composer stays enabled until the user reloads or navigates away. Boot-time recover still works, because the page's initial `GET` sees the fenced row.

**Suggestion:** Drive the banner and composer disablement from `digest.lifecycle === "fenced"` (and refetch the session for `fence_reason` / the Reap copy), or push `fence_reason` on the digest. The launch-time path in `runtime.rs` has the same live-update gap if `reap` itself hits `ResumeLost`.

crates/tidebreak-harness/src/codex/session.rs
```diff
@@ -443,7 +475,17 @@ impl HarnessSession for CodexSession {
     }
 
     fn resume_ref(&self) -> Option<String> {
-        self.resume_ref.lock().expect("codex resume").clone()
+        // Codex 0.147.0 does not persist a thread that never ran a turn:
```
naingthet, line 478: **[suggestion]** The eight-line comment on `resume_ref()` restates the Codex 0.147.0 rollout story and the change's before/after. The non-obvious part is only why a launch-time ref is still reported when no turn has run.

**Suggestion:** Keep one sentence: report the mutex only after a turn has been acknowledged; otherwise return the spec's already-persisted ref.

naingthet (commented):
## Summary

The change matches the stated Codex wedge: a thread id from `thread/start` is no longer reported as a resume ref until a turn is acknowledged, and a later `thread not found` / opencode 404 becomes `HarnessError::ResumeLost` so the server can fence and drop the dead ref. The attach-time Codex test and the server fence-then-conflict test cover the two halves of that contract, and reap already re-attaches with whatever `harness_resume_ref` the fence left behind.

Posted after merge (this PR is already on main). No correctness bugs. Two non-blocking suggestions.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 0

PR #2168, opened 2026-08-17
feat(code): map Allow all onto each harness permission mode

naingthet:
## Summary

Code mode only offered Plan, Ask, and Auto. Chat already has Allow, and every shipped harness already has an allow-everything posture. This adds `Allow` as a first-class session mode and maps it explicitly per engine.

- Claude Code → `--dangerously-skip-permissions` (plus `--allow-dangerously-skip-permissions` so print mode honors it)
- Codex CLI → `sandbox=danger-full-access`, `approvalPolicy=never`
- opencode → `build` with every permission rule `allow`
- Grok CLI → `--always-approve` (Plan and Ask stay refused)

`allow_mode` is its own capability flag, same as Auto. The bypass denylist still rejects those flags on Plan / Ask / Auto, including extra argv. Allow is the explicit per-session choice reserved by decision 0033, stated where the mode is chosen. Schema epoch 29 widens the `code_session.permission_mode` check constraint.

See [decision 0039](docs/decisions/0039-allow-is-a-first-class-code-permission-mode.md).

## Test plan

- [x] Adapter mapping tests: each Allow plan composes the documented native setting; Plan / Ask / Auto still contain no bypass flag
- [x] Server refuses Allow when `allow_mode` is unsupported even if `auto_mode` is supported
- [x] CLI default-mode test still defaults to Ask / Plan / Auto, never Allow
- [x] Clippy on the touched crates
- [ ] UI vitest (node_modules not installed in this worktree)
- [ ] Drive a session in Allow on Claude Code and confirm the skip flags land and no approval cards appear

### DIFF
--- a/crates/tidebreak-cli/src/code.rs
+++ b/crates/tidebreak-cli/src/code.rs
@@ -48,7 +48,7 @@ usage: tidebreak code doctor [--refresh]
        tidebreak code ws list [--repo <id|path>]
        tidebreak code ws show <id>
        tidebreak code ws archive <id> [--force]
-       tidebreak code session start --ws <id> --harness <kind> [--mode plan|ask|auto]
+       tidebreak code session start --ws <id> --harness <kind> [--mode plan|ask|auto|allow]
        tidebreak code session show <id>
        tidebreak code session reap <id>
        tidebreak code run (--session <id> | --ws <id>) [<message>]
@@ -799,6 +799,9 @@ async fn resolve_start_mode(
         }
         CodePermissionMode::Auto => Some(
             "starting in auto mode — this engine has no approval channel; every action proceeds without asking",
+        ),
+        CodePermissionMode::Allow => Some(
+            "starting in allow mode — this engine's permission system is off; every action runs without asking",
         ),
         CodePermissionMode::Ask => None,
     };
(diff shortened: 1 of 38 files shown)


crates/tidebreak-cli/src/code.rs
```diff
@@ -800,6 +800,9 @@ async fn resolve_start_mode(
         CodePermissionMode::Auto => Some(
             "starting in auto mode — this engine has no approval channel; every action proceeds without asking",
         ),
+        CodePermissionMode::Allow => Some(
```
naingthet, line 803: **[suggestion]** `resolve_start_mode` adds an Allow warning, but an explicit `--mode allow` returns at the top of the function with `note: None`, and `default_create_permission_mode` never selects Allow. The new arm is unreachable. Decision 0039 requires the “permission system is off” statement wherever Allow is offered; desktop create surfaces do this, the CLI does not on the only path that can choose Allow.

**Suggestion:** Emit the Allow note for an explicit `--mode allow` (or drop the dead arm). The implicit-default match is the wrong place for a mode that is never a default.

naingthet (commented):
## Summary

Allow is added as a first-class `CodePermissionMode`, gated on its own `allow_mode` cap, and mapped onto each harness’s documented bypass posture without changing existing Plan/Ask/Auto flags. Schema, wire, UI picker, and server refusal all stay in lockstep, and Allow is never a composed default.

Two leftovers: the CLI warning cannot fire, and `docs/code-mode.md` still says Grok honors Auto only.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 1

## Issues outside the diff

- **[nit]** docs/code-mode.md:200 — The intro now says Grok honors Auto and Allow, but the process-model bullet still says it “honors Auto only.” Update that bullet to match the intro and decision 0039.

PR #2169, opened 2026-08-17
fix(server): keep Ended from being overwritten by the outgoing worker

naingthet:
Follow-up to #2160. That PR fenced `save_session` on `spawn_epoch` so a superseded worker could not regress the row, and it waited for the outgoing worker during archive/reap. The wait is still same-epoch: archive never bumps `spawn_epoch`, so the worker that is shutting down can persist `Running` or `Idle` (and a pid) over `Ended` while `end_workspace_sessions` waits. The archived workspace then looks live.

`save_session` now treats `Ended` as terminal. A persist that is not itself `Ended` cannot overwrite that lifecycle, even at the same epoch. `end_workspace_sessions` reloads the row after the wait and re-asserts `Ended` (clearing pid and fence) so a late write cannot stick. Re-asserting `Ended` at the same epoch still succeeds; the epoch fence is unchanged.

`an_ended_session_cannot_be_revived_by_a_same_epoch_persist` covers the CAS: same-epoch `Running`/`Idle` writes are rejected, the row stays `Ended` with no pid, and a same-epoch `Ended` persist still lands.

Verified locally: `cargo fmt --all -- --check` and `cargo test -p tidebreak-core --lib an_ended_session_cannot_be_revived`.

### DIFF
--- a/crates/tidebreak-core/src/db/ops/code/session.rs
+++ b/crates/tidebreak-core/src/db/ops/code/session.rs
@@ -1,5 +1,6 @@
 use sea_orm::{
-    ActiveModelTrait, ColumnTrait, EntityTrait, QueryFilter, QueryOrder, Set, TransactionTrait,
+    ActiveModelTrait, ColumnTrait, Condition, EntityTrait, QueryFilter, QueryOrder, Set,
+    TransactionTrait,
 };
 
 use crate::code::{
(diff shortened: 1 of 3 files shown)


naingthet (commented):
## Summary

This follow-up does what #2160 left open. `save_session` still requires a non-decreasing `spawn_epoch`, and a non-`Ended` caller can no longer overwrite a stored `Ended` row; `end_workspace_sessions` reloads after `shut_down_worker` and writes `Ended` again, failing the archive with `session_not_ended` if that write does not land. The store test covers same-epoch Running/Idle revival, pid restoration, and the post-wait re-assert.

No correctness issues in this diff. The red `test` lane is an unrelated harness flake (`codex::session::tests::a_stale_resume_ref_reports_a_lost_resume`), not this change.

Looks good to merge once `test` is green.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

PR #2170, opened 2026-08-17
ci: do not invoke Release Drafter without a published baseline

naingthet:
## Summary

#2167 cleaned up after Release Drafter already created a `v0.0.1` draft. This stops that draft from being created.

The job now lists GitHub Releases and `v*` tags (retrying a few times) and refuses to invoke Release Drafter when version tags exist but no published release is visible. That is the outage that produced the extra draft. The existing post-step still deletes a fallback if Release Drafter loses the race after the check.

## Test plan

- [x] `node --test scripts/require-release-baseline.test.mjs scripts/reconcile-release-drafts.test.mjs scripts/workflow-security.test.mjs scripts/workflow-security-mutations.test.mjs`
- [ ] After merge, the next **Release draft** run on `main` should still update the single `v0.47.0` draft

### DIFF
--- a/.github/workflows/release-draft.yml
+++ b/.github/workflows/release-draft.yml
@@ -77,6 +77,43 @@ jobs:
       pull-requests: read
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+      - name: Require a published release baseline
+        env:
+          GH_TOKEN: ${{ github.token }}
+        run: |
+          attempt=1
+          max_attempts=5
+          while (( attempt <= max_attempts )); do
+            if gh api --paginate \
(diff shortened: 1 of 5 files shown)


naingthet (commented):
## Summary

The draft job now refuses to invoke Release Drafter when `v*` version tags exist but the Releases API lists no non-draft release, which is the outage that mints a `v0.0.1` fallback. The planner reuses `normalizeReleaseList` and the same `draft === false` published definition as reconcile (including in-flight prereleases), still allows the empty-history first-release path, and leaves the existing post-drafter cleanup as a second line of defense. Retrying both fetch failures and a missing baseline is appropriate.

No correctness issues. Looks good to merge once required checks are green.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

PR #2171, opened 2026-08-17
feat(desktop): put code-mode git review in a sidebar

naingthet:
## What

Code-mode workspaces now keep the conversation in the center and move git work off it.

- **Review sidebar** (⌘I / Ctrl+I): commit, push, PR create, PR state, and a comments section. Preference is remembered.
- **Terminal drawer** (⌘J / Ctrl+J): the auxiliary shell opens under the conversation instead of as a side-panel tab. Existing `?tabs=terminal` links still open it.
- Files and diff stay in the existing side-panel region.

The comments section is the home for review discussion. There is still no comments API, so a PR links out and an empty workspace says comments appear after one exists.

## Why

Git sync and PR state were sitting on top of the transcript. The conversation should stay readable; review belongs in chrome that can be toggled.

## Tests

- Shortcut table: ⌘I / ⌘J fire only in code mode
- Layout helper: terminal is lifted out of the side region
- Workspace page: PR/comments render in the review rail; Terminal opens a drawer, not a tab

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeInspector.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeInspector.tsx
@@ -1,0 +1,135 @@
+import { GitPullRequest, MessageSquare, X } from "lucide-react";
+
+import type { ApiClient } from "../api/client";
+import type {
+  CodeWorkspaceSnapshot,
+  PullRequestDigest,
+} from "../api/types";
+import { Badge } from "@/components/ui/badge";
+import { Button } from "@/components/ui/button";
+import { WithTooltip } from "@/components/ui/tooltip";
+import { openExternal } from "@/host";
+import { PrCard } from "./PrCard";
+import { useCodeUpdatesStore } from "./CodeUpdatesStore";
+
+/**
+ * Workspace review rail: git sync, pull-request state, and comments.
+ *
+ * The conversation stays the center of the workspace. Commit, push, the PR,
+ * and review discussion live here so they are always one chord away without
+ * sitting on top of the transcript.
+ */
+export function CodeInspector({
+  client,
+  workspaceId,
+  workspace,
+  contentRevision,
+  shortcutHint,
+  onClose,
+}: {
+  client: ApiClient;
+  workspaceId: string;
+  workspace: CodeWorkspaceSnapshot | null;
+  contentRevision: number;
+  shortcutHint?: string;
+  onClose: () => void;
+}) {
+  const digest = useCodeUpdatesStore((state) => state.byWorkspace[workspaceId]);
+  const pr = digest?.pr_state ?? workspace?.pr;
+  const active = workspace?.status !== "archived";
+
+  return (
+    <aside
+      className="flex w-80 shrink-0 flex-col overflow-hidden border-l"
+      aria-label="Review"
+      data-testid="code-inspector"
+    >
+      <header className="flex h-10 shrink-0 items-center gap-2 border-b px-3">
+        <GitPullRequest className="text-muted-foreground size-3.5 shrink-0" />
+        <h2 className="min-w-0 flex-1 truncate text-sm font-medium">Review</h2>
+        <WithTooltip label={shortcutHint ? `Hide sidebar ${shortcutHint}` : "Hide sidebar"}>
+          <Button
+            type="button"
+            variant="ghost"
+            size="icon-xs"
+            aria-label="Hide review sidebar"
+            onClick={onClose}
+          >
+            <X />
+          </Button>
+        </WithTooltip>
+      </header>
+      <div className="flex min-h-0 flex-1 flex-col gap-5 overflow-y-auto px-3 py-3">
+        {active && (
+          <PrCard
+            client={client}
+            workspaceId={workspaceId}
+            contentRevision={contentRevision}
+            framed={false}
+          />
+        )}
+        <ReviewComments pr={pr} />
+      </div>
+    </aside>
+  );
+}
+
+function ReviewComments({ pr }: { pr?: PullRequestDigest }) {
+  return (
+    <section className="flex flex-col gap-2" aria-label="Comments">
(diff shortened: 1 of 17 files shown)


crates/tidebreak-desktop/ui/src/code/CodeWorkspacePage.tsx
```diff
@@ -305,62 +356,91 @@ function CodeWorkspaceBody({ workspaceId }: { workspaceId: string }) {
...
-            {fenced && session?.fence_reason && (
-              <div className="border-warning-border bg-warning-background text-warning-foreground mx-4 mt-3 flex flex-col gap-2 rounded-md border px-3 py-2 text-sm">
-                <p>{fenceReasonText(session.fence_reason)}</p>
-                <Button type="button" size="sm" className="self-start" onClick={() => void reap()}>
-                  Reap
-                </Button>
+      <div className="flex min-h-0 flex-1 overflow-hidden">
+        <div className="flex min-h-0 min-w-0 flex-1 flex-col overflow-hidden">
+          <PanelLayout
+            layout={chrome.panels}
```
naingthet, line 362: **[bug]** `PanelLayout` is given `chrome.panels` (terminal stripped) but still calls `usePanelNav()`'s `focusTab` / `closeTab`, which index the unsplit URL layout. `openPanel` always appends, so Files then Terminal then Diff yields `tabs=[files,terminal,diff]`. The strip shows `[files, diff]`; clicking Diff calls `focusTab(1)` and selects the terminal, and closing Diff calls `closeTab(1)` and removes the terminal. Same mismatch if the drawer is already open and the reader then opens Files or Diff.

**Suggestion:** Drive tab focus/close from the filtered strip and re-merge the terminal into the URL, keep the terminal in its own search bit so it is not a strip index, or make `PanelLayout` accept focus/close handlers that understand the split.

crates/tidebreak-desktop/ui/src/code/codeChrome.ts
```diff
@@ -0,0 +1,69 @@
...
+
+/** Open the terminal drawer, or close it if it is already in the layout. */
+export function toggleTerminalLayout(layout: LayoutState): LayoutState {
+  const index = layout.tabs.findIndex((tab) => tab.type === "terminal");
+  if (index === -1) {
+    return {
+      ...layout,
+      tabs: [...layout.tabs, { type: "terminal" }],
+      activeIndex: layout.tabs.length,
+    };
```
naingthet, line 52: **[bug]** Opening the drawer via `toggleTerminalLayout` or `openPanel({ type: "terminal" })` sets `activeIndex` to the new terminal tab. `splitCodeChromeLayout` then treats the terminal as the active tab and remaps to `min(terminalIndex, remaining-1)`. With Files and Diff open and Files showing, opening Terminal leaves Diff selected. Closing the drawer does not restore Files.

**Suggestion:** When inserting the terminal, leave `activeIndex` on the still-visible panel. Only remap when the URL actually named the terminal as active (reload / shared link).

crates/tidebreak-desktop/ui/src/code/CodeInspector.tsx
```diff
@@ -0,0 +1,135 @@
...
+            onClick={onClose}
+          >
+            <X />
+          </Button>
+        </WithTooltip>
+      </header>
+      <div className="flex min-h-0 flex-1 flex-col gap-5 overflow-y-auto px-3 py-3">
+        {active && (
+          <PrCard
+            client={client}
```
naingthet, line 65: **[suggestion]** `PrCard` keeps the commit message in component state. Closing the review rail (button or Cmd+I, including while the commit box is focused) unmounts the card and drops the draft; reopening refills the server suggestion.

**Suggestion:** Lift the draft to workspace- or store-scoped state so hiding the rail does not discard typed commit text.

crates/tidebreak-desktop/ui/src/code/CodeInspector.tsx
```diff
@@ -0,0 +1,135 @@
...
+          </div>
+          <p className="text-muted-foreground text-xs leading-relaxed">
+            Review comments live on the pull request. Open it to read and reply.
+          </p>
+          {pr.url && (
+            <Button
+              type="button"
+              size="sm"
+              variant="secondary"
+              className="self-start"
```
naingthet, line 104: **[suggestion]** `pr.url!` is used even though the click path already sits behind `pr.url &&`. Same pattern on `closeTab(chrome.terminal!)` in `CodeWorkspacePage.tsx`.

**Suggestion:** Capture the narrowed value (`const url = pr.url`, `const terminal = chrome.terminal`) and pass that, so the assertion is unnecessary.

naingthet (commented):
## Summary

The review rail and terminal drawer are a sound chrome split: git/PR leave the transcript, the terminal stays URL-addressable, and the new Cmd+I / Cmd+J bindings are scoped to code mode. The implementation still treats the terminal as a tab in the same strip `PanelLayout` indexes, so files/diff tab clicks and closes go to the wrong entry whenever the terminal is not last. Opening the drawer also activates that hidden tab and can switch the visible files/diff panel.

**Do not merge yet** — three bugs in tab indexing / shortcut delivery.

GitHub will not accept `REQUEST_CHANGES` from the PR author; treating this as a blocking comment instead.

## Issue counts by severity

- bugs: 3
- suggestions: 2
- nits: 0

## Issues outside the diff

- **[bug]** crates/tidebreak-desktop/ui/src/code/TerminalPane.tsx:96 — The pane captures `keydown` on the host and `stopPropagation()`s any shell shortcut so xterm does not eat it. That capture runs before the target and stops the event, so the window bubble listener in `useShellShortcuts` never runs. Cmd+J (hide terminal) and Cmd+I (hide review) therefore do nothing while the PTY is focused — the moment the new drawer shortcut is meant to fire.

  **Suggestion:** Do not stop the event from reaching the window listener (handle the shortcut here and `preventDefault`, or register the shell listener in capture on `window`).

PR #2172, opened 2026-08-17
fix(desktop): keep review strip indexes independent of the terminal drawer

naingthet:
Follow-up to #2171. Three bugs that landed with the review sidebar:

1. **Tab strip indexed the unsplit URL layout.** `PanelLayout` received the filtered files/diff tabs but called `focusTab` / `closeTab` against the URL, so Files then Terminal then Diff made clicking or closing Diff select or remove the terminal.
2. **Opening the drawer remapped the visible files/diff tab.** `toggleTerminalLayout` / `openPanel({ type: "terminal" })` made the terminal the active tab; `splitCodeChromeLayout` then remapped that onto another side tab. With Files showing, opening Terminal left Diff selected, and closing the drawer did not restore Files.
3. **Cmd+J / Cmd+I were dead while the PTY was focused.** `TerminalPane` captured the chord and `stopPropagation()`'d it so xterm would not eat it, which also stopped the window bubble listener in `useShellShortcuts`.

The strip now focuses and closes against the filtered tabs and re-merges the terminal into the URL. Opening the drawer leaves the still-visible side tab selected; remapping happens only when the URL actually named the terminal. The shell shortcut listener registers on window capture so a focused PTY cannot swallow the chords.

## Tests

- Strip click/close with the terminal in the middle of the URL tabs
- Opening the terminal drawer keeps the previously visible files/diff tab
- Cmd+J / Cmd+I still fire while the terminal host is focused

### DIFF
--- a/crates/tidebreak-desktop/ui/src/ShellShortcuts.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/ShellShortcuts.dom.test.tsx
@@ -1,0 +1,118 @@
+// @vitest-environment jsdom
+import { cleanup, render, screen } from "@testing-library/react";
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+import {
+  SHELL_SHORTCUTS,
+  useShellShortcuts,
+  usesCommandModifier,
+  type ShellShortcutAction,
+  type ShellShortcutHandlers,
+} from "./ShellShortcuts";
+import { TerminalPane } from "./code/TerminalPane";
+
+vi.mock("@xterm/xterm", () => ({
+  Terminal: class {
+    cols = 80;
+    rows = 24;
+    write(_data: string, cb?: () => void) {
+      cb?.();
+    }
+    loadAddon() {}
+    open() {}
+    dispose() {}
+    onData() {
+      return { dispose() {} };
+    }
+  },
+}));
+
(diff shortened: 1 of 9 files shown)


naingthet (commented):
Follow-up that actually fixes the three review-sidebar regressions.

The strip now talks to `focusCodeChromeTab` / `closeCodeChromeTab` against the filtered files/diff tabs and writes the remapped URL, so Files–Terminal–Diff no longer treats Diff as the drawer. `toggleTerminalLayout` stops promoting the terminal to `activeIndex`, which is what made opening the drawer steal the visible files/diff tab. Cmd+J / Cmd+I move to window capture so a focused PTY cannot swallow them; `resolveShellShortcut` still honors editable/modal context, and TerminalPane still keeps the chord out of xterm.

Tests pin the two layout cases and the capture delivery. No blocking correctness issue.

Hold merge until the desktop UI lane is green (Rust lanes correctly skipped).

naingthet (commented):
## Summary

Follow-up to #2171. Strip clicks/closes now map through the filtered files/diff tabs; opening the drawer no longer remaps the visible side tab; `useShellShortcuts` listens on window capture so Cmd+J / Cmd+I still fire while the PTY is focused.

No correctness issues. Posted after merge.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

PR #2173, opened 2026-08-17
feat(desktop): ship Windows and Linux packages

naingthet:
Prerequisite trusted-policy transition: #2174.

## Summary

- require production desktop releases to include universal macOS, Windows x86_64 NSIS, and Linux x86_64 AppImage and Debian packages
- unpark native Windows CI and release jobs, and add cache-write-free Linux packaging on Ubuntu 22.04
- publish stable GitHub download aliases, immutable hosted artifacts, checksums, recovery inputs, and format-specific signed Linux updater entries
- keep staging and runtime automatic updates macOS-only while documenting unsigned Windows behavior, platform capability limits, and deferred ARM64/update rehearsal work
- record the cross-platform release contract in decision 40

## Why

Tidebreak's desktop shell already has Windows and Linux platform implementations, but production releases currently require users on those systems to build from source. This establishes an all-or-nothing cross-platform package contract without claiming that every optional macOS-native capability is available elsewhere.

The Linux job builds from the validated release tag with compiler and dependency caches in read-only mode. It does not enable pnpm caching or save compiler caches, and the updater private key enters the environment only after both package formats have been built.

## Validation

- `actionlint .github/workflows/ci.yml .github/workflows/release.yml`
- `git diff --check`
- `node --test scripts/*.test.mjs` — 174 tests passed
- PR #2174's transitional trusted policy against this workflow — 35 tests passed
- `pnpm --dir docs-site lint`
- `pnpm --dir docs-site types:check`
- `pnpm --dir docs-site build`
- post-rebase focused release/security tests — 41 tests passed

- `cargo test -p tidebreak-server --lib desktop_schema --locked` — 13 tests passed
- Unix-only server suites remain covered on Unix hosts — 26 targeted tests passed

## Not verified locally

The development host is macOS, so actual Windows/Linux package generation and clean-machine install, launch, deep-link, sidecar, persistence, and uninstall smoke tests require native hosted runners and test machines.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -188,13 +188,14 @@ is the whole rule, and it is coarse on purpose:
   change's scope reports a successful skip. The always-running change detector
   rejects impossible narrower-scope combinations before conditional jobs
   consume them. There is no serial aggregate wrapper after the slowest job.
-- The native Windows lane is **parked**, along with the release workflow's
-  Windows build — see [`docs/deferred.md`](docs/deferred.md). `windows-check`
-  and `prepare_windows`/`build_windows` are each gated behind a literal `false`
-  and never run; the `windows-ci` label currently does nothing. Releases are
-  macOS-only until those literals flip back. Nothing about the Windows boundary
-  has been declared unsupported, so a change that touches Windows-gated code
-  still deserves the same care — it just gets no CI signal right now.
+- The native Windows lane runs automatically for Rust-scoped pushes to `main`,
+  schedules, and manual dispatches. Pull requests opt in with the `windows-ci`
+  label when they touch a Windows-native boundary. It stays non-required because
+  it is the slowest lane; the post-merge and scheduled runs are the backstop.
+- Production releases require macOS, Windows x86_64, and Linux x86_64 packages.
+  Windows ships NSIS; Linux ships AppImage and `.deb`. Windows Authenticode,
+  ARM64 packages, and automatic updates outside macOS remain parked in
+  [`docs/deferred.md`](docs/deferred.md).
 
 **The remaining trap is a PR that ran nothing.** A conflicting PR runs nothing
 but trivial policy checks; rebase it before believing its status. Judge a PR by
--- a/docs-site/content/docs/installation.mdx
+++ b/docs-site/content/docs/installation.mdx
@@ -1,6 +1,6 @@
 ---
 title: Installation
-description: Download Tidebreak for macOS, or build it from source.
+description: Download Tidebreak for macOS, Windows, or Linux, or build it from source.
 ---
 
 ## macOS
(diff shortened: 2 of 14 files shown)


naingthet (commented):
Unparks Windows packaging and adds Linux x86_64 AppImage/deb under an all-or-nothing production contract (ADR 0040). Staging/runtime auto-update stays macOS-only; Windows/Linux installer signing is explicitly unsigned.

**Blocking — CI will not go green on this head as structured.**

The required `release policy` job runs **main’s** `.trusted-workflow-security.test.mjs` against this `release.yml`. Four of those assertions fail because the trusted graph does not know about Linux:

1. Cache warming only allows `macos-release-cargo-registry-v2` / `windows-release-cargo-registry-v1`. `linux-release-cargo-registry-v1` does not match.
2. Windows prepare/build restore-key count (`restorableKeys.length >= 3`) is now false — check whether a Windows cache restore was dropped, not only that Linux was added.
3. `publish.needs` must be exactly `[validate, inspect_hosted, build_macos, build_windows, source_sbom, finalize_release]`. Adding `build_linux` fails the trusted test.
4. GitHub asset attach-before-immutable-publish has the same graph mismatch.

Updating `scripts/workflow-security.test.mjs` in this PR does not move that job; it restores the trusted copy from `main`. Landing Linux means a first change that teaches the trusted tests the new graph (or an agreed two-step), then this workflow.

CodeQL also flags `prepare_linux`: `pnpm install` plus `actions/cache/save` after checkout of `needs.validate.outputs.sha` on `workflow_dispatch` — same cache-poisoning class as the existing macOS/Windows prepare jobs. If those are accepted only because they already match the trusted tests, Linux has to be added the same way, not by weakening isolation.

`windows-check` on PRs is still label-gated (`windows-ci`); it is no longer behind a literal `false`, which is the real unpark.

Do not merge until the trusted policy job is green and the Windows restore-key regression is explained or fixed.

naingthet (commented):
## Summary

Unparks Windows CI and release jobs, adds a credential-free Linux prepare/build pair, and makes a fresh production release all-or-nothing across macOS, Windows x86_64 NSIS, and Linux x86_64 AppImage+`.deb`. The secret/cache split matches the existing macOS/Windows model. Decision 0040, `docs/deferred.md`, and `docs/releases.md` agree that Authenticode, ARM64, and automatic updates outside macOS stay parked.

Two non-blocking suggestions below. CI is currently red: `release policy` (workflow-security / restorable-keys assertions), CodeQL (2 high alerts on this PR), and `test` (`codex::session::tests::a_thread_that_ran_no_turn_is_not_a_resume_ref` — likely the same harness flake seen on unrelated PRs). Do not merge until those are green.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 0

CLAUDE.md
```diff
@@ -188,13 +188,14 @@ is the whole rule, and it is coarse on purpose:
...
   rejects impossible narrower-scope combinations before conditional jobs
   consume them. There is no serial aggregate wrapper after the slowest job.
-- The native Windows lane is **parked**, along with the release workflow's
-  Windows build — see [`docs/deferred.md`](docs/deferred.md). `windows-check`
-  and `prepare_windows`/`build_windows` are each gated behind a literal `false`
-  and never run; the `windows-ci` label currently does nothing. Releases are
-  macOS-only until those literals flip back. Nothing about the Windows boundary
-  has been declared unsupported, so a change that touches Windows-gated code
-  still deserves the same care — it just gets no CI signal right now.
+- The native Windows lane runs automatically for Rust-scoped pushes to `main`,
```
naingthet, line 191: **[suggestion]** `CLAUDE.md` now says Windows CI is live and production releases require all three OS packages, but `AGENTS.md` is unchanged and still says the native Windows lane and release Windows jobs are parked behind `false`. After merge, sessions that load `AGENTS.md` will follow the old parking story and can reintroduce the literals this PR removes.

**Suggestion:** Apply the same Windows-CI / three-platform release paragraph to `AGENTS.md`, or make one file the source of truth and have the other include it.

docs-site/content/docs/installation.mdx
```diff
@@ -24,15 +24,33 @@ shasum -a 256 -c Tidebreak-macos-universal.dmg.sha256
...
+may warn on first run. The release includes a `.sha256` sidecar for independent
+download verification.
+
+## Linux
+
+Choose the portable AppImage or the Debian package. Both are x86_64 builds.
+
+- [Download AppImage](https://github.com/brightwave-inc/tidebreak/releases/latest/download/Tidebreak-linux-x86_64.AppImage)
+- [Download Debian package](https://github.com/brightwave-inc/tidebreak/releases/latest/download/Tidebreak-linux-x86_64.deb)
+
```
naingthet, line 41: **[suggestion]** The new Linux download section tells users to `chmod +x` and launch the AppImage, with no runtime baseline. A Tauri AppImage/`.deb` built on `ubuntu-22.04` still typically needs WebKitGTK 4.1 (and a matching glibc) on the target. Calling the AppImage portable without that caveat will produce a launch failure that looks like a broken download.

**Suggestion:** State the WebKitGTK 4.1 / glibc baseline next to the download commands, and point at the `.sha256` sidecars the release already attaches (the Windows section already mentions them).

naingthet (commented):
Follow-up on 550d0f6: Linux packaging dropped `prepare_linux` and `actions/cache/save`. `build_linux` now needs only `[validate, inspect_hosted]`, rust-cache is `save-if: false`, sccache is `READ_ONLY`, and there is no pnpm cache. That matches the graph #2174 authorizes.

Still do not merge until #2174 lands (the required `release policy` job runs `main`'s trusted tests) and this head's lanes are green. The earlier Windows restore-key / publish.needs issues are the ones #2174 widens.

naingthet (commented):
Rebased onto #2174 so `main`'s trusted policy now recognizes the Windows/Linux release graph; release-policy is green on this head. The follow-up commit gates Unix-only server tests (`checkpoint`/`gh` module tests plus `code_clone`/`code_git`) behind `cfg(unix)` / `cfg(all(test, unix))` so the unparked Windows cargo-check lane can compile `tidebreak-server` lib tests without `PermissionsExt`/`set_mode`.

Linux packaging remains cache-write-free and loads updater material after both package formats. Staging/runtime auto-update stays macOS-only. Decision 40 records the all-or-nothing package contract.

No blocking issue in the policy/test gating. Hold merge until Windows cargo check and workspace test finish on this head.

PR #2174, opened 2026-08-18
test(release): authorize Windows and Linux packages

naingthet:
## Summary

- authorize the current release workflow and the upcoming Windows/Linux release graph in the trusted workflow policy
- allow the Windows signing job to tighten its compiler-cache restore to the exact validated release SHA and version
- require the Linux packaging job to remain cache-write-free before updater signing material is loaded

## Why

The release-policy CI job intentionally runs `main`'s trusted tests against pull-request workflow changes. PR #2173 cannot add Linux to the release graph until the trusted policy recognizes that graph. This is the same prerequisite-policy transition used for universal macOS packaging and immutable releases.

The transition keeps every current assertion active while admitting the stricter follow-up workflow. PR #2173 will tighten the tests to the final Windows/Linux-only shape after this prerequisite lands.

## Validation

- `node --test scripts/workflow-security.test.mjs` — 35 tests passed
- `node --test scripts/workflow-security-mutations.test.mjs` — 43 tests passed
- `git diff --check`

Prerequisite for #2173.

### DIFF
--- a/scripts/workflow-security.test.mjs
+++ b/scripts/workflow-security.test.mjs
@@ -107,6 +107,14 @@ const DESKTOP_SIGNING_JOBS = [
   },
 ];
 
+if (/^  build_linux:/m.test(workflows["release.yml"])) {
+  DESKTOP_SIGNING_JOBS.push({
+    file: "release.yml",
+    name: "build_linux",
+    validate: "Verify and collect Linux artifacts",
+  });
+}
+
 function desktopSigningJobs() {
   return DESKTOP_SIGNING_JOBS.map((spec) => ({
     ...spec,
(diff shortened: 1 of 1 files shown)


naingthet (commented):
Prerequisite trusted-policy update so #2173 can teach `main` a Linux graph. Current assertions stay in force; Linux-only checks are gated on `build_linux` existing.

The authorized Linux shape is stricter than #2173 as filed: no `prepare_linux`, no `actions/cache/save`, `save-if: false` on rust-cache, no pnpm cache, packages built before updater material is loaded. That is the right isolation (and matches the CodeQL cache-poisoning note on #2173). After this lands, #2173 still has to drop `prepare_linux` or the trusted job will fail again.

Windows restore-key assertion now accepts either today's multi-key fallback or a single SHA+version pin. The pin is stricter; the current graph still has to keep `hashFiles('Cargo.lock', …)` on every restorable key.

`publish.needs` / attach `needs` optionally include `build_linux`. `linux-release-cargo-registry-v1` is added to the allowed cargo-registry keys.

No blocking issue in this policy-only change. Hold merge until desktop test / clippy / workspace test finish (Rust-scope jobs are still running). Then #2173 must be rewritten to the authorized graph before it can go green.

naingthet (commented):
## Summary

This is the intended prerequisite-policy transition for #2173: `main`'s trusted workflow-security tests stay green on today's release graph, and they will also accept the upcoming Windows/Linux graph instead of failing the release-policy lane. Current assertions remain in force. Linux checks are gated on `build_linux` existing and lock that job to a cache-write-free shape.

No correctness issues.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

PR #2175, opened 2026-08-18
feat(code): pin harness binaries and rework workspace review

naingthet:
## Summary

Code mode now drives pinned harness installs and the workspace review chrome is no longer a second copy of Files/Diff.

- Install each engine under `{data_dir}/tools/harnesses/{kind}/{version}` at an exact npm version. Probe and launch use that binary; PATH is not the engine. Decision [0041](docs/decisions/0041-pinned-harness-binaries.md) records the choice and supersedes the "no bundling in v1" clause of 0034.
- Persist `CodeSession.model` (desktop schema epoch 30). Create and submit accept an optional model; adapters pass it through (`--model`, Codex/opencode session start).
- Model options come from the live harness catalog, or only model-gateway models when the gateway is in use. The picker shows real names plus source, with ⌘1–9. Workspace create no longer picks a model — the harness default is what opens.
- Composer reuses the chat control. Files, Diff, and the PR live in the inspector; stale Files/Diff URL tabs are not promoted into the conversation strip. PR checks are spelled out (passing/pending/failing) and the PR number links out.
- Claude model listing reads `~/.claude/settings.json` instead of launching the `models` TUI.

## Test plan

- [ ] First code-mode visit installs the pins under the data directory; doctor shows a failed install instead of falling through to PATH
- [ ] Claude 2.1.234 reports Ask, Auto, and Allow as supported
- [ ] Starting a workspace without choosing a model uses the harness default, and that name appears in the composer immediately
- [ ] Changing the model in the picker actually changes the next turn
- [ ] With model-gateway configured, the picker only lists gateway models
- [ ] Files/Diff stay in the inspector; opening the terminal does not create a Files/Diff strip tab
- [ ] An existing PR shows title, status-colored icon, outbound link, and spelled-out checks
- [ ] Pre-v1 local DB is wiped by the epoch 30 bump


naingthet (commented):
Pins each code-mode harness to an exact npm version under `{data_dir}/tools/harnesses/{kind}/{version}` with an `installed.json` marker. Probe prefers that binary when `data_dir` is set and does **not** fall through to PATH (listing must not install or silently drive an unpinned CLI). Create-session and doctor refresh call `ensure_installed`. Tests without a data directory still use the login-shell shim.

Decision 0041 records this and supersedes 0034's no-bundling clause. Capability flags describe the pin (Claude 2.1.234). Session `model` is persisted (desktop schema epoch 30; pre-v1 DBs wipe). Workspace create uses the harness default; the composer picker lists live harness or gateway-only models. Files/Diff/PR move into the inspector.

Install is `npm install pkg@exact` with a version+package marker, not a digest pin like managed Node. That matches the decision's written rule. Windows pinning is explicitly parked.

No blocking issue in the pin/probe contract. Hold merge until clippy, desktop test, UI, and workspace test finish on this head.

naingthet (commented):
Follow-up folds Grok print-mode launch into `PrintLaunch` so `--model` does not trip the 8-arg clippy cap. Probe listing uses `and_then(normalize_model_id)` and a char-array trim. AppShell's mock now stubs `listCodeHarnessModels` so catalog refresh after Cmd+N does not reject.

Pin/probe contract from the previous head is unchanged: managed binary wins, no PATH fallback when `data_dir` is set. Desktop UI already passed on this head.

No blocking issue. Hold merge until clippy and workspace test finish.

naingthet (commented):
## Summary

The pin + epoch-30 pairing is the right pre-v1 shape: `code_session.model` is a baseline column, so the wipe is required. Probe no longer falls through to PATH once `data_dir` is set, which matches 0041.

**Do not merge yet.** Six bugs: npm install still depends on the GUI PATH; doctor refresh swallows pin-install failures; create now defaults to Allow against 0033; Cmd+W is unscoped and preventDefaults the window close; persisted session model is ignored by live Codex/OpenCode children; pin probe can launch with an empty env if shell capture fails.

## Issue counts by severity

- bugs: 6
- suggestions: 4
- nits: 0

crates/tidebreak-harness/src/pin.rs
```diff
@@ -0,0 +1,236 @@
...
+    .map_err(|err| format!("could not write harness install marker: {err}"))?;
+    managed_binary(data_dir, kind).ok_or_else(|| {
+        format!(
+            "npm install {spec} finished but {} was not executable",
+            pin.bin
+        )
+    })
+}
+
+fn find_npm(data_dir: &Path) -> Option<PathBuf> {
```
naingthet: **[bug]** `ensure_installed` finds npm by scanning `{data_dir}/tools/node/*/bin/npm` and then PATH, and spawns that binary with no PATH rewrite. Official Node's `bin/npm` is `#!/usr/bin/env node`, so even a managed npm will look up `node` on the process PATH. A Tidebreak desktop server is a GUI child: that PATH usually has neither Homebrew npm nor the sibling `node`. First-run doctor/create then fails with "no npm available". 0041's "user PATH is not the engine" rule is applied to the harness binary, then the installer reintroduces GUI-PATH discovery.

**Suggestion:** Resolve npm via the managed Node pin, put that `bin` directory first on the child PATH (or invoke `node npm-cli.js`), and fail with a doctor-facing "install the managed Node runtime" reason when that pin is missing.

crates/tidebreak-server/src/routes/code/harnesses.rs
```diff
@@ -21,10 +21,47 @@ pub async fn list_harnesses(
...
     State(state): State<AppState>,
 ) -> Result<Json<HarnessDoctorReport>, ServerError> {
-    require_code(&state)?.invalidate_probes();
+    let runtime = require_code(&state)?;
+    #[cfg(not(test))]
+    {
+        let installs = HarnessKind::ALL.iter().map(|kind| {
+            let data_dir = runtime.data_dir.clone();
+            async move {
+                let _ = tidebreak_harness::ensure_installed(&data_dir, *kind).await;
```
naingthet: **[bug]** Doctor refresh is the documented install trigger, but each `ensure_installed` error is discarded (`let _ = ...`). After a failed npm/Node lookup the probes are invalidated and the doctor still says "refresh to install the pinned {kind} binary". The user can loop on Refresh forever with no failure text.

**Suggestion:** Collect install errors per kind and put them on the doctor row. Do not advertise "refresh to install" when the last refresh already failed.

crates/tidebreak-desktop/ui/src/code/labels.ts
```diff
@@ -124,15 +125,15 @@ export function autoIsUnsupervised(caps: ModeCaps): boolean {
...
 
 /**
- * Create default: Ask when the engine carries approvals, else Plan when it
- * has one, else Auto when that is the only posture it honors. An engine
- * honoring nothing still answers Plan — create is disabled before it posts.
+ * Create default: Allow all when the engine can honor it. Otherwise the
+ * most autonomous posture it actually has.
  */
 export function defaultCreatePermissionMode(caps: ModeCaps): CodePermissionMode {
+  if (caps.allow_mode === "supported") return "allow";
```
naingthet: **[bug]** Create default is now Allow whenever `allow_mode` is supported. Decision 0033 still says Ask is the default and "Tidebreak never launches a harness with its permission-bypass flag as a composed default"; 0038 only moved Grok-tier to unsupervised Auto, not Allow.

**Suggestion:** Keep Ask (or the most autonomous *non-bypass* honorable mode) as the create default. Leave Allow as an explicit opt-in.

crates/tidebreak-desktop/ui/src/ShellShortcuts.ts
```diff
@@ -171,6 +172,14 @@ export const SHELL_SHORTCUTS: readonly ShellShortcutDef[] = [
     scope: "code",
     allowInEditable: true,
   },
+  {
+    id: "close-tab",
```
naingthet, line 176: **[bug]** `close-tab` is Cmd+W with no `scope: "code"`. `useShellShortcuts` always preventDefaults a matched definition. In chat, and in a code workspace with no URL tabs, the handler returns immediately, so Cmd+W no longer closes the window.

**Suggestion:** Scope the binding to `code`, skip preventDefault when there is no closable strip tab, and close via `closeCodeChromeTab`.

crates/tidebreak-desktop/ui/src/code/CodeWorkspacePage.tsx
```diff
@@ -151,6 +152,7 @@ function CodeWorkspaceBody({ workspaceId }: { workspaceId: string }) {
       });
       catalog.rememberSession(created);
       setSession(created);
+      await client.submitCodeTurn(created.id, message);
```
naingthet, line 165: **[bug]** Session `model` is persisted and the composer offers a picker, but the running engine often never sees the choice. Codex only puts `model` on `thread/start`; OpenCode only on `session_create_body`. After that, `onModelChange` is local state plus a later submit — the long-lived engine does not pick it up.

**Suggestion:** Pass the selected model into `createCodeSession`. For Codex/OpenCode, refuse mid-session model changes or tear down/recreate the engine session.

crates/tidebreak-harness/src/probe.rs
```diff
@@ -99,6 +102,22 @@ pub async fn probe_shell(host: &HostEnv, name: &str) -> Result<ProbeCapture, Pro
     if name.is_empty() || name.contains(['/', '\\', '\0', '\'', '"', ';', '|', '&']) {
         return Err(ProbeError::NotFound(name.to_owned()));
     }
+    if let Some(data_dir) = &host.data_dir {
+        if let Some(kind) = kind_for_command(name) {
+            if let Some(binary) = crate::managed_binary(data_dir, kind) {
+                let env = capture_shell_env(host).await.unwrap_or_default();
```
naingthet: **[bug]** When a pin exists, probe captures the login-shell environment and launches with `env_clear()` plus that snapshot. Capture failure becomes `unwrap_or_default()`, i.e. an empty env. The child then has no HOME, no user PATH, and none of the gateway/profile variables 0041 says still come from the shell.

**Suggestion:** On capture failure, use the process environment (minus Tidebreak internals), and put the shell stderr on the doctor row.

crates/tidebreak-harness/src/pin.rs
naingthet: Addressed in 2220250c42c23465ff6f996f75dca476fd762f85: npm install now uses `{data_dir}/tools/node/<ver>/bin/{node,npm}` only. That bin dir is prepended to the child PATH so the shebang finds the sibling node. Missing managed Node is a doctor-facing "install the managed Node runtime" error; GUI PATH is not consulted.

crates/tidebreak-server/src/routes/code/harnesses.rs
naingthet: Addressed in 2220250c42c23465ff6f996f75dca476fd762f85: refresh collects each `ensure_installed` result onto the runtime. A failed pin is written on that kind's doctor remediation and the "refresh to install" copy is not shown again. Success clears the stored error.

crates/tidebreak-desktop/ui/src/code/labels.ts
naingthet: Addressed in 2220250c42c23465ff6f996f75dca476fd762f85: create default is Ask when structured approvals are supported, otherwise Auto (Grok-tier, 0038) or Plan. Allow remains in the picker as an explicit opt-in.

crates/tidebreak-desktop/ui/src/ShellShortcuts.ts
naingthet: Addressed in 2220250c42c23465ff6f996f75dca476fd762f85: `close-tab` is scoped to code. The handler returns false (no preventDefault) when there is no workspace or no closable strip tab, and otherwise closes via `closeCodeChromeTab`.

crates/tidebreak-desktop/ui/src/code/CodeWorkspacePage.tsx
naingthet: Addressed in 2220250c42c23465ff6f996f75dca476fd762f85: create now passes the inferred catalog default. Claude and Grok still honor a later picker change on the next turn. Codex and OpenCode lock the picker after start because they only take the model on thread/session create.

crates/tidebreak-harness/src/probe.rs
naingthet: Addressed in 2220250c42c23465ff6f996f75dca476fd762f85: a failed login-shell capture no longer yields an empty env. The child gets the process environment minus Tidebreak internals, and the capture error is written to probe stderr for the doctor row.

naingthet (commented):
## Summary

`83b937e` wraps the Cmd+J / Cmd+I test handlers so they return `void` instead of `number`. That matches the `() => boolean | void` handler type from the Cmd+W preventDefault change and is the desktop UI `tsc` failure.

## Blocking

None in this commit. The earlier pin / doctor / default / Cmd+W / model-lock fixes on `2220250` still look right.

Waiting on the new desktop UI / clippy / test lanes.

naingthet (commented):
## Summary

`fb391ee` folds the extra `drive_turn` argument into `QueuedFollowUp` so clippy's arg cap stays satisfied after session model was added, and marks `record_pin_install` unused-ok in test builds where pin install is skipped.

## Blocking

None. Same payload, fewer parameters. Waiting on the new clippy / desktop UI / test lanes.

crates/tidebreak-desktop/ui/src/code/CodeCatalogStore.ts
```diff
@@ -43,22 +60,51 @@ export const useCodeCatalogStore = create<CodeCatalogStore>()((set, get) => ({
...
     } catch (error) {
       set({
         loaded: true,
         error: error instanceof Error ? error.message : String(error),
       });
+      return;
+    }
+    const extras: Promise<void>[] = [
+      client
+        .getHarnessDoctor()
```
naingthet, line 89: **[bug]** Server `refresh_harnesses` now records install errors, but Code home Refresh still calls `catalog.refresh`, which only `getHarnessDoctor()`s. `list_harnesses` never installs, so `pin_install_error` stays empty and the empty state still says "Refresh to install the engines this build drives". Only Settings (`refreshHarnessDoctor`) records an install error.

**Suggestion:** Point the doctor list's Refresh at `refreshHarnessDoctor`, then write the report into the catalog. Keep catalog page-load on GET so listing workspaces does not wait on npm.

crates/tidebreak-cli/src/code.rs
```diff
@@ -762,19 +762,14 @@ async fn resolve_run_session(
...
 }
 
-/// Desktop create default: Ask when this engine's structured approvals are
-/// Supported; else Plan when it has a plan posture; else Auto when that is
-/// the only posture it honors (grok-tier — unsupervised, and said so). A
-/// missing doctor row defaults to Plan and lets the server refuse.
+/// Desktop create default: Allow all when the engine can honor it.
 fn default_create_permission_mode(caps: Option<&HarnessCaps>) -> CodePermissionMode {
     match caps {
+        Some(caps) if caps.allow_mode == CapLevel::Supported => CodePermissionMode::Allow,
```
naingthet: **[bug]** Desktop `defaultCreatePermissionMode` now returns Ask, then Auto, then Plan, and never Allow. The CLI helper this PR still presents as the "Desktop create default" does the opposite: Allow when `allow_mode` is supported, and Allow again when the doctor row is missing. `tidebreak code start` with no `--mode` therefore composes the bypass flag by default, which decision 0033 still forbids.

**Suggestion:** Mirror `labels.ts`: Ask when structured approvals are supported, else the most autonomous non-bypass honorable mode, else Plan. Keep Allow as an explicit `--mode`.

crates/tidebreak-desktop/ui/src/code/CodeWorkspacePage.tsx
```diff
@@ -142,15 +140,21 @@ function CodeWorkspaceBody({ workspaceId }: { workspaceId: string }) {
...
 
-  async function startSession(harness: HarnessKind, permissionMode: CodePermissionMode) {
+  async function startSession(
+    harness: HarnessKind,
+    permissionMode: CodePermissionMode,
+    message: string,
+  ) {
     setStarting(true);
     try {
       const created = await client.createCodeSession(workspaceId, {
```
naingthet, line 158: **[bug]** Codex/OpenCode now omit `onModelChange` after attach, and create *can* send `inferredHarnessModel`. That id is filled only after catalog listing succeeds, and doctor GET usually wins that race, so New Workspace and the start prompt can post `model: undefined`. Codex still puts the id only on `thread/start`, OpenCode only on `session_create_body`. The start composer and new-workspace dialog still have no model picker — the only time those two engines would honor a choice.

**Suggestion:** Await the listing before create, and put the picker on the start prompt / new-workspace dialog. After attach is too late for Codex/OpenCode.

naingthet (commented):
## Summary

Rechecked the iteration (`2220250c` plus the clippy/typecheck follow-ups on this head). Three of the six earlier bugs are actually fixed: managed Node bin is first on PATH, Cmd+W is code-scoped and only preventDefaults when a strip tab closes, and a failed shell capture uses the process environment.

**Still do not merge.** Three leftovers:

1. Code-home Refresh still only GETs the doctor. It never hits `refresh_harnesses`, so first-run install still does not run from the empty-state button.
2. Desktop create now defaults to Ask; `tidebreak code start` still defaults to Allow, including when the doctor row is missing.
3. Codex/OpenCode lock the picker after start, but create often sends no model and the start surfaces still have no picker.

## Issue counts by severity

- bugs: 3
- suggestions: 0
- nits: 0

crates/tidebreak-desktop/ui/src/code/CodeCatalogStore.ts
naingthet: Addressed in c5ed1081f933959763894ff062b46634fd5ff0dd: the empty-state Refresh button now calls `refreshDoctor`, which POSTs `/code/harnesses/refresh` and writes the report into the catalog. Page load still GETs so listing workspaces does not wait on npm.

crates/tidebreak-cli/src/code.rs
naingthet: Addressed in c5ed1081f933959763894ff062b46634fd5ff0dd: `tidebreak code start` now uses the same default as the desktop — Ask when structured approvals are supported, otherwise Auto or Plan. Missing doctor row is Plan, not Allow. Allow stays an explicit `--mode`.

crates/tidebreak-desktop/ui/src/code/CodeWorkspacePage.tsx
naingthet: Addressed in c5ed1081f933959763894ff062b46634fd5ff0dd: the start prompt and new-workspace dialog now list models. Create awaits `ensureHarnessModels` (or gateway models) and posts that id, so Codex/OpenCode see the choice on thread/session create.

naingthet (commented):
## Summary

Rechecked `c5ed1081` against the three leftovers from `fb391ee`.

1. Code-home Refresh now calls `refreshDoctor` (POST `/code/harnesses/refresh`). Page load still GETs. First-run install from the empty-state button actually runs.
2. `tidebreak code start` now mirrors desktop: Ask when structured approvals are supported, else Auto, else Plan. Allow is only `--mode`. The missing-doctor fallback is Plan, not Allow.
3. Start prompt and New Workspace list models, wait on `ensureHarnessModels`, and pass the chosen/default id into `createCodeSession`. Codex/OpenCode can receive the model at session start. The live picker still locks after attach.

Those three bugs look fixed. No new correctness issues in this commit. Hold only for CI on this head.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

naingthet (commented):
## Summary

`95771c7` boxes `CodeUpdateNotice::Digest.pr_state` as `Option<Box<PullRequestDigest>>` and switches the CLI renderer to `as_deref()`. That is the `large_enum_variant` clippy failure from `PullRequestDigest` growing checks/title. Serde of `Box<T>` is transparent, so the wire shape is unchanged.

## Blocking

None. Waiting on the new clippy / desktop / test lanes.

PR #2178, opened 2026-08-18
test(server): stabilize container worker service test

naingthet:
## Why

The August 18, 2026 `main` CI run failed in `sandbox_container_run::tests::container_worker_service_drives_queued_work_over_loopback` after the test exhausted its 30-second outer timeout. The same path completes in roughly 0.1 seconds in isolation, and 968 sibling tests passed in the failed run, identifying this as parallel-suite contention rather than a product regression.

The test was still using the aggressive 250 ms durable-fence cadence and two service drive lanes even though it only proves discovery and completion of one queued run. Under the full parallel suite, those extra pollers and writes contend for the fixture SQLite database's single writer and can consume the hang guard.

## What changed

- use the existing quiet drive cadence because this test does not assert heartbeat or fence timing
- run one service drive lane for the singular queued fixture
- retain periodic maintenance and the state-based completion assertion
- give the integration-heavy path a 60-second outer hang guard and a 45-second diagnostic state wait

No production code changed and no tests were removed.

## Validation

- `cargo test -p tidebreak-server --locked sandbox_container_run -- --test-threads 4 --skip docker`
- 20/20 repeated container-run module iterations at `--test-threads 4`
- `cargo test -p tidebreak-server --lib --locked` — 969 passed, 3 ignored

---

> [!NOTE]
> **Low Risk**
> Only test configuration and timeouts in one integration test; no production code changes.
> 
> **Overview**
> **Stabilizes** `container_worker_service_drives_queued_work_over_loopback`, which was timing out under full parallel CI while passing quickly in isolation.
> 
> The worker is configured with the existing **`quiet_drive_config`** (slower heartbeat/fence cadence) and **`max_concurrency: 1`** so the test does not assert timing behavior and reduces SQLite single-writer contention from extra pollers and aggressive 250 ms fences. Outer hang guard moves to **60s** and the state **`wait_until`** to **45s**; completion assertions are unchanged.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e11e0291bee0098df5cd4a764a71873edd674bd0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/tidebreak-server/src/sandbox_container_run/tests.rs
+++ b/crates/tidebreak-server/src/sandbox_container_run/tests.rs
@@ -1407,7 +1407,7 @@ async fn wait_until_lease_expired(store: &Arc<dyn Store>, run_id: AgentRunId) {
 /// sandbox agent over loopback to its committed result.
 #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
 async fn container_worker_service_drives_queued_work_over_loopback() {
-    tokio::time::timeout(Duration::from_secs(30), async {
+    tokio::time::timeout(Duration::from_secs(60), async {
         let (_dir, store, chat) = store().await;
         let run_id = admit_container_run(&store, chat.id, "driven by the service").await;
         let backend = MockBackend::spawning();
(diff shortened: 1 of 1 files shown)


naingthet (commented):
The test now uses `quiet_drive_config` and a single worker lane, so it no longer contends on the fixture SQLite writer under the parallel suite. Completion still asserts discovery plus teardown discharge. Timeouts widen to 60s / 45s without changing the product path.

No blocking issue.

PR #2180, opened 2026-08-18
test(server): stabilize the cancellation usage-accounting test

naingthet:
`attached_cancellation_accounts_usage_observed_before_reverse_stream_drop` flaked on this PR's own CI run by exhausting its 30-second outer guard with a bare `Elapsed` — the stall predates the inner waits, whose own bounds (a 30-second flag wait, a 15-second join) could never surface their diagnostics under a 30-second outer limit. Same parallel-suite SQLite contention class as #2178 and #2181.

Cancellation quiescing and usage accounting are under test, not lease or fence timing:

- The quiet cadence stops the 250 ms durable fence writing against the fixture database's single writer for the whole gated window.
- The long lease removes the 2-second expiry a stalled heartbeat could cross under load; nothing asserted the short lease.
- The heartbeat stays sub-second (250 ms) so nothing riding a periodic tick waits out a quiet period.
- The outer guard rises to 120 seconds, above the sum of the inner waits, and the join keeps its diagnostic panic at 30 seconds.

Assertions are unchanged.

This PR originally also stabilized the heartbeat lease test; #2181 landed on the same test first and has the floor, so that half was dropped here — the two PRs are now disjoint and complementary.

## Validation

- 10/10 repeated `sandbox_container_run` module iterations at `--test-threads 4`
- `cargo test -p tidebreak-server --lib --locked` — 973 passed

---

> [!NOTE]
> **Low Risk**
> Test-only timing and configuration changes in integration tests; no production code paths modified.
> 
> **Overview**
> Stabilizes **`attached_cancellation_accounts_usage_observed_before_reverse_stream_drop`** against parallel-suite SQLite contention—the same flake class as recent sandbox container run test fixes.
> 
> The runner now uses **`quiet_drive_config()`** with a **250 ms** heartbeat instead of **`fast_config()`**’s short lease and **250 ms** durable fence, so cancellation quiescing and usage accounting stay under test without hammering the fixture DB’s single writer or racing lease expiry under load. The outer test timeout rises to **120 s** (above inner waits so diagnostics surface), and the drive **join** timeout rises to **30 s** with the existing diagnostic panic unchanged.
> 
> **Assertions are unchanged.**
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c3e266358bf20d68d95bb08777646691feee56d3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/tidebreak-server/src/sandbox_container_run/tests.rs
+++ b/crates/tidebreak-server/src/sandbox_container_run/tests.rs
@@ -3779,7 +3779,9 @@ async fn a_noop_heartbeat_does_not_revoke_a_live_container_drive() {
 /// immutable cancelled result.
 #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
 async fn attached_cancellation_accounts_usage_observed_before_reverse_stream_drop() {
-    tokio::time::timeout(Duration::from_secs(30), async {
+    // Above the sum of the inner waits, so a slow provision surfaces their
+    // diagnostics instead of the outer guard's bare Elapsed.
+    tokio::time::timeout(Duration::from_secs(120), async {
         let (_dir, store, chat) = store().await;
         let run_id = admit_container_run(&store, chat.id, "account before cancellation").await;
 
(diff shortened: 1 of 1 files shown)


naingthet (commented):
Calms the durable fence via `quiet_drive_config` and widens the outer guard to 120s so it covers the inner waits. Heartbeat stays under a lease period. Assertions are unchanged.

Competes with #2181 on the same test. This head's CI failed on a different flake (`attached_cancellation_accounts_usage_observed_before_reverse_stream_drop`), not on the heartbeat proof. Do not land both.

naingthet:
Deconflicted with #2181 per the review: this PR no longer touches `heartbeats_the_lease_so_a_long_run_is_not_reaped` — #2181 has the floor there. It now carries only the `attached_cancellation_accounts_usage_observed_before_reverse_stream_drop` stabilization, which is the flake that fired on this PR's first CI run and which #2181 does not cover. The two are disjoint and can land in either order.

naingthet (commented):
The cancellation usage-accounting test now uses `quiet_drive_config` and a 120s outer guard so a slow provision surfaces the inner diagnostics instead of a bare `Elapsed`. Cancellation and usage assertions are unchanged. The heartbeat-lease half is gone, so this no longer competes with #2181.

No blocking issue.

PR #2177, opened 2026-08-18
fix(server): stop reading merge stderr naming an author as a sign-out

naingthet:
The merge-error classifier treated any stderr containing the bare substring `auth` as gh being signed out. `author` contains `auth`, so a blocked merge whose host message names the pull request author ("review is required from someone other than the pull request author") was misrouted to the sign-in remediation, hiding the real reason.

The sign-out branch now matches specific markers only — `not logged`, `not signed`, `authenticat…`, `gh auth login`, `HTTP 401` — and the classifier test pins an author-naming blocked message to `MergeBlocked` and an author-naming unknown message to the `User` fallthrough.

Follow-up to the non-blocking review note on #2176.

---

> [!NOTE]
> **Low Risk**
> Localized change to merge stderr string matching and unit tests; no auth flow or merge execution changes.
> 
> **Overview**
> **Merge error classification** no longer treats any stderr containing `auth` as “gh signed out.” Host messages that mention the pull request **author** (e.g. review rules naming the author) were misclassified as sign-in remediation instead of the real merge block reason.
> 
> Sign-out detection in `classify_merge_error` now uses specific markers only: `not logged`, `not signed`, `authenticat…`, `gh auth login`, and `HTTP 401`. Tests pin an author-related blocked message to `MergeBlocked` and a generic author-related message to the `User` fallthrough.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 80a9e41216d217da1f1dbc3863808770baf88170. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #2181, opened 2026-08-18
test(server): stabilize container heartbeat lease test

naingthet:
## Why

The August 18, 2026 CI run on #2177 failed in
`sandbox_container_run::tests::heartbeats_the_lease_so_a_long_run_is_not_reaped`
after the test exhausted its 60-second outer timeout. The same path
completes in roughly 0.2 seconds in isolation, and 972 sibling tests
passed in the failed run, identifying this as parallel-suite SQLite
contention rather than a product regression.

This is the same class of flake #2178 fixed for
`container_worker_service_drives_queued_work_over_loopback`. Unlike that
test, this one *does* prove the attached heartbeat: the run must stay
`running` and its lease expiry must move forward while the model call is
held. The 100 ms heartbeat stays so the proof does not sleep through a
full lease period.

The 250 ms durable fence is not under test. Under the full parallel
suite those extra writes contend for the fixture SQLite database's
single writer. Setup can then consume most of the hang guard before
post-release completion has time to finish, and the outer expect reports
only `Elapsed(())`.

## What changed

- keep the 100 ms attached heartbeat and the state assertions that the
  lease moved, the run stayed live, and the drive completed
- use a 120-second lease and a quiet 30-second fence so provision and
  attach do not fight the suite for SQLite's writer or expire before the
  first tick
- give the integration-heavy path a 120-second outer hang guard, a
  60-second wait for the first model call, and a 20-second wait for the
  observed heartbeat
- if the drive is still outstanding after release, print provision,
  destroy, provider-call, and heartbeat counts instead of a bare elapsed
  expect

No production code changed and no tests were removed.

## Validation

- `cargo test -p tidebreak-server --lib --locked sandbox_container_run -- --test-threads 4 --skip docker`
- 10/10 repeated focused runs of
  `heartbeats_the_lease_so_a_long_run_is_not_reaped`

---

> [!NOTE]
> **Low Risk**
> Test-only timing and configuration in one integration test; no server or driver behavior changes.
> 
> **Overview**
> CI flake in **`heartbeats_the_lease_so_a_long_run_is_not_reaped`**: the 60s outer hang guard could expire under parallel **`sandbox_container_run`** SQLite writer contention while provision/attach ran, even though the heartbeat proof itself is fast in isolation.
> 
> The test still holds the first model step on a gate and asserts the run stays **`running`** and **`lease_expires_at`** moves after at least one observed heartbeat tick. It now uses a **120s lease**, keeps the **100ms** heartbeat, and sets **`durable_fence_interval` to 30s** so setup does not fight the suite for extra durable writes (the fence is not what this test proves). Waits are lengthened (**60s** for first model call, **20s** for heartbeat), the outer guard is **120s**, and if the drive task is still outstanding after release it panics with provision/destroy/provider/heartbeat counts instead of a bare elapsed timeout.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2d9eda44d64fbbb11f20df91a9ddf05b281a2bac. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/crates/tidebreak-server/src/sandbox_container_run/tests.rs
+++ b/crates/tidebreak-server/src/sandbox_container_run/tests.rs
@@ -2070,7 +2070,7 @@ async fn recovered_drive_keeps_failed_provider_attempts_in_the_spend_budget() {
 /// move forward before the model call is released.
 #[tokio::test(flavor = "multi_thread", worker_threads = 4)]
 async fn heartbeats_the_lease_so_a_long_run_is_not_reaped() {
-    tokio::time::timeout(Duration::from_secs(60), async {
+    tokio::time::timeout(Duration::from_secs(120), async {
         let (_dir, store, chat) = store().await;
         let run_id = admit_container_run(&store, chat.id, "takes a while").await;
 
@@ -2081,7 +2081,8 @@ async fn heartbeats_the_lease_so_a_long_run_is_not_reaped() {
         // the container is driving — so a short lease expires under CI
         // contention before the first tick and this loop never observes an
         // extension. A short heartbeat still proves the expiry moved without
-        // sleeping through a full lease period.
+        // sleeping through a full lease period. The durable fence is not under
+        // test here; a quiet interval avoids extra SQLite writes during setup.
         let provider = Arc::new(ScriptedProvider::gated(
             vec![
                 "use-tool:write_file:{\"path\":\"note.txt\",\"content\":\"a b\"}".to_owned(),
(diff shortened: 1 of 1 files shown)


naingthet (commented):
Keeps the 100 ms heartbeat the test actually proves, quiets only the unused fence, and replaces a bare elapsed panic with provision/destroy/provider/heartbeat counts. That is the tighter of the two competing fixes (#2180).

This head's CI failed in `codex::session::tests::a_thread_that_ran_no_turn_is_not_a_resume_ref`, unrelated to the sandbox change. Do not land both #2180 and #2181.

PR #2183, opened 2026-08-18
fix(release): retry newly created draft lookup

naingthet:
## What changed

- treat a just-created Release Drafter ID that is temporarily absent from the releases list as retryable
- retry release-list reconciliation with bounded backoff before failing
- keep failing immediately when the reported release is visible but is not a normal draft
- add regression and workflow-policy coverage for the retry path

## Root cause

The failed `v0.48.0` publish left that release as an in-flight prerelease, so Release Drafter correctly created the next `v0.48.1` draft. The action returned the new release ID before GitHub's paginated releases-list endpoint included it. Reconciliation interpreted that transient omission as a permanent invariant violation and failed the release-draft workflow.

## Validation

- `node --test scripts/*.test.mjs`
- `actionlint .github/workflows/release-draft.yml`
- `git diff --check`

---

> [!NOTE]
> **Low Risk**
> Changes only GitHub Actions release-draft automation and reconciliation scripts; no runtime product or auth paths.
> 
> **Overview**
> Fixes **release-draft** failures when Release Drafter returns a new release ID before GitHub’s paginated releases API lists that release.
> 
> **`planDraftReconciliation`** now returns **`action: "retry"`** when the reported ID is missing from the fetched list (transient lag). It still **throws** if the ID appears but is not a normal non-prerelease draft.
> 
> The **“Keep exactly one native release draft”** workflow step mirrors the baseline step: up to **5 attempts**, paginated fetch + `jq` flattening, bounded backoff (`attempt * 5` seconds), then fail with the last reason. Deletes extra drafts only when the plan is not `retry`.
> 
> Tests cover the retry path, the “listed but wrong kind” path, and workflow policy checks for `max_attempts=5` and retry handling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f5fc8fa2d161503748a96435b503c82015ec3690. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/release-draft.yml
+++ b/.github/workflows/release-draft.yml
@@ -122,24 +122,46 @@ jobs:
           RELEASE_ID: ${{ steps.release_drafter.outputs.id }}
           RESOLVED_VERSION: ${{ steps.release_drafter.outputs.resolved_version }}
         run: |
-          gh api --paginate \
-            -H 'Accept: application/vnd.github+json' \
-            "repos/$GITHUB_REPOSITORY/releases?per_page=100" | \
-            jq -s '[.[][]]' > "$RUNNER_TEMP/releases.json"
-          plan="$(node scripts/reconcile-release-drafts.mjs \
-            --releases "$RUNNER_TEMP/releases.json" \
-            --resolved-version "$RESOLVED_VERSION" \
-            --release-id "$RELEASE_ID")"
-          echo "$plan"
-          while IFS= read -r id; do
-            gh api --method DELETE \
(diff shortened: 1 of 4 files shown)


naingthet (commented):
A missing Release Drafter ID is now `retry` with five list-and-backoff attempts, so transient pagination lag no longer fails the draft job. A listed ID that is not a normal draft still throws. Extra-draft deletes stay off the retry path.

No blocking issue.

PR #2182, opened 2026-08-18
feat(desktop): ship ARM64 packages and cross-platform updates

naingthet:
## Summary

- fix the Windows release prerequisite by invoking pnpm through the Windows command interpreter when generating third-party notices
- add native ARM64 Windows and Linux release matrices, artifacts, immutable manifest entries, updater keys, stable downloads, and recovery paths
- enable the existing signed, user-confirmed Tauri updater flow on production Windows and Linux builds
- supersede decision 40 with decision 43 and update the release contract, installation docs, deferred scope, and agent guidance

## Root cause

The v0.48.0 Windows prerequisite reached `generate-third-party-notices.mjs` from Git Bash, but Node attempted to spawn the extensionless `pnpm` command directly. On Windows pnpm is a command shim, so the process failed with `spawnSync pnpm ENOENT` before compilation began.

The generator now uses `cmd.exe /d /c pnpm ...` on Windows and keeps the direct shell-free invocation on Unix hosts.

## Release impact

Production releases now require:

- universal macOS
- Windows x86_64 and ARM64 NSIS installers
- Linux x86_64 and ARM64 AppImages and Debian packages

Windows ARM64 uses `windows-11-arm`; Linux ARM64 uses `ubuntu-22.04-arm`. Publication remains all-or-nothing across the declared platform and architecture set. Windows and Linux release builds use the same authenticated background-download and explicit restart-to-install flow as macOS; debug builds remain disabled.

## Validation

- `actionlint .github/workflows/release.yml`
- `node scripts/generate-third-party-notices.mjs --check`
- `node --test scripts/create-release-manifests.test.mjs scripts/third-party-notices.test.mjs scripts/workflow-security.test.mjs` (48 passed)
- `cargo test -p tidebreak-desktop updater:: --locked` (9 passed)
- `pnpm exec vitest run src/settings/UpdatesPanel.test.tsx` (3 passed)
- `pnpm build`
- `git diff --check`

## Remaining native verification

GitHub Actions is the authoritative native build check for the Windows and Linux package matrices. This PR should not be considered release-ready until those native jobs and required PR checks are green.

---

> [!NOTE]
> **Medium Risk**
> Changes production release gates, hosted manifest shape, and cross-platform updater behavior; a partial or mis-keyed ARM64 matrix could block or mis-publish releases, though workflow-policy tests pin the contract.
> 
> **Overview**
[...]

PR #2185, opened 2026-08-18
test(release): authorize ARM64 packages and cross-platform updates

naingthet:
## Summary

- authorize the current release workflow and the upcoming ARM64 Windows/Linux release graph in the trusted workflow policy
- allow Linux packaging to select `ubuntu-22.04` through `${{ matrix.runner }}` while keeping that runner as the x86_64 baseline
- allow immutable publication to attach native `aarch64` Windows and Linux assets
- allow the packaged updater to enable signed production updates on Windows and Linux as well as macOS

## Why

The release-policy CI job intentionally runs `main`'s trusted tests against pull-request workflow changes. PR #2182 cannot add ARM64 matrices or enable the signed updater on Windows and Linux until the trusted policy recognizes that graph. This is the same prerequisite-policy transition used for Windows/Linux packaging in #2174.

The transition keeps every current assertion active while admitting the stricter follow-up workflow. PR #2182 will tighten the tests to the final ARM64 and cross-platform-updater shape after this prerequisite lands.

## Validation

- `node --test scripts/workflow-security.test.mjs` — 35 tests passed against `main`
- `node --test scripts/workflow-security-mutations.test.mjs` — 43 tests passed against `main`
- the same trusted tests passed against #2182's files
- `git diff --check`

Prerequisite for #2182.

---

> [!NOTE]
> **Low Risk**
> Test-only changes to trusted workflow policy matchers; no release workflow or application code is modified.
> 
> **Overview**
> Updates **`scripts/workflow-security.test.mjs`** so release-policy CI can pass on **`main`** today while also accepting the follow-up workflow from PR #2182 (ARM64 matrices and Windows/Linux production updaters).
> 
> **Linux `build_linux`** checks still require `ubuntu-22.04` and the existing packaging guards, but **`runs-on`** may be pinned literally or via **`${{ matrix.runner }}`**. When an ARM runner appears in the job, the test also expects both **`x86_64-unknown-linux-gnu`** and **`aarch64-unknown-linux-gnu`** matrix targets.
> 
> **`attach_downloads`** keeps current x86_64 asset names while allowing Debian signature paths to use **`${arch}`** instead of a fixed **`x86_64`**. If Windows ARM64 artifact names are present, it additionally requires matching Linux ARM64 packages and signatures.
> 
> **Packaged updater policy** still demands macOS in the production **`cfg!`**, with a looser matcher for nested **`all(...)`** blocks. If **`updater.rs`** mentions Windows or Linux, both **`target_os`** entries must appear—without requiring them on **`main`** until the packaging PR lands.
> 
> This is the same prerequisite pattern as #2174: current assertions stay active; stricter final shapes can tighten tests after this merges.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bdb9025599865b20a25557d887d1ef68af06e1e6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/scripts/workflow-security.test.mjs
+++ b/scripts/workflow-security.test.mjs
@@ -1511,7 +1511,17 @@ test("Linux packaging writes no shared cache before loading updater material", (
   assert.doesNotMatch(release, /^  prepare_linux:/m);
   const buildJob = workflowJob(release, "build_linux");
   assert.match(buildJob, /needs: \[validate, inspect_hosted\]/);
-  assert.match(buildJob, /runs-on: ubuntu-22\.04/);
+  // Current workflow pins the x86_64 job to ubuntu-22.04. The upcoming ARM64
+  // matrix keeps that runner and selects it through ${{ matrix.runner }}.
+  assert.match(buildJob, /ubuntu-22\.04/);
+  assert.match(
+    buildJob,
+    /runs-on: (?:ubuntu-22\.04|\$\{\{ matrix\.runner \}\})/,
+  );
+  if (/runner: ubuntu-22\.04-arm/.test(buildJob)) {
+    assert.match(buildJob, /target: x86_64-unknown-linux-gnu/);
+    assert.match(buildJob, /target: aarch64-unknown-linux-gnu/);
+  }
   assert.match(buildJob, /libwebkit2gtk-4\.1-dev/);
   assert.match(buildJob, /SCCACHE_GHA_RW_MODE: READ_ONLY/);
   assert.match(buildJob, /version: 10\.18\.3/);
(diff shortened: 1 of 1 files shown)


naingthet (commented):
## Summary

Same two-step as #2174. Trusted tests still pin today's x86_64 / macOS-only literals, and additionally admit the ARM64 matrix and cross-platform updater when those strings appear. Release policy is already green on this head.

No blocking issue. Hold only until clippy / desktop test / workspace test finish (scripts-only change). Merge this before #2182.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

PR #2192, opened 2026-08-18
refactor(desktop): extract shared transcript primitives from chat

naingthet:
Groundwork for the code-mode transcript overhaul. ADR 0030 asks code mode to
share UI primitives with chat rather than grow a second implementation of the
same transcript, and today those primitives are private to `MessageList` and
`ChatView`. This moves them out. Chat is the only consumer in this change, and
the DOM it renders is unchanged — verified by rendering a user turn with
attachments and invoked skills before and after and diffing the markup.

## What moved

- `src/UserMessage.tsx` — the user branch of `MessageBubbleImpl`: the
  `.message-user-frame` wrapper, the `.message.message-user` article with its
  `aria-label` and transcript anchor, the Markdown body, and the footer. Chat's
  image and file attachments pass as `leading`, its invoked skills as
  `trailing`, because those are chat's and code mode has neither.
- `src/AssistantMessageBody.tsx` — moved verbatim from `MessageList.tsx`.
- `src/TranscriptSkeleton.tsx` — moved verbatim from `MessageList.tsx`.
- `src/useTranscriptFollow.ts` — `ChatView`'s scroll-follow machine, moved as
  it stood: `followsLatest` with drift-disarm, programmatic-scroll marking, the
  `scrollend` re-pin, the `--transcript-viewport` variable, the edge-fade class,
  and both ResizeObservers. It composes the pure helpers in `ChatScroll.ts`,
  which stays where it is. `ChatView` keeps the parts that are its own — the
  deep-link reveal, the `?at=` anchor jump, per-chat re-arming — and drives them
  through the hook's `armFollow` / `disarmFollow` / programmatic-scroll controls
  rather than reaching into refs.

The edge fades stay wrapper-overlay pseudo-elements (`.message-view::before` /
`::after`). No `mask-image` goes anywhere near the scrolling layer: masking a
composited scroll layer is what WKWebView occasionally failed to repaint after
the mount-time catch-up scroll, and `styles.css` says so at the rule.

## What is new

- `src/ToolOutputPreview.tsx` — a shared mono block clamped to the first lines
  of a tool's output, with an honest "Show N more lines" expander and a copy
  control revealed on hover or focus. Nothing adopts it yet; it lands here so
  the code transcript can stop rendering raw `pre` blocks in the next change.
- `ToolCardShell` gives a truncated title a native `title` attribute when the
  title is a string, so a long command is readable on hover. Chat benefits too.

## Tests

The existing chat DOM suite is the regression net and passes unchanged — no
test needed an import update. One new DOM test covers `ToolOutputPreview`
(clamp, expand, copy present). No tests were deleted.

`pnpm test` (1129 passing) and `pnpm build` are green locally.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/AssistantMessageBody.tsx
+++ b/crates/tidebreak-desktop/ui/src/AssistantMessageBody.tsx
@@ -1,0 +1,23 @@
+import { MessageMarkdown } from "./MessageMarkdown";
+import { useStreamingTypewriter } from "./useStreamingTypewriter";
+
+/**
+ * Assistant prose driven by the typewriter: while the bubble is the live
+ * streaming turn its text is typed in, and a settled or rehydrated message
+ * renders at once. Block-level memoization inside {@link MessageMarkdown} keeps
+ * each tick's re-parse confined to the trailing block.
+ */
+export function AssistantMessageBody({
+  text,
+  streaming,
+  containerRef,
+}: {
+  text: string;
+  streaming: boolean;
+  containerRef?: React.Ref<HTMLDivElement>;
+}) {
+  const displayed = useStreamingTypewriter(text, streaming);
+  return (
+    <MessageMarkdown containerRef={containerRef}>{displayed}</MessageMarkdown>
+  );
+}
--- a/crates/tidebreak-desktop/ui/src/ChatView.tsx
+++ b/crates/tidebreak-desktop/ui/src/ChatView.tsx
@@ -9,11 +9,8 @@ import {
 } from "react";
 import type { ApiClient, Chat } from "./api";
 import type { RendererTurnUsage } from "./generated/wire";
-import {
-  followScrollBehavior,
-  isNearBottom,
-  scrollToLatest,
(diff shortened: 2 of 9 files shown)


naingthet (commented):
## Summary

Extracts the chat transcript pieces code mode will need (ADR 0030): `UserMessage` with leading/trailing slots, `AssistantMessageBody`, `TranscriptSkeleton`, and `useTranscriptFollow`. Chat stays the only consumer. The follow machine is the same one that lived in `ChatView` — drift-disarm, programmatic-scroll marking, `scrollend` re-pin, `--transcript-viewport`, edge-fade class, both ResizeObservers. Fades stay wrapper overlays, not a mask on the scroll layer.

`ToolOutputPreview` is unused on purpose and clamps by line with an honest remainder count. `ToolCardShell` adds a native `title` when the heading is a string.

No blocking issue. Chat DOM suite is the regression net. Hold merge until the desktop UI lane finishes.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

PR #2193, opened 2026-08-18
feat(desktop): default code sessions to Allow and speed up workspace creation

naingthet:
Opening a code workspace used to cost five decisions and a round of approvals before any work started. This makes the create flow one keystroke deep.

## Permission mode defaults to Allow

`defaultCreatePermissionMode` now walks the scale down from the top — Allow, then Auto, then Ask, then Plan — and stops at the first mode the harness's own capability flag says it honors. Nothing else about decision 0039 changes: the flags, the adapter mapping, and the closed bypass denylist for every mode but Allow all stand.

The statement rule stands too. Allow and unsupervised Auto still say what they are next to the control that selects them; now that they can arrive by default, that statement renders on open rather than only after a deliberate pick.

Decision 0039 recorded `Ask` as the default, so it carries a dated amendment (the record's status line points at it) covering the new default, why supervision cost dominated the create flow, and what would revisit it: an incident caused by an unsupervised default.

## The dialog remembers

`NewWorkspaceDialog` preselects the repo of the newest workspace in the catalog, the harness of the most recently started session, and the model through the existing default-model plumbing. A small `lastCreate` record in `CodeUiStore` (localStorage, like `reviewSidebarOpen`) is the tie-breaker for a fresh window, written after a successful create. Focus starts on the repo field.

## Cmd+Enter

Cmd+Enter (Ctrl+Enter off macOS) submits from anywhere in the dialog with the values on screen, disabled while creating and when there is no repo to create against. Plain Enter keeps its field-local behavior. The same binding starts the session from `StartSessionPrompt`, routed through the composer's own send button so it obeys one set of disabled rules. The Create button carries the keycap hint the project dialog already uses.

## Harness picker

Rows are logo and product name. The vendor blurbs ("Anthropic's coding agent", …) told a reader choosing an engine nothing they did not already know. The subtitle slot stays for the functional unusable-reason line on disabled rows.

## Fix found on the way

The model select's hidden native input answers an async model list — rows and value landing in the same commit — with an empty change event, which wiped the preselected model. An empty id is never a model, so it is no longer treated as a choice.

## Tests

Added one DOM test (`NewWorkspaceDialog.dom.test.tsx`): the dialog opens prefilled from a seeded catalog (most-recent repo and harness, listed default model, Allow stated) and Cmd+Enter posts those values.

Replaced, with the reason for each:

- `labels.test.ts` "defaults to Ask when the engine can honor it, never Allow" — asserted the superseded default; rewritten to walk the new scale one flag at a time.
- `labels.test.ts` grok case, the `defaultCreatePermissionMode(auto + allow) === "auto"` assertion — that combination now defaults to Allow, and the rewritten first test covers it.
- `StartSessionPrompt.dom.test.tsx` "defaults to Ask and posts Ask…" — same superseded default; replaced by one that asserts the Allow default, its statement, and the new Cmd+Enter start in a single run.
- `StartSessionPrompt.dom.test.tsx` "states Allow all when that mode is chosen" — its statement assertion is now duplicate coverage of the default; kept only the mode-menu wiring it uniquely covered.
- `HarnessPicker.dom.test.tsx` — extended the existing row test with an exact-name match, so a subtitle creeping back onto a ready row fails.

`pnpm vitest run` 1128 passed, `pnpm build` clean.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeUiStore.ts
+++ b/crates/tidebreak-desktop/ui/src/code/CodeUiStore.ts
@@ -1,8 +1,44 @@
 import { create } from "zustand";
 
+import type { HarnessKind } from "../api/types";
 import { useCodeCatalogStore } from "./CodeCatalogStore";
 
 const REVIEW_SIDEBAR_OPEN_KEY = "tidebreak.code-review-sidebar-open";
+const LAST_CREATE_KEY = "tidebreak.code-last-create";
+
+/** What the reader picked the last time they created a workspace. */
+export type CodeCreateDefaults = {
+  repoId?: string;
+  harness?: HarnessKind;
+  model?: string;
+};
+
+function readStoredCreateDefaults(): CodeCreateDefaults | null {
+  try {
+    const raw = window.localStorage.getItem(LAST_CREATE_KEY);
+    if (!raw) return null;
+    const parsed: unknown = JSON.parse(raw);
+    if (!parsed || typeof parsed !== "object") return null;
+    const record = parsed as Record<string, unknown>;
+    const text = (value: unknown) =>
+      typeof value === "string" && value.length > 0 ? value : undefined;
+    return {
+      repoId: text(record.repoId),
+      harness: text(record.harness) as HarnessKind | undefined,
+      model: text(record.model),
+    };
+  } catch {
+    return null;
+  }
+}
+
+function storeCreateDefaults(defaults: CodeCreateDefaults): void {
+  try {
+    window.localStorage.setItem(LAST_CREATE_KEY, JSON.stringify(defaults));
+  } catch {
+    // Preference persistence is best-effort.
+  }
+}
 
 function readStoredReviewSidebarOpen(): boolean {
   try {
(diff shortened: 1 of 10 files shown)


naingthet (commented):
Create flow is now one keystroke deep, and the default posture matches the 0039 amendment.

- `defaultCreatePermissionMode` walks Allow → Auto → Ask → Plan and stops at the first `supported` flag. Allow and unsupervised Auto still render their statement on open, not only after a pick.
- Dialog seeds repo / harness / model from newest catalog activity, then `lastCreate` in localStorage. Invalid stored harness kinds are ignored because `recentHarness` only accepts a ready doctor entry.
- Cmd+Enter submits from anywhere; the start-session prompt clicks the composer's own send button so disabled rules stay in one place.
- Harness rows drop the vendor gloss; unusable rows still show the reason.

ADR 0039 stays Proposed with a dated amendment and a revisit trigger (incident caused by an unsupervised default). Tests pin the new default, the remembered-create path, and the shortcut. No blocking issue. Merge once the UI lane is green.

PR #2194, opened 2026-08-18
feat(desktop): render code-mode turn outcomes and adopt shared transcript primitives

naingthet:
## What changed

Code-mode turn boundaries now render. A completed turn closes with a quiet seam that shows duration, a `DiffstatBadge`, and a token summary. A failed turn is a critical `role=alert` block that shows the engine error — previously `TurnReviewCard` returned `null`, so a failed turn showed nothing. An interrupted turn is a warning one-liner.

The transcript adopts the shared chat primitives from #2192, per [ADR 0030](docs/decisions/0030-code-mode-separate-surface.md) convergence: `UserMessage`, `AssistantMessageBody`, and `TranscriptSkeleton` from `src/`. The pane uses `useTranscriptFollow` so code mode scrolls, fades, and returns to the tail the same way chat does. A session shows a skeleton until the durable snapshot settles (`hydrated` on the reducer, `onHydrateSettled` on the controller), including when the snapshot cannot be read, so an existing session does not flash its empty state. Streaming uses `AssistantWorkingIndicator` in the gaps between live cards, and each row is wrapped in `isolatedCard` so one throw does not take the transcript with it. User items carry `createdAt` from the turn's `started_at` so the footer timestamp is the server's, not the client's.

A session whose harness reports zero models no longer refetches `listCodeHarnessModels` forever. An empty remembered list is a finished fetch.

## Validation

- `pnpm vitest run src/code` — 117 passed
- `pnpm vitest run` — 1136 passed (one `ManagedGate` pairing-nudge case failed once under load and passed on retry; unrelated to this change)
- `pnpm build` — `tsc` and Vite production build passed

## Compatibility and risk

None. No persisted or wire-format change. The reducer already carried turn status, duration, usage, error, and diffstat; this change only renders them. `createdAt` is new on the in-memory user item and is filled from `turn.started_at`.

## Tests deleted

- `expect(screen.getByText("list the files")).toBeInTheDocument()` — asserted the old plain-div user bubble; the prompt is now a `UserMessage` article with markdown and a timestamped footer.
- `expect(screen.queryByText("Turn completed")).toBeNull()` — asserted that a turn boundary renders nothing; that silent boundary is the bug this PR removes.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeSessionController.ts
+++ b/crates/tidebreak-desktop/ui/src/code/CodeSessionController.ts
@@ -26,6 +26,12 @@ export type CodeSessionControllerOptions = {
    */
   hydrateTurns?: () => Promise<CodeTurnSnapshot[]>;
   onHydrate?: (turns: CodeTurnSnapshot[]) => void;
+  /**
+   * The snapshot settled, whether it arrived or failed. The transcript hangs
+   * its skeleton on this rather than on `onHydrate`, so a session whose history
+   * could not be read still reaches a state the reader can send from.
+   */
+  onHydrateSettled?: () => void;
 };
 
 /**
@@ -69,6 +75,7 @@ export class CodeSessionController {
         if (this.disposed) return;
       }
     }
+    this.options.onHydrateSettled?.();
     this.connect();
   }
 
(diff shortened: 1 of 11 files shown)


naingthet (commented):
## Summary

Failed turns now render as a critical `role=alert` with the engine error, which is the actual bug this PR fixes. Completed turns close with a quiet seam (duration, diffstat, usage); interrupted turns stay a one-line warning. Hydration is a separate `markCodeSessionHydrated` flip so a snapshot that never arrives still drops the skeleton. The empty-models loop is gone: a remembered `[]` is a finished fetch.

Deleted tests are justified — they asserted the silent boundary and the old plain-div bubble. No blocking issue. Auto-merge is already on; admin-merge once this COMMENT is the only remaining gate.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

PR #2195, opened 2026-08-18
fix(release): keep in-flight notes as a draft prerelease

naingthet:
## Summary
Keep the in-flight desktop release as a **draft prerelease** until assets attach.

The previous freeze published the notes immediately (`draft=false`, `prerelease=true`). That did stop Release Drafter from appending later merges, but GitHub then locked the release as immutable before `gh release upload` ran. Turning the repo setting off does not unlock an already-immutable release, which is why `v0.48.0`–`v0.50.0` stayed empty.

This change still freezes the notes: Release Drafter only updates non-prerelease drafts. The in-flight object stays unpublished so assets can attach, then finalize publishes it after the upload.

## Root cause
`.github/workflows/release.yml` validate PATCHed the snapshot to a published prerelease. GitHub immutable releases then rejected later asset uploads with `HTTP 422: Cannot upload assets to an immutable release.`

## What this does not do
- It does not retry or rewrite the empty immutable prereleases `v0.48.0`, `v0.49.0`, `v0.49.1`, or `v0.50.0`.
- After this lands, the next ship should be the existing `v0.51.0` notes draft.

## Test plan
- [x] `node --test scripts/workflow-security.test.mjs scripts/reconcile-release-drafts.test.mjs scripts/require-release-baseline.test.mjs`
- [ ] After merge, dispatch **Publish desktop release** from `main` with a blank tag so it takes `v0.51.0`
- [ ] Confirm validate marks that release `draft=true, prerelease=true` and does **not** publish it
- [ ] Confirm attach can `gh release upload`
- [ ] Confirm Linux ARM still has `xdg-mime` and Windows ARM still uses clang-cl

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -123,20 +123,20 @@ jobs:
             echo "tag=$RELEASE_TAG"
             echo "version=${RELEASE_TAG#v}"
           } >> "$GITHUB_OUTPUT"
-      - name: Convert the draft to a prerelease while it builds
+      - name: Mark the in-flight draft as a prerelease
         if: ${{ steps.release.outputs.draft == 'true' }}
         env:
           GH_TOKEN: ${{ github.token }}
           RELEASE_ID: ${{ steps.release.outputs.release_id }}
         run: |
-          # A draft keeps collecting release notes from new merges while the
-          # build runs, making the live draft drift away from the frozen notes.
-          # Mark the in-flight release as a prerelease so Release Drafter treats
-          # it as already-published and starts the next draft for any PRs that
-          # merge during this build.
+          # Keep the frozen notes on this draft, but do not publish it yet.
+          # Publishing here makes GitHub treat the release as immutable, so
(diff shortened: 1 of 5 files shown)


naingthet (commented):
## Summary

Validate now PATCHes the in-flight notes to `draft=true, prerelease=true` instead of publishing a prerelease. That keeps GitHub from locking the release before `gh release upload`. Release Drafter still only rewrites non-prerelease drafts, so later merges start a new notes draft. Tests cover draft-prerelease ignore in reconcile, baseline (a draft prerelease is not published), and the workflow contract (`-f draft=true`, no `-f draft=false` in validate).

Does not rewrite the empty immutable `v0.48.0`–`v0.50.0` objects. Next ship is `v0.51.0`.

No blocking issue in the freeze/upload contract. Hold merge until release-policy and the test lane finish. If release-policy still runs main's trusted tests, this may need the same two-step as the clang-cl/xdg-utils policy PRs.

## Issue counts by severity

- bugs: 0
- suggestions: 1
- nits: 0

PR #2196, opened 2026-08-18
test(release): accept JSON booleans when freezing in-flight notes

naingthet:
## Summary
Loosen the trusted release-policy assertions so an in-flight notes freeze may use either `gh -f draft=true` or a JSON body `{draft: true, prerelease: true}`.

This is the first half of the durable fix. `gh -f` stringifies booleans, and GitHub then publishes the in-flight release. The next PR will send JSON booleans and verify the object stayed a draft.

## Test plan
- [x] `node --test scripts/workflow-security.test.mjs`
- [ ] After this lands, follow with the workflow PATCH that uses `--input` JSON and asserts `.draft == true`

### DIFF
--- a/scripts/workflow-security.test.mjs
+++ b/scripts/workflow-security.test.mjs
@@ -1065,8 +1065,11 @@ test("release builds freeze a draft tag from the trusted main workflow", () => {
   assert.match(validateJob, /git merge-base --is-ancestor "\$RELEASE_SHA" origin\/main/);
   assert.match(validateJob, /release-snapshot\.json/);
   assert.match(validateJob, /Mark the in-flight draft as a prerelease/);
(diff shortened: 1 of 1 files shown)


naingthet (commented):
## Summary

Trusted release-policy tests now accept either `gh -f draft=true` / `-f prerelease=true` or a JSON body `{draft: true, prerelease: true}`. That is the right first half of the two-step: main will admit the follow-up PATCH that uses real JSON booleans (`gh -f` stringifies them and GitHub publishes). `draft=false` is still forbidden in validate.

No blocking issue. Hold merge until clippy / test finish, then land this before the workflow PATCH.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

PR #2199, opened 2026-08-18
fix(server): keep attached container heartbeats off the drain

naingthet:
## Summary
Stop a live container drive from awaiting lease renewal inside the same `select!` as the connection drain.

The flaky CI test `heartbeats_the_lease_so_a_long_run_is_not_reaped` already observed a heartbeat, then hung after `gate.release()` with `provider_calls=1`. That is the attached-drive path: heartbeat and model-step accounting both take the shared agent-run claim lock. Awaiting the heartbeat inline parked the drain, so the second model step never completed.

Pre-attach already kept at most one durable fence in flight. The attached loop now does the same.

## Test plan
- [x] `cargo test -p tidebreak-server --lib --locked heartbeat -- --test-threads=1`
- [x] `cargo clippy -p tidebreak-server --lib --locked -- -D warnings`

### DIFF
--- a/crates/tidebreak-server/src/sandbox_container_run.rs
+++ b/crates/tidebreak-server/src/sandbox_container_run.rs
@@ -1682,33 +1682,70 @@ impl SandboxContainerRunner {
                 .max(Duration::from_millis(1));
             let mut heartbeat = tokio::time::interval(heartbeat_interval);
             heartbeat.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
+            // The first interval tick is immediate. Consume it so attach does
+            // not issue a redundant write, then keep at most one durable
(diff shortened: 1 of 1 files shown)


naingthet (commented):
## Summary

Attached-drive lease renewal no longer sits in the same `select!` as the connection drain. The first interval tick is consumed; at most one `renew_or_validate_execution` is in flight. A finished drive waits for that in-flight write, then exits. That matches the pre-attach fence and is the right fix for `heartbeats_the_lease_so_a_long_run_is_not_reaped` parking on the shared claim lock.

No blocking issue. Hold merge until clippy and the workspace `test` lane finish on this head.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

PR #2198, opened 2026-08-18
feat(desktop): redesign code-mode tool and approval cards

naingthet:
## What changed

An approval now leads with the normalized kind, not the harness JSON. A command shows the command in a mono block with a cwd caption. A file write lists the paths. A network request shows the summary. `requested_at` is stamped with `formatMessageTimestamp`; a decided card also stamps `decided_at`. Approved is a success caption. Denied is a warning caption and still shows the feedback.

The raw harness payload stays available behind a collapsed **Harness payload** disclosure, using the same `ScrollableContainer` as before. That keeps [ADR 0033](docs/decisions/0033-code-mode-approvals.md) honesty — the engine's request is still there, it just does not occupy the card before the reader can decide. Approve, Deny, and deny-feedback are unchanged. The card stays inside the `isolatedCard` row PR 2 added.

Tool cards drop the raw `max-h-40` `<pre>` for `ToolOutputPreview`: a clamped mono block, a Show-N-more expander, and a hover copy button. The cwd/detail line stays. Failed tools get a `border-critical-border` tint. Running tools stay expanded by default.

The header pending-approval badge is a button. It scrolls the first `[data-code-approval-state="pending"]` card into view. The workspace page diff is limited to `PendingApprovalBadge` so it does not collide with other work on that file.

## Validation

- `pnpm vitest run src/code` — 122 passed
- `pnpm vitest run` — 1141 passed
- `pnpm build` — `tsc` and Vite production build passed

## Compatibility and risk

None. No persisted or wire-format change. The approval snapshot already carried `kind`, `requested_at`, `decided_at`, and `harness_raw_json`; this change only changes what the card puts first.

## Tests deleted

- `it("renders the harness payload rather than a paraphrase")` — the card no longer leads with the raw JSON.
- `expect(screen.getByText(/"tool_name": "Write"/)).toBeInTheDocument()` — asserted the Write payload was visible by default.
- `expect(screen.getByText(/"file_path": "\/workspace\/probe.txt"/)).toBeInTheDocument()` — asserted the same payload's path field was visible by default.

Replaced by tests that the command (or path list) leads, the disclosure starts closed, deny-feedback still submits, and both timestamps render.

PR #2200, opened 2026-08-18
feat(desktop): tighten code-mode card visual language

naingthet:
## What changed

A pass over #2198 against the code-mode visual bar. The first PR shipped the hierarchy; this one removes the extra signals and puts the remaining elements on one type ramp.

**Removed**
- Failed-tool `border-critical-border`. The Failed badge is already the status. A tint on the same fact is a second signal.
- A second decided caption under the body. Approved / Denied now sits once, as 11px meta beside the title, in the success or warning quad. Denial feedback stays as uncolored body copy so it is not a second warning.
- Native `<details>` for the harness payload. It popped open. The payload is now a button plus a 140ms ease-out height reveal that respects `prefers-reduced-motion`.

**What each remaining element answers**
- Title: the question being asked.
- Kind body (mono command + cwd, path list, or network summary): the thing being decided, in the engine's own terms.
- Timestamps: when it was requested, and when it was decided. 11px meta, no status color.
- Harness payload disclosure: ADR 0033 honesty. The raw request is still there; it is not the first thing you read.
- Approve / Deny: the decision. Unchanged.
- Tool badge: the only status color on a tool card (info / success / warning / critical).
- `ToolOutputPreview` clamp, expander, and copy: long output stays glanceable and copyable.
- Middle-truncation on commands and paths: the tail is what you need when the line runs out of room.
- Header pending-approval button: jump to the first parked card. Scrolls with `block: nearest` so it does not yank the transcript.

**Motion and layout**
- Reveals (payload, chevron, copy hover) are 140ms ease-out and `motion-reduce:transition-none`.
- Expanding a tool body sets `overflow-anchor: none` so the browser does not steal the scroll.
- Status updates do not retint or rebadge the same fact. Running stays `defaultExpanded`; completing does not collapse.

**Type**
- 13px body (title, command, network summary, feedback).
- 11px meta (cwd, timestamps, state caption, payload, tool header, detail line).
- Mono for command, path, cwd, and output.

## Validation

- `pnpm exec tsc --noEmit` passed
- `pnpm vitest run src/code` passed
- `pnpm vitest run` — 1141 passed

## Compatibility and risk

None. No persisted or wire-format change. Follow-up to #2198.

## Tests deleted

None in this pass. The #2198 deletions stand. Assertions now target `aria-expanded` on the payload button instead of a `<details open>` attribute, matching the disclosure that no longer pops.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/ToolCardShell.tsx
+++ b/crates/tidebreak-desktop/ui/src/ToolCardShell.tsx
@@ -61,10 +61,10 @@ export function ToolCardShell({
         aria-controls={bodyId}
         onClick={() => setExpanded((current) => !current)}
       >
-        <span className="text-muted-foreground flex min-w-0 items-center gap-1.5 text-xs font-medium">
+        <span className="text-muted-foreground flex min-w-0 items-center gap-1.5 text-[11px] font-medium">
           <ChevronDown
             className={cn(
-              "size-3.5 shrink-0 transition-transform",
+              "size-3.5 shrink-0 transition-transform duration-[140ms] ease-out motion-reduce:transition-none",
               !expanded && "-rotate-90",
             )}
             aria-hidden="true"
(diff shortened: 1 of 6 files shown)


naingthet:
Closing so the boxed tool-call restyle does not land. The coordinator replaced that treatment with a boxless inline line. Approvals stay a card. A new PR will follow after the spec is confirmed.

PR #2204, opened 2026-08-18
feat(desktop): render code-mode tools as inline lines

naingthet:
## What changed

Code-mode tool calls are a boxless line. Each call is one row: a glyph, a bold constant verb (`Command run`, `File edited`, `File read`, `Search`, or `Command denied`), and a muted mono subject with middle truncation. The right slot holds duration (live elapsed while running), an exit code when the preview reports one, one status glyph, and a quiet chevron. The whole line is the click target and exposes `aria-expanded`.

Failed and denied calls open. Success stays closed. A running call shows a spinner plus elapsed time and streams the last twelve lines of output at a fixed height, so new lines do not grow the row.

Expanded output is plain indented mono. There is no container and no left rail. Long output clamps at 12 lines behind `· · · N more lines`, with a hover copy control.

Approvals stay a card. The kind leads. The harness JSON stays behind **Harness payload**. Approved and Denied are one caption. The header badge jumps to the first pending card.

## What each element answers

- Constant verb: what kind of call this is, without changing tense as status changes.
- Muted mono subject: the command or path being decided or run.
- Duration / elapsed: how long the live call has been running, or how long it ran.
- Exit code: only when the preview states one. The wire `tool_completed` event does not carry `exit_code`.
- Status glyph: the only color on the line (info / success / warning / critical).
- Chevron: whether the output is open.
- 12-line clamp: output stays glanceable.
- Approval card: a decision surface, not a log line.

## Follow-ups

- Chat exec calls (`ToolCallCard`) stay on the boxed shell. Applying this line there is a separate change.
- Per-call `+N −M` is not on `tool_completed`. It waits on a per-call diffstat on the wire or a reducer join from `file_changed`.

## Validation

- `pnpm vitest run src/code` — 142 passed
- `pnpm vitest run` — 1162 passed
- `pnpm exec tsc --noEmit` passed

## Compatibility and risk

None on the wire. `startedAt` and `durationMs` are in-memory on the tool item. Replay leaves them null so a reopened session does not invent a duration.

## Tests deleted

- `it("renders the harness payload rather than a paraphrase")` — the approval card no longer leads with raw JSON.
- `expect(screen.getByText(/"tool_name": "Write"/)).toBeInTheDocument()` — asserted the Write payload was visible by default.
- `expect(screen.getByText(/"file_path": "\/workspace\/probe.txt"/)).toBeInTheDocument()` — asserted the same payload's path field was visible by default.
- `getByRole("status", { name: "Bash succeeded" })` — the line is labeled by the constant verb, not the harness tool name.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/ToolCardShell.tsx
+++ b/crates/tidebreak-desktop/ui/src/ToolCardShell.tsx
@@ -61,10 +61,10 @@ export function ToolCardShell({
         aria-controls={bodyId}
         onClick={() => setExpanded((current) => !current)}
       >
-        <span className="text-muted-foreground flex min-w-0 items-center gap-1.5 text-[11px] font-medium">
+        <span className="text-muted-foreground flex min-w-0 items-center gap-1.5 text-xs font-medium">
           <ChevronDown
             className={cn(
-              "size-3.5 shrink-0 transition-transform duration-[140ms] ease-out motion-reduce:transition-none",
+              "size-3.5 shrink-0 transition-transform",
               !expanded && "-rotate-90",
             )}
             aria-hidden="true"
@@ -74,10 +74,7 @@ export function ToolCardShell({
               repeat it. The native tooltip gives the whole line back to anyone
               who hovers, for the titles that are plain text to begin with. */}
           <span
-            className={cn(
-              typeof title === "string" ? "truncate" : "min-w-0",
-              titleClassName,
-            )}
+            className={cn("truncate", titleClassName)}
             title={typeof title === "string" ? title : undefined}
           >
             {title}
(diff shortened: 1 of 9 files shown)


naingthet (commented):
## Summary

Tool calls become a boxless line: constant verb, muted mono subject, duration/elapsed, optional exit code, one status glyph, chevron. Failed/denied open; success stays closed; running streams a 12-line tail. Approvals stay a card with kind leading and harness JSON behind **Harness payload**. Duration is in-memory only so replay does not invent it. Tests that asserted default-visible JSON payload were correctly deleted.

No blocking issue. Hold merge until desktop UI finishes.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

PR #2205, opened 2026-08-18
feat(desktop): theme the code terminal from app tokens and add keyboard resize

naingthet:
## Summary

The code-mode terminal now follows the app theme and is resizable from the keyboard.

xterm used hardcoded zinc hex values, so the pane stayed dark-zinc in light mode and ignored theme flips. The drawer height lived in `useState` and reset on every remount, and the separator was pointer-only.

## Theme

`readXtermTheme()` reads `--background`, `--foreground`, `--muted-foreground`, and the status quads from `getComputedStyle(document.documentElement)`. ANSI red/green/yellow/blue come from `--critical` / `--success` / `--warning` / `--info`. Magenta has no token, so it reuses the critical foreground; cyan reuses info.

A `MutationObserver` on `document.documentElement`'s `class` reapplies `terminal.options.theme` when `theme.ts` toggles `.dark`.

### oklch in the xterm renderer

Headless Chrome 151 (macOS) accepts `oklch(...)` as a canvas `fillStyle` and paints the correct pixel. `getComputedStyle` still returns the `oklch(...)` string, not `rgb()`. xterm's parser uses that same canvas path, so oklch would work in this Chromium. WKWebView was not available in this checkout.

`resolveCssColor()` still converts tokens to `#rrggbb` before they reach xterm: hex and `rgb()` pass through, other functions go through a live element's computed `color`, then a 1×1 canvas `fillStyle` plus `getImageData` when the fill style stays an `oklch(...)` string. That keeps the renderer on hex if a WebKit build rejects oklch.

## Drawer

The separator is focusable (`tabIndex={0}`) with `role="separator"` and `aria-valuenow` / `min` / `max`. ArrowUp / ArrowDown change height by 32px. Height is stored per workspace in `CodeUiStore` (`tidebreak.code-terminal-drawer-heights` in localStorage).

Open and close clip the reserved band with a 140ms ease-out `max-height` transition. The inner xterm host keeps a fixed height so streamed output does not reflow. `prefers-reduced-motion` skips the reveal.

## Chrome added

| Element | Need it answers |
| --- | --- |
| Focusable resize separator | Keyboard-first: the pointer drag already existed; ArrowUp / ArrowDown is the key path. |
| `aria-label` on the xterm host | The host is an unlabeled region once the drawer header names the surface. |

No second accent, gradient, or shadow. Status color comes only from the existing quads.

## Tests

- Theme derives from stubbed CSS variables and updates when the root class flips (`TerminalPane.dom.test.tsx`).
- Keyboard resize changes `aria-valuenow` and persists per workspace (`TerminalDrawer.dom.test.tsx`).

`pnpm vitest run src/code`, full `pnpm vitest run`, and `pnpm build` passed locally after rebasing onto `origin/main`.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeUiStore.ts
+++ b/crates/tidebreak-desktop/ui/src/code/CodeUiStore.ts
@@ -10,6 +10,7 @@ import {
 const REVIEW_SIDEBAR_OPEN_KEY = "tidebreak.code-review-sidebar-open";
 const LAST_CREATE_KEY = "tidebreak.code-last-create";
 const WORKSPACE_SORT_KEY = "tidebreak.code-workspace-sort";
+const TERMINAL_DRAWER_HEIGHTS_KEY = "tidebreak.code-terminal-drawer-heights";
 
 /** What the reader picked the last time they created a workspace. */
 export type CodeCreateDefaults = {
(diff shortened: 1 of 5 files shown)


naingthet (commented):
## Summary

Token-driven `readXtermTheme()` plus a class MutationObserver is the right theme path. `oklch` resolution through computed style then canvas is the honest xterm constraint. Height lives in `CodeUiStore` / localStorage per workspace. The separator is a real slider (`tabIndex`, `aria-valuenow`, ArrowUp/Down). 140ms clip honors reduced motion. Tests pin keyboard persist and theme flip.

Hold merge until desktop UI finishes on this head.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

PR #2202, opened 2026-08-18
feat(desktop): redesign the code-mode sidebar into a command-center rail

naingthet:
## Summary

Code mode's rail is now a dense command-center list: compact cards, a shared Chat/Code switch, and a right-click command list. Viewing a workspace no longer reshuffles the list.

## Card anatomy

Each expanded workspace card answers one question per mark. Compact mode keeps the existing flat buttons.

| Element | Need it answers |
| --- | --- |
| Attention dot on the title | Is this workspace asking for you? Failure is the same critical dot; the prompt lives in the tooltip. The old "the engine turn failed" text pill is gone. |
| Title at 13.5px medium | Which workspace is this? |
| Repo chip (swatch + name) | Which repository does this worktree belong to? The swatch is identity, not status. |
| Mono middle-truncated branch | Which branch? The tail of `tidebreak/…` stays readable. |
| PR glyph (`PR_ICON_TONE_CLASSES`) | Is there a pull request, and is it open, draft, merged, or closed? Number and title stay on the tooltip. |
| Terminal glyph | Is the terminal drawer open on the workspace you are viewing? |
| Pending-approval glyph | Is a structured approval waiting? Distinct from a lifecycle failure. |
| Nested session row | Is a session live? Harness mark + "Running" / "Needs you" + compact age from `created_at`. The chevron is reserved for later subagent nesting. |

Status color comes only from the semantic quads (and the existing merged purple on the PR icon). Reveals use 150ms ease-out and honor `prefers-reduced-motion`. Data updates do not animate. Every pointer control is a button, radio, or menu item.

## Sort modes

The Workspaces header cycles three persisted modes (`CodeUiStore` / localStorage):

- **By repo** — catalog repo order, `created_at` ascending inside each repo.
- **By status** — needs you → running → PR open → done-unreviewed → idle → archived. A digest may move a row only when the rank itself changes.
- **By created** — flat, newest first. Archived rows stay off this list.

## Reshuffle fix

Opening a workspace used to prepend it in `CodeCatalogStore.upsertWorkspace`. The rail also trusted that array order. Presentation now sorts only on `created_at`, repo catalog order, and status rank. `upsertWorkspace` updates in place. Tests pin "selecting a workspace does not reorder" at both the catalog and `arrangeWorkspaces` layers.

## Segmented Chat/Code switch

[...]

PR #2206, opened 2026-08-18
feat(desktop): finish the code workspace chrome and composer effort control

naingthet:
## Summary

Finish the code workspace header and mid-session composer chrome. Each new mark answers one question. Nothing here is decorative.

## What each element answers

| Element | Need it answers |
| --- | --- |
| Header skeleton bars | Is this page still loading? The old fallback printed `Workspace` and the raw repo id, which reads as a finished title. Two bars hold the title and repo slots until those strings exist. |
| Centered error + **Retry** | Did the loader fail, and can you try again? A permanent `p` was a dead end. Retry re-runs the same load. |
| Warning dot on the lifecycle cluster | Did the engine drop journal frames? `unrecognized_event_count > 0` is a warning-quad dot, not a text pill. The count and "transcript may be incomplete" live in the lifecycle tooltip (decision 0031). The stray comment that used to sit on `CodeSessionPane` now lives on this mark. |
| Mono worktree-path chip | Which directory is this session in? Middle-truncated so the tail stays readable. Click copies, using the same copy-state pattern as `ClipboardCopyButton`. Reveal / open-in-editor needs a host capability and is a follow-up. |
| Review-sidebar toggle | Is the inspector open? `PanelRight`, `aria-pressed`, tooltip with the `⌘I` keycap. Same one-keystroke rule as the terminal (`⌘J`). |
| Harness version in the lifecycle tooltip | Which engine build is this session? Visible chrome stays `Idle`; the tooltip adds `Claude Code 2.1.233`. |
| Overflow `...` menu | Where do rename, pin/clear, quick actions, and archive live? The bare **Archive** button was one action pretending to be the whole menu. Archive is last, destructive, and separated. |
| Disabled permission-mode tooltip | Why can you not change the mode mid-session? There is no `POST /code/sessions/{id}/permission-mode` route. The trigger is wrapped so a disabled button can still host the tooltip. |

Status color comes only from the semantic quads. Reveals use 150ms ease-out and honor `prefers-reduced-motion`. Data updates do not animate.

## Shared command list

`origin/thet/code-sidebar-command-center` already landed as #2202, including `ApiClient.patchCodeWorkspace`. This PR extends that exportable list rather than adding a second copy:

- `workspaceCommands()` — card context menu. Adds pin/clear when a session is present.
- `workspaceHeaderCommands()` — header overflow: Rename…, pin/clear, `Run: <name>` per repo `quick_actions`, Archive last.
- `useWorkspaceCardCommands()` runs both surfaces. A quick-action toast is success or error from the exit code, with **View output** opening stdout/stderr in `ToolOutputPreview`.

## Reasoning effort (PR 4b)

Read first: `POST /code/sessions/{id}/turns` (`SubmitTurnBody` in `routes/code/types.rs`) accepts only `message` and `model` (`deny_unknown_fields`). `TurnInput` is `{ text, model }`. Claude, Codex, and Grok adapters pass a model flag only. Grok advertises `reasoning_levels: supported` and documents `--reasoning-effort`, but `compose_print_plan` does not compose that flag.

Nothing on the wire can carry effort today. Shipping a gated selector would be dead UI, so this PR does not add the control. A later ADR can add a turn field (or an honest adapter argv) before the composer offers one.

## Follow-ups

- Host capability to reveal the worktree or open it in an editor.
- Turn-payload field (or adapter argv) for reasoning effort, then the composer control.

## Tests

Replaced, not piled on:

- Header loading asserts skeleton bars and the absence of the literal `Workspace` / repo-id fallback.
- `unrecognized_event_count > 0` asserts the warning dot.
- Review-sidebar toggle flips `CodeUiStore.reviewSidebarOpen`.
- Quick action `lint` surfaces `lint exited 1` on an error toast with **View output**.
- Archive still leaves for the repo page, now through the overflow menu.
- Composer: disabled permission trigger is hoverable and states why.

No effort-control test: the control is not shipped.

## Validation

- `pnpm vitest run src/code` — 148 passed
- `pnpm vitest run` — 1166 passed
- `pnpm build` — `tsc` and Vite production build passed

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeComposer.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeComposer.dom.test.tsx
@@ -31,6 +31,25 @@ describe("CodeComposer", () => {
     expect(screen.getByRole("button", { name: "Permissions: Ask" })).toBeInTheDocument();
   });
 
+  it("explains why the session permission mode cannot change", async () => {
+    const user = userEvent.setup();
+    render(
+      <CodeComposer
+        running={false}
+        permissionMode="ask"
+        onSend={vi.fn()}
+        onInterrupt={vi.fn()}
+      />,
+    );
+
+    const trigger = screen.getByRole("button", { name: "Permissions: Ask" });
+    expect(trigger).toBeDisabled();
+    await user.hover(trigger.parentElement!);
+    expect(await screen.findByRole("tooltip")).toHaveTextContent(
+      "Set when the session started — start a new session to change it",
+    );
+  });
+
   it("queues a follow-up written while a turn is running", async () => {
     const onSend = vi.fn().mockResolvedValue(QUEUED);
     render(
--- a/crates/tidebreak-desktop/ui/src/code/CodeComposer.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeComposer.tsx
@@ -27,6 +27,7 @@ import {
   groupCodeModelOptions,
   type CodeModelOption,
   PERMISSION_MODE_UNAVAILABLE_REASON,
+  SESSION_PERMISSION_MODE_LOCKED,
 } from "./labels";
 
 const MODES: CodePermissionMode[] = ["plan", "ask", "auto", "allow"];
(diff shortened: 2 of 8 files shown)


naingthet (commented):
## Summary

Header chrome is the right finish: skeleton instead of a fake title, Retry on load failure, warning dot for dropped journal frames, worktree chip with copy, `⌘I` inspector toggle, overflow menu with Archive last. Permission mode stays disabled mid-session with a tooltip because there is no mode-change route. Effort control is correctly *not* shipped — the turn body cannot carry it.

No blocking issue. Hold merge until desktop UI finishes.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

naingthet:
Rebased onto `origin/main` after #2205 (`1a606ccc`). One commit replayed cleanly; no conflicts.

Head is now `193225b9`. Desktop UI already passed on the previous head; this is a replay of that same change.

PR #2208, opened 2026-08-18
test(server): stabilize sandbox container timing tests

### LINKED ISSUE #2191: Flaky: sandbox_container_run heartbeat-lease and attached-cancellation tests
Both failed on an unrelated docs/assets-level PR (#2190, run 32155582785), 974 others passed.

```
---- sandbox_container_run::tests::heartbeats_the_lease_so_a_long_run_is_not_reaped ----
panicked at crates/tidebreak-server/src/sandbox_container_run/tests.rs:2157:
lease expiry 2026-08-18 15:45:41.348999 UTC did not move past 2026-08-18 15:45:41.348999 UTC after an observed heartbeat

---- sandbox_container_run::tests::an_attached_cancellation_is_acknowledged_and_torn_down ----
panicked at crates/tidebreak-server/src/sandbox_container_run/tests.rs:3194:
test completed within its time bound: Elapsed(())
```

The heartbeat test was previously stabilized in #2181 and has recurred. Note the failure shows the expiry *equal* to the prior value: if the assertion is strict-greater while a heartbeat landing within the clock's resolution rewrites the same microsecond expiry, the equal case is legitimate liveness, not a missed heartbeat — a lead for whoever picks this up. The cancellation test looks like a plain time-bound squeeze under CI load.

naingthet:
## Summary

- wait for a heartbeat completed after the lease snapshot instead of accepting a stale absolute heartbeat count
- reduce unrelated lease/fence write contention in cancellation and terminal-accounting tests
- replace polling of a transient cancellation-finalization state with an explicit successful-renewal signal
- give diagnostic inner waits enough outer hang-guard headroom under parallel CI load

## Root cause

The heartbeat test could sample the lease after a heartbeat had already completed, then accept that same old heartbeat count and compare the lease value against itself. Several cancellation tests also combined short leases, high-frequency SQLite writes, and outer timeouts shorter than their diagnostic state waits.

A high-parallelism stress audit additionally reproduced `cancellation_finalization_survives_accounting_failure_past_execution_lease`: it depended on polling a transient durable state within a two-second scheduler window rather than observing the renewal operation directly.

## Validation

- `git diff --check`
- entire `sandbox_container_run::tests` module
- 30 consecutive module runs with `--test-threads=32` (1,230 executed tests, zero failures)

Closes #2191

### DIFF
--- a/crates/tidebreak-server/src/sandbox_container_run/tests.rs
+++ b/crates/tidebreak-server/src/sandbox_container_run/tests.rs
@@ -305,6 +305,8 @@ struct TerminalFaultStore {
     fail_accounting: AtomicBool,
     accounting_failure_observed: Notify,
     accounting_calls: AtomicUsize,
+    cancellation_finalization_renewed: Notify,
+    cancellation_finalization_renewals: AtomicUsize,
 }
 
 #[derive(Clone, Copy, PartialEq, Eq)]
@@ -344,6 +346,8 @@ impl TerminalFaultStore {
             fail_accounting: AtomicBool::new(false),
             accounting_failure_observed: Notify::new(),
             accounting_calls: AtomicUsize::new(0),
+            cancellation_finalization_renewed: Notify::new(),
+            cancellation_finalization_renewals: AtomicUsize::new(0),
         })
     }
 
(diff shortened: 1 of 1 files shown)


naingthet (commented):
## Summary

`00f33f9a` is test-only. The heartbeat contract now waits for a tick after the lease snapshot instead of accepting a stale absolute count. Cancellation tests wait on an explicit successful-renewal notify, drop unrelated lease/fence write contention, and give the outer hang guards enough room under parallel CI.

Hold merge until desktop test, clippy, test, and postgres state machine finish on this head.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

PR #2209, opened 2026-08-18
feat(server): carry rich turn input behind harness capabilities

naingthet:
Implements ADR 0045 (`docs/decisions/0046-rich-turn-input-rides-harness-capabilities.md`).

## Scope

- Add `image_input` and `slash_commands` to `HarnessCaps`. Every adapter states them: Claude Code, Codex, OpenCode, and Grok declare `Unknown`. The scripted adapter can declare `Supported` for route tests only.
- Add optional `HarnessProbe.commands` (`name`, one-line description). Every adapter returns an empty list.
- Accept optional `attachments: [{blob_id, media_type}]` on `POST /code/sessions/{id}/turns`. The route rejects with `unsupported_attachment` naming the harness unless that session's adapter declares `image_input: Supported`. Accepted turns journal bounded `{blob_id, media_type, byte_len}` and pin those blobs through `code_turn_attachment`.
- Serve `GET /code/workspaces/{id}/tree?query=&limit=` from the worktree layer: git-tracked plus untracked-unignored paths, name-matched, default 50 / max 200, never file contents.
- Composer `@` completion uses the tree route through the existing mention popup and inserts a plain relative path. The `/` popup lists `HarnessProbe.commands` when that list is non-empty; it does not appear today. Free-typed `/` text still submits verbatim.

## Exclusions

- No attachment UI affordance. Caps are `Unknown` everywhere, and the ADR requires refusal-with-reason rather than a hidden dead end. This PR ships the wire, route, and gating only.
- No Tidebreak-owned slash vocabulary, no command interception, no client-side `@` content resolution, no non-image attachment types.

## Validation

- Route rejection/acceptance against the scripted adapter.
- Adapter honesty: no `image_input: Supported` without a fixture directory that contains an image round-trip capture.
- Tree route bounds, ignore rules, and no contents.
- Composer verbatim-slash test.


naingthet (commented):
## Summary

`536bd6ac` implements ADR 0046 (already on main; no ADR-number collision). `HarnessCaps` stays exhaustive: `image_input` and `slash_commands` have no `Default`, and Claude Code, Codex, OpenCode, and Grok all declare `Unknown`. The turn route rejects attachments unless the probed session is `Supported`, with `unsupported_attachment` naming the harness. Probe commands stay empty on every production adapter.

Hold merge until rustfmt/clippy, desktop UI, desktop test, test, and postgres state machine finish on this head.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

PR #2210, opened 2026-08-18
ci: run required checks on merge queue groups

naingthet:
## Summary

Prepare Tidebreak for a batched GitHub merge queue on `main`.

CI now listens for `merge_group`. The queue builds a temporary branch of ready pull requests, and required checks have to report on that branch or the queue never drains.

- Diff merge groups against their base the same way a pull request diffs against `main`, so a docs-only group still skips Rust.
- Keep the `semantic PR title` check name on merge groups. The group has no title payload, so the job succeeds after the pull request already passed that check.
- Document the queue in `CONTRIBUTING.md`.

This pull request does **not** turn the queue on. Merge it first. After it is on `main`, enable a squash / all-green queue with a short grouping window (about 2 minutes, max 8). Leave `HEADGREEN` off.

Do not stack pull requests that target each other into this queue. Every queued pull request must target `main`.

## Test plan

- [x] `node --test scripts/*.test.mjs` (includes the new merge-queue workflow assertion)
- [ ] GitHub CI on this pull request
- [ ] After merge: enable the `main` merge queue and land a pair of ready pull requests through it

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -6,6 +6,10 @@ on:
   pull_request:
     types:
       [opened, reopened, synchronize, edited, labeled, unlabeled, ready_for_review]
+  # Speculative merge groups from the main merge queue. Required checks must
+  # report here or the queue never drains. Scope detection diffs the group
+  # against its base the same way a pull request diffs against main.
+  merge_group:
   # Backstop the deliberately scoped post-merge capability lanes even during a
   # week with no relevant merge. The non-PR full-validation path below makes a
   # scheduled run exercise every lane.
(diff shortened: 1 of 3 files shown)


naingthet (commented):
## Summary

`merge_group` on `ci.yml` is the right prerequisite. Scope detection diffs the group against its base the same way a PR diffs `main`, so a docs-only group still skips Rust. The semantic-title job keeps the required check name and no-ops on groups. The workflow-security test pins the trigger and the SHA wiring.

This PR does not enable the queue. Hold merge until clippy / test / desktop UI finish, then admin-squash. Enable the queue only after this is on `main`.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 1

### Nits

`CONTRIBUTING.md` states `main` already uses a batched merge queue. That is true only after the settings flip. Fine if this lands immediately before you turn it on; until then the page is ahead of the repo.

PR #2219, opened 2026-08-18
feat(desktop): hide the sidebar entirely when collapsed

naingthet:
Collapsing the sidebar now removes it from the layout instead of shrinking it to an icon rail. A single expand button stays pinned beside the macOS traffic lights (at the plain left edge elsewhere) so the way back never disappears with the rail.

## What changes

- The rail unmounts on collapse rather than going icon-only. The compact width, the per-component `isCompact` branches, and the compact-rail CSS overrides all come out with it.
- The shell now owns the expanded-width CSS var (`useSyncSidebarWidthCssVar`) and renders a new `SidebarExpandStrip` — both have to outlive the rail's unmount. The strip doubles as the window's drag region where the macOS overlay titlebar sits.
- `SegmentedControl` loses its `compact` prop (only the sidebar used it).

## Verification

- `pnpm --dir ui exec tsc --noEmit` clean.
- `pnpm --dir ui test` — 1182/1182 pass.

## Notes

- Net −122 lines; the icon-rail path was the larger half of the primitives.
- The `toggle-sidebar` shortcut (Cmd/Ctrl+B) still works; it now flips between rail-on and rail-off.

Auto-merge is not enabled; this is a visual change worth a look before it lands.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/AppShell.tsx
+++ b/crates/tidebreak-desktop/ui/src/AppShell.tsx
@@ -43,6 +43,8 @@ import { useInterfaceZoom } from "./InterfaceZoom";
 import { Logomark } from "./Logomark";
 import { ManagedGate } from "./ManagedGate";
 import { resolvedRoleKey } from "./ModelSelection";
+import { SidebarExpandStrip } from "./sidebar/SidebarExpandStrip";
+import { useSyncSidebarWidthCssVar } from "./sidebar/primitives";
 import { Titlebar } from "./Titlebar";
 import { WindowDragStrip } from "./WindowDragStrip";
 import { useActiveChatId } from "./useActiveChatId";
@@ -130,6 +132,10 @@ export function AppShell() {
   const desktopUpdates = useDesktopUpdates();
   const desktopNavigation = useDesktopNavigation();
   const zoom = useInterfaceZoom();
+  // The expanded rail width is published by the shell, not the rail itself, so
+  // the value survives a collapse — the expand strip and the titlebar both
+  // depend on it.
+  useSyncSidebarWidthCssVar();
   // Read at keydown rather than subscribed to: which mode a shortcut fires in
   // is the route's answer, and the shell has no other reason to re-render on
   // every navigation.
@@ -696,6 +702,7 @@ export function AppShell() {
               navigation={desktopNavigation}
             />
           )}
+          <SidebarExpandStrip macOverlay={macOverlayTitlebar} />
           <ComputerUseIndicator />
           {/* Each route renders its own rail beside its content — see RouteFrame. */}
           <div className="app-body">
--- a/crates/tidebreak-desktop/ui/src/code/CodeModeSwitch.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeModeSwitch.tsx
@@ -2,7 +2,6 @@ import { FolderGit2, MessageSquare } from "lucide-react";
 import { useNavigate, useRouterState } from "@tanstack/react-router";
 
 import { SegmentedControl } from "@/components/ui/segmented";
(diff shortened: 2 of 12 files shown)


naingthet (commented):
## Summary

The compact-rail deletion is the right cut: `Sidebar` unmounts on collapse, the compact CSS/tooltip path goes with it, and `useSyncSidebarWidthCssVar` correctly moves to the shell so the var outlives the rail.

Do not merge this head.

1. **Titlebar still mounts when the rail is gone.** `AppShell` always renders `Titlebar` under `nativeTitlebar`. `.titlebar` is `position: absolute; width: var(--sidebar-expanded-width); background: var(--page-background)`. On collapse the hook *removes* the inline var, so the titlebar falls back to the stylesheet default of 280px and paints a 40px bar over the conversation. `SidebarExpandStrip` (120px, same `z-index: 40`) sits on top of that leftover chrome. Either stop rendering `Titlebar` while collapsed, or give the collapsed titlebar a width that is not the expanded rail.

2. **desktop UI is still running.** Wait for it. The PR is also 5 commits behind `main` (`#2213`, `#2214`, `#2217`, `#2218`, plus whatever else landed). Rebase after the titlebar fix.

Cmd/Ctrl+B still flipping `sidebarCollapsed` is fine. `Sidebar`'s `if (collapsed) return null` is after its hooks.

## Issue counts by severity

- bugs: 1
- suggestions: 0
- nits: 0

naingthet (commented):
Rebased onto `origin/main` (resolved `CodeSidebar` keeping hide-on-collapse) and unmounted `Titlebar` while the rail is collapsed.

The leftover chrome was the absolute 40px titlebar sized to `--sidebar-expanded-width`. Collapse cleared that var, so it fell back to 280px and painted over the conversation under the expand strip.

Verified on `a69dd066`:
- `Titlebar` mounts only when the native host is up and the sidebar is expanded.
- `Sidebar` still unmounts on collapse; `SidebarExpandStrip` stays as the way back.
- Cmd/Ctrl+B still flips `sidebarCollapsed` and brings the rail and titlebar back.

Did not merge.

naingthet (commented):
## Summary

The titlebar bug is fixed on this head. `AppShell` now renders `Titlebar` only when `nativeTitlebar && !sidebarCollapsed`, `SidebarExpandStrip` returns null while the rail is up, and `useSyncSidebarWidthCssVar` still clears `--sidebar-expanded-width` on collapse so the 280px stylesheet fallback cannot paint leftover chrome. The new `AppShell` test covers unmount plus Cmd/Ctrl+B bringing the rail and Back button back.

Rebase onto `main` succeeded (`mergeable`). Do not merge until **desktop UI** finishes green.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

naingthet (commented):
## Summary

This head (`013d8997`) dropped the titlebar-unmount fix that was on `a69dd066`. Auto-merge is off until that is restored.

`AppShell` again mounts `Titlebar` whenever `nativeTitlebar` is true:

```
{nativeTitlebar && (
  <Titlebar macOverlay={macOverlayTitlebar} navigation={desktopNavigation} />
)}
```

`.titlebar` is still `position: absolute; width: var(--sidebar-expanded-width)`. Collapse still *removes* that inline var, so the bar falls back to the 280px stylesheet default and paints a 40px strip over the conversation. `SidebarExpandStrip` (120px, same `z-index: 40`) sits on top of leftover chrome. Restore `nativeTitlebar && !sidebarCollapsed` (or size the collapsed titlebar to something that is not the expanded rail).

The AppShell test that covered unmount plus Cmd/Ctrl+B bringing the rail and Back button back is also gone from this rebase. Put it back with the render-gate.

Do not merge this head.

## Issue counts by severity

- bugs: 1
- suggestions: 0
- nits: 0

PR #2211, opened 2026-08-18
feat(desktop): add nested files explorer and fix OpenCode session create

naingthet:
## Summary

OpenCode `POST /session` rejects `{ modelID }`. This maps the selected model to `{ providerID, id }` so a new workspace session starts.

Opening a file or a per-turn diff now adds a tab in the center pane. **Chat** stays first and stays open. The **Files** inspector is a nested explorer with name search (Cmd+F / Ctrl+F) and include/exclude globs.

The tree route stays paths-only (`{ paths, query }`). File text is a separate `GET /code/workspaces/{id}/blob` that Monaco 0.56.0 (`@monaco-editor/react` 4.7.0) reads. Third-party notices are regenerated.

## Also in this change

- The new-workspace repo picker stays enabled when you already have a repo.
- You can search the model list.
- The composer stays at the bottom. Reasoning effort sits next to the model.
- Image attach is in the composer; images stay local until a harness declares `image_input`.
- Package-cache warmup runs after bind so startup is shorter.

## Test plan

- [ ] Create an OpenCode workspace session. It should not 400 on `POST /session`.
- [ ] Open a file from **Files**. It should open as a center tab in Monaco.
- [ ] Open a turn diff. It should open as a center tab.
- [ ] Search files with Cmd+F and include/exclude globs.
- [ ] Confirm the repo picker is enabled when you open **New workspace** from a repo.

PR #2220, opened 2026-08-18
refactor(desktop): normalize code-mode spacing and type ramp

naingthet:
R1 of the code-mode refinement stack. It was built on #2211 (`thet/fix-engine-session-launch`) and audits that PR's new surfaces too; #2211 has since merged, so this targets `main` directly. One concern only: spacing and typography. No behavior, no restructuring, no new UI elements — the diff is classes, one `title` helper, and hover-recovery attributes.

> Replaces #2218, which was opened against #2211's branch and merged there after that branch had already squashed into `main`, leaving the work stranded. These are the same four commits, cherry-picked onto `main`. Nothing to clean up on `thet/fix-engine-session-launch`; it is a merged branch.

## The ramp this settles on

Code mode was carrying six text sizes with no rule for which was which (`text-[10px]`, `text-2xs`, `text-[11px]`, `text-xs`, `text-[13px]`, `text-[13.5px]`, `text-sm`). Four now, each with a job:

- **14px `text-sm`** — prose and panel titles: messages, dialog copy, form labels, "Files", "Diff", "Terminal", "Git".
- **13.5px `text-[13.5px]`** — dense row body: transcript tool lines and their output, what an approval is asking for, the sidebar card title.
- **12px `text-xs`** — secondary: captions, notes, empty states, explorer and diff rows, header sub-labels.
- **11px `text-[11px]`** — meta: durations, counts, timestamps, path and branch chips, diff gutters, disclosure toggles.
- 10px `text-2xs` survives for keycap chips only, which is the app-wide convention shared with chat's dialogs.

`gap-1.5` and `py-0.5` stay where they express tight internal leading. The rule applied here is the Tailwind scale with no arbitrary pixel values, plus equal spacing between sibling rows.

## What was off, surface by surface

**CodeTranscript** — the tool row, its streaming tail, and the expanded output preview sat at 13px while the rest of the transcript body ran 14px. All three now share 13.5px, so a tool card does not change size when it expands. Per-file `+N −M` counts were sitting wherever the path left them; they now right-align on the row's meta axis with the durations above them.

**TurnReviewCard** — durations and diffstat counts were proportional figures, so a turn's elapsed time jittered as it counted. Both are tabular now.

**CodeApprovalCard** — the title, the denial feedback, the command block, and the network summary were 13px; the file paths under a `file_write` request were 11px, two steps below the command block that asks the same question. Everything the reader is deciding on now reads at 13.5px.

**CodeSidebar workspace cards** — the card's three rows sat 2px apart, except the session row, which added its own 2px margin on top of the gap. All three are 4px apart now. Group headers stacked `mt-1` on `pt-1` on the list gap; that is one 8px/4px band, and the header carries a `title` so a truncated group name is recoverable.

**CodeWorkspacePage header** — the worktree chip middle-truncated to 28 characters inside a 44-unit box that then end-truncated it again, losing the tail the middle-truncation existed to protect; the budget now fits the box. The title and repo name are hover-recoverable, and the approvals badge is tabular.

**CodeInspector** — check summary icons were 12px above 14px icons in the rows they summarize; one glyph width now. Check counts and comment timestamps are tabular. Check rows sat 4px inside the column's left edge; they align with everything else and keep the hover bleed. The scope chip, PR title, check names, inline comment paths, and the branch line are all hover-recoverable, and the branch line is mono like every other git reference.

**DiffPanel** — diff lines were 11px, below the ramp's floor for anything you actually read; they are 12px now with the gutters left at 11px meta, which is the usual editor relationship. The scope caption and per-file path headers gained titles.

**FilesPanel** — the filter stack was 6px apart; 8px, on the scale.

**CodeCenterTabs** — a tab shows only the file name and truncates at 160px, with nothing behind it. The full path is now the tab's title.

**FileViewer, NewWorkspaceDialog, StartSessionPrompt, HarnessPicker, workspaceActions, TerminalPane** — audited, already on the ramp, unchanged.

**TerminalDrawer** — its title was 12px while every other panel title in the same product is 14px.

**CodeComposer** — the model-menu keycap used a raw `text-[10px]` where the rest of the app uses `text-2xs` for the same chip; same size, one token. Truncated model names are hover-recoverable.

**ToolOutputPreview** (shared location, code mode is its only consumer) — its `bare` variant renders inside the code tool card, so it has to match the row it belongs to; it moved with the row to 13.5px. The boxed variant stays 12px, matching the panel mono size.

## Tests

`pnpm vitest run src/code` (179 passed), `pnpm vitest run` (1200 passed), `pnpm build`: all green. No test assertions changed — nothing in the suite pins these classes.

## Logged, not fixed (bigger than spacing and type)

1. **No shared fluid middle-truncation.** `MiddleTruncate` is duplicated privately in `CodeTranscript.tsx` and `CodeApprovalCard.tsx`, so surfaces without a copy (diff scope caption, per-file diff headers, PR branch line) end-truncate paths and lose the tail. They get `title` attributes here. A shared primitive is the real fix.
2. **Panel headers do not share a height.** `DiffPanel` has a title plus a caption; `FileViewer` has a single path line. Switching center tabs shifts the content below by the difference.
3. **The files explorer is sans, not mono.** Its rows are names, not paths, so this may be right — but it is the one place a path-shaped thing is not mono, and it deserves a decision at the screenshot gate.
4. **Monaco's font size (13px) is set as a prop, outside the ramp**, near but not equal to the 12px diff body next to it.
5. **`DoctorList` uses a 16px card title and shows harness paths in sans.** It is shared with Settings, so its ramp is a wider question than code mode.
6. **The diff body moving from 11px to 12px is the one density change here** and the item most worth a look in the screenshot sweep.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/ToolOutputPreview.tsx
+++ b/crates/tidebreak-desktop/ui/src/ToolOutputPreview.tsx
@@ -55,7 +55,7 @@ export function ToolOutputPreview({
           aria-label={label}
           className={
             bare
-              ? "text-muted-foreground overflow-x-auto pr-7 font-mono text-[13px] break-words whitespace-pre-wrap [overflow-anchor:none]"
+              ? "text-muted-foreground overflow-x-auto pr-7 font-mono text-[13.5px] break-words whitespace-pre-wrap [overflow-anchor:none]"
               : "bg-muted text-muted-foreground overflow-x-auto rounded-md p-2 pr-9 font-mono text-xs break-words whitespace-pre-wrap"
           }
         >
--- a/crates/tidebreak-desktop/ui/src/code/CodeApprovalCard.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeApprovalCard.tsx
@@ -42,7 +42,7 @@ export function CodeApprovalCard({
       data-testid="code-approval-card"
     >
       <div className="flex items-baseline justify-between gap-3">
-        <h3 className="text-[13px] font-medium break-words">
+        <h3 className="text-[13.5px] font-medium break-words">
           {approvalTitle(approval)}
         </h3>
         <ApprovalState approval={approval} />
@@ -53,7 +53,7 @@ export function CodeApprovalCard({
         decidedAt={approval.decided_at}
       />
       {approval.state === "denied" && approval.feedback && (
-        <p className="text-[13px] break-words">{approval.feedback}</p>
+        <p className="text-[13.5px] break-words">{approval.feedback}</p>
       )}
       <div>
         <button
@@ -151,7 +151,7 @@ function ApprovalKindBody({ approval }: { approval: CodeApprovalSnapshot }) {
     case "command":
       return (
         <div className="flex flex-col gap-1">
-          <pre className="bg-muted overflow-x-auto rounded-md p-2 font-mono text-[13px] break-words whitespace-pre-wrap">
+          <pre className="bg-muted overflow-x-auto rounded-md p-2 font-mono text-[13.5px] break-words whitespace-pre-wrap">
             {approval.kind.cmd || "Command"}
           </pre>
           {approval.kind.cwd && (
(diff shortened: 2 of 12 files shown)


naingthet (commented):
## Summary

Class and `title` work only — no behavior, no new surfaces. The four-step ramp (14 / 13.5 / 12 / 11, plus `text-2xs` keycaps) is applied consistently. Hover-recovery on truncated paths and `tabular-nums` on durations/counts are the right cheap fixes.

The `centerTabTitle` helper is the one new function and it matches the tab labels. `MiddleTruncate` remaining private is already logged.

Do not merge until **desktop UI** finishes. This PR and #2219 both touch `CodeSidebar.tsx`; #2219 also has an unfixed titlebar overlay on collapse, so land this first if both stay open.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

PR #2221, opened 2026-08-18
refactor(desktop): polish code-mode interactions and restore tool-line detail

naingthet:
R2 of the code-mode refinement stack, on top of R1 (#2220). One concern:
how the surfaces respond to a pointer and a keyboard. No new components, no
new data, no restructuring — the diff is interaction states, one motion
tempo, two gate findings, and the plumbing that keeps an expansion from
moving the row under the reader's cursor.

## The two gate findings

**Inline tool lines rendered a bare verb.** The approved anatomy is glyph +
bold verb + muted mono target + meta + status glyph + chevron, and the
target was reaching the DOM but disappearing in two ways. Harness adapters
open a call before its arguments finish streaming and fill the missing
field with an empty string (`claude/parse.rs` `unwrap_or("")`, same in the
other three), so `File read` and `Command run` rendered with nothing after
them; the detail is fixed at `tool_started` and never restated, so the row
stayed empty for the life of the transcript. `toolSubject` now falls back
to the tool's own name, which is the floor the reader can always act on.
The subject also had no width floor: in a flex row between a `shrink-0`
verb and a `shrink-0` meta slot it was the only shrinkable item, so a
narrow column (rail plus review sidebar open) squeezed it to nothing. It
now takes the free space with a `6ch` floor, so it truncates instead of
vanishing. The server-side gap — a detail that arrives empty and is never
corrected — is worth its own issue; this is the honest client behaviour
either way.

**"Turn finished · 0s" reads as a broken clock.** `formatTurnDuration` now
returns `<1s` under a second, a tenth through the first ten seconds
(`2.4s`), whole seconds to a minute, then `1m 12s` and `1h 5m`. The tenth
rounds up to ten before the millisecond count does, so 9,999 ms reads
`10s`, not `10.0s`. A live elapsed counter reads its own tenth as jitter,
so `formatElapsedDuration` keeps the running tool line on whole seconds
while sharing the same `<1s` floor. Both stay tabular.

## Zero-jolt expansion

Expanding anything at the tail used to yank the transcript. The follow
machine watches the content box and scrolls to the bottom whenever it
grows, so opening a tool line at the bottom of the viewport moved that
line off screen — the one row the reader was reading. `useTranscriptFollow`
gains `pauseFollow`: stop following without claiming the reader scrolled
away, which `disarmFollow` does and which would flash the jump-to-latest
pill for a frame. Every in-place reveal in the transcript (tool line, tool
output "N more lines", file-activity row, approval payload disclosure)
reports through one `onReveal` prop, and the workspace page answers it with
`pauseFollow`. The next content measurement reports where the reader
actually ended up, so the pill appears only if they really are away from
the tail.

## Per surface

- **CodeTranscript** — tool line is one full-width target with a bleeding
  hover tint and a focus ring; the file-activity disclosure gets the same;
  both report reveals. Subject fallback and width floor as above. The
  running spinner keeps spinning under reduced motion (see below).
- **TurnReviewCard** — the diffstat button had a focus ring but no hover
  state, so it read as a label until you tabbed to it; it now tints and
  uses the shared ring with an offset.
- **CodeSidebar** — the sort control, both "add" buttons, and the workspace
  card share one ring and tempo. The card already had them; it now says so
  through the same constants.
- **CodeInspector** — the turn-scope chip, the check-summary toggle, the
  check rows, and the PR title link were all reachable by keyboard with no
  visible focus; the scope chip also had no hover fill.
- **FilesPanel** — explorer rows had a hover tint and no focus ring; the
  directory chevron animated at the browser default with no reduced-motion
  guard. The include/exclude toggle had neither ring nor cursor.
- **DiffPanel** — per-file header, the "Show diff" expander, and the "Open"
  affordance all gain rings; the chevron joins the one tempo.
- **CodeCenterTabs** — chat, file, and diff tabs plus their close buttons
  had hover only, so keyboard tab-switching was invisible.
- **CodeWorkspacePage** — the jump-to-latest pill was the only motion in
  code mode on `ease-in-out` at 150 ms; the worktree-path chip and the
  pending-approvals badge gain rings.
- **CodeApprovalCard** — the harness-payload disclosure gains a ring and
  reports its reveal.
- **CodeHome / CodeRepoPage** — repo and workspace list rows were
  hover-only.
- **TerminalDrawer** — the resize separator's focus fill now eases like
  every other state change.

New `src/code/interactive.ts` holds the three strings these share:
`FOCUS_RING` (`--ring`, offset 2), `FOCUS_RING_TIGHT` (offset 0, for dense
rows where 2px would collide with the row above), and `HOVER_TINT`
(`motion-safe` colour transition at the reveal tempo). Writing them out per
control is how they drift; this is why the same file has one ring in three
shapes today.

## Reduced-motion audit

Every animation in `src/code` is now 140 ms ease-out and guarded by
`motion-safe`/`motion-reduce`, with one deliberate exception: the running
tool line's spinner had `motion-reduce:animate-none`, which froze the only
signal that a call is still going and made a running tool read as a hung
one. The shared `Spinner` — used by every other panel in the product —
never had that guard. Progress is the content, not decoration.

## Tests

`pnpm vitest run src/code` (185 passed), `pnpm vitest run` (1206 passed),
`pnpm build`: all green.

- Extended the existing `CodeTranscript` tool-line assertions: the settled
  duration now reads `1.2s` and `4.0s` (the rounding change), the expand
  test also pins that a reveal reaches the host, and one new case covers a
  detail that arrives empty falling back to the tool name.
- New `TurnReviewCard.test.ts`: the duration boundaries, including the
  9,999 ms rounding edge and the elapsed variant's whole seconds.
- No tests deleted.

Not verified: nothing was driven in the live app, so hover, focus, and the
expand-at-the-tail behaviour are for the screenshot gate.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/ToolOutputPreview.tsx
+++ b/crates/tidebreak-desktop/ui/src/ToolOutputPreview.tsx
@@ -17,6 +17,11 @@ type ToolOutputPreviewProps = {
    * "· · · N more lines" instead of "Show N more lines".
    */
   bare?: boolean;
+  /**
+   * The reader expanded or collapsed the block. Hosts that follow the tail of
+   * a transcript use it to stop, so the block stays where it was clicked.
+   */
+  onToggle?: () => void;
 };
 
 /**
@@ -34,6 +39,7 @@ export function ToolOutputPreview({
   collapsedLines = DEFAULT_COLLAPSED_LINES,
   label = "Output",
   bare = false,
+  onToggle,
 }: ToolOutputPreviewProps) {
   const [expanded, setExpanded] = useState(false);
   const bodyId = useId();
@@ -76,10 +82,13 @@ export function ToolOutputPreview({
       {hiddenCount > 0 && (
         <button
           type="button"
-          className="text-muted-foreground hover:text-foreground text-[11px]"
+          className="text-muted-foreground hover:text-foreground ring-offset-background focus-visible:ring-ring cursor-pointer rounded-sm text-[11px] focus-visible:ring-2 focus-visible:ring-offset-2 focus-visible:outline-none motion-safe:transition-colors motion-safe:duration-[140ms] motion-safe:ease-out"
           aria-expanded={expanded}
           aria-controls={bodyId}
-          onClick={() => setExpanded((current) => !current)}
+          onClick={() => {
+            onToggle?.();
+            setExpanded((current) => !current);
+          }}
         >
           {expanded
             ? "Show less"
--- a/crates/tidebreak-desktop/ui/src/code/CodeApprovalCard.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeApprovalCard.tsx
@@ -7,6 +7,7 @@ import { WithTooltip } from "@/components/ui/tooltip";
 import { formatMessageTimestamp } from "@/MessageFooter";
 import { cn } from "@/lib/utils";
 import { ScrollableContainer } from "@/ScrollableContainer";
+import { FOCUS_RING, HOVER_TINT } from "./interactive";
 
 /**
  * Parked engine approval. The normalized kind leads so the reader can decide;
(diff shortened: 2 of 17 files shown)


naingthet (commented):
Conflicts with `main` after #2219. Rebase before this can merge.

The interaction tokens, duration clock, and subject fallback look right. `pauseFollow` is the right primitive: it drops follow without flashing the jump-to-latest pill, and the next content measurement reports where the reader actually landed.

Hold merge until the rebase is clean and desktop UI is green. Hover, focus, and expand-at-the-tail still need a look in the running app.

naingthet (commented):
Rebase onto main after #2219 is in (`8149f0c8`). Same review as before: tokens, `pauseFollow`, subject fallback, and the duration clock look right.

Still hold for a green desktop UI lane. This PR does not restore `{nativeTitlebar && !sidebarCollapsed && <Titlebar />}` — that remains missing on `main` after #2219, and it is out of scope here.

PR #2222, opened 2026-08-19
fix(desktop): unmount the titlebar when the sidebar is collapsed

naingthet:
#2219 hid the sidebar on collapse but dropped the follow-up that unmounted the titlebar with it.

`.titlebar` is `position: absolute; width: var(--sidebar-expanded-width)`. Collapse removes that CSS var, so the bar falls back to the stylesheet default of 280px and paints a 40px strip over the conversation. `SidebarExpandStrip` (same `z-index: 40`) then sits on leftover chrome.

## What changes

- `AppShell` renders `Titlebar` only when `nativeTitlebar && !sidebarCollapsed`.
- `SidebarExpandStrip` stays as the collapsed expand control.
- Cmd/Ctrl+B still flips `sidebarCollapsed` and brings the rail and titlebar back.
- An `AppShell` DOM test covers the unmount and the shortcut restore.

## Verification

- `pnpm --dir crates/tidebreak-desktop/ui test` — 1201/1201 pass.

Do not merge until desktop UI is green.

PR #2223, opened 2026-08-19
refactor(desktop): design code-mode empty states and harden content edges

naingthet:
Third of the code-mode refinement stack, after #2220 (spacing and type) and
#2221 (interaction polish). Two concerns only: the zero state of every
code-mode surface, and what happens to those surfaces under hostile content.

## Empty states, one decision each

| Surface | Zero state | Why this treatment |
| --- | --- | --- |
| `/code` home, no repos | Shared `Empty` block: "No repos yet", how to add one, and the Add repo button | The whole page is the zero state, so it earns the full block. It also replaces two stacked sections ("Add a repo" with its own button, then a "Repos" heading over "None registered yet.") that carried one message between them. |
| `/code` home, repos listed | Repos section with the Add repo button on the heading row | The instruction is no longer news once a repo exists; the action stays. |
| Repo page, no workspaces | `Empty` block naming what a workspace is, no button | The page header already carries "New workspace" a few centimetres away. A second copy of it would be chrome; the missing information is what the thing is. |
| Rail, no repos | The empty line becomes a quiet "Add a repo" button | A full block cannot fit a 240px rail. The empty slot is where the reader looks, and the section's own `+` is a 14px target beside a heading. This replaces the dead label rather than adding to it. |
| Rail, no workspaces | Same treatment, "New workspace" — **suppressed entirely when no repo is registered** | With no repo there is nothing to open a workspace on, and the line above already says so. Two dead-end lines stacked read as a broken rail. |
| Files explorer, no files | One quiet line: "This worktree has no tracked files yet." | Narrow rail; nothing is actionable. |
| Files explorer, no search or filter matches | Names the query in mono and offers "Clear search and filters" | "No matching files." beside three inputs leaves the reader guessing which one is the culprit. |
| Diff panel, empty | Three distinct lines: no changes in this file / "Turn 4 changed no files." / "The worktree matches its base branch." | These are three different facts. The per-turn case is the one a reader who just clicked a turn diffstat is actually asking about, and it used to answer "No diff." |
| Inspector PR tab, no PR | Title, one line of what will land here, and an "Open Source control" button that switches tabs | Checked the gh-absent path: the remediation copy lives in the git card on Source control (warning block, exact commands in mono, copy control) and presents correctly. Restating it here would be a second copy to keep true — a way over to it is not. Trimmed the copy that narrated the tab's own quietness. |
| Transcript, hydrated and empty | Two lines, centred: the instruction at foreground weight, then what a turn produces | The engine, mode, and model are already chosen by the time a reader is here, so this is not a welcome screen. The old single muted sentence floating in the column was the default, not a decision. |
| Terminal drawer, before first output | "Opening a shell in the worktree…" over the host, gone as soon as the shell answers | An empty black rectangle looks the same as a failed one. It clears on the first read rather than on the first byte: a live shell showing nothing is a cleared screen, not a pending one. Positioned over the host so the terminal does not jump a row. |
| Doctor list, all healthy | One line: "Every engine this build drives is installed and signed in." | Otherwise the reader checks each card to learn that nothing needs doing. Also shows in Settings, which is where the all-ready case is most often read. |

## Content edges

Fixed:

- **Long workspace titles in the header.** The title now takes the free space and yields it first (`flex-1`), the repo name is capped and `shrink-0`, and the worktree-path chip stops shrinking. Before, a long title could squeeze the repo name to nothing — the two facts that say which checkout you are looking at. The header's right-hand controls got `shrink-0` so the left side is what gives.
- **Long branch names on rail cards.** The branch already middle-truncated to 22 characters but had no CSS floor, so a narrow rail overflowed the row. Added `truncate` under the middle truncation.
- **Long repo paths on the home and repo pages.** Both were raw `root_path` in a wrapping inline span. Now middle-truncated with the full path on hover, in a row that truncates.
- **Diff tabs in the centre strip.** `name.ts (diff)` truncated to `name.ts…`, which made a diff tab indistinguishable from the file tab for the same path. The suffix is now a separate `shrink-0` span, so only the file name truncates. The accessible name is unchanged.
- **Deep nesting in the explorer.** The indent was `8 + depth * 12` with no ceiling; twenty levels indented past the rail. Extracted `treeIndentPx`, which stops growing at eight levels.
- **Huge approval payloads.** Two problems. The disclosure nested a `pre` inside `ScrollableContainer` (itself a `pre`), so the browser's own `white-space: pre` won and the wrapping the class list asked for never applied — a single long JSON line scrolled sideways instead. And the payload was rendered whole: an engine that attaches a file's contents put megabytes into the transcript's DOM. Now one `pre` that wraps, capped at 20,000 characters with the remainder stated.
- **Long harness paths in the doctor cards.** No break opportunity in a path, so a deep install location ran past the card. `break-words`.

Verified intact, no change needed:

- Tool output with 500-character single lines wraps (`whitespace-pre-wrap break-words`) in both the clamped preview and the running tail; the tail stays height-capped so a streaming line cannot grow the row.
- Diffs with very long lines keep per-file horizontal scroll (`overflow-x-auto` on each file's `pre`, `min-w-max` rows) — R1/R2 left that intact.
- Command approvals and file-write path lists already wrap or middle-truncate.

**200-item transcripts.** Rows were *not* memoized — `isolatedCard` wraps each in an error boundary, which is not the same thing. Measured with a throwaway fixture: changing one item's text re-rendered all 200 rows. Two changes fix it without virtualization: `TranscriptItem` is now `memo`, and `onDecide` moved into a `useCallback` in `CodeSessionPane` (it was written inline, so it was a fresh identity on every streamed delta and would have defeated the memo for every row). Re-measured: 1 of 200. The reducer already keeps untouched item objects identical across an update, which is what makes this work; the doc comment on `TranscriptItem` records the dependency so a future inline callback in the host is a visible mistake.

## Tests

Added five, each for a failure worth acting on:

- `fileTree` — the tree indent stops growing (the rule the fix installs).
- `FilesPanel` — a search that matches nothing names the query, and clearing restores the tree.
- `DiffPanel` — a turn-scoped empty diff says the turn changed no files, and an unscoped one says the worktree matches its base.
- `CodeApprovalCard` — a payload twice the cap renders capped rather than whole.
- `TerminalPane` — the opening line shows before the shell answers and clears once it does, including when it answers with nothing.

No tests removed. `pnpm vitest run src/code` (28 files, 189), full `pnpm vitest run` (182 files, 1211), and `pnpm build` all pass.

Not verified: driving the running app. Screenshots for the state × theme sweep are the coordinator's gate.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeApprovalCard.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeApprovalCard.dom.test.tsx
@@ -2,7 +2,7 @@
 import { cleanup, fireEvent, render, screen } from "@testing-library/react";
 import { afterEach, describe, expect, it, vi } from "vitest";
 
-import { CodeApprovalCard } from "./CodeApprovalCard";
+import { CodeApprovalCard, MAX_PAYLOAD_CHARS } from "./CodeApprovalCard";
 import type { CodeApprovalSnapshot } from "../api/types";
 
 afterEach(() => {
@@ -77,6 +77,22 @@ describe("CodeApprovalCard", () => {
       "deny",
       "no — use the fixtures directory instead",
     );
+  });
+
+  it("caps a payload an engine sent a whole file in", () => {
+    const huge = "x".repeat(MAX_PAYLOAD_CHARS * 2);
+    render(
+      <CodeApprovalCard
+        approval={{
+          ...pendingWrite,
+          harness_raw_json: JSON.stringify({ content: huge }),
+        }}
+        onDecide={vi.fn()}
+      />,
+    );
+    fireEvent.click(screen.getByRole("button", { name: "Harness payload" }));
+    const shown = screen.getByText(/more characters not shown/);
+    expect(shown.textContent!.length).toBeLessThan(MAX_PAYLOAD_CHARS + 200);
   });
 
   it("tones an approved card as success and stamps both times", () => {
--- a/crates/tidebreak-desktop/ui/src/code/CodeApprovalCard.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeApprovalCard.tsx
@@ -76,8 +76,14 @@ export function CodeApprovalCard({
           Harness payload
         </button>
         <Reveal open={payloadOpen}>
-          <ScrollableContainer className="bg-muted text-muted-foreground mt-2 max-h-48 rounded-md p-3 text-[11px] break-words whitespace-pre-wrap">
-            <pre className="font-mono">{prettyRaw(approval.harness_raw_json)}</pre>
+          {/*
+            One `pre`, not two: the scroll container is itself a `pre`, and a
+            nested one carried the browser's own `white-space: pre`, so the
+            wrapping asked for here never applied and a single long JSON line
+            scrolled sideways instead.
+          */}
+          <ScrollableContainer className="bg-muted text-muted-foreground mt-2 max-h-48 rounded-md p-3 font-mono text-[11px] break-words whitespace-pre-wrap">
+            {prettyRaw(approval.harness_raw_json)}
           </ScrollableContainer>
         </Reveal>
       </div>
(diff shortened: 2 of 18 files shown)


naingthet (commented):
## Summary

The empty-state table is the right cut: one `Empty` block when `/code` has no repos, quiet one-liners in the narrow rail, and a filtered Files miss that quotes the query plus a way out. Hostile content is handled with `min-w-0` / `truncate` / `break-all` on names and search text, and tab labels split so the path can truncate without eating the suffix.

Hold merge until desktop UI finishes. Branch is one behind `main` after #2222 — rebase if GitHub reports it dirty.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

PR #2224, opened 2026-08-19
fix(harness): carry completed tool arguments so transcripts name their subject

naingthet:
## The gap

Engines open a tool call before its arguments finish streaming. Adapters built
the `ToolDetail` from that not-yet-complete payload and filled the missing
fields with `""`, so the journaled `tool_started` carried an empty command or
path — and nothing ever corrected it. The transcript's inline tool line falls
back to the tool's name when its subject is empty, which is why a real call
rendered as "Command run Bash" instead of "Command run cargo test -p
tidebreak-server". The line is the only place a reader learns what the engine
actually did, so it lived or died on that subject.

The fixtures show it directly. Claude Code's `content_block_start` opens the
call with `input: {}` and streams `input_json_delta` after it; opencode's
`pending` tool part carries `input: {}` and the arguments land on the frames
that follow.

## The correction channel

`CodeEvent::ToolCompleted` gains an optional `detail`, rebuilt from the
complete arguments when the call resolves. Additive and wire-compatible: the
field is `skip_serializing_if = "Option::is_none"` and `#[ts(optional)]`, so
existing rows read back unchanged and the generated TypeScript declares it
optional rather than nullable.

It is a correction, not a replacement. An adapter that can already see complete
arguments when the call opens still emits a complete `tool_started`. The
reducer merges a non-null completed detail into the existing tool item and
takes it only when it says at least as much as the line already shows, so a
correction never downgrades a populated line to a bare tool name. `ToolDetail`
gained `subject()` and `specificity()` in `tidebreak-core` to state that
ordering once; `CodeSessionReducer.ts` mirrors it.

Equality matters in one real case rather than being a tie-breaking detail:
opencode's `glob` start falls back to the tool name as the query, and the
completion carries the actual pattern. Both score the same, so the rule is
"the correction wins unless it says less", not "unless it says more".

## Per-adapter fixture coverage

Behavior is derived from the checked-in fixtures only.

| Adapter | What its fixtures show | Result |
|---|---|---|
| claude-code 2.1.233 | `content_block_start` opens with `input: {}`; the following `assistant` message repeats the block with the assembled input, before the `tool_result` | Records the later, more specific view and rides it on `ToolCompleted` |
| opencode 1.18.18 | `pending` part has `input: {}`; the `running` and terminal (`completed`/`error`) parts carry the full input | Builds the correction from the terminal part |
| codex 0.147.0 | `item/started` already carries the whole command; `item/completed` repeats it | Start was never empty; the completion carries the same detail |
| grok 1.0.4 | `tool_call` carries the whole `rawInput`; `tool_call_update` repeats no arguments at all | Start was never empty, and there is nothing to correct with — stays `None` |

Grok is the one adapter that emits no completed detail, and it is left that way
deliberately rather than synthesized from a frame that does not carry it.

## Tests

- `fixture_replay_tool_use` for claude-code and opencode now assert both halves
  of the bug: the started detail names nothing (`specificity() == 0`) and the
  completed detail names the real path. Codex's asserts its start is already
  populated and its completion repeats the command.
- `CodeSessionReducer.test.ts` covers the merge: empty start upgraded to a
  populated correction, a populated start kept against a weaker one, and an
  absent correction leaving the line alone.
- `parsers.test.ts` covers the validator: `tool_completed` parses with and
  without `detail`, and a malformed detail rejects the event. Without the new
  key in the `onlyKeys` allowlist every completion would be dropped silently,
  which nothing else catches.
- Harness expected sequences, `code-journal-events.json`, and
  `generated/wire.ts` regenerated.

## Schema epoch

`DESKTOP_SCHEMA_EPOCH` moves 31 to 32. The field is additive and old rows
deserialize unchanged, but ADR 0002 makes a persisted-format change and the
epoch bump one change, and the pinned journal fixture names the bump as the
fix. Pre-v1 local data is disposable by that record's design.

## Not verified

The change was not driven in the running app; the transcript rendering is
covered by the reducer tests rather than by a live session.

### DIFF
--- a/crates/tidebreak-core/fixtures/code-journal-events.json
+++ b/crates/tidebreak-core/fixtures/code-journal-events.json
@@ -34,7 +34,11 @@
     "type": "tool_completed",
     "call_id": "toolu_1",
     "outcome": "succeeded",
-    "preview": "demo"
+    "preview": "demo",
+    "detail": {
+      "kind": "file_read",
+      "path": "README.md"
+    }
   },
   {
     "type": "file_changed",
--- a/crates/tidebreak-core/src/code/event.rs
+++ b/crates/tidebreak-core/src/code/event.rs
@@ -51,6 +51,41 @@ pub enum ToolDetail {
     },
 }
 
+impl ToolDetail {
+    /// The subject a transcript line names: the command, path, or query.
+    #[must_use]
+    pub fn subject(&self) -> &str {
+        match self {
+            Self::Command { cmd, .. } => cmd,
+            Self::FileEdit { path } | Self::FileRead { path } => path,
+            Self::Search { query } => query,
+            Self::Other { summary } => summary,
+        }
+    }
+
+    /// How much this detail says about the call, for the correction channel.
+    ///
+    /// An engine can open a tool call before its arguments finish streaming,
+    /// so the detail on [`CodeEvent::ToolStarted`] may name nothing. A later
+    /// detail built from the complete arguments rides
+    /// [`CodeEvent::ToolCompleted`] and replaces the first one only when it
+    /// scores higher, so a correction never downgrades a line that already
+    /// names its subject.
+    ///
+    /// Zero is a detail with no subject, one is a bare tool name, and two is
+    /// a real command, path, or query.
+    #[must_use]
+    pub fn specificity(&self) -> u8 {
+        if self.subject().trim().is_empty() {
+            0
+        } else if matches!(self, Self::Other { .. }) {
+            1
+        } else {
+            2
+        }
+    }
+}
+
 /// How a tool call finished.
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, TS)]
 #[serde(rename_all = "snake_case")]
@@ -212,6 +247,16 @@ pub enum CodeEvent {
         outcome: ToolOutcome,
         /// Bounded preview of the result.
         preview: String,
+        /// Classification rebuilt from the call's complete arguments.
+        ///
+        /// Engines open a tool call before its arguments finish streaming, so
+        /// the detail on [`CodeEvent::ToolStarted`] can name nothing. This is
+        /// the correction: adapters that see the final arguments fill it in,
+        /// and renderers merge it into the started call. It is `None` when
+        /// the engine's completion payload carries no arguments.
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        #[ts(optional)]
+        detail: Option<ToolDetail>,
     },
     /// A file changed. The diff body is loaded from a bounded GET route.
     FileChanged {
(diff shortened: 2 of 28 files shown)


naingthet (commented):
## Summary

The correction channel is the right contract. Engines open a tool call with empty input; adapters were journaling that as `\"\"` and never replacing it, so the transcript named the tool instead of the command. `ToolCompleted.detail` plus `specificity` (0 empty / 1 other / 2 real subject) lets the complete arguments upgrade the start line without downgrading one that already names a path or command.

Fixtures and the reducer test pin that merge. Hold until clippy/test finish. Branch is behind `main` — rebase if it goes dirty.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

naingthet:
Known limitation worth naming: the correction rides the completion, so a call parked on an approval keeps its started detail until it resolves. Claude Code's `approval-request` fixture is exactly this — the `assistant` message carries the complete arguments, but no `tool_result` follows while the request is parked, so the line still reads "File edit Write" until the decision lands and the continuation delivers the correction. Closing that gap would need a second correction channel on the approval event, which is wider than this change.

naingthet (commented):
## Summary

The correction channel is the right cut. `ToolCompleted.detail` is additive (`skip_serializing_if`, `#[ts(optional)]`), adapters that already see complete arguments at start keep emitting a populated `tool_started`, and the reducer takes a correction only when it says at least as much as the line already shows. That last rule is the one that matters: opencode's glob start falls back to the tool name and the completion carries the pattern at the same score, so `>=` rather than `>` is the contract, not a tie-break.

Per-adapter fixture coverage matches what the frames actually carry. Claude and opencode assert the empty start plus the named completion; Codex asserts a populated start that the completion repeats; Grok stays `None` because `tool_call_update` has no arguments. The parser allowlists `detail` and rejects a malformed one instead of dropping the whole completion, which is the failure that would have gone silent.

Schema epoch 31 → 32 is the right companion to the journal fixture change.

Hold merge until the Rust lanes (clippy, desktop test, test, postgres) and desktop UI finish. Branch is one behind `main` after #2223 — MERGEABLE, so admin squash is fine unless GitHub reports it dirty.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

naingthet (commented):
## Summary

Clippy, desktop test, desktop UI, and postgres are green. The first `test` attempt failed in an unrelated flake: `tests::code_terminals::create_write_read_and_two_cursors` at `assert_eq!(late_page["cursor"], cursor)` (`left: 177`, `right: 26`). The shell had already echoed `TERM_TWO_READERS_ok`; more PTY bytes arrived before the late read, so the cursor advanced. That path is not the tool-detail change.

A re-run is already in progress. Hold the admin squash until `test` is green. Branch is still `BEHIND` after #2223 but `MERGEABLE` (not dirty), so no rebase unless GitHub reports a conflict.

## Issue counts by severity

- bugs: 0
- suggestions: 0
- nits: 0

PR #2226, opened 2026-08-19
refactor(desktop): complete the code-mode accessibility contract

naingthet:
R4 of the code-mode refinement stack: roles and names, keyboard paths, live-region placement, and focus behaviour across `src/code/**`. Three shared primitives change too, each because code mode is the surface that exposed the gap.

No visual change is intended. Every fix below is a semantic or keyboard one.

## Live regions on streaming

**The firehose, fixed.** Every `CodeToolCard` wrapped itself in `role="status"` — an atomic polite live region — and a running call streams its output *inside* that region. A screen reader re-read the whole tool line, output and all, on every streamed byte, once per tool line in the transcript. The wrapper is now a plain container. The outcome moved to where it belongs: the line's own accessible name, beside the status glyph it mirrors.

**Turn lifecycle, added.** With the tool lines silenced, nothing announced that a turn started or ended. `CodeTranscript` now carries exactly one live region — an `sr-only` `role="status"` fed by `codeTurnAnnouncement(items, busy)`: "Turn running", then "Turn finished · 2m 14s" or "Turn interrupted". It mounts empty and only fills in on a later change, so reopening a finished session does not read its last outcome at the reader unprompted. A failed turn is deliberately *not* announced here: `TurnReviewCard` already raises it as an `alert`, and saying it twice is worse than saying it once.

**Verified already correct**, no change made:
- Failed turn → `role="alert"`; harness error notice → `alert`, warning/info → `status` (`HarnessNotice`).
- Reconnect bar, queued-follow-up pill, composer image notice, "Opening a shell…" → polite `role="status"`.
- Composer send failure and merge failure → `role="alert"`.
- Copy-to-clipboard announcements → `sr-only role="status" aria-live="polite"`.

## Roles and names

- **Workspace card** (`CodeSidebar`) — the card announced only its title. `aria-label={title}` overrode everything below it, so the glyph rail (attention dot, PR state, terminal, pending approval) was pointer-only information: a triage read could not hear "needs you". New pure `workspaceCardLabel()` builds the name in card order — title, attention, PR number and state, terminal, repo, branch — and the glyphs it now speaks for are `aria-hidden`.
- **Attention dot** — `<span aria-label>` with no role names nothing. Now `role="img"`.
- **Unread-engine-event dot** (workspace header) — same fix.
- **Tool output** (`StreamingTail`, and shared `ToolOutputPreview`) — `<pre aria-label="Output">` is a generic element; the label was dropped. Now `role="group"`.
- **Terminal host** — `<div aria-label="Terminal output">` around xterm's own tree, unnamed for the same reason. Now `role="group"`.
- **`Spinner`** (shared) — its doc comment tells callers to pass `aria-label` "when the glyph is the only signal", but lucide renders a bare `<svg>` and the label never reached anyone. It now sets `role="img"` when a caller labels it. Four code-mode surfaces relied on this: Files, Diff, FileViewer, PrCard.
- **Diff file section** — the "Open" control was `role="link" tabIndex={0}` *nested inside* the disclosure button, and an `<h3>` sat inside it too. A control inside a control is reachable by neither, and the disclosure's name swallowed "Open" and the ± counts. They are siblings now; the heading wraps the disclosure (the standard disclosure-heading shape), so heading navigation still lands per file. "Open" and "Show diff" take a label naming their file — a multi-file diff otherwise lists a column of identical controls.
- **Pending-approval badge** — named "2 approvals", a state rather than an action. Now "Jump to 2 pending approvals".
- **File-activity rows** — the A/M/D/R letter was `aria-label`ed on a bare span and read as the letter. The letter is `aria-hidden`; the word carries the change kind.
- **Verified already correct**: inspector icon tabs, terminal drawer controls, composer model/effort/permission menus, `WorkspaceOverflowMenu`, context-menu items, repo and sort-mode controls, `HarnessPicker`, `NewWorkspaceDialog` fields.

## Keyboard paths

- **Explorer tree** — it advertised `role="tree"` but was a list of buttons: no arrow keys, no `aria-level`, focus on a button *inside* each `treeitem` (which announced as a button, in a tree). Rebuilt to the ARIA tree pattern: the row is the treeitem, one roving tab stop for the whole tree, Up/Down between drawn rows, Right to open a folder or step into an open one, Left to close or climb out, Home/End, Enter/Space to activate. New pure `flattenVisibleTree()` derives the row order the keys walk. Each treeitem takes an explicit `aria-label` of its own name — a treeitem otherwise draws its name from everything it contains, which for an open folder is the whole subtree.
- **Center tabs** — a `tablist` whose tabs sat inside untyped wrapper `<div>`s, with no arrow keys and no panel linkage. Wrappers are `role="presentation"`; Left/Right/Home/End move and select with a roving tab stop; each tab `aria-controls` its panel and each panel is `aria-labelledby` its tab.
- **Session lifecycle mark** — its tooltip carries the engine version and the dropped-event warning, and nothing else on the page does. A `<span>` cannot be tabbed to, so it was hover-only. It now takes a tab stop and the focus ring.
- **Collapsed file-activity list** — `grid-rows-[0fr]` + `overflow-hidden` hides pixels, not semantics: every changed file of every turn was readable as though the summary were always open. Now `aria-hidden` + `inert` when closed, matching the `Reveal` shape already used in `CodeApprovalCard`.
- **`aria-controls`** added to the file-activity, harness-payload, and diff-file disclosures, which had `aria-expanded` and nothing to point at.
- **Verified already correct**, no change made: tool-line expand is a real button; the Chat/Code segmented switch is a proper radiogroup with arrows, Home/End and roving tabindex; Radix Tabs gives the inspector arrow keys (now pinned by a test); Radix ContextMenu opens from Shift+F10 / the Menu key, because it listens for the plain `contextmenu` event a keyboard fires, and returns focus to the card on close; the drawer separator keeps `tabIndex`, `aria-valuenow`, and ±32px arrows from #2205; dialogs keep Radix's focus trap untouched.

## Focus management

- Expanding or collapsing a tool line leaves focus on the line — the toggle is the row itself.
- The context menu returns focus to its card on Escape (Radix default, now asserted).
- The jump-to-latest pill is a real button that leaves the tab order (`tabindex="-1"`, `aria-hidden`) until it is on screen, rather than sitting there invisibly focusable.
- The explorer's tab stop follows the reader and falls back to the first row whenever the row they left stops being drawn.

## Tests

Assertions replaced, not piled on. Net +7 tests, all pinning a contract that can regress silently.

Replaced:
- `CodeTranscript` — `getByRole("status", { name: "Command run succeeded" })` asserted the firehose this PR removes; it is now a button-name assertion plus a test that the streaming output has no live-region ancestor.
- `FilesPanel` — the `aria-current` assertion moved from `button` to `treeitem`.
- `CodeSidebar` — the card is found by its full name; the context-menu test now goes through the keyboard path and checks focus return.
- `DiffPanel` — "Show diff" is matched by its file-qualified name.

Added: tree arrow-key navigation end to end; `workspaceCardLabel` unit coverage; turn-lifecycle announcement and the failed-turn exception; center-tab arrows and panel linkage; jump-to-latest tab order; inspector tab arrows; terminal host naming; the diff disclosure and its Open control not nesting.

## Checks

`pnpm vitest run` — 1220 passed (182 files). `pnpm build` clean. `tsc` clean. Full lanes on the PR.

Not verified: no screen reader was driven against a live session. The contracts above are asserted at the DOM level; the coordinator's screenshot sweep and any VoiceOver pass are still owed.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/ToolOutputPreview.tsx
+++ b/crates/tidebreak-desktop/ui/src/ToolOutputPreview.tsx
@@ -58,6 +58,10 @@ export function ToolOutputPreview({
       <div className="group relative w-full">
         <pre
           id={bodyId}
+          // A bare `aria-label` on a `pre` names nothing: the element is
+          // generic. The group role is what carries the label — and the copy
+          // control beside it uses the same word.
+          role="group"
           aria-label={label}
           className={
             bare
--- a/crates/tidebreak-desktop/ui/src/code/AttentionBadge.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/AttentionBadge.tsx
@@ -35,6 +35,10 @@ export function AttentionBadge({
           tone === "info" && "bg-info-foreground-muted",
           className,
         )}
+        // A dot with no text is only a state if something says which one. On a
+        // bare `span` the label would be dropped; `img` is the role that takes
+        // a name and has no children to read.
+        role="img"
         aria-label={label}
         title={tooltip}
         data-attention={attention.state.type}
--- a/crates/tidebreak-desktop/ui/src/code/CodeApprovalCard.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeApprovalCard.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useState, type ReactNode } from "react";
+import { useEffect, useId, useState, type ReactNode } from "react";
 
 import type { CodeApprovalSnapshot } from "../api/types";
 import { Button } from "@/components/ui/button";
@@ -32,6 +32,7 @@ export function CodeApprovalCard({
   const [denying, setDenying] = useState(false);
   const [feedback, setFeedback] = useState("");
   const [payloadOpen, setPayloadOpen] = useState(false);
+  const payloadId = useId();
   const decided = approval.state !== "pending";
 
   useEffect(() => {
@@ -68,6 +69,7 @@ export function CodeApprovalCard({
             HOVER_TINT,
           )}
           aria-expanded={payloadOpen}
+          aria-controls={payloadOpen ? payloadId : undefined}
           onClick={() => {
             onReveal?.();
             setPayloadOpen((current) => !current);
@@ -82,7 +84,10 @@ export function CodeApprovalCard({
             wrapping asked for here never applied and a single long JSON line
             scrolled sideways instead.
           */}
-          <ScrollableContainer className="bg-muted text-muted-foreground mt-2 max-h-48 rounded-md p-3 font-mono text-[11px] break-words whitespace-pre-wrap">
+          <ScrollableContainer
+            id={payloadId}
+            className="bg-muted text-muted-foreground mt-2 max-h-48 rounded-md p-3 font-mono text-[11px] break-words whitespace-pre-wrap"
+          >
             {prettyRaw(approval.harness_raw_json)}
           </ScrollableContainer>
         </Reveal>
--- a/crates/tidebreak-desktop/ui/src/code/CodeCenterTabs.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeCenterTabs.tsx
@@ -1,8 +1,20 @@
+import { useRef, type KeyboardEvent } from "react";
 import { FileCode, FileDiff, MessageSquare, X } from "lucide-react";
 
 import { cn } from "@/lib/utils";
 import { panelKey, type PanelContent } from "@/panel/panelTypes";
 import { FOCUS_RING_TIGHT, HOVER_TINT } from "./interactive";
+
+/** Ids the strip and the two center panels agree on, so tabs name panels. */
+export const CHAT_TAB_ID = "code-center-tab-chat";
+export const CHAT_PANEL_ID = "code-center-panel-chat";
+export const EDITOR_PANEL_ID = "code-center-panel-editor";
+const editorTabId = (index: number) => `code-center-tab-editor-${index}`;
+
+/** Which tab labels the editor panel right now. */
+export function centerEditorTabId(index: number): string {
+  return editorTabId(index);
+}
 
 /**
  * Center strip: Chat is always first and cannot close. File and diff tabs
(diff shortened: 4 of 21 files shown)


naingthet (commented):
## Summary

R4 does what the title says: it stops treating code mode as a visual surface with leftover ARIA and actually implements the contracts. The live-region firehose is gone, the turn announcer is one polite status that stays quiet on first mount, the explorer is a real tree, the center strip is a real tablist, and the names that used to die on generic elements now have roles that carry them.

Verified on this head (`9364609`):

- `codeTurnAnnouncement` returns `"Turn running"` / `"Turn finished · …"` / `"Turn interrupted"` and `""` on failed, so `TurnReviewCard`'s alert is not doubled.
- `TurnLifecycleAnnouncer` skips the first effect, so reopening a finished session does not speak the last outcome.
- `flattenVisibleTree` plus the ul-level key handler matches the APG tree keys (Up/Down, Right expand-or-enter, Left collapse-or-climb, Home/End, Enter/Space).
- Diff disclosure and Open are siblings; the heading wraps the disclosure rather than sitting inside the button.
- Workspace card name is composed in card order and the glyph rail is `aria-hidden`.
- Shared `Spinner` only takes `role="img"` when a caller labeled it.

No ADR collision. UI-only, so Rust lanes correctly skip. Hold merge until **desktop UI** is green. Own PR, so this is comment-only. VoiceOver against a live session is still owed, as the body says.

## Issue counts by severity

- bugs: 0
- suggestions: 2
- nits: 1

### Suggestions

1. **Terminal host `role="group"` + `aria-label="Terminal output"`.** That names the wrapper around xterm's own tree. Some ATs stop at the group name and never enter the inner textbox. Confirm with VoiceOver that focus still lands in the terminal and the buffer is readable; if the group swallows the surface, drop the wrapper name and let xterm's widget speak.

2. **`aria-controls` is omitted while collapsed** on the approval payload, the diff file body, and (when `hasOutput` is true but the tool line is closed) the tool card. APG keeps the relationship even when the target is not shown. The tool card is the sharper case: `aria-controls={bodyId}` stays set whenever `hasOutput` is true, but neither body is mounted on a collapsed success line, so the id can dangle.

### Nits

1. **Close buttons stay independently tabbable** inside the center tablist. Fine, and common, but it means Tab after the selected tab walks every close control rather than leaving the strip. Mentioning so a later pass does not "fix" it by accident.

PR #2227, opened 2026-08-19
refactor(desktop): close code-mode theme parity and logged visual debt

naingthet:
The last pass of the code-mode refinement stack: read every code-mode surface in
both themes, and close the visual debt earlier layers logged.

R4 (#2226) merged while this was in flight, so it targets `main`.

## Light/dark parity

Contrast was measured, not eyeballed: every token pairing below was converted
from oklch to sRGB and scored against WCAG 2.1. The pattern the sweep found is
one mistake repeated — the design system has two inks per tone, and the mark
colour (`--success`) was being used where the readable ink (`--success-foreground`)
belongs. The mark colour is tuned to clear 3:1, which an icon needs; in the dark
theme it happens to clear 10:1 as well, so nothing looked wrong there while the
light theme sat at 3.2:1.

| Surface | Was | Now |
| --- | --- | --- |
| Diff added/removed lines (`DiffPanel`) | 3.22 / 4.39 light | 8.44 / 9.26 both themes |
| Per-file `+n −n` counts (`DiffPanel`) | 3.22 light | 9.07 / 10.06 |
| Turn file-activity counts and A/M/D/R letters (`CodeTranscript`) | 3.01 light | 9.07+ |
| PR check counts (`CodeInspector`) | 3.22 light | 9.07 / 10.06 |
| "Reconnecting to the session…" (`CodeWorkspacePage`) | 3.04 light | 9.47 |
| Workspace-card branch name (`CodeSidebar`) | 3.49 light | 4.83 |
| Focused destructive menu item (`context-menu`, `dropdown-menu`) | 1.86 dark, 3.90 light | 8.24 both |
| `--ring` focus indicator | 2.63 light | 4.28 light |

Two of these reach outside `src/code/`, deliberately:

- **`--ring`** in the light theme cleared 2.6:1 against the page and 2.9:1
  against a hovered row, so the focus indicator R4 spent a PR wiring up was
  close to invisible in one of the two themes. `0.704 → 0.58` on the same hue
  clears 4.3:1 on the page, 4.1:1 on a card, 3.4:1 on `--muted`.
- **The destructive menu item** paired `text-destructive` with
  `focus:bg-critical-background`: dark red on dark red, 1.9:1, at the moment
  the reader lands on Archive or Delete. Both menu primitives now take the ink
  the critical tint was built for. `context-menu` is code-mode-only;
  `dropdown-menu` carries the identical defect on chat's delete actions and is
  fixed with it rather than left to diverge.

No new colour is introduced and no hardcoded colour is added. The two raw
palette usages in `workspaceCards.ts` stay: the repo-chip swatches are identity,
not status, and GitHub's merged-PR purple has no token — both already carry
explicit light and dark values and a comment saying why.

Nothing regressed in the terminal (#2205); its theme derivation and its
light/dark test are untouched.

## Logged debt from earlier layers

1. **Shared middle-truncation primitive.** `MiddleTruncate` was duplicated
   byte-for-byte in `CodeTranscript` and `CodeApprovalCard`. Extracted to
   `src/code/MiddleTruncate.tsx` — fluid, no character cap, `title` built in —
   and adopted at both old sites plus the diff scope caption, the per-file diff
   headers, the file-viewer path, and the PR branch line (where an end-truncate
   ate the base branch, the half a reader is checking).
2. **`DiffPanel` and `FileViewer` header heights.** The diff header was two
   lines and the file header one, so switching center tabs moved the body about
   20px. `FileViewer` now carries the same structure: heading, mono caption,
   same padding, same fixed spinner slot. Both spinner slots are fixed-size, so
   a refresh no longer nudges the diffstat sideways either.
3. **Explorer paths in sans.** The files tree was the only path-shaped surface
   set in the UI face, which made one file name look like two different strings
   between the tree and the tab it opened. Now mono at 12px, matching every
   other path in code mode.
4. **Monaco vs the diff body.** Chosen: **13px/20px for both**, by moving the
   diff body up from 12px rather than moving Monaco down. 13px is on the design
   ramp (13/14 body) where 12px is not, it sits directly under the transcript's
   13.5px mono, and the two views open the same file from the same tab strip —
   a reader flipping between them now sees the same lines in the same places.
   Monaco gains an explicit `lineHeight: 20` to match `leading-5`.
5. **`DoctorList`.** The 16px card title drops to the ramp's 14px, and harness
   paths and versions render mono. This shows in Settings too and reads
   coherently there.
6. **Approval-parked tool lines.** The card and the still-running line above it
   render the same command, 8px apart, as two unrelated blocks. An approval that
   directly follows a running tool line is now drawn hanging off it: the column
   gap closes and a rail drops from the tool line's own icon column into the
   card. Adjacency only — no wire change, and the card's command block already
   matched the tool line's 13.5px mono.

## Tests

One added: a parked approval renders attached to the tool line above it, and an
approval that follows anything else does not. Nothing on the wire links the two,
so the ordering rule is the contract worth pinning.

One updated: `CodeSidebar.dom.test.tsx` asserted the archive item's destructive
tone by class name; the class changed with the fix above, the assertion still
pins the same fact.

No theme-toggling DOM tests added. Colour classes are not a contract worth
pinning per surface, and the terminal — where the computed theme genuinely is a
contract — already has its own.

## Not done

- `badge.tsx` pairs `-foreground-muted` ink with the tone tint at roughly 2.9:1
  in the light theme. It is applied consistently app-wide and looks like a
  deliberate softening rather than a code-mode defect, so changing every badge in
  the product does not belong in this PR. Worth its own look.
- Chat's own `text-destructive` body copy sits at 3.7:1 in the dark theme, same
  root cause as the menu item. Out of scope here.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeApprovalCard.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeApprovalCard.tsx
@@ -8,6 +8,7 @@ import { formatMessageTimestamp } from "@/MessageFooter";
 import { cn } from "@/lib/utils";
 import { ScrollableContainer } from "@/ScrollableContainer";
 import { FOCUS_RING, HOVER_TINT } from "./interactive";
+import { MiddleTruncate } from "./MiddleTruncate";
 
 /**
  * Parked engine approval. The normalized kind leads so the reader can decide;
@@ -246,29 +247,6 @@ function Reveal({ open, children }: { open: boolean; children: ReactNode }) {
     >
       <div className="overflow-hidden">{children}</div>
     </div>
-  );
-}
-
-function MiddleTruncate({
-  text,
-  className,
-}: {
-  text: string;
-  className?: string;
-}) {
-  const tail = Math.min(28, Math.max(12, Math.ceil(text.length / 3)));
-  if (text.length <= 40) {
-    return (
-      <span className={cn("block truncate", className)} title={text}>
-        {text}
-      </span>
-    );
-  }
-  return (
-    <span className={cn("flex min-w-0", className)} title={text}>
-      <span className="truncate">{text.slice(0, -tail)}</span>
-      <span className="shrink-0">{text.slice(-tail)}</span>
-    </span>
   );
 }
 
--- a/crates/tidebreak-desktop/ui/src/code/CodeInspector.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeInspector.tsx
@@ -41,6 +41,7 @@ import { useCodeUiStore } from "./CodeUiStore";
 import { DiffPanel } from "./DiffPanel";
 import { FilesPanel } from "./FilesPanel";
 import { FOCUS_RING, FOCUS_RING_TIGHT, HOVER_TINT } from "./interactive";
+import { MiddleTruncate } from "./MiddleTruncate";
 import { PrCard } from "./PrCard";
 import { useCodeUpdatesStore } from "./CodeUpdatesStore";
 import { PR_ICON_TONE_CLASSES, prTone, prToneLabel } from "./workspaceCards";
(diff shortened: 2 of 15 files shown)


naingthet (commented):
COMMENT (author is naingthet — cannot APPROVE own PR).

Read the token sweep, the shared `MiddleTruncate`, the FileViewer/DiffPanel header alignment, and the approval-attachment heuristic.

No merge-blocking defect.

The contrast story is consistent: mark colours stay on glyphs, readable `-foreground` inks go on words and numerals, and the two menu primitives stop pairing `text-destructive` with `focus:bg-critical-background`. `--ring` moving to `0.58` is the one token that leaves `src/code/`; the measured 2.6:1 on the light page is enough reason.

`parksTheCallAbove` is a journal-order heuristic, not a wire contract, and the new test pins the only rule that can regress (attached after a running tool line, standalone otherwise). Fine.

One nits-only note, not blocking: the `itemSignature` JSDoc in this diff looks jammed (`*/function itemSignature`). If that landed in the file, prettier/desktop UI will catch it; if it is just a diff artifact, ignore.

Admin-merge when desktop UI is green.

PR #2228, opened 2026-08-19
test: stabilize terminal cursor and admission lease tests

### LINKED ISSUE #2225: Flaky: terminal cursor read and turn-admission lease tests race wall-clock on slow runners
Two pre-existing wall-clock races surfaced on #2224's lanes (both green on rerun, neither related to the change):

- `tests::code_terminals::create_write_read_and_two_cursors` — PTY output arrived between two reads, moving the cursor (saw 177 vs 26).
- `db::tests::turn_admission_reservation_is_global_exact_and_recoverable` — takes a 50 ms lease then asserts the next call is still `Pending`; a slow runner expires the lease first.

Same family as the since-fixed sandbox timing races (#2191 → #2199/#2208): assertions racing real clocks. Candidates for the same treatment (mock clocks or comfortable bounds).

naingthet:
## Summary

- end the terminal test's shell producer before asserting that a cursor remains stable
- wait for both the output marker and terminal termination before taking the final cursor snapshot
- replace the turn-admission test's 50 ms wall-clock lease with an explicitly expired database reservation

## Root cause

Both tests raced real time. The terminal test could observe its marker before trailing PTY output arrived, allowing the cursor to advance between reads. The admission test expected several database assertions to complete before a 50 ms lease expired, which was not reliable on slow runners.

## Validation

- `git diff --check`
- focused terminal cursor test
- focused turn-admission reservation test
- 10 consecutive runs of each focused test

Closes #2225

### DIFF
--- a/crates/tidebreak-core/src/db/tests.rs
+++ b/crates/tidebreak-core/src/db/tests.rs
@@ -2927,7 +2927,7 @@ async fn turn_admission_reservation_is_global_exact_and_recoverable() {
     };
     let first_token = uuid::Uuid::new_v4();
     let first_lease = match store
-        .begin_turn_admission(&request, first_token, chrono::Duration::milliseconds(50))
+        .begin_turn_admission(&request, first_token, chrono::Duration::seconds(30))
         .await
         .unwrap()
     {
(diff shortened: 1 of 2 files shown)


naingthet (commented):
COMMENT (author is naingthet — cannot APPROVE own PR).

Both races are the same family as #2191: the assertion waited on a wall clock instead of the state the product actually uses.

The terminal test now ends the producer (`echo …; exit`) and waits for the marker *and* `ended` before snapshotting the cursor. That is the right contract — trailing PTY output, including the next prompt, cannot move the cursor after the shell has exited. The 3s deadline is unchanged; if `exit` never reports `ended`, the new assertion fails closed instead of passing a moving cursor.

The admission test stops asking a loaded runner to finish every conflict assertion inside a 50 ms lease. A 30 s lease covers the Pending/IdentityConflict checks, then the row is expired with `database_now - 1s` so takeover uses the same clock the store consults. Good.

No merge-blocking defect. Merge when required CI is green. Rebase if it stays behind `main`.

PR #2229, opened 2026-08-19
chore(desktop): sync the refreshed component set to the design system

naingthet:
Adds the six shared components the code-mode overhaul introduced to the design-system entry, the sync component map, and the authored preview set (screenshot-verified). The design project itself is already refreshed from this build; this PR lands the reproducible inputs. 63 components total, render check clean.

### DIFF
--- a/.design-sync/config.json
+++ b/.design-sync/config.json
@@ -7,28 +7,88 @@
   "srcDir": "src",
   "tsconfig": "tsconfig.json",
   "cssEntry": ".ds-compiled.css",
-  "extraEntries": ["sonner"],
+  "extraEntries": [
+    "sonner"
+  ],
   "readmeHeader": ".design-sync/conventions.md",
   "overrides": {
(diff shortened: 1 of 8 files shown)


naingthet (commented):
The six new shared pieces land in the right three places: `design-system.ts` re-exports, the sync `componentSrcMap`, and authored previews.

Source paths match files already on `main`, and the preview imports use the exported names (`SegmentedControl`, `ContextMenu*`, `ToolOutputPreview`, `UserMessage`, `TranscriptSkeleton`, `MiddleTruncate`). ContextMenu fires a mount-time `contextmenu` so the card can show the open menu; that matches how Radix menus work.

Config overrides for ContextMenu (single card, 480x420, `WorkspaceMenu`) and MiddleTruncate (column) look right. The rest of the override block is a format-only expand.

`.design-sync/config.json` lost its trailing newline. Not a merge blocker.

PR #2230, opened 2026-08-19
test: stabilize Windows process-tree and sandbox cancellation timeouts

naingthet:
## Summary

Yes, the original failure was a flake. This PR also covers the other loaded-runner timeouts that already failed on `main`.

### Windows process-tree pid publication

`run_action` and the harness Job Object tests waited for a descendant pid after a short spawn budget. On a loaded Windows runner, PowerShell can take longer than that to start the child and write the file.

[CI run 32208633478](https://github.com/brightwave-inc/tidebreak/actions/runs/32208633478/job/95936714206) failed:

```
code::gh::windows_tests::quick_action_timeout_terminates_its_descendant
quick action ended before publishing its descendant pid
```

[CI run 32199033165](https://github.com/brightwave-inc/tidebreak/actions/runs/32199033165) failed the same race in the harness:

```
child::windows_tests::dropping_a_process_tree_kills_its_descendant
child::windows_tests::terminating_a_process_tree_kills_its_descendant
PowerShell descendant pid was not published within 10s
```

The tests now spawn through the production helper, wait up to 30s for the pid, fail fast if the parent exits, and use `ping.exe -t` instead of nested PowerShell.

### Sandbox cancellation accounting

[CI run 32197409775](https://github.com/brightwave-inc/tidebreak/actions/runs/32197409775) failed:

```
sandbox_agent_run_worker::tests::local_cancellation_finalization_survives_accounting_failure_past_execution_lease
test completed within its time bound: Elapsed(())
```

That test already spends up to 9s in inner waits inside a 10s outer bound. The outer bound is now 30s.

The earlier interrupt flakes (`interrupt_stops_a_running_scripted_turn`) were already fixed in #2217.

## Validation

- `cargo fmt -p tidebreak-harness -p tidebreak-server`
- `git diff --check`
- focused sandbox cancellation test, 5 consecutive runs
- Windows process-tree tests need a Windows runner

### DIFF
--- a/crates/tidebreak-harness/src/child.rs
+++ b/crates/tidebreak-harness/src/child.rs
@@ -445,6 +445,7 @@ mod windows_tests {
     use super::{spawn_process_tree, ProcessTreeChild};
 
     const DESCENDANT_TIMEOUT: Duration = Duration::from_secs(10);
+    const PID_PUBLISH_TIMEOUT: Duration = Duration::from_secs(30);
 
     #[tokio::test]
     async fn terminating_a_process_tree_kills_its_descendant() {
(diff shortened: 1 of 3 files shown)


naingthet (commented):
The flake is real: `run_action` starts the 5s budget at spawn, and the old test waited for the descendant pid after that. On a loaded Windows runner PowerShell can miss the window, so the action finished before publishing a pid.

This test now uses `spawn_workspace_script` and only then cancels `wait_with_output`. That is the same drop path `run_action` uses — Job Object close with `KILL_ON_JOB_CLOSE` still has to reap `ping.exe`. The 30s pid wait no longer races the action budget.

`ActionOutcome` flags are no longer asserted because the test no longer goes through `run_action`. The remaining contract is the one that failed: timeout teardown kills the descendant.

Windows cargo check is parked, so CI will not execute this lifecycle test. Squash-merge when the required Linux lanes are green.

naingthet (commented):
The second commit covers the other loaded-runner timeouts that already failed on `main`.

Harness Job Object tests now wait 30s for the pid, fail fast if the parent exits, and spawn `ping.exe -t` instead of nested PowerShell. That is the same descendant shape the quick-action test already used.

The sandbox cancellation test already spends up to 9s inside a 10s outer bound. Opening that bound to 30s is slack, not a new contract.

Windows cargo check is parked, so the process-tree tests still will not run in CI. Squash-merge when the required Linux lanes are green.

PR #2232, opened 2026-08-19
fix(code): classify engine approvals and auto-accept them in Allow

naingthet:
## Summary

Codex and OpenCode already send the command or path on an approval request. The parser dropped that payload, so the card showed **unknown** and parked a file read until the turn was interrupted.

Allow all still parked those leftover asks. OpenCode only pre-granted bash, edit, and read, so grep, glob, and reads outside the worktree still prompted.

This change:

- Keeps the engine payload and classifies Codex commands, OpenCode bash, and file reads so the card names what it is asking.
- Accepts leftover approval requests immediately when the session is Allow all, so the turn does not hang.
- Grants the extra OpenCode permission rules in Allow so the engine asks less often.

## Test plan

- [x] `cargo test -p tidebreak-server --lib code::session_worker::tests`
- [x] `cargo test -p tidebreak-harness --lib fixture_replay_approval`
- [x] `cargo test -p tidebreak-harness --lib permission_mode_mapping`
- [x] `pnpm exec vitest run src/code/CodeApprovalCard.dom.test.tsx`
- [ ] Start a new Allow all code session and confirm a file read does not park a card.
- [ ] In Ask, confirm a Codex or OpenCode shell approval shows the command instead of **unknown**.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeApprovalCard.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeApprovalCard.dom.test.tsx
@@ -116,6 +116,22 @@ describe("CodeApprovalCard", () => {
     ).not.toBeNull();
   });
 
+  it("does not render an unknown engine summary as the decision", () => {
+    render(
+      <CodeApprovalCard
+        approval={{
+          ...pendingCommand,
+          kind: { type: "other", summary: "unknown" },
+          harness_raw_json: "null",
+        }}
+        onDecide={vi.fn()}
+      />,
+    );
+    expect(screen.getByText("Allow this?")).toBeInTheDocument();
+    expect(screen.getByText("The engine needs approval")).toBeInTheDocument();
+    expect(screen.queryByText("unknown")).not.toBeInTheDocument();
+  });
+
   it("tones a denial as a warning and shows the feedback", () => {
     render(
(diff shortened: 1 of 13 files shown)


naingthet (commented):
Comment. The parser was dropping the engine payload, so the card said unknown and leftover Allow asks parked the turn. Keeping `raw`, classifying Codex `command` / OpenCode `metadata.command` / file reads, and auto-accepting in Allow matches how those engines already speak.

Verified:

- Codex and OpenCode parsers now clone the request params into `ApprovalRequested.raw`. Fixture expectations and replay assertions require the command / `permission: bash` fields.
- `kind_from_raw` reads Codex `command`+`cwd` first, then OpenCode `metadata.command`, then tool/permission names. Empty or `unknown` becomes a readable other-summary; the card no longer prints `unknown`.
- Allow sessions decide Approve and skip the sink emit so a leftover ask cannot park a card. OpenCode Allow also pre-grants grep, glob, list, external_directory, websearch, webfetch, and task.

One failure-path note, not a merge block: both session loops do `let _ = self.decide(...).await` then `continue`. If that approve fails, the UI never sees the request and the engine is still waiting. Happy path is the one this PR is for; a later change should emit or surface the error instead of swallowing it.

No correctness issue on the intended path. Merge when required CI is green.

PR #2234, opened 2026-08-19
feat(desktop): keep PR status above the inspector tabs

naingthet:
## Summary

The pull-request state lived only on its own tab. The inspector now shows a strip above **Files**, **Source control**, and **Pull request**: number, status, check count, and **Merge** / **Fix errors** / **Resolve conflicts**.

Each action writes a prompt into the composer and focuses it. The primary button follows the host state (ready, failing checks, or conflicts). The menu still offers the other two.

## Test plan

- [x] `pnpm exec vitest run src/code/prActions.test.ts src/code/CodeInspector.dom.test.tsx src/code/CodeComposer.dom.test.tsx`
- [ ] Open a workspace with an open PR. Confirm the strip stays visible when you switch inspector tabs.
- [ ] Click **Merge**. Confirm the composer fills with a merge prompt.
- [ ] On a PR with failing checks, confirm the primary button is **Fix errors**.
- [ ] On a conflicting PR, confirm the primary button is **Resolve conflicts**.

PR #2235, opened 2026-08-19
feat(desktop): tighten code-mode load, images, and transcript order

naingthet:
## Summary

Code mode first load, images, and transcript order were fighting the user.

- `/code` keeps a centered **Loading…** empty until the catalog or doctor can decide what to show. Repos list as soon as they exist.
- The OpenCode mark matches model-gateway's window icon.
- Token use sits on the composer context ring, not `11k in / 12 out` on the turn seam.
- Claude Code turns can carry published images. Codex, opencode, and Grok still hide attach until they declare support.
- OpenCode leftover `session.idle` no longer closes the next prompt as a 21ms empty turn. Replayed tools and assistant text stay above that turn's seam.
- **Copy debug JSON** is on the workspace overflow menu. It copies the session, turns, and journal events.

## Test plan

- [ ] Open `/code` with no catalog yet. Confirm the loading empty, then **No repos yet** or the repo list.
- [ ] Confirm the OpenCode mark in the harness picker.
- [ ] Finish a code turn. Confirm the composer ring, not in/out text on the seam.
- [ ] On a Claude Code session, paste an image and send. Confirm it publishes, rides the turn, and renders on the user bubble.
- [ ] On OpenCode, send two prompts in a row. Confirm the second turn does not close empty, and a reload keeps tools above **Turn finished**.
- [ ] Use **Copy debug JSON** from the workspace menu. Confirm the clipboard has session, turns, and events.


naingthet (commented):
Comment. Code home no longer flashes an empty register form. Claude Code can carry published images. OpenCode leftover `session.idle` no longer closes the next prompt as a 21ms empty turn.

Verified:

- `/code` shows Loading until repos, the empty register, the doctor, or an error can render. Repos list as soon as they exist.
- OpenCode leftover idle is ignored when `turn_terminal || !turn_open`. The parse test is idle, idle, busy, idle → started then completed once.
- Replay inserts tools and assistant text before that turn's seam (`insertBeforeTurnBoundary`), so a reload keeps tools above Turn finished.
- Claude stdin encodes images as one stream-json user line (ADR 0046). Text-only stays plain bytes. Attach stays hidden unless the doctor says the engine consumes images.
- Token use moves to the composer context ring. Copy debug JSON is a session overflow action and warns before share.

No correctness issue on the intended path. Merge when required CI is green.

naingthet (commented):
CI failed on this head. Do not merge.

- clippy `-D warnings`: `drain_queued` and `drive_turn` in `session_worker.rs` now have 8 arguments (limit 7). Fold the extra image/follow-up state into the existing structs rather than allowing the lint.
- desktop UI: `CodeSessionRegistry` tests still expect a user item without `attachments`. Hydrate now always sets `attachments: []`. Update those two equals. `StartSessionPrompt` cannot find Permissions / Harness / Model controls — likely the first-load empty or composer image/context change. Fix the tests or the empty gate.

Hold until clippy and desktop UI are green.

naingthet (commented):
Comment. `53a1620c` updates the two `CodeSessionRegistry` hydrations so a user item always carries `attachments: []`. That matches hydrate and is the remaining desktop UI failure from the last head.

`a1a45b94` already folded `db`/`bus` into `LiveSink`, so clippy's 7-arg cap holds without an allow. Session-prompt tests wrap `AppContextProvider`.

No remaining correctness issue on the intended path. Merge when required CI is green on this head.

naingthet (commented):
Comment. Desktop UI failed again on `53a1620c`: `tsc` rejected a search-tool fixture (`pattern` is not on `{ kind: "search"; query: string }`) and `backing.previews.set(id, previewUrl)` when `createObjectURL` is missing (`string | null` into `Map<string, string>`).

Pushed `147235e2`: the fixture uses `query`, and a null preview is not stored. Merge when required CI is green on that head.

naingthet (commented):
Comment. Rebased onto #2234. The only conflict was `CodeComposer.dom.test.tsx` afterEach: keep both `pendingComposerPrompt: null` and `attachments: {}`.

The type-cycle store fix from #2234 is already on main. This head still has the clippy fold, attachments on hydrate, `query` on search details, and a null preview that is not stored.

Merge when required CI is green on `a687f53a`.

PR #2238, opened 2026-08-19
test: assemble secret fixtures so scanners do not match

naingthet:
## What changed

Test fixtures no longer contain a contiguous private-key PEM or `sk-ant-` token that secret scanners treat as live.

Debug-export scrubber tests still need real-shaped values. Those strings are assembled at runtime so the source never contains a complete match. Secret-rehome tests that only need an opaque stored value now use `test-anthropic-key`.

## Validation

- `cargo fmt --check` on the changed files
- `cargo test -p tidebreak-desktop --lib -- chat_debug`
- `cargo test -p tidebreak-server --lib -- secret_rehome`

## Compatibility and risk

None. Test-only. Runtime behavior is unchanged.

### DIFF
--- a/crates/tidebreak-desktop/src/chat_debug.rs
+++ b/crates/tidebreak-desktop/src/chat_debug.rs
@@ -1529,9 +1529,9 @@ mod tests {
     #[test]
     fn credential_shaped_material_never_survives_into_the_bundle() {
         let cases = [
-            // Vendor-issued prefixes, wherever they appear.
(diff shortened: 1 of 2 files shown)


naingthet (commented):
Comment. Scrubber tests that need a real-shaped token now assemble it at runtime. Rehome tests that only need an opaque stored value use `test-anthropic-key`. Runtime paths are untouched.

Gitleaks already passed on this head. Merge when required CI is green.

PR #2241, opened 2026-08-19
fix(code): recognize harness protocol telemetry

naingthet:
## What changed

- Recognize current lifecycle and telemetry events emitted by Claude, Codex, opencode, and Grok harnesses.
- Keep protocol-gap counts focused on genuinely unknown events instead of known non-content events.
- Add a real Codex 0.147.0 hook-lifecycle fixture and expected parse result.

## Why

Live Code-mode exercises showed successful harness sessions being reported with protocol gaps because known hook, lifecycle, and telemetry records were falling through the parsers. That made otherwise healthy runs look unreliable and obscured real compatibility regressions.

## Impact

Developers get more trustworthy harness diagnostics across all four supported harnesses. Content extraction and session behavior are unchanged; this only improves event recognition and accounting.

## Backwards compatibility

Existing parsed content remains compatible. Truly unrecognized records are still counted, so forward-compatibility monitoring remains intact.

## Safety and risk

Low risk. The changes are parser classification updates with fixture coverage and do not alter process execution, permissions, or provider configuration.

## Validation

- `cargo test -p tidebreak-harness` — 100 tests passed
- `cargo fmt --check`
- `git diff --check`
- Live Code-mode exercises across Claude, Codex, opencode, and Grok using varied non-Fable models
- Fresh Codex Luna session completed with `unrecognized_event_count: 0`

This PR is independent and targets `main` directly.


naingthet (commented):
The classification change matches the live failure: hook, lifecycle, and catalog frames were inflating protocol-gap counts on healthy turns. Treating them as recognized no-ops, while still counting genuinely unknown kinds, is the right split.

Each harness keeps a leftover that should stay counted (`plan` / `max_turns_reached` on Grok, unknown subtypes elsewhere). The Codex `hook-lifecycle` fixture plus the flipped `unrecognized == 0` assertions on the captured plain-text streams pin the new contract. Content extraction is unchanged.

Not a merge blocker. Waiting on required Rust CI.

PR #2242, opened 2026-08-19
fix(desktop): synchronize document viewer themes

naingthet:
## What changed

- Make DOCX and spreadsheet viewers react to live dark/light theme changes.
- Refresh PDF and presentation viewer surfaces when the application theme changes.
- Centralize theme-aware viewer surface extension behavior.
- Add DOM coverage for switching themes without reloading the document manually.

## Why

The embedded document viewers captured theme state during initialization and did not consistently react when the application theme changed. Users had to refresh the document viewer before its colors matched the rest of Tidebreak.

## Impact

All supported document surfaces now follow the application theme immediately, keeping document inspection visually coherent while working in Code mode or elsewhere in the desktop app.

## Backwards compatibility

Document loading and rendering APIs are unchanged. The update only synchronizes existing viewer instances with the current theme.

## Safety and risk

Low risk. The change is isolated to document-viewer presentation and is covered by focused DOM tests plus the complete UI suite and production build.

## Validation

- UI tests — 188 files, 1,264 tests passed
- Production UI build passed
- `cargo fmt --check`
- `git diff --check`

This PR is independent and targets `main` directly.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/document/DocxViewer.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/document/DocxViewer.dom.test.tsx
@@ -2,6 +2,7 @@
 import { cleanup, render, waitFor } from "@testing-library/react";
 import { afterEach, beforeEach, expect, it, vi } from "vitest";
 
+import { setThemeMode } from "@/theme";
 import { clearFileDownloadCache, type FileBytesSource } from "./useFileDownload";
 
 const docxMocks = vi.hoisted(() => ({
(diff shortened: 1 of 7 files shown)


naingthet (commented):
The bug was real: DOCX and spreadsheet viewers pinned `isDark={false}` and a light canvas, so a theme switch left them stale until remount. Wiring `useTheme().resolved` through and replacing `LIGHT_DOCUMENT_SURFACE` with a live `bg-background` chrome is the right fix.

PDF and presentation only refresh the surrounding surface, not an `isDark` prop. That matches authored page/slide colors staying put. The DOM tests now flip the theme after load and assert the blob is not recreated, which is the regression that would have been easy to miss.

Not a merge blocker. Waiting on desktop UI CI.

PR #2245, opened 2026-08-19
fix(desktop): restore sidebar width on refresh

naingthet:
## Summary

- publish the restored sidebar width on the app shell during its first render
- remove the timing-sensitive effect that queried for `.app-shell`
- add a regression test covering a custom saved width after the boot and managed-policy gate

## Root cause

On refresh, the sidebar-width synchronization effect could run while the boot screen or managed-policy gate was still mounted. Because `.app-shell` did not exist yet and the stored width did not change afterward, the effect never retried. The sidebar rendered at its saved width while the overlaid titlebar kept the 280px CSS fallback, producing a mismatched notch over the content.

## User impact

The sidebar and native titlebar now use the same restored width on the first shell render, so refreshing no longer leaves the content partially covered. Existing collapse and resize behavior is unchanged.

## Compatibility and risk

This is a renderer-only layout change. It does not alter persisted preference keys or backend APIs. The CSS variable remains driven by the same Zustand state, but is now applied synchronously instead of through a DOM query.

## Validation

- `pnpm exec vitest run src/AppShell.dom.test.tsx` — 22 tests passed
- `pnpm exec tsc --noEmit`
- `git diff --check`

### DIFF
--- a/crates/tidebreak-desktop/ui/src/AppShell.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/AppShell.dom.test.tsx
@@ -593,6 +593,20 @@ describe("app shell", () => {
     expect(window.localStorage.getItem("tidebreak.sidebar-collapsed")).toBe("true");
   });
 
+  it("publishes the restored sidebar width when the shell appears after boot", async () => {
+    hasNativeHost.mockReturnValue(true);
+    useUiStore.setState({ sidebarWidth: 216 });
+
+    await mountApp();
+    await screen.findByText("Welcome to Tidebreak");
+
+    const shell = document.querySelector<HTMLElement>(".app-shell");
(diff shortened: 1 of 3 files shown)


naingthet (commented):
The refresh mismatch is a real first-paint race: the old `querySelector('.app-shell')` effect can fire while the boot or managed-policy gate is still mounted, then never retry because `sidebarWidth` does not change again.

Publishing `--sidebar-expanded-width` on the shell's first render is the right fix. The titlebar still unmounts with the rail (`nativeTitlebar && !sidebarCollapsed`), so leaving the var set on collapse does not revive the leftover chrome the old hook was clearing. Collapse and resize stay on the existing Zustand path.

The new DOM test covers the actual bug: a custom saved width is visible on `.app-shell` and `[data-sidebar]` after boot. No correctness issues.

PR #2243, opened 2026-08-19
feat(code): graduate and polish code mode

naingthet:
## What changed

- Graduate Code mode from the experimental settings path and enable it by default.
- Replace the dense diff landing view with a changed-file overview; selecting a file opens its readable diff in the center pane.
- Improve added/removed/context coloring and the center-pane diff treatment.
- Add Monaco language detection and syntax highlighting for repository files.
- Add VS Code-style text search with wildcard support while reserving centered `Cmd+P` quick open for filename navigation.
- Add a right-click tab action menu for closing and managing open tabs.
- Improve PR/commit synchronization, status presentation, worktree behavior, and server search/diff support.
- Update Code-mode guidance and historical protocol-gap wording.

## Why

Code mode had accumulated the core workflow but still felt experimental: entry was gated, navigation mixed filename and content search, diffs opened as an undifferentiated wall of content, tabs lacked expected desktop actions, and several server/UI state transitions made the idea-to-PR loop harder to understand than it should be.

## User and developer impact

Code mode is now a supported default surface with a clearer path from workspace creation through editing, review, commit, push, and pull request. The UI follows familiar editor conventions and the underlying worktree/search/diff contracts are more reliable for all supported harnesses.

## Backwards compatibility

Existing Code workspaces and harness configuration continue to work. Experimental settings routes and the Code feature gate are removed intentionally; Code is now available by default. API additions are additive except for removal of the obsolete experimental toggle behavior.

## Safety and risk

Medium risk because this touches desktop routing, Code-mode UI, worktree operations, and server endpoints. The changes are covered by focused DOM/unit/server tests, the complete UI suite, a production build, generated binding verification, and live CLI/UI exercises. No Fable models were used.

## Validation

- UI tests — 188 files, 1,264 tests passed
- Production UI build passed
- `cargo test -p tidebreak-harness` — 100 tests passed
- Server settings round-trip coverage passed
- Generated wire bindings verified
- `cargo fmt --check`
- `git diff --check`
- Live idea-to-PR exercises across Claude, Codex, opencode, and Grok
- Model coverage included Claude Sonnet variants, Codex Luna and Sol, opencode GLM 5.2, DeepSeek V4 Flash, and Grok 4.5

This PR is independent and targets `main` directly.

PR #2246, opened 2026-08-19
fix(desktop): spotlight a specific first-task row

naingthet:
## What changed

The first-task walkthrough was ringing the composer trigger. That hid the menu row the copy describes.

Each step now spotlights the box that opened:

- Choose a model: the current model row
- Set internet access: Network
- Choose a permission level: Ask
- Add attachments: Attach files, or Attach folder if files is missing
- Start a real task: the first starter card

If the specific row is late to mount, the ring starts on the open menu and moves to that row.

## Validation

- `pnpm exec vitest run src/FirstTaskWalkthrough.dom.test.tsx src/WelcomeState.test.tsx src/WelcomeState.dom.test.tsx src/ComposerToolsMenu.dom.test.tsx src/PermissionModeMenu.dom.test.tsx src/ModelMenu.dom.test.tsx` — 30 tests passed
- `git diff --check`

I did not click through the live desktop tour.

## Compatibility and risk

None. Visual-only React change. No API, persistence, or permission changes.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/ComposerToolsMenu.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/ComposerToolsMenu.dom.test.tsx
@@ -158,6 +158,12 @@ it("opens when the first-task walkthrough is on the tools step", () => {
     />,
   );
 
-  expect(screen.getByRole("menuitem", { name: "Attach files" })).toBeInTheDocument();
-  expect(screen.getByRole("menuitem", { name: /Network/ })).toBeInTheDocument();
+  expect(screen.getByRole("menuitem", { name: "Attach files" })).toHaveAttribute(
+    "data-first-task-target",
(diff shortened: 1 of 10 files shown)


naingthet (commented):
The walkthrough was lighting the composer trigger, which hid the row the copy describes. Preferring the opened box (`model-choice`, `network`, `permissions-ask`, `attach-files` / `attach-folder`, `starter-choice`) and falling back to the menu until that row mounts is the right targeting.

The retry loop that upgrades from the menu once the preferred target appears matches the late-portal case, and the DOM tests pin both the specific-row ring and the late-mount upgrade. Trigger attributes are gone so the ring cannot lock onto chrome that is already visible.

No correctness issues. Waiting on required desktop UI CI, and the branch is one commit behind `main` after #2243.

PR #2248, opened 2026-08-19
docs(decisions): gateway-linked hosting and one interaction model (47, 48)

naingthet:
Two proposed records.

**Decision 47 — Gateway-linked hosting.** A gateway-linked deployment is a
self-host deployment whose identity derives from a gateway roster: the token
file is generated from the roster first, and a gateway authenticator behind
decision 6's credential-to-principal seam is the end state. The record adopts
machine/client vocabulary (a tidebreak-server instance is a machine; the
desktop app, then web and mobile surfaces, are clients that attach), makes a
desktop remote connection mode the first client with host authority degrading
legibly, gates shared-machine code mode on owner-scoping the code tables, and
narrows the durable-self-host posture to a disposable regime for this
deployment class until 1.0.

**Decision 48 — One interaction model for chat and code.** The convergence
record decision 30 and deferred.md anticipate, as five independently useful
steps: code owner scoping, one permission vocabulary, one attention surface,
one turn-submission contract, then the entity merge with the internal loop as
one engine behind the decision-31 adapter. The merge direction is explicit —
code-mode structures survive; chat contributes the content model — and
whether the internal engine later becomes an external harness process is
deliberately deferred.

Decision 47 also names a gap no doc stated before: `code_*` tables carry no
owner column and `/code/*` sits outside the owner-scoped regime, so shared
deployments cannot enable code mode until that closes.

### DIFF
--- a/docs/decisions/0047-gateway-linked-hosting.md
+++ b/docs/decisions/0047-gateway-linked-hosting.md
@@ -1,0 +1,166 @@
+# 47. Gateway-linked hosting: machines, clients, and roster-derived identity
+
+- Status: Proposed
+- Date: 2026-08-19
+- Owners: server, desktop
+- Related: [`0006-self-host-deployment-plane-authorization.md`](0006-self-host-deployment-plane-authorization.md),
+  [`0017-gateway-capability-negotiation.md`](0017-gateway-capability-negotiation.md),
+  [`0030-code-mode-separate-surface.md`](0030-code-mode-separate-surface.md),
+  [`docs/gateway-boundary.md`](../gateway-boundary.md),
+  [`docs/self-hosting.md`](../self-hosting.md),
+  [`docs/deferred.md`](../deferred.md),
+  [`0048-one-interaction-model.md`](0048-one-interaction-model.md)
+- Supersedes: none
+
+## Context
+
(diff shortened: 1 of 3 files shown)


naingthet (commented):
47 and 48 are the next free numbers. No other open PR claims them, so this is not an ADR-number hold.

The records themselves hold up. 47 unparks the member client as a desktop remote connection mode, keeps the token file as the decision-6 seam for v1, and gates shared-machine code mode on owner-scoping the `code_*` tables. The disposable regime is scoped to gateway-linked deployments only, so ordinary operator-managed self-host keeps the back-up-before-upgrade posture. 48 makes convergence the plan of record in five independently useful steps, with code-mode structures as the merge survivor and chat contributing the content model. Translation layers stay rejected.

One gap before merge: `docs/deferred.md` still parks both items. It still says auth beyond the token file "waits on gateway-derived identity," and it still calls chat–code convergence "a future record on top of two proven models." These two records are that work. Update those sections in this PR so the parked list and the plan of record do not disagree.

naingthet:
Pushed `a60bd6f3`: `docs/deferred.md` now points the member-client and chat–code items at decisions 47 and 48. Remaining later surfaces (hosted web, mobile, implementation, engine packaging) stay listed.

PR #2250, opened 2026-08-19
feat(server): owner-scope code mode (decision 48 step 1)

naingthet:
Code mode was the last surface outside the owner-scoped regime decision 6 built for chats, documents, and apps. The `code_*` tables carried no owner, and the `/code/*` routes answered any token-holder about any row — which is why a shared deployment cannot enable code mode today.

[Decision 47](docs/decisions/0047-gateway-linked-hosting.md) makes closing that gap decision point 4, and gates hosted multi-user code mode on it. [Decision 48](docs/decisions/0048-one-interaction-model.md) makes it convergence step 1, "because it is paid once". This is that step. Steps 2 through 5 — permission vocabulary, attention surface, turn submission, entity merge — are out of scope, as are the remote connection mode and gateway identity.

## What changed

**Owner columns, as a baseline edit plus an epoch bump.** Every `code_*` table gains an `owner` column, denormalized onto child rows the way chat rows carry theirs, rather than reached through a join — so an owner-scoped query exists for each table. The `DESKTOP_SCHEMA_EPOCH` bump ships in the same change, per [decision 2](docs/decisions/0002-pre-v1-schema-and-persisted-format-mutability.md). Existing self-host databases take the column in place through the durable-baseline branch in `migration.rs`, following the `ensure_app_owner` precedent its module documentation prescribes. On the desktop profile everything belongs to the single local owner and keeps the column default.

**A scoped view, not scoped handlers.** `ScopedCode` is the code-mode counterpart of `ScopedStore`: an extractor that binds the process runtime to the requesting principal and never lets the unscoped handle escape. Route code therefore cannot express a cross-owner query — another owner's repository, workspace, session, turn, event, or approval is indistinguishable from one that does not exist. This follows decision 6's rule that enforcement is a router property rather than a handler habit, applied to data scoping.

System paths are not request paths and keep unscoped access, through store functions named `_all_owners` that say so in their names and documentation: boot recovery re-attaching workers, the stall sweep, and a session worker re-reading the row it was spawned against.

**Per-owner repositories and clone paths.** Nothing in code mode is shared across owners. Repository registration is unique per owner rather than per install, so a second user can register a path the first user already has. The shared clone parent directory gains an owner segment, so two users cloning the same remote land in separate directories instead of colliding.

**A partitioned updates channel, not a filtered one.** The bus holds one broadcast channel per principal. A subscriber's receiver carries only its own owner's digests and terminal notices, so a cross-owner notice is not dropped on the way out — it never arrives. Decision 47 names the alternative, filtering an install-wide stream at the route or in the client, as the wrong implementation.

**Deployment-plane code routes.** Installing pinned harness binaries (`POST /code/harnesses/refresh`) and the shared clone-parent-directory setting (`GET /code/repos/clone-defaults`) move to the deployment-plane router behind `require_admin`; both change what the machine is. The doctor read and cloning stay on the member plane, where they produce owner-scoped rows and owner-keyed checkouts.

Host paths and pids stay machine-private: `worktree_path` and `child_pid` are reachable only through owner-scoped lookups, and `child_pid` is still absent from every renderer-facing response.

## Validation

Matching the Validation sections of decisions 47 and 48:

- `a_second_user_sees_neither_code_rows_nor_updates_of_another_owner` drives two real principals against the assembled router on a shared store. Bob's listings are empty rather than filtered, every by-id read answers `404`, writes are refused the same way, and his `/code/updates` socket receives neither a connect snapshot naming Alice's session nor a live notice when her turn runs — while Alice's own socket still gets both.
- `owner_scoped_code_queries_partition_every_table` exercises every table rather than a sample, including that a save carrying another owner's key changes no row.
- `every_code_table_carries_an_owner_column` inspects the built baseline, so a new `code_*` table added without an owner fails here rather than in review.
- `code_store_queries_are_owner_scoped_or_say_they_are_not` and `code_routes_go_through_the_owner_scoped_view` are the repository checks, in the idiom of the existing id-space check. Both were mutation-tested: adding an unscoped store query makes the first fail with an actionable message.
- `code_deployment_plane_routes_refuse_a_member` asserts member `403` and admin non-`403` on the moved routes, and that the member plane is untouched.
- `clone_targets_are_keyed_by_owner` covers the collision case and the hostile-key cases (an owner key may contain separators and dots; neither escapes the parent).
- The id-space checks from decision 30 still pass — they retire at step 5, not here.

## Checks run

- `cargo fmt --all --check` — clean.
- `cargo clippy -p tidebreak-core -p tidebreak-server --all-targets -- -D warnings` — clean.
- `cargo test -p tidebreak-core --lib` — 677 passed.
- `cargo test -p tidebreak-server --lib` — 990 passed. One unrelated failure, `code_terminals::create_write_read_and_two_cursors`, reproduces identically on unmodified `main` under full-suite load and passes in isolation and on rerun; it is a pre-existing PTY timing flake, not a regression here.
- PostgreSQL, both lanes, against a real server: the new `postgres_code_owner` target (2 passed) and the existing `postgres_turn_state` target (29 passed). The new target is added to the durable-turn CI job beside the existing one.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -716,6 +716,10 @@ jobs:
         run: >-
           cargo test -p tidebreak-core --no-default-features --features postgres
           --test postgres_turn_state --locked
+      - name: Exercise code-mode owner scoping on PostgreSQL
+        run: >-
+          cargo test -p tidebreak-core --no-default-features --features postgres
+          --test postgres_code_owner --locked
 
   ui:
     name: desktop UI
--- a/crates/tidebreak-core/src/code/mod.rs
+++ b/crates/tidebreak-core/src/code/mod.rs
@@ -497,6 +497,10 @@ pub enum CodeApprovalKind {
 pub struct CodeRepo {
     /// Stable id.
     pub id: RepoId,
+    /// Principal this repository belongs to. Repositories are never shared
+    /// across owners: two users may register or clone the same remote and
+    /// each gets a checkout of their own.
+    pub owner: crate::OwnerId,
     /// Canonical git toplevel.
     pub root_path: String,
     /// Display name.
@@ -520,6 +524,8 @@ pub struct CodeRepo {
 pub struct CodeWorkspace {
     /// Stable id.
     pub id: WorkspaceId,
+    /// Principal this workspace belongs to, denormalized from its repo.
+    pub owner: crate::OwnerId,
     /// Owning repo.
     pub repo_id: RepoId,
     /// Display title.
@@ -545,6 +551,8 @@ pub struct CodeWorkspace {
 pub struct CodeSession {
     /// Stable id.
     pub id: CodeSessionId,
+    /// Principal this session belongs to, denormalized from its workspace.
+    pub owner: crate::OwnerId,
     /// Owning workspace.
     pub workspace_id: WorkspaceId,
     /// Engine this session is bound to.
--- a/crates/tidebreak-core/src/db/entities.rs
+++ b/crates/tidebreak-core/src/db/entities.rs
@@ -1658,6 +1658,7 @@ pub mod code_repo {
     pub struct Model {
         #[sea_orm(primary_key, auto_increment = false)]
         pub id: Uuid,
+        pub owner: String,
         pub root_path: String,
         pub display_name: String,
         pub default_base_ref: String,
@@ -1683,6 +1684,7 @@ pub mod code_workspace {
     pub struct Model {
         #[sea_orm(primary_key, auto_increment = false)]
         pub id: Uuid,
+        pub owner: String,
         pub repo_id: Uuid,
         pub title: String,
         pub worktree_path: String,
@@ -1709,6 +1711,7 @@ pub mod code_session {
     pub struct Model {
         #[sea_orm(primary_key, auto_increment = false)]
         pub id: Uuid,
+        pub owner: String,
         pub workspace_id: Uuid,
         pub harness_kind: String,
         pub harness_version: Option<String>,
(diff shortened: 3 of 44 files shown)


naingthet (commented):
Decision 48 step 1 is implemented as specified: owner columns denormalized onto every `code_*` table, `DESKTOP_SCHEMA_EPOCH` 32→33, self-host backfill through `ensure_code_owner`, and request paths go through `ScopedCode` so a cross-owner query is not expressible in route code. System paths keep `_all_owners` names. The updates bus is partitioned per principal rather than filtered. Clone targets take an owner segment (local profile unchanged). Harness refresh and `GET /code/repos/clone-defaults` sit on the deployment plane.

The two-principal router test, per-table owner queries, baseline owner-column check, and route-source tripwires match the validation sections of decisions 47 and 48.

Two follow-ups, neither a merge blocker for this step (hosted multi-user code mode is still gated):

- `get_clone_job` still looks up the in-memory job map by UUID only. Progress events are owner-partitioned, but a leaked job id would still return another owner's target path and `repo_id`. Worth scoping the job map the same way as the durable rows.
- `GET /code/repos/clone-defaults` is now admin-only while `POST /code/repos/clone` stays on the member plane and still requires `parent_dir` from the client. Members on a shared host cannot discover the clone parent. Intentional for this step if the UI remains single-owner; call it out before hosted code mode turns on.

SQLite cannot drop the old unique-on-`root_path` constraint in place — documented honestly. Hold for required CI (currently behind / still running).

naingthet:
Workspace `test` hit the 20m job cap during `Install headless system deps` (apt), before rust-cache or the suite. Re-running the failed job. The PR is still behind `main`.

naingthet:
Rebased onto `main` (only #2249, the DOMPurify pin). Required CI was green on the previous head; re-running after the rebase.

PR #2251, opened 2026-08-19
feat(code): refine the workspace workflow

naingthet:
## What changed

- Make the main agent history a real default center tab and add first-class tab creation, selection, closing, reordering, and persisted split-pane layout.
- Allow existing tabs to be dragged into adjacent pane groups, with keyboard and context-menu actions that follow the focused tab.
- Move the pull-request workflow into the workspace header and simplify the surrounding repository/status chrome.
- Tighten quick-open, panel navigation, session restoration, and workspace-action behavior so reopening a Code workspace feels immediate and stable.
- Expose Codex model discovery through the harness path used by Code mode and add focused coverage for the new workspace/session contracts.

## Why

Code mode still behaved like an experimental collection of panels: the main agent history was not presented as a durable tab, the center area could not be organized naturally, and important repository actions competed with low-value path/status chrome. This change establishes the core workspace model needed for a production-ready idea-to-PR flow.

## User impact

Opening a Code workspace now lands on a clear main-agent tab. Users can create and rearrange tabs, split the center area by dragging, recover the same layout after reopening, and reach the PR actions from the top bar. Existing panel routes and stored sessions continue to resolve through compatibility parsing.

## Compatibility and safety

- No database migration or wire-format change.
- Persisted Code chrome is additive and validated before restoration; invalid or obsolete state falls back to the default main-agent layout.
- Existing panel URLs remain accepted while new tab/group state is normalized at the workspace boundary.
- The harness change only improves Codex model discovery and does not change execution defaults or credentials.

## Risk

The primary risk is center-pane state synchronization across drag/drop, restoration, and session changes. The implementation keeps one normalized chrome state, scopes actions to the focused group, and adds DOM/unit coverage for the workspace, session store, registry, routing, quick-open, and chrome transitions.

## Validation

- Full Code mode UI suite: 36 files, 263 tests passed
- Focused Codex harness tests: 19 passed
- TypeScript typecheck passed
- Production desktop UI build passed
- `cargo fmt --all`
- `git diff --check`

CI is expected to run the complete workspace test matrix on the rebased `main` merge candidate.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeCenterTabs.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeCenterTabs.tsx
@@ -1,12 +1,22 @@
-import { useRef, type KeyboardEvent, type ReactNode } from "react";
+import {
+  useRef,
+  type DragEvent,
+  type KeyboardEvent,
+  type ReactNode,
+} from "react";
 import {
   ArrowRightFromLine,
   CircleX,
+  Columns2,
   Copy,
   FileCode,
   FileDiff,
   ListX,
   MessageSquare,
+  MoveLeft,
+  MoveRight,
+  PanelRightClose,
+  Plus,
   X,
 } from "lucide-react";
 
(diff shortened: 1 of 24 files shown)


naingthet (commented):
The workspace chrome is now a real tab model: main-agent history is the default center tab, file/diff tabs can split into a persisted right group, and PR actions move to the header. Restoration is additive — invalid chrome falls back to the main-agent layout, and older panel URLs still parse.

`applyEvents` is the right replay path: one Zustand publish for a journal chunk instead of a render per historical frame. Codex `list_models` uses the app-server stdio protocol rather than inventing a CLI subcommand, and failures return an empty list.

Title was `improve(code)`, which is not an allowed type. Retitled to `feat(code)` so the semantic-title and release-label lanes can pass.

No correctness blocker. Hold for required CI after the title change, and keep this behind #2250 if the two collide on Code workspace routing.

naingthet:
Desktop UI failed on `ManagedGate.dom.test.tsx` — the pairing-nudge case never left \"the open product\" for \"Connect to your model gateway\". This PR does not touch ManagedGate; the chrome/router change only adds `split*` search keys. Re-running the failed job. If it fails again, treat it as a real routing-isolation issue rather than a flake.

#2250 (owner-scope code mode) has now merged. Rebase onto main before merge if GitHub reports behind.

naingthet:
Rebased onto main after #2250. CI re-running.

naingthet (commented):
Rebased head `49a617a3` is on current main after #2250.

The workspace chrome is a real tab model: main-agent history is the default center tab, file/diff tabs can split into a persisted right group, and PR actions live in the header. Restoration is additive — invalid chrome falls back to the main-agent layout, and older panel URLs still parse. `split*` search keys are Code-workspace only in the test router; production `panelSearchFrom` accepts them on home/chat as well and ignores them when absent.

`applyEvents` remains the right replay path. Codex `list_models` uses the app-server stdio protocol, and failures return an empty list.

No correctness blocker. Hold for required CI. If desktop UI fails again on `ManagedGate.dom.test.tsx` pairing-nudge, treat it as a routing-isolation bug rather than a flake.

PR #2252, opened 2026-08-19
ci: fail Linux apt hangs instead of waiting out the job

naingthet:
## What changed

Linux release and CI jobs no longer call bare `apt-get update` against the hosted runner's Azure Ubuntu mirror. They now run `scripts/install-linux-apt-packages.sh`, which rewrites `azure.archive.ubuntu.com` to `archive.ubuntu.com` on x86_64 and `ports.ubuntu.com` on ARM, sets 20s apt timeouts with three retries, and is bounded by an 8-minute step timeout.

v0.54.0's Linux x86_64 job sat on **Install Linux packaging dependencies** for ~58 minutes until it was cancelled. Linux aarch64 in the same run finished that step in 34 seconds. v0.53.0 already needed the public-archive fallback after a wave of Azure `Ign:` lines; this just makes that the first path and fails fast if the public archive is also unreachable.

## Validation

- `node --test scripts/install-linux-apt-packages.test.mjs`
- `node --test scripts/workflow-security.test.mjs scripts/workflow-security-mutations.test.mjs`
- `bash -n scripts/install-linux-apt-packages.sh`

Did not rerun the production release workflow from this branch. The next Linux x86_64 package install is the live proof.

## Compatibility and risk

No product or wire-format change. The helper still installs the same packages. If both Azure and the public Ubuntu archive are down, the step now fails in 8 minutes instead of burning the 90-minute job timeout.

## Tracking

Follow-up to the stuck v0.54.0 job: https://github.com/brightwave-inc/tidebreak/actions/runs/32277870160/job/96149615899

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -417,9 +417,9 @@ jobs:
         with:
           version: v0.16.0
       - name: Install system deps (Tauri)
+        timeout-minutes: 8
         run: |
-          sudo apt-get update
-          sudo apt-get install -y --no-install-recommends \
+          scripts/install-linux-apt-packages.sh \
             cmake \
             libwebkit2gtk-4.1-dev \
             libayatana-appindicator3-dev \
@@ -471,9 +471,9 @@ jobs:
         with:
           version: v0.16.0
       - name: Install system deps (Tauri)
+        timeout-minutes: 8
         run: |
-          sudo apt-get update
-          sudo apt-get install -y --no-install-recommends \
+          scripts/install-linux-apt-packages.sh \
             libwebkit2gtk-4.1-dev \
             libayatana-appindicator3-dev \
             librsvg2-dev \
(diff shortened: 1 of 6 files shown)


naingthet (commented):
The helper is the right fix for the v0.54.0 hang: rewrite `azure.archive.ubuntu.com` first, then `apt-get` with 20s timeouts, 3 retries, and an 8-minute step bound. amd64 goes to `archive.ubuntu.com`, arm64 to `ports.ubuntu.com`. Dry-run fixtures cover both, plus a missing `sources.list`.

`release.yml` only swaps the packaging-deps install. Secrets, environments, checkout refs, and cache writers are untouched, so this is compatible with the base-branch trusted policy. The new assertions live in the PR copy of `workflow-security.test.mjs` and do not weaken that policy.

The follow-up commit escapes the full regex metacharacter set, including backslash, so the CodeQL `js/incomplete-sanitization` alert on the step-name matcher should clear.

No correctness blocker. Hold for required CI. The branch is one commit behind `#2251` with no file overlap; squash onto current main when green.

naingthet:
### Merge activity

* **Aug 19, 5:56 PM UTC**: A user started a stack merge that includes this pull request via [Graphite](https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2252).
* **Aug 19, 5:57 PM UTC**: [Graphite](https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2252) rebased this pull request as part of a merge.
* **Aug 19, 6:02 PM UTC**: @naingthet merged this pull request with [Graphite](https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2252).

PR #2253, opened 2026-08-19
fix(code): compact replay deltas

naingthet:
## What changed

- batches historical Code-mode replay frames before publishing them to the session store
- preserves immediate delivery for live event frames
- adds focused controller coverage for replay batching, ordering, and live-frame behavior

## Why

Large harness replays were publishing one renderer state update per historical frame. That made completed sessions feel noisy and expensive to restore, especially after high-volume Claude Code, Codex, OpenCode, and Grok runs. The replay path now coalesces those deltas into bounded updates without delaying live work.

## Compatibility and risk

This is renderer-only and does not change the server journal or harness protocol. Live frames still publish synchronously; only frames marked as replayed are batched. The main risk is ordering at the replay/live boundary, covered by the focused controller tests.

## Validation

- desktop UI suite: 190 files, 1,280 tests passed
- `git diff --check main...HEAD`

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeSessionController.test.ts
+++ b/crates/tidebreak-desktop/ui/src/code/CodeSessionController.test.ts
@@ -1,0 +1,79 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+import type { SequencedCodeEventFrame } from "../api/types";
+import { CodeSessionController } from "./CodeSessionController";
+
+class FakeSocket {
+  onopen: (() => void) | null = null;
+  onerror: (() => void) | null = null;
+  onclose: (() => void) | null = null;
+
+  constructor(
+    readonly emit: (frame: SequencedCodeEventFrame) => void,
+  ) {}
+
+  close() {}
+}
+
+afterEach(() => {
+  vi.useRealTimers();
+});
+
+describe("CodeSessionController", () => {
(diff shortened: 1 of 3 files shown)


naingthet (commented):
Adjacent same-type replay deltas are folded in place, and the compacted frame keeps the newest seq so reconnect still resumes after every durable row. Live frames still flush pending replay first (`[...replay, frame]`), so the replay/live boundary stays ordered.

The 10k-character Grok-style run becoming one `reasoning_delta` is the right regression. The PR body also claims live-frame coverage; that path is unchanged and is not in the new test. Not a blocker.

Renderer-only, no journal or harness change. Hold for required desktop UI.

naingthet (commented):
Second commit is the merge candidate: replay compaction now happens on both sides.

Server `replay_after` folds consecutive `AssistantDelta`/`ReasoningDelta` rows and keeps the last seq, so reconnect still resumes after every durable row. Live frames stay uncompacted. A gap (`seq` 5 then 7) is not merged. The unit test covers those boundaries.

The renderer keeps the same fold, but buffers chunks and joins once so a 10k-character Grok replay is linear. `takeReplay` still flushes the open run before a live frame, so the replay/live boundary stays ordered.

Postgres failed on `postgres_turn_admission_serializes_processes_and_recovers_expiry`, which this PR does not touch. Treat that as a flake unless it repeats. Hold for required CI.

PR #2254, opened 2026-08-19
perf(code): cache quick-open file lists

naingthet:
## What changed

- cache the workspace filename list across repeated quick-open invocations
- invalidate the cache when the worktree content revision changes
- clear stale results and close the picker when switching workspaces
- cover cache reuse, revision invalidation, and workspace transitions

## Why

Opening Cmd/Ctrl+P repeatedly should be instant and should not issue the same large workspace-tree request every time. The picker also must never show filenames from the previously selected workspace.

## Compatibility and risk

This is a UI-only behavior change with no API or persisted-data changes. Cache invalidation continues to use the existing content revision, and workspace changes explicitly discard all cached state.

## Validation

- `pnpm --dir crates/tidebreak-desktop/ui exec vitest run src/code/CodeQuickOpen.dom.test.tsx` — 5 passed
- `git diff --check`

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeQuickOpen.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeQuickOpen.dom.test.tsx
@@ -85,4 +85,94 @@ describe("CodeQuickOpen", () => {
     ).toHaveFocus();
     expect(client.listCodeWorkspaceTree).toHaveBeenCalledTimes(1);
   });
+
+  it("reuses the file list until the worktree revision changes", async () => {
+    const user = userEvent.setup();
+    const client = {
+      listCodeWorkspaceTree: vi.fn(async () => ({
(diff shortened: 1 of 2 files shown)


naingthet (commented):
Cache key is `workspaceId:contentRevision`, which is the right invalidation: reopen is a hit until the worktree revision moves. Switching workspaces clears paths, closes the picker, and drops `loadedKeyRef`, so `old.ts` cannot linger. `openWorkspaceId` keeps a fetch from applying to a workspace the dialog was not opened for.

Title was `improve(code)`, which is not an allowed type. Retitled to `perf(code)` so the semantic-title and release-label lanes can pass.

No correctness blocker. Hold for required desktop UI.

naingthet:
Rebased onto main after #2253. Cache change does not overlap the replay compaction. CI re-running.

PR #2255, opened 2026-08-19
feat(code): clarify source-control flow

naingthet:
## What changed

- present one clear next action for commit, push, or pull-request creation
- keep existing pull-request state visible while local changes or unpushed commits remain
- tighten GitHub CLI remediation and external pull-request linking
- expand the PrCard DOM coverage for each source-control state

## Why

The previous card showed all three actions at once, including disabled actions that were not yet relevant. The revised flow makes the next required step obvious while preserving context for an existing pull request.

## Validation

- `pnpm --dir crates/tidebreak-desktop/ui exec vitest run src/code/PrCard.dom.test.tsx` (12 passed)
- `git diff --check`
- full UI Vitest suite also passed during dependency verification (189 files, 1282 tests)

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/PrCard.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/PrCard.dom.test.tsx
@@ -60,28 +60,32 @@ describe("PrCard", () => {
   it("shows the no-commits state", () => {
     renderState(BASE);
     expect(screen.getByText("No commits")).toBeInTheDocument();
-    expect(screen.getByRole("button", { name: "Commit" })).toBeDisabled();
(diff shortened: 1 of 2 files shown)


naingthet (commented):
The card now shows one next action: commit when dirty, push when unpushed, create PR when the branch is pushed and has no PR, otherwise the existing PR digest. PR state stays visible as a chip while local work remains. GitHub CLI remediation is limited to the create-PR step, which is the only place it is needed.

`feat(code)` is an allowed type. No correctness blocker. Hold for required desktop UI.

naingthet:
Rebased onto main after #2253. No file overlap. CI re-running.

PR #2256, opened 2026-08-19
fix(code): reveal restored transcripts atomically

naingthet:
## What changed

- keeps restored Code transcripts behind one skeleton until both the durable turn snapshot and initial event replay have settled
- publishes replay frames and the hydration transition in one session-store update
- reveals genuinely empty sessions after the same bounded quiet window
- falls back to a usable transcript while the event socket reconnects

## Why

Reopened sessions could briefly show user turns, then visibly reconstruct assistant text and tool activity underneath them. The replay is now revealed atomically, so restoring a workspace feels like loading a completed document rather than replaying old typing.

## Stack

This is the next focused slice after #2253. Its branch contains the replay-compaction commits until #2253 lands; the PR diff will shrink to this atomic-reveal commit when the base work is on `main`.

## Compatibility and risk

The event protocol is unchanged. The renderer still applies live frames immediately after initial settlement. The main edge cases—empty history, replay history, snapshot failure, and initial socket failure—have focused coverage.

## Validation

- 5 focused UI test files, 45 tests passed
- TypeScript typecheck passed
- `git diff --check`

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeSessionController.test.ts
+++ b/crates/tidebreak-desktop/ui/src/code/CodeSessionController.test.ts
@@ -19,6 +19,88 @@ afterEach(() => {
 });
 
 describe("CodeSessionController", () => {
+  it("reveals an empty transcript after the initial socket stays quiet", async () => {
+    vi.useFakeTimers();
+    const sockets: FakeSocket[] = [];
+    const batches: Array<{
+      events: readonly SequencedCodeEventFrame[];
+      settled: boolean;
+    }> = [];
+    const controller = new CodeSessionController({
+      openSocket: (_after, onFrame) => {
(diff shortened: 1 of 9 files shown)


naingthet (commented):
The first `onEvents` delivery after the replay quiet window now carries `initialViewSettled`, so the store can publish history and drop the skeleton in one update. Empty sessions still wait for that window instead of flashing an empty transcript. A reconnect after settlement does not re-hide the view.

Title was `improve(code)`; retitled to `fix(code)`. The branch is already rebased onto main after #2253. Hold for required desktop UI.

naingthet:
Rebased onto main after #2254. CI re-running.

naingthet (commented):
:white_check_mark: g2g

PR #2257, opened 2026-08-19
perf(code): coalesce model catalog requests

naingthet:
## What changed

- cache completed empty model catalogs so opening another picker does not rerun the harness CLI
- share in-flight per-harness model probes across concurrent consumers
- coalesce catalog refreshes mounted by the sidebar and route body
- keep explicit refreshes authoritative by reprobing cached harness catalogs

## Why

Code mode can mount multiple model consumers together. Empty catalogs were treated as cache misses, and simultaneous mounts issued duplicate native model discovery requests. This avoids repeated CLI work while preserving explicit refresh behavior.

## Validation

- `pnpm --dir crates/tidebreak-desktop/ui exec vitest run src/code/CodeCatalogStore.test.ts` (6 passed)
- `git diff --check`

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeCatalogStore.test.ts
+++ b/crates/tidebreak-desktop/ui/src/code/CodeCatalogStore.test.ts
@@ -1,4 +1,4 @@
-import { afterEach, describe, expect, it } from "vitest";
+import { afterEach, describe, expect, it, vi } from "vitest";
 
 import type { CodeWorkspaceSnapshot } from "../api/types";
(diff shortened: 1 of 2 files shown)


naingthet (commented):
Empty catalogs are now a hit, not a miss, so opening another picker does not rerun the harness CLI. In-flight probes are shared per kind, and sidebar plus route body share one `catalogRefresh`. Explicit `refresh` still forces a reprobe. `reset` clears both maps.

No correctness blocker. Hold for required desktop UI.

naingthet (commented):
:white_check_mark: g2g

PR #2260, opened 2026-08-19
fix(desktop): hide idle sidebar resize rule

naingthet:
## Summary

- hide the sidebar resize hairline while the rail is idle
- center the rule on the resize target boundary
- reveal it on hover, keyboard focus, and active dragging

## Root cause

The resize handle rendered its full-height hairline continuously. Because the handle sits beside rounded content surfaces, the idle rule remained visible through the rounded corner cutout on first render and looked like it crossed the curve.

## User impact

The sidebar and content card now meet cleanly at rest. The resize affordance remains visible whenever the reader interacts with it, including keyboard focus.

## Compatibility and risk

This is a renderer-only styling change in the existing sidebar resize handle. It does not alter resize behavior, persisted widths, layout bounds, or backend APIs.

## Validation

- `pnpm exec tsc --noEmit`
- `pnpm exec vitest run src/AppShell.dom.test.tsx` — 23 tests passed
- `pnpm run build`
- `git diff --check`

### DIFF
--- a/crates/tidebreak-desktop/ui/src/sidebar/primitives.tsx
+++ b/crates/tidebreak-desktop/ui/src/sidebar/primitives.tsx
@@ -121,7 +121,10 @@ function SidebarResizeHandle({
         }
       }}
     >
-      <div className="absolute top-0 right-1 bottom-0 w-[0.5px] bg-border transition-all group-hover:w-px group-hover:translate-x-[0.5px] group-hover:bg-foreground/35 group-hover:delay-75 group-data-[dragging]:bg-foreground/75" />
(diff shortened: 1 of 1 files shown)


naingthet (commented):
The idle hairline was `right-1` on a handle that already straddles the rail (`translate-x-1/2`), so the rule sat on the content-card side and showed through the rounded cutout. Centering it on the hit target and keeping `opacity-0` until hover, `:focus-visible`, or drag matches the cause. Keyboard resize still works; this only changes when the rule paints.

Renderer-only. No resize math, persisted width, or layout bounds change. Hold for required desktop UI.

naingthet (commented):
:white_check_mark: g2g

PR #2261, opened 2026-08-19
fix(code): recover workspace startup after harness switches

naingthet:
## What changed

This PR preserves two coupled, focused commits:

1. `improve(code): reset models when switching harnesses`
   - clears the previous harness model immediately
   - shows an explicit loading state while the next harness catalog resolves
   - applies the behavior in both the new-workspace dialog and empty-workspace session prompt

2. `fix(code): recover newly created workspaces`
   - separates workspace creation failures from initial session startup failures
   - keeps and opens a successfully created workspace when the first session cannot start
   - reports the session error so the user can retry from the recovered workspace

## Why

A model selected for one harness could remain visible and be submitted while another harness catalog was still loading. Separately, a session startup failure after successful workspace creation left the workspace created but did not navigate the user to it, making recovery unclear.

## Compatibility and risk

- No API or persistence schema changes.
- Existing successful workspace/session creation behavior is unchanged.
- Scope is limited to five Code-mode UI and test files.
- The final files match the previously validated preservation commit `1a7f7d7f` for this slice.

## Validation

- Focused Vitest: 2 files, 12 tests passed
- Full desktop UI suite (triggered by the package test wrapper): 190 files, 1,284 tests passed
- TypeScript: `tsc --noEmit` passed
- `git diff --check` passed

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeComposer.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeComposer.tsx
@@ -126,6 +126,7 @@ export function HarnessModelMenu({
   value,
   onChange,
   disabled,
+  loading = false,
   variant = "composer",
 }: {
   harness: HarnessKind;
@@ -133,6 +134,7 @@ export function HarnessModelMenu({
   value?: string;
   onChange?: (model: string) => void;
   disabled?: boolean;
+  loading?: boolean;
   /** Composer sits above the draft; field fills a form row. */
   variant?: "composer" | "field";
 }) {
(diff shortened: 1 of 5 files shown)


naingthet (commented):
Harness switch now clears the previous model and shows a disabled loading control in both the new-workspace dialog and the empty-session prompt, so a foreign model cannot be submitted while the next catalog is in flight. Workspace create and first-session start are split: a created workspace is upserted, opened, and reported when the session fails.

`improve` is not an allowed type. Retitle to `fix(code): …` and add `semver:patch` plus `release-note:fix`. `ensureHarnessModels` still has no rejection path, so a failed catalog probe can leave `modelLoading` stuck; not a merge blocker if the store always resolves.

Hold for the title/labels and required desktop UI.

naingthet (commented):
:white_check_mark: g2g

PR #2262, opened 2026-08-19
fix(code): cover pull request blockers while watching

naingthet:
## What changed

- makes Watch and fix associate status with the current head SHA on every cycle
- treats conflicts and a behind-base branch as repair work, including fetch/rebase/conflict resolution and renewed monitoring after push
- handles draft PRs by marking them ready when authorized or reporting the exact authorization blocker

## Why

The initial action could otherwise wait forever on merge-state blockers or a draft even after checks passed.

## Compatibility and safety

This only strengthens the generated agent instruction. It does not add a background worker, silently mutate PR state, or claim task-fork durability that Tidebreak does not yet provide.

## Validation

- `pnpm --dir crates/tidebreak-desktop/ui exec vitest run src/code/CodeWorkspacePage.dom.test.tsx src/code/prActions.test.ts` — 25 passed
- `git diff --check`


naingthet (commented):
:white_check_mark: g2g

PR #2263, opened 2026-08-19
fix(code): preserve replay journal sequences

naingthet:
## What changed

- remove server-side replay compaction from the Code session WebSocket endpoint
- emit every durable journal event with `seq > after` in ascending sequence order
- replace the compaction unit test with an integration contract test covering adjacent assistant deltas

## Root cause

The replay endpoint merged adjacent text-delta journal rows into a rewritten WebSocket frame carrying only the final sequence number. Reconnecting API and CLI consumers therefore observed a different stream from the durable journal and could interpret the missing intermediate sequence as data loss.

## Backward compatibility and safety

This restores the documented row-for-row replay contract and makes replay match the live/debug journal stream again. It does not change the journal schema, live event delivery, or renderer-local batching and compaction. Existing clients receive more replay frames only where the server previously collapsed adjacent deltas; frame shapes and cursor semantics remain unchanged.

## Validation

- `cargo test -p tidebreak-server ws_replay -- --nocapture` (5 passed)
- `git diff --check`

### DIFF
--- a/crates/tidebreak-server/src/routes/code/session_events.rs
+++ b/crates/tidebreak-server/src/routes/code/session_events.rs
@@ -6,7 +6,7 @@ use axum::response::Response;
 use tokio::sync::broadcast::error::RecvError;
 
 use tidebreak_core::db::code::list_events;
-use tidebreak_core::{CodeEvent, CodeSessionId, OwnerId, SequencedCodeEvent};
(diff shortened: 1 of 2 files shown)


naingthet (commented):
:white_check_mark: g2g

PR #2264, opened 2026-08-19
feat(code): show subscription usage in sidebar

naingthet:
## Summary

- add a fixed subscription-usage meter to the bottom of the Code mode sidebar
- show provider/account quota windows, reset timing, warning states, refresh controls, and progressive disclosure for shared accounts
- read rich usage data from `modelctl --json usage`, with a direct Codex app-server fallback when Model Gateway is unavailable
- preserve an honest unavailable state for harnesses without a stable machine-readable quota interface

## Implementation

The server exposes a new owner-scoped `GET /code/usage` route. Model Gateway is preferred because it provides normalized multi-provider and shared-account data. If it is unavailable, Tidebreak queries Codex through `account/rateLimits/read` and maps its primary and secondary windows into the same response shape.

The renderer validates the response before displaying it. Raw Model Gateway account labels, binding identifiers, and owner emails are not returned to the renderer; accounts receive local `Personal` and `Shared` labels instead. Collection failures degrade to diagnostics or an unavailable report rather than breaking Code mode.

The sidebar keeps the compact meter in the fixed footer above Theme and Settings. The detailed popover is viewport-bounded and scrollable, and automatic refreshes cannot overlap an in-flight request.

## Compatibility and risk

- no schema or persistent-data migration
- no behavior change for Chat mode or other sidebar consumers
- installations without Model Gateway continue to work through direct Codex usage when available
- Claude Code and Grok usage remain unavailable rather than relying on unstable shell scraping
- provider commands are time-limited and response sizes are bounded

## Validation

- `cargo test -p tidebreak-server routes::code::usage::tests --lib`
- `pnpm exec tsc --noEmit`
- `pnpm exec vitest run src/code/parsers.test.ts src/code/CodeSidebar.dom.test.tsx`
- live visual verification with Model Gateway data in light and dark themes
- verified default and minimum-width sidebars, warning/critical meters, bounded scrolling, and shared-account disclosure

### DIFF
--- a/crates/tidebreak-desktop/ui/src/api/client.ts
+++ b/crates/tidebreak-desktop/ui/src/api/client.ts
@@ -98,6 +98,7 @@ import {
   type CodeWorkspaceSnapshot,
   type CodeCloneDefaults,
   type CodeCloneJobSnapshot,
+  type CodeSubscriptionUsage,
   type HarnessDoctorReport,
   type HarnessKind,
   type SequencedCodeEventFrame,
@@ -145,6 +146,7 @@ import {
   parseHarnessDoctorReport,
   parseSequencedCodeEvent,
   parseCodeUpdateNotice,
+  parseCodeSubscriptionUsage,
 } from "../code/parsers";
 
 const WS_HANDSHAKE = "tidebreak-v1";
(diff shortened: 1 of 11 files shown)


naingthet (commented):
:white_check_mark: g2g

PR #2265, opened 2026-08-19
feat(desktop): attach the client to a remote machine

naingthet:
Decision record 47 point 3 gives the desktop app a remote connection mode: the same webview, attached to a `tidebreak-server` running somewhere else, reached with a base URL and a token you supply. This is the connection plumbing for it. It is the first of three changes; host-authority refusals and the connect UI follow.

## What this adds

A `remote` module in `tidebreak-desktop` holds the attachment. The address lives in `remote-machine.json` under the profile data dir. The token lives in the OS credential store, keyed by the channel-scoped keychain service, so a staging build never reads a release build's token. An address with no token behind it reads as *not attached* — half an attachment would boot the renderer against a machine it cannot authenticate to.

Three commands carry it: `remote_machine_state`, `connect_remote_machine`, and `disconnect_remote_machine`, each allowlisted in `native-only-commands.txt` because reaching the OS credential store is the native shell's job.

## Two rules govern the address

**TLS unless loopback.** The token is a bearer credential for somebody's whole account, so cleartext to anywhere but this machine hands it to whoever is on the path. That is the pairing rule in `docs/gateway-boundary.md`. The provision-link check now calls the same function rather than keeping a second copy, so the cleartext exception has one definition.

**Nothing is stored until the machine answers.** Connecting probes the machine with the token before writing anything. A refused connect leaves the previous attachment exactly as it was, and a stored address is one that worked at least once.

## `ServerInfo` says which machine

`ServerInfo` gains an `attachment` field, `local` or `remote`. Callers that reach host authority read it there instead of inspecting the base URL: a developer machine on loopback is still remote, and "is this URL loopback" is not the question any caller has. `server_info` answers from the attachment first and does not wait for the embedded server — a local boot failure must not keep a remote client off a machine that is running fine.

## Refusals carry a reason, not prose

`RemoteConnectError` is a stable reason plus an optional `detail` for logs. The renderer owns the copy and switches on the reason, the way it already does for an export failure. The reasons:

- `remote_machine_url_invalid`
- `remote_machine_requires_tls`
- `remote_machine_unreachable`
- `remote_machine_token_refused`
- `remote_machine_not_a_machine`
- `remote_machine_token_storage_failed`

## What does not change

The embedded server still boots and still holds host authority; what changes is which machine the renderer talks to. The desktop profile's loopback lockdown in `tidebreak-server/src/auth.rs` is untouched — a remote client talks to a server whose profile is not desktop, so that gate has nothing to relax, and relaxing it would reopen the origin and DNS-rebinding refusals it exists to hold.

Host authority is not gated yet. That is the next change in the stack.

## Checks

- `cargo fmt --all --check`
- `cargo test -p tidebreak-desktop --lib` — 163 pass, including 7 new tests covering the TLS rule, the loopback exception, addresses refused before any probe, probe-status mapping, and the token/address pairing.
- `cargo clippy -p tidebreak-desktop --all-targets` — clean
- `tsc --noEmit`, plus `AppShell.dom.test.tsx` and `api.test.ts`

### DIFF
--- a/crates/tidebreak-desktop/Cargo.toml
+++ b/crates/tidebreak-desktop/Cargo.toml
@@ -37,7 +37,7 @@ cap-std.workspace = true
 futures.workspace = true
 flate2 = "=1.1.9"
 tidebreak-code-execution = { path = "../tidebreak-code-execution" }
-tidebreak-core = { path = "../tidebreak-core", default-features = false }
+tidebreak-core = { path = "../tidebreak-core", default-features = false, features = ["keychain"] }
 tidebreak-host-broker = { path = "../tidebreak-host-broker" }
 tidebreak-server = { path = "../tidebreak-server" }
 reqwest = { workspace = true }
--- a/crates/tidebreak-desktop/native-only-commands.txt
+++ b/crates/tidebreak-desktop/native-only-commands.txt
@@ -26,6 +26,11 @@
 server_info: hands the renderer the embedded server's address and per-launch token — the handshake that makes every HTTP route reachable, so it cannot itself be one.
 request_user_attention: native window attention hint when a parked question needs the user.
 
+# Attaching this client to a machine it does not host (ADR 0047)
+remote_machine_state: reports which machine this client is attached to, from shell-held state that exists before any machine is reachable.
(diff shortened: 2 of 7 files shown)


naingthet (commented):
:white_check_mark: g2g

PR #2266, opened 2026-08-19
fix(providers): preserve auth enablement across updates

naingthet:
## Summary

- restore missing Anthropic and OpenAI provider rows at startup when update-surviving authentication is present
- treat a valid ChatGPT OAuth sign-in like an API key when determining initial provider enablement
- preserve explicit provider disables and reject orphaned OAuth tokens without the provider credential marker

## Root cause

The boot migration only restored Anthropic enablement from a surviving legacy/API key. ChatGPT OAuth credentials live independently in the OS credential store, so an update could retain the sign-in while losing the OpenAI provider row. Since the boot-default model is OpenAI, the catalog then reported no usable default model even though the user was still signed in.

## Compatibility and risk

The migration only writes a provider row when no row exists. Existing settings, including an explicit `enabled: false`, remain untouched. Managed profiles continue to skip BYOK auto-enablement.

## Validation

- `cargo test -p tidebreak-server providers::tests::upgrade_ --lib`
- existing focused credential/model tests passed before publication:
  - `cargo test -p tidebreak-server writing_a_provider_credential_enables_the_provider --lib`
  - `cargo test -p tidebreak-server chatgpt_auth_marks_api_only_openai_models_unavailable --lib`

### DIFF
--- a/crates/tidebreak-server/src/lib.rs
+++ b/crates/tidebreak-server/src/lib.rs
@@ -1265,19 +1265,19 @@ async fn bind_inner(
     // below would read the unmigrated profile as unmanaged and clear the
     // very session the import exists to preserve.
     managed_policy::import_legacy_setting(&*provisioned_policy, &*store).await?;
-    // The legacy Anthropic auto-enable is gated on one policy read. A resolution
+    // Legacy credential auto-enable is gated on one policy read. A resolution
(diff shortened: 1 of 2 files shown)


naingthet (commented):
:white_check_mark: g2g

PR #2267, opened 2026-08-19
feat(desktop): refuse host authority while attached remotely

naingthet:
Stacked on #2265. Review that one first; this diff is only the second commit.

Four capabilities exist only on the machine this process runs on: the folder broker, the client executor, native export, and computer use. All four reach your own filesystem, screen, or input devices through credentials the renderer never holds. While the client is attached to a remote machine, none of them apply to the conversation you are looking at.

## What this refuses, and why separately

Decision record 47 names the failure mode to prevent: a remote-attached client that half-works, performing a folder operation or a native export against the wrong host — the server's filesystem, or this machine's, for a conversation on neither. Each authority refuses instead, with its own stable reason:

| Authority | Reason |
|---|---|
| Folder broker | `folder_broker_authority_unavailable` |
| Client executor | `client_executor_authority_unavailable` |
| Native export | `native_export_authority_unavailable` |
| Computer use | `computer_use_authority_unavailable` |

Four reasons rather than one shared one. They are lost for the same cause but have four different consequences, and a client that can only learn "something host-shaped was refused" cannot tell you which capability it lost.

## Where the gate sits

`HostAccess` holds the attachment, and every host-authority command opens with `require_local`. That is the chokepoint: all seventeen commands already take `State<HostAccess>`, so the gate costs one line each and nothing reaches the broker, the control plane, or a save dialog ahead of it.

The error is the bare reason code with no prose around it, following `output_writeback_authority_unavailable`. These commands otherwise return free-text errors, so `hostAuthorityRefusal` in the renderer tells a refusal from a failure by comparing against the code rather than reading the message. The copy stays the renderer's to write, which is how export failures already work.

## One deliberate exception

`computer_use_state` is left ungated. It reports the native control state this process holds, which is genuinely empty while you are working on another machine. Refusing a read that already answers correctly would only cost the indicator its ability to say nothing is happening.

## Checks

- `cargo fmt --all --check`
- `cargo test -p tidebreak-desktop --lib` — 166 pass, including the decision-47 validation case: attached remotely, every authority refuses with its own reason rather than succeeding against the wrong host.
- `cargo clippy -p tidebreak-desktop --all-targets` — clean
- `tsc --noEmit`, plus a new `host.test.ts` that pins the four codes on the renderer side so a rename on either side fails a test instead of passing silently.

### DIFF
--- a/crates/tidebreak-desktop/src/client_execution.rs
+++ b/crates/tidebreak-desktop/src/client_execution.rs
@@ -79,6 +79,9 @@ pub(crate) async fn resolve_folder_access_request(
     state: State<'_, HostAccess>,
     request: ResolveFolderAccessRequest,
 ) -> Result<(), String> {
+    state
+        .require_local(crate::host_authority::Authority::ClientExecutor)
+        .await?;
     if request.chat_id.is_nil() || request.call_id.is_nil() {
         return Err("invalid folder-access request identity".to_owned());
     }
--- a/crates/tidebreak-desktop/src/client_execution/computer_use.rs
+++ b/crates/tidebreak-desktop/src/client_execution/computer_use.rs
@@ -484,6 +484,9 @@ pub(crate) async fn stop_computer_use_control(
     app: AppHandle,
     state: State<'_, HostAccess>,
 ) -> Result<(), String> {
+    state
+        .require_local(crate::host_authority::Authority::ComputerUse)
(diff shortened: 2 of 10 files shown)


naingthet (commented):
:white_check_mark: g2g

naingthet (commented):
:white_check_mark: g2g

PR #2269, opened 2026-08-19
feat(desktop): connect to a machine from settings

naingthet:
Stacked on #2267, which is stacked on #2265. Review those first; this diff is only the third commit.

The last piece of decision record 47 point 3: a place to attach this window to another machine, and copy for what that costs.

## The Machine section

A new settings section takes an address and a token and says which machine you are on. It sits last among the substantive sections, because attaching changes what every section above it is configuring — you meet them in their default meaning first.

Connecting and disconnecting reload the window. The API client, the event stream, and every cached read were built against the machine this window opened on; rebuilding them piecemeal is how a client ends up half-attached.

The panel states the trade before you make it, rather than letting you find it when a folder picker refuses. It lists the four capabilities that stay on this computer — connected folders, tool calls that run here, saving files here, and computer use — and strikes them through while you are attached elsewhere.

## One place owns the words

`remoteMachine.ts` holds every sentence. Both boundaries send stable reasons and no prose, so the wording changes here without touching the shell: the six connect refusals from #2265 and the four host-authority refusals from #2267.

Connect reasons are recognized from a table rather than accepted as any string. A reason nobody has worded yet surfaces as an ordinary failure instead of as a blank.

## Refusals reach the reader

Wherever a host command can fail, a refusal now says which capability is unavailable and why, in place of the raw code: the folder-access and write-back decision cards, the folders panel, the permissions panel, and output export. Output export routes through `friendlyOutputError`, which was already the single chokepoint for that surface.

## Checks

- `cargo fmt --all --check`; `cargo test -p tidebreak-desktop --lib` — 166 pass, unchanged by this commit
- `tsc --noEmit`
- Full UI suite: 1293 tests across 192 files pass, including new tests that every connect reason has copy, that the https refusal says so, and that a host refusal names the capability rather than printing its code.

One existing test double needed widening: `useFolderAccessRequests.test.ts` mocked the whole host module, so the real refusal mapping is now spread in rather than stubbed out — otherwise the test would exercise a path the app never runs.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/FoldersView.tsx
+++ b/crates/tidebreak-desktop/ui/src/FoldersView.tsx
@@ -42,7 +42,7 @@ import {
 } from "./FolderAccess";
 import { verbLabel } from "./settings/PermissionsPanel";
 import { useRefreshSignals } from "./RefreshSignals";
-import { friendlyErrorMessage } from "@/lib/utils";
+import { hostErrorMessage } from "@/remoteMachine";
 
 /**
  * A chat's connected folders: the directories the native host may reach on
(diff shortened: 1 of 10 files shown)


naingthet (commented):
:white_check_mark: g2g

naingthet (commented):
:white_check_mark: g2g

PR #2272, opened 2026-08-19
fix(code): compact subscription usage panel

naingthet:
## Summary

- reduce the subscription-usage popover width, header height, padding, and section spacing
- collapse each usage window into one aligned row with its label, meter, percentage, and reset duration
- move freshness and exceptional account state into the account header while hiding redundant `Available` labels
- retain full percentage/reset details in titles and expose progress values to assistive technology

## Why

The original provider/account/window layout repeated a label row, meter, reset row, and account metadata for every quota. With multiple Anthropic accounts this made the panel visually busy and tall enough to require significant scrolling.

The compact table-like treatment preserves the same information while making provider and account hierarchy easier to scan. Representative three-account data fits in roughly 390 × 363 px without internal scrolling.

## Compatibility and risk

- presentation-only change; subscription usage fetching and refresh behavior are unchanged
- shared-account progressive disclosure is unchanged
- warning and critical thresholds retain their existing colors and logic
- responsive width remains bounded by the viewport

## Validation

- `pnpm exec vitest run src/code/CodeSidebar.dom.test.tsx`
- `pnpm exec tsc --noEmit`
- `pnpm build`
- visual verification with the real component, production theme CSS, and representative multi-account data

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeSidebar.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeSidebar.dom.test.tsx
@@ -169,7 +169,15 @@ describe("CodeSidebar", () => {
     expect(await screen.findByText("Subscription usage")).toBeInTheDocument();
     expect(screen.getByText("Model Gateway")).toBeInTheDocument();
     expect(screen.getByText("Weekly (Fable)")).toBeInTheDocument();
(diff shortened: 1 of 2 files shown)


naingthet (commented):
:white_check_mark: g2g

PR #2274, opened 2026-08-19
ci: require writer approval on outsider pull requests

naingthet:
## What changed

Outsider pull requests cannot enter the merge queue until someone with write access approves the current head commit. Maintainer pull requests stay green without a second review.

GitHub's native code-owner rule cannot do that split: it also blocks Engineering authors, who cannot approve their own pull requests. This check runs from `pull_request_target` against the pull request's **base** SHA, so a fork cannot replace the job with one that always passes.

An approval of an older commit does not count. A later `CHANGES_REQUESTED` review from the same writer replaces their approval.

## Validation

- `node --test scripts/check-engineering-approval.test.mjs scripts/workflow-security.test.mjs`
- `node --test scripts/workflow-security-mutations.test.mjs`

After this lands on `main`, the next step is to add `engineering approval` as a required status check and require this workflow from `refs/heads/main`. The check cannot be required before the workflow exists on the default branch, or every open pull request stalls.

## Compatibility and risk

Fork pull requests fail closed until a writer approves the current head. Dependabot and other bot authors need the same approval. Maintainers are unchanged.

## Tracking

None.

### DIFF
--- a/.github/workflows/engineering-approval.yml
+++ b/.github/workflows/engineering-approval.yml
@@ -1,0 +1,48 @@
+name: Engineering approval
+
+on:
+  # Trusted default-branch workflow. Forks cannot delete this job. Checkout is
+  # always the pull request's base SHA, never the head.
+  pull_request_target:
+    types:
+      [opened, reopened, synchronize, ready_for_review, converted_to_draft]
+  pull_request_review:
+    types: [submitted, dismissed, edited]
+  # Required checks must report on speculative merge groups or the queue stalls.
+  merge_group:
+
+permissions:
+  contents: read
+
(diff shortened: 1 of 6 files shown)


naingthet (commented):
:white_check_mark: g2g

naingthet:
The first `engineering approval` failure is this bootstrap PR checking out `main`, which does not have the script yet. The follow-up commit checks out `github.workflow_sha` so the workflow and script stay on the same commit. After this merges, later pull requests run the check from `main`.

naingthet (commented):
:white_check_mark: g2g

PR #2275, opened 2026-08-19
fix(desktop): restore categorical icon colors

naingthet:
## Summary

- restore categorical color to Tidebreak icons after the Gateway design-system pass
- use Vivian’s Tailwind color convention: `600` steps in light mode and brighter `400` steps in dark mode
- map the existing blue, cyan, violet, amber, rose, and green icon categories without changing component markup or adding colored tiles

## Root cause

The initial design-system port mapped every categorical icon token to the foreground color. That preserved the neutral shell but removed useful visual differentiation across welcome prompts, settings, tabs, tools, inbox, apps, plugins, and code mode.

## Compatibility and risk

- CSS-token-only change
- no behavioral, layout, data, or backend changes
- existing semantic status colors remain unchanged

## Validation

- `pnpm build`
- confirmed the generated CSS contains each referenced Tailwind `400` and `600` color token
- `git diff --check`

### DIFF
--- a/crates/tidebreak-desktop/ui/src/styles.css
+++ b/crates/tidebreak-desktop/ui/src/styles.css
@@ -77,14 +77,14 @@
   --citation-mark: oklch(0.87 0 0);
   --citation-mark-active-ink: var(--foreground);
 
-  /* The Gateway system is deliberately monochrome. Semantic colors remain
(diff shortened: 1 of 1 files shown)


naingthet (commented):
:white_check_mark: g2g

PR #2276, opened 2026-08-19
docs: document required reviews without a merge queue

naingthet:
## What changed

Branch protection is back to a simple policy: every pull request needs an Engineering / code-owner approval. `@brightwave-inc/engineering` can merge without waiting for that review. The merge queue is off.

This PR only updates the docs and deletes the custom `engineering approval` workflow. The GitHub settings are already applied.

## Validation

## Compatibility and risk

Outsider pull requests stay blocked until an Engineering member approves. Engineering members can use the admin / bypass merge path again.

## Tracking

None.


naingthet (commented):
Release policy is red for the trusted-policy two-step, not for the docs.

Main's `workflow-security.test.mjs` (from #2274) still requires `engineering-approval.yml` as a trusted base-branch check. This job runs that copy against the PR head, so deleting the workflow here throws `Cannot read properties of undefined` in `workflowJob`.

Land a first PR that drops the trusted assertion (and the mutation) while the workflow still exists. Then this deletion can pass. CONTRIBUTING change itself is fine.

PR #2277, opened 2026-08-19
fix(code): scope modelctl usage to paired gateway

naingthet:
## Summary

- gate Model Gateway subscription usage on Tidebreak's resolved managed policy
- require a `modelctl` profile whose normalized endpoint matches Tidebreak's paired gateway
- invoke `modelctl usage` with the matching profile explicitly instead of using its unrelated default profile
- clarify the subscription-usage empty-state copy

## Root cause

The `/code/usage` route preferred `modelctl --json usage` whenever a `modelctl` executable was discoverable on the user's shell PATH. It did not consult Tidebreak's gateway policy, so an unmanaged local profile could display usage from a globally installed and logged-in modelctl profile. It also allowed modelctl's default profile to supply data even when Tidebreak was paired to a different gateway.

## Behavior and compatibility

Unmanaged Tidebreak profiles now skip modelctl entirely and continue to query Codex directly. Managed profiles use modelctl only when one of its configured profiles names the same normalized gateway URL; otherwise they fall back to direct Codex usage. No persistent data or API response shape changes.

## Validation

- `cargo test -p tidebreak-server routes::code::usage::tests --lib` (3 passed)

The frontend change is copy-only. Frontend tests were unavailable in the worktree because JavaScript dependencies were not installed.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeSubscriptionUsage.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeSubscriptionUsage.tsx
@@ -302,9 +302,10 @@ function UnavailableUsage({ diagnostics }: { diagnostics: string[] }) {
     <div className="flex flex-col gap-2 py-2">
       <p className="text-sm font-medium">No machine-readable limits found</p>
       <p className="text-muted-foreground text-xs leading-5">
-        Tidebreak reads Model Gateway when modelctl is available, then falls
-        back to Codex directly. Claude Code and Grok do not currently expose a
-        stable non-interactive subscription-usage command.
+        Tidebreak reads Model Gateway when this profile is paired to the same
+        gateway configured in modelctl, then falls back to Codex directly.
+        Claude Code and Grok do not currently expose a stable non-interactive
+        subscription-usage command.
       </p>
       {diagnostics[0] && (
         <p className="text-muted-foreground text-[11px]">{diagnostics[0]}</p>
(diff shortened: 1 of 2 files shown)


naingthet (commented):
Title fails Conventional Commits: `improve` is not an allowed type. Rename to `fix(code): scope modelctl usage to paired gateway` (this is a real scoping bugfix).

The gate itself is right: unmanaged profiles skip modelctl, managed ones only invoke a profile whose normalized endpoint matches Tidebreak's paired gateway, and an unrelated default profile can no longer leak usage. Matching test covers the default-vs-paired case.

Postgres lane also failed on this head; treat that as independent of the title until the log says otherwise.

naingthet (commented):
Thet's review agent

:white_check_mark: g2g on the pairing gate. modelctl usage only runs when a profile endpoint matches this Tidebreak gateway after `validated_gateway_url`; unrelated PATH installs stay out. Test covers match vs miss.

Title `improve(code)` fails semantic-title / version-label. Retitle to `fix(code)` (or `feat(code)`).

naingthet (commented):
Thet's review agent

Hold — not g2g. Clippy is red on this head: `needless_borrow` at `crates/tidebreak-server/src/routes/code/usage.rs:148` (`command_from_probe(&probe)` → `probe`). Title still fails Conventional Commits (`improve` is not an allowed type). Semantic version label is also red because of that title.

Postgres `postgres_turn_admission_serializes_processes_and_recovers_expiry` failed. This PR does not touch turn admission, so treat that as a flake unless it reproduces.

The pairing gate itself is still right.

naingthet (commented):
Thet's review agent

:white_check_mark: g2g

PR #2278, opened 2026-08-19
fix(code): prefer local modelctl with direct fallback

naingthet:
## Summary

- restore local `modelctl --json usage` as the preferred subscription-usage source whenever `modelctl` is available
- fall back to the Codex app-server rate-limit API when `modelctl` is missing, signed out, errors, or returns no usable providers
- add focused regression coverage for installations without `modelctl`

## Context

PR #2277 scoped Model Gateway usage to Tidebreak’s managed gateway policy. That hides usage in local development when requests are routed through a locally configured `modelctl` but Tidebreak is not explicitly tethered to that gateway.

This follow-up removes the policy/profile coupling. It preserves graceful behavior for users who do not install Model Gateway by retaining the direct-provider fallback and making that selection logic directly testable.

## Compatibility and risk

The `/code/usage` response schema is unchanged. Users with working local `modelctl` installations again see Model Gateway usage. Users without `modelctl` continue to receive direct Codex usage when available. This only changes usage-source selection; request routing and authentication are unaffected.

## Validation

- `cargo test -p tidebreak-server routes::code::usage::tests --lib`
- `cargo fmt -- crates/tidebreak-server/src/routes/code/usage.rs`
- `git diff --check`

### DIFF
--- a/crates/tidebreak-server/src/routes/code/usage.rs
+++ b/crates/tidebreak-server/src/routes/code/usage.rs
@@ -1,18 +1,16 @@
 //! Subscription quota usage for code mode.
 //!
 //! Model Gateway is the richest source and already exposes a stable JSON CLI
-//! contract, but it belongs to this surface only when the Tidebreak profile is
-//! paired to that same gateway. An unrelated `modelctl` installation on PATH
-//! must not change an open profile's usage accounting. Otherwise Codex's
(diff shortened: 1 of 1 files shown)


naingthet (commented):
Thet's review agent

:white_check_mark: g2g

PR #2280, opened 2026-08-19
fix(desktop): brand code-mode models by family

naingthet:
## Summary

The code-mode model picker grouped DeepSeek and GLM under the right rail tab, but each row fell back to the engine mark when the id had no first-party vendor. On Codex CLI that painted DeepSeek and GLM with the OpenAI blossom.

`CodeModelMark` now uses the family icon when the model id names one. An unrecognized id still uses the engine mark.

## Test plan

- [x] `pnpm --dir crates/tidebreak-desktop/ui exec vitest run src/code/CodeComposer.dom.test.tsx`
- [ ] In New workspace, pick Codex CLI, then open Model. Confirm DeepSeek shows the whale and GLM shows the Z.ai mark, not the OpenAI blossom.


naingthet (commented):
Thet's review agent
:white_check_mark: g2g

PR #2284, opened 2026-08-20
fix(code): stop Unix harness process groups

naingthet:
## Summary

- launch Unix harness children in dedicated process groups
- send interrupt and termination signals to the whole group so wrapper descendants cannot keep sessions alive
- harden cancellation, drop, reaping, and inherited-pipe behavior against stale PGIDs and hanging descendants
- add Unix regression coverage for interrupt, terminate, drop, cancellation, and retained stdout descendants

## Compatibility and risk

- Windows Job Object behavior is unchanged
- this is a process-group containment boundary, not a claim of full OS-level ownership; descendants that deliberately call `setsid()` or move process groups can escape
- the change targets the observed Grok/shell-wrapper stop hang while preserving the existing grace-period behavior

## Validation

- `rustfmt --edition 2021 --check crates/tidebreak-harness/src/child.rs`
- `git diff --check`
- broad Rust tests and clippy intentionally left to CI

### DIFF
--- a/crates/tidebreak-harness/src/child.rs
+++ b/crates/tidebreak-harness/src/child.rs
@@ -9,10 +9,11 @@ use std::io;
 use std::process::{ExitStatus, Output};
 use std::time::Duration;
 
+use tokio::io::AsyncReadExt;
 use tokio::process::{Child, ChildStderr, ChildStdin, ChildStdout, Command};
 use tokio::sync::watch;
 #[cfg(unix)]
-use tokio::time::timeout;
+use tokio::time::sleep;
 
 use crate::TurnOutcome;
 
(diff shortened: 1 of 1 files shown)


naingthet (commented):
Thet's review agent

:white_check_mark: g2g

naingthet (commented):
Thet's review agent

:white_check_mark: g2g

naingthet (commented):
Thet's review agent

Not g2g. Clippy `-D warnings` is red on this head: two `clippy::needless_return` in `interrupt` / `terminate`. Drop the explicit `return`s.

naingthet (commented):
Thet's review agent

:white_check_mark: g2g

PR #2288, opened 2026-08-20
feat: authenticate hosted machines through gateway (ADR 0049)

naingthet:
Implements decision 49: hosted Tidebreak machines authenticate people through their existing Model Gateway session instead of a provisioned token file.

## Server
- A self-host server selects exactly one principal authenticator at boot: `TIDEBREAK_AUTH_GATEWAY_URL` or `TIDEBREAK_AUTH_TOKENS_FILE`. Both or neither is a boot error (decision 1).
- Gateway mode resolves `mg_at_` bearers through `GET /api/v1/tidebreak/principal`, mapping the stable user UUID to `user:<uuid>` ownership and the live administrator bit to the deployment-plane role (decision 2). Refusal, timeout, oversized or malformed responses admit nobody (decision 3).
- `TIDEBREAK_PUBLIC_URL` is required in gateway mode. The server hashes the canonical URL into the OAuth resource (`tidebreak:<sha256>`), binding every user credential to this one machine; a token captured from another Tidebreak host does not replay here.
- `/auth/discovery` advertises the mode, Gateway URL, and the machine-bound resource (decision 4).
- Gateway-authenticated WebSockets revalidate the original bearer every 60 seconds and close on identity or role change, so revocation reaches long-lived sockets (decision 5 / consequences).

## Desktop
- "Connect with Model Gateway" derives the machine resource locally from the validated base URL; discovery must echo it but cannot choose it. Discovery and probe refuse redirects and read bounded bodies.
- Machine bearers stay process-only: Tidebreak resources bypass the durable per-resource cache, entries persisted by older builds are scrubbed on the next vault save, and the rotating refresh token remains the only durable credential (validation item 3).
- The two new Tauri commands are recorded in `native-only-commands.txt`: both mint from the keychain-held OAuth session, which only the native shell can reach.

## Static tokens
Static-token attachment is unchanged for ordinary self-host servers (decision 6).

## Checks
- `cargo test -p tidebreak-core -p tidebreak-server -p tidebreak-desktop` — pass (one unrelated PTY test, `code_terminals::create_write_read_and_two_cursors`, is flaky under parallel load and passes in isolation)
- `cargo clippy --all-targets` on the three crates — clean
- `cargo fmt --check` — clean
- UI: `vitest run` 1349 pass, `tsc` clean

### DIFF
--- a/crates/tidebreak-core/src/config.rs
+++ b/crates/tidebreak-core/src/config.rs
@@ -16,6 +16,7 @@ use std::net::{Ipv4Addr, SocketAddr};
 use std::path::PathBuf;
 
 use serde::{Deserialize, Serialize};
+use sha2::{Digest, Sha256};
 
 use crate::error::{AgentError, Result};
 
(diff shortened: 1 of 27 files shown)


naingthet (commented):
Thet's review agent

Do not land this against current Model Gateway. `crates/tidebreak-server/src/connectors/gateway.rs` now puts `tidebreak:access` on the desktop authorize `SCOPE`. MG `canonical_scope` requires every requested token to be a subset of `SUPPORTED_SCOPE` (`openid profile offline_access models:read inference:invoke mcp:invoke runtime:execute control_plane:read`). That list has no `tidebreak:access`, so a new Tidebreak sign-in against today's gateway is `invalid_scope`.

The hosted attach path is also unimplemented on the gateway side. `resource_scope` has no `tidebreak` or `tidebreak:<sha256>` audience, so refresh mint returns `invalid_resource`. `GET /api/v1/tidebreak/principal` is not on MG `main`, and there is no open MG PR for it.

Keep the current authorize `SCOPE` until MG admits `tidebreak:access`, the machine-bound resource, and the principal lookup. Then land this with that MG change, or behind a Gateway version that already has them.

naingthet (commented):
Thet's review agent

:white_check_mark: g2g

naingthet (commented):
Thet's review agent

Not g2g. Retracting the earlier check.

Current Model Gateway `SUPPORTED_SCOPE` is `openid profile offline_access models:read inference:invoke mcp:invoke runtime:execute control_plane:read`. It does not include `tidebreak:access`, so a desktop authorize with this PR's `SCOPE` is `invalid_scope`. MG main also has no `GET /api/v1/tidebreak/principal` and no `tidebreak:<sha256>` resource mint.

Hold until those Gateway surfaces exist. The Tidebreak-side fail-closed / machine-bound resource work can stay as written.

naingthet:
The gateway hold is resolved. brightwave-inc/model-gateway#1099 is merged: `SUPPORTED_SCOPE` admits `tidebreak:access`, `resource_scope` resolves the machine-bound `tidebreak:<sha256>` audience (gated to the `tidebreak` OAuth client), and `GET /api/v1/tidebreak/principal` is live.

459aa6cfd also removes `tidebreak:access` from the desktop authorize scope. The Gateway derives that scope from the requested resource at refresh time, so sign-in stays compatible with Gateways that predate the Tidebreak audience — the deploy-ordering coupling the review flagged is gone.

naingthet (commented):
Thet's review agent

:white_check_mark: g2g

PR #2289, opened 2026-08-20
feat(code): make watch-and-fix a durable forked task

naingthet:
## Summary

- replace the prompt-only Watch and fix with a durable server-side watch task (new decision record 0049)
- add `code_watch` table and a try-based sweep that reads active rows every tick, so restarts resume watches with no extra recovery state
- give sessions a `kind` (`interactive` | `watch`): the watch forks a dedicated auto-mode session in the same worktree, and fix turns submit only while no other session in the workspace runs a turn
- classify the fresh PR digest each sweep and take one step: bounded fix turn (failing checks, conflicts, behind, requested changes), wait, park with NeedsYou, or finish
- add `head_sha` to the PR digest so a fix turn that does not move the head parks the watch instead of looping
- honor decision 42: the watch never merges, arms auto-merge, or marks drafts ready — the old prompt's auto-merge instruction is gone
- routes `POST`/`DELETE /code/workspaces/{id}/watch`; watch state rides the PR snapshot; watch digests carry their kind and never displace the interactive session in the updates store
- workflow control starts/stops the watch and shows Watching / Fixing / Blocked; Storybook stories cover the states
- CLI mirrors `kind` and never picks a watch session for `code run --ws`

## Validation

- TypeScript typecheck clean; full desktop UI suite passes (199 files, 1385 tests)
- Storybook build passes
- standalone formatting review and `git diff --check` (the two wire.ts trailing-space notes match the ts-rs generator's own output)

Broad Rust build, clippy, and the wire-type parity test are delegated to CI per repository guidance.

PR #2291, opened 2026-08-20
fix(code): bump the desktop schema epoch for the watch baseline

naingthet:
## Summary

- #2289 edited the schema baseline (`code_session.kind`, `code_watch`) without the `DESKTOP_SCHEMA_EPOCH` bump decision 2 requires, so an existing pre-v1 desktop database kept its old schema
- symptom on a live dev profile: every stall sweep, watch sweep, and workspace digest fails with `no such column: code_session.kind` / `no such table: code_watch`
- bump the epoch to 34 so pre-v1 databases rebuild on next launch

Pre-v1 lifecycle applies: the rebuild deletes the local SQLite database by design.

## Validation

- one-line constant change; `git diff --check` clean
- the epoch round-trip tests in `desktop_schema.rs` cover the reset path in CI

### DIFF
--- a/crates/tidebreak-server/src/desktop_schema.rs
+++ b/crates/tidebreak-server/src/desktop_schema.rs
@@ -33,7 +33,7 @@ const VECTOR_DIRECTORY: &str = "vectors";
 const MAX_MARKER_BYTES: u64 = 1_024;
 
 /// Bump this whenever the pre-v1 schema baseline changes incompatibly.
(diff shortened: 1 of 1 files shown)


naingthet (commented):
Thet's review agent

:white_check_mark: g2g

#2289 changed the pre-v1 baseline (`code_session.kind`, `code_watch`) without bumping `DESKTOP_SCHEMA_EPOCH`. 33 → 34 is the rebuild marker decision 2 requires. Do not merge from here.

PR #2292, opened 2026-08-20
feat(desktop): offer the center's openers from one + menu

naingthet:
## Summary

- the tab strip's `+` jumped straight into the file picker, and the browser opener was a second unlabeled glyph beside it
- one `+` now drops the list of what the center can open — Open file…, All changes (diff), New browser tab, Terminal — so a bare click says what it will do
- the two new openers are optional props; a region without them lists only the file picker
- the browser-obscure contract holds: the menu is a Radix overlay, so native webviews yield while it is open
- Storybook covers the strip: conversation-only, files open, browser tab, split's secondary group, and the minimal menu

## Validation

- desktop UI typecheck clean; CodeWorkspacePage dom suite passes (24 tests)
- Storybook build passes

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeCenterTabs.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeCenterTabs.tsx
@@ -12,6 +12,7 @@ import {
   FileCode,
   FileDiff,
   Globe2,
+  SquareTerminal,
   ListX,
   MessageSquare,
   MoveLeft,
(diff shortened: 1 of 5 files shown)


naingthet (commented):
Thet's review agent

The `+` menu is an opener, but `onNewTerminal` calls `toggleTerminalLayout`. That helper closes the drawer when the terminal is already in the layout — it is the status-bar control with `aria-pressed`. Choosing Terminal from New tab while the drawer is open therefore dismisses it instead of focusing it.

Open (or no-op if already present) from this menu; leave the toggle on the status-bar button. Both the primary and secondary strips wire the same handler.

naingthet:
Good catch — fixed: the menu now uses an open-only `openTerminalLayout` (no-op when the terminal is already in the layout); the status-bar control keeps the toggle. Typecheck and the page + chrome suites pass.

naingthet (commented):
Thet's review agent

:white_check_mark: g2g

One `+` menu instead of a bare file-picker click plus a second unlabeled browser glyph. Optional openers stay off the menu when a region does not pass them. Tests click through the menu items. Do not merge from here.

naingthet (commented):
Thet's review agent

:white_check_mark: g2g

naingthet (commented):
Thet's review agent

:white_check_mark: g2g

Follow-up is right: `openTerminalLayout` is open-only, so the + menu cannot close an already-open terminal. Do not merge from here.

PR #2296, opened 2026-08-20
feat(ci): publish the tidebreak-server image on release

naingthet:
## What this adds

`.github/workflows/publish-server-image.yml` builds the self-host server from
`deploy/self-host/Dockerfile` and publishes it to
`ghcr.io/brightwave-inc/tidebreak-server`.

The image is the artifact `deploy/self-host/docker-compose.yml` builds locally
as `tidebreak-self-host:local`: the `tidebreak` binary with tidebreak-server's
`postgres` feature, running `Profile::SelfHost`. Publishing it means operators
no longer compile the workspace to run a server, and model gateway deployments
that install Tidebreak as an add-on get a digest they can pin.

## Triggers

- **A published, non-prerelease release.** This is the primary path, and every
  release publishes a versioned image. `release` events load this file from the
  default branch, so a stale or hostile tag YAML cannot publish.
- **A weekly schedule**, Tuesday 05:17 UTC, off the on-the-hour scheduling
  spikes. It rebuilds the current latest release so Debian security patches in
  the runtime base flush through even when nothing in-repo changes.
- **Manual dispatch**, for retries and backfills. It takes an optional release
  tag; with none, it rebuilds the current latest release.

## Tag scheme

For a release tag `vX.Y.Z`:

- `X.Y.Z` — the version without the `v`, so Helm-style `image.tag` values read
  as plain semver.
- `latest` — moved only when the built tag is the repository's current latest
  release, so a backfill of an older version never demotes `latest`.

Release publishes are immutable: a version tag that already exists on GHCR
fails the run. Scheduled and dispatched rebuilds deliberately repoint the same
version tag, because flushing base-image patches into `X.Y.Z` is what that lane
is for. To get an immutable reference, pin the digest. Every run prints it in
its step summary, both as a line and as a copy-paste `repository@sha256:...`
block.

## Conventions this follows

The workflow mirrors `publish-sandbox-image.yml`:

- A `resolve` job that validates the publication source, refuses a revision
  that is not an ancestor of the default branch, and requires dispatch and
  schedule runs to use the default-branch workflow.
- Untrusted `github.event.*` values reach shell through `env:`, never through
  inline interpolation.
- Multi-arch amd64 plus arm64, built on native runners, pushed as per-arch tags
  and joined with `docker buildx imagetools create`.
- SHA-pinned actions, top-level `permissions: contents: read`, `packages: write`
  only on the jobs that push, and `persist-credentials: false` on checkout.
- GHCR login with the workflow's own `GITHUB_TOKEN`. No repository secret is
  consumed.

## What I validated

- `actionlint` (1.7.12) passes on the new file.
- `node --test scripts/workflow-security.test.mjs` passes, 34 passed and 2
  skipped, where the skips are the Docker context probes. The suite picks up
  every file in `.github/workflows` automatically, so the new workflow is
  covered by its SHA-pinning, digest-pinning, and no-secrets rules.
- `docker build --check --file deploy/self-host/Dockerfile .` from the repo
  root reports no warnings, confirming the build context matches the
  `context: ../..` that docker-compose.yml uses.
- The full image build is deliberately not run here; it compiles the whole Rust
  workspace.

## One thing to confirm after merge

GHCR creates a new package as private. Making
`ghcr.io/brightwave-inc/tidebreak-server` publicly pullable and linking it to
this repository is a one-time manual toggle in the package settings, the same
as for the sandbox images.

### DIFF
--- a/.github/workflows/publish-server-image.yml
+++ b/.github/workflows/publish-server-image.yml
@@ -1,0 +1,335 @@
+name: Publish server image
+
+run-name: >-
+  ${{ github.event_name == 'release'
+    && format('Publish server image {0}', github.event.release.tag_name)
+    || github.event_name == 'workflow_dispatch' && inputs.release_tag != ''
+    && format('Publish server image {0}', inputs.release_tag)
+    || 'Publish server image (latest release)' }}
+
+# Builds the self-host server from deploy/self-host/Dockerfile and publishes it
+# to GHCR:
+#
+#   ghcr.io/brightwave-inc/tidebreak-server
+#
+# The image is the same artifact deploy/self-host/docker-compose.yml builds
+# locally as `tidebreak-self-host:local`: the `tidebreak` binary with
+# tidebreak-server's `postgres` feature, running `Profile::SelfHost`. It exists
+# as a published image so operators do not have to compile the workspace to run
+# a server, and so model gateway deployments can install Tidebreak as an add-on
+# against a digest they pin.
+#
+# Publishes happen on three triggers:
+#
+#   - a published, non-prerelease GitHub Release, which is the release cadence
+#     and the primary path. `release` events run this file from the default
+#     branch, not from the tag, so a hostile or stale tag YAML cannot publish;
+#   - a weekly scheduled rebuild, so Debian security patches in the runtime
+#     base flush through even when nothing in-repo changes;
+#   - manual dispatch, for retries and backfills. Dispatch takes an optional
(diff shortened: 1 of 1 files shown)


naingthet (commented):
Thet's review agent

:white_check_mark: g2g

Mirrors the sandbox publish path: default-branch workflow, ancestor check, event values only via `env:`, SHA-pinned checkout, `persist-credentials: false`, `packages: write` only on push jobs. Release tags are immutable; schedule/dispatch rebuilds may retag. Pin the digest. Do not merge from here.

PR #2297, opened 2026-08-20
feat(providers): fall back to environment variables for provider base URLs

naingthet:
## What this changes

Provider credentials already fall back to environment variables, because a
container has no OS keychain. The provider endpoint had no such fallback: it
lived only in the deployment plane, so a hosted machine could not be fully
configured from its chart or compose file. An admin had to attach after first
boot and set the base URL by hand.

This adds the matching endpoint fallbacks, at the same seam and with the same
precedence rule.

## Environment variables

| Variable | Provider |
| --- | --- |
| `ANTHROPIC_BASE_URL` | `anthropic` |
| `OPENAI_BASE_URL` | `openai` |
| `OPENAI_COMPATIBLE_BASE_URL` | `openai_compatible` |
| `OLLAMA_BASE_URL` | `ollama` |

The names are bare and vendor-standard, matching the API-key fallbacks
(`ANTHROPIC_API_KEY`, `OPENAI_API_KEY`) rather than the `TIDEBREAK_` namespace
the server keeps for its own configuration. `ANTHROPIC_BASE_URL` and
`OPENAI_BASE_URL` are also what the upstream SDKs read, so a deployment that
already sets them for other tooling gets the same endpoint here.

Only providers that accept a stored base URL get a variable. Gemini and xAI use
fixed first-party endpoints, the hosted presets (Fireworks, Together, OpenRouter)
pin their own, and the model gateway takes its URL from policy. All six ignore
the environment exactly as they ignore a stored value.

## Precedence and failure

- A stored base URL wins. The variable is read only when nothing is stored,
  which is the API-key rule.
- An unusable value is ignored, never an error. Empty, unparseable, embedded
  credentials, a non-HTTPS scheme, or cleartext to a non-loopback host all leave
  the provider on its built-in endpoint instead of failing the boot or dropping
  the route. This matches the API-key posture, where an empty variable reads as
  unset.
- Validation reuses `base_url_is_allowed`, the same helper the stored write path
  uses. A provider that needs no credential — Ollama — may be pointed at
  loopback HTTP, as it may from a stored value.
- A managed profile is unaffected. Route collection skips every BYOK provider
  before reading anything, so these variables are as inert there as the stored
  keys.

Enabling `openai_compatible` now accepts an endpoint from the environment, where
it previously required a stored one.

## Tests

Unit tests cover stored-wins precedence, the fallback answering when nothing is
stored, loopback HTTP for Ollama, each unusable-value shape, and the fixed-endpoint
providers refusing to be redirected. They inject the lookup rather than mutating
the process environment. One end-to-end test uses the existing `ScopedEnv` harness
to confirm the real variable reaches the router route.

`cargo test -p tidebreak-server`, `cargo fmt --check`, `cargo clippy -p
tidebreak-server --all-targets`, and `cargo machete` are clean.

### DIFF
--- a/crates/tidebreak-server/src/providers.rs
+++ b/crates/tidebreak-server/src/providers.rs
@@ -397,16 +397,66 @@ impl ProviderKind {
         !matches!(self, Self::Ollama)
     }
 
-    /// The URL this kind actually calls: a stored override, else the default,
-    /// with hosted presets ignoring any stored value.
+    /// Environment variable consulted for this kind's base URL when the
+    /// deployment plane holds none.
+    ///
+    /// Only kinds that accept a stored base URL have one. Hosted presets and
+    /// the first-party Gemini and xAI endpoints are fixed, so no variable can
+    /// redirect them.
+    pub const fn base_url_env_var(self) -> Option<&'static str> {
+        match self {
+            Self::Anthropic => Some("ANTHROPIC_BASE_URL"),
+            Self::Openai => Some("OPENAI_BASE_URL"),
+            Self::Ollama => Some("OLLAMA_BASE_URL"),
+            Self::OpenaiCompatible => Some("OPENAI_COMPATIBLE_BASE_URL"),
+            Self::Xai
+            | Self::Gemini
+            | Self::Fireworks
+            | Self::Together
+            | Self::Openrouter
+            | Self::ModelGateway => None,
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


naingthet (commented):
Thet's review agent

:white_check_mark: g2g

Stored URL still wins. Env fallback only for kinds that already accept a stored endpoint; unusable values are ignored via `base_url_is_allowed`. Fixed-endpoint providers stay fixed. Do not merge from here.

PR #2298, opened 2026-08-20
feat(desktop): make the watch fork a navigable child task

naingthet:
## Summary

The durable watch (decision 50) ran invisibly: the workflow control showed its state, but the fork's transcript was unreachable. This makes the fork a real child task, per the pane-centric design notes:

- the rail shows a nested **Fix PR #N** row under the workspace with the watch's attention dot; clicking opens the fork
- `?task=<session>` attaches a child task's transcript instead of the conversation; layout writes preserve it
- the watch task's seat carries the watch controls — state, **Stop watching**, **Back to main task** — instead of a composer, since the sweep drives its turns
- the header segment reads **Watching in "Fix PR #N"** and links into the fork; Stop moves to the details popover
- the updates store tracks watch digests per workspace beside the conversation's, dropping them when the watch session ends
- Storybook: the workflow control's Watching story shows the link form

## Validation

- desktop UI typecheck clean; code + panel suites pass (46 files, 374 tests), including a new dom test that attaches the fork via `?task`, asserts the composer yields to the watch bar, and navigates back
- Storybook build passes

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeSidebar.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeSidebar.tsx
@@ -3,6 +3,7 @@ import { useNavigate, useRouterState } from "@tanstack/react-router";
 import {
   ChevronRight,
   CircleAlert,
+  CircleDotDashed,
   FolderGit2,
   GitPullRequest,
   Plus,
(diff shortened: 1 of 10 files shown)


naingthet (commented):
Thet's review agent

:white_check_mark: g2g

Watch fork is a real child task: rail row, `?task=` attach, composer yields to the watch bar, layout writes keep the task. Store keeps watch digests beside the conversation and drops them on end. Do not merge from here.

PR #2300, opened 2026-08-20
docs(decisions): on-behalf-of inference for hosted machines (51)

naingthet:
Decision 49 gave hosted machines live gateway identity, but inference still rides one deployment-scoped token: coarse attribution, a standing secret, manual rotation. This record makes per-user on-behalf-of exchange the default — the server exchanges each caller's machine-bound token at the gateway for a short-lived inference token and drives that caller's turns with it. Stored provider configuration always wins, static-token servers are unchanged, and refusals fail closed with the sign-in-required shape clients already handle.

Pairs with the gateway-side record (add-on principals and capability exchange). Design only; implementation follows review.


naingthet:
Gateway-side pair: brightwave-inc/model-gateway#1112.

PR #2301, opened 2026-08-20
feat(desktop): add optional starting prompt to new workspace

naingthet:
## Summary

- Reorder the new-workspace dialog to repo, harness, model, then an optional starting prompt, then title, base ref, and permission mode.
- If you fill in a starting prompt, create still opens the workspace, and that text is inserted into the composer so you can send it as the first turn.

## Test plan

- [x] `pnpm --dir crates/tidebreak-desktop/ui exec vitest run src/code/NewWorkspaceDialog.dom.test.tsx`
- [ ] Open **New workspace**, confirm field order, leave the starting prompt blank, and create. The composer stays empty.
- [ ] Fill in a starting prompt and create. The workspace composer shows that text, trimmed, ready to send.


naingthet (commented):
Thet's review agent

:white_check_mark: g2g

PR #2303, opened 2026-08-20
feat(code): workspace rail overhaul with hover detail panel and restore

naingthet:
## What changed

The code-mode workspace rail is rebuilt around three layers with one job each:

- **Rows are plain.** One button that opens the workspace: attention dot, PR mark, title, view-dependent meta. No inline action buttons or hover overlays.
- **Detail and actions live in a hover panel** (Radix HoverCard, new `components/ui/hover-card.tsx`) that opens beside the row: full branch, state line, checks summary, and the pull request with its open action. `detailDefaultOpen` renders it for stories.
- **Right-click is the row's one menu** (Radix ContextMenu) and the keyboard path.

Around the card:

- The Repos section is gone: by-repo group headers link to their repo page, and a searchable repo-switcher popover in the toolbar (with "Add repo…") covers every sort mode.
- A rail settings popover controls sort mode, card density (compact/detailed), repo/branch visibility, and an archived shelf. Prefs persist as one `tidebreak.code-rail-prefs` blob, migrating the legacy sort key.
- Watch sessions (ADR 0050) render as live child rows under their card, from digests the updates store previously dropped; clicking opens the watch transcript read-only at `?session=<id>`. The interactive digest never leaves its slot.
- Session digests now carry `watch_state`/`watch_detail`/`watch_cycles` so watch rows say "Fixing ×3" without any per-card `gh` round trip.
- Archive gains an explicit "Force archive (discard changes)" command (the error-driven escalation is unchanged, `ARCHIVE_FORCE_KINDS` untouched).
- New `POST /code/workspaces/{id}/restore` reactivates an archived workspace Conductor-style: worktree re-created at the same path from the kept branch, setup script mirroring create's semantics, journals and session rows intact. 409 kinds `branch_missing` (UI offers a new-workspace fallback) and `worktree_path_occupied`. Restore is reachable from archived cards and the repo page.
- ADR 0052 records the harness-subagent contract (renumbered from 0051 after a collision on main; parent_call_id, subagents as derived spans, never sessions) ahead of implementation.

## Why

Design review of the rail: hover actions clashed with row text, repo info repeated per view, no card density control, watch tasks were invisible on the rail, and archived work had no way back.

## Testing

- `pnpm test` — 1402 tests green (card, sidebar, stores, parsers, workspace page rewritten/extended).
- `pnpm storybook:build` — stories for row states, detail panels, archived shelf.
[...]

PR #2305, opened 2026-08-20
feat(code): show harness subagents on the rail and share PR workflow actions

naingthet:
## Summary

Second pass on the code-mode rail, following #2303. Three things the design
review asked for:

- **The row no longer glows.** The focus ring was drawn outside the row box, so
  a focused card read as a lit-up inner border. New `FOCUS_RING_INSET` in
  `code/interactive.ts` keeps the ring inside the box.
- **PR actions in the hover panel.** The panel's footer now offers the same
  one-click action the workspace header's control offers — Merge on an approved
  green PR, Resolve conflicts on a conflicting one, Fix CI on failing checks.
  It runs off the shared `workspaceWorkflowModel` / `workspaceWorkflowActionLabel`
  / `prWorkflowPrompt` machinery rather than a second copy, driven from the
  digest alone so the rail never shells out to `gh`. While a watch task is
  driving the worktree, agent actions stay suppressed, same as the header.
- **Subagents and their status.** The Claude-first slice of decision 52: a
  subagent is the span of its parent `Task` call. The card renders one row per
  subagent under the watch child rows, with Running / Done / Failed.

## Subagent slice, by layer

- `CodeEvent` and `HarnessEvent` tool/assistant variants gain an optional
  `parent_call_id`. Old journals replay byte-identically — the pinned journal
  fixture is unchanged.
- The Claude parser threads the stream's top-level `parent_tool_use_id` into
  produced events and names `Task` calls from their description and subagent
  type. A synthesized `subagent-task` fixture pins the mapping.
- Sessions carry a bounded subagent list (cap 8, `Done` dropped first) in a new
  nullable JSON column. `set_session_subagents` is a targeted write and
  `save_session` deliberately leaves the column alone, so the worker's full-row
  saves from a stale in-memory copy cannot clobber the sink. A db test covers
  that race.
- `LiveSink` flips Running/Done/Failed on `Task` boundaries and restates the
  digest, which carries `subagents` through the bus, the wire types, and the
  updates socket.

## Schema epoch

Bumped to **35**. Main is at 34 for the watch baseline (#2291); this adds
another baseline column, so a pre-v1 database already rebuilt at 34 would
otherwise keep the old schema and fail with `no such column:
code_session.subagents` — the same symptom #2291 fixed.

## Follow-ups

Deliberately out of scope, per decision 52: the filtered sub-transcript view
(events whose `parent_call_id` matches — clicking a subagent row opens the
workspace for now) and the Grok `spawn_subagent` mapping.

## Test plan

- `cargo test -p tidebreak-core --lib` — 681 pass
- `cargo test -p tidebreak-harness` — 123 pass
- `cargo test -p tidebreak-server --lib` — 1049 pass, 1 pre-existing failure
  (`tests::lifecycle::client_resolution_publishes_cancellation_and_wakes_resumable_turns`,
  verified failing identically on clean `main` at 719fa9729)
- `pnpm test` — 201 files, 1407 pass
- `tsc --noEmit`, `pnpm storybook:build`, `cargo fmt --check` clean
- New stories: `WithSubagents`, `DetailPanelReadyToMerge`, `DetailPanelConflicts`


naingthet (commented):
Thet's review agent

:white_check_mark: g2g

Epoch 35 for the new `code_session.subagents` baseline column is the right follow-on to #2291. `save_session` leaving that column alone and `set_session_subagents` as the targeted write is the correct race story against the worker's full-row saves. `parent_call_id` is optional and skipped when absent, so old journals stay byte-identical. Claude `Task` as the subagent span, Done-first cap of 8, and sharing the header's workflow action from the digest (no `gh` from the rail) all match decision 52. Grok `spawn_subagent` and the filtered sub-transcript view staying out of this PR is the right cut.

**Verdict: g2g**

PR #2306, opened 2026-08-20
fix(release): dispatch the server image build when a release publishes

naingthet:
## What broke

`publish-server-image.yml` declares `on: release: [published]` and its header calls that "the release cadence and the primary path." That trigger cannot fire for a release `release.yml` publishes. The `PATCH` that flips the draft runs on `GITHUB_TOKEN`, and GitHub does not start a workflow run from an event that token created.

v0.57.0 published this morning and no image build ran. Nothing failed, so nothing reported it. The release had no `ghcr.io/brightwave-inc/tidebreak-server` image until a manual dispatch built one, and every future release would have been silent in the same way.

## The fix

`finalize_release` dispatches `publish-server-image.yml` with the tag it just published. `workflow_dispatch` is one of the two events `GITHUB_TOKEN` may raise, and the same job already dispatches `release-draft.yml` this way, so the mechanism is proven in this repository and the job already holds `actions: write`.

The step carries the same `draft == 'true'` guard as its neighbor. Re-running the release workflow against an already-published release must not rebuild an image, because a version tag that already exists on GHCR fails the publish.

The `release` trigger stays declared. A release someone publishes by hand in the UI carries a user token and does fire it. Both routes land as the same build, and the per-tag concurrency group serializes them.

## Test

`workflow-security.test.mjs` gains a case asserting the dispatch, its tag argument, its draft guard, and that the declared trigger survives. Removing the dispatch line fails it. The header comment in `publish-server-image.yml` now describes the trigger that actually works.

37 tests pass.

### DIFF
--- a/.github/workflows/publish-server-image.yml
+++ b/.github/workflows/publish-server-image.yml
@@ -21,18 +21,25 @@ run-name: >-
 #
 # Publishes happen on three triggers:
 #
-#   - a published, non-prerelease GitHub Release, which is the release cadence
-#     and the primary path. `release` events run this file from the default
-#     branch, not from the tag, so a hostile or stale tag YAML cannot publish;
+#   - a release, which is the release cadence and the primary path. release.yml
+#     dispatches this workflow with the tag it just published, because a release
+#     that job publishes raises no `release` event: the PATCH runs on
+#     GITHUB_TOKEN, and GitHub does not start a workflow run from an event that
+#     token created. The `release` trigger below stays declared for a release
+#     someone publishes by hand in the UI, which carries a user token and does
+#     fire. Both land here as the same build, and the per-tag concurrency group
+#     serializes them if a release ever arrives by both routes;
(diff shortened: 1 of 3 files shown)


naingthet (commented):
Thet's review agent

:white_check_mark: g2g

The diagnosis is right: a release `release.yml` publishes with `GITHUB_TOKEN` does not fire `on: release`, so `publish-server-image.yml` never ran for v0.57.0. Dispatching it from `finalize_release` is the same mechanism the job already uses for `release-draft.yml`, and `workflow_dispatch` is one of the two events that token may raise. The `draft == 'true'` guard matches the neighbor and avoids rebuilding a GHCR tag that already exists. Leaving the `release` trigger for a hand-published UI release, pinning `--ref main`, and asserting the dispatch in `workflow-security.test.mjs` are the right holds. This is an additive step in `release.yml`, not a required-workflow deletion, so the trusted-policy two-step trap does not apply.

**Verdict: g2g**

PR #2307, opened 2026-08-20
fix(code): stabilize Windows code-mode lifecycle tests

naingthet:
## Summary

- Windows `cargo check` on main failed in the code-mode lifecycle suite (deterministic restore CRLF mismatch; flaky mid-turn steer races under load).
- Park supported/stalled steer tests on a pending approval so the active-turn window is event-driven instead of a short wall-clock delay.
- Pin test git repos to `core.autocrlf=false` and compare restored file contents with normalized newlines.

## Test plan

- [x] `cargo test -p tidebreak-server --lib supported_steer_reaches`
- [x] `cargo test -p tidebreak-server --lib stalled_native_steering`
- [x] `cargo test -p tidebreak-server --lib restore_reactivates`
- [x] `cargo test -p tidebreak-server --lib steer` (22 passed)
- [ ] CI Windows cargo check green on this PR

### DIFF
--- a/crates/tidebreak-server/src/tests/code.rs
+++ b/crates/tidebreak-server/src/tests/code.rs
@@ -100,6 +100,9 @@ fn init_git_repo_named(dir: &std::path::Path, name: &str) -> std::path::PathBuf 
         ["git", "init", "-b", "main"].as_slice(),
         ["git", "config", "user.email", "dev@example.com"].as_slice(),
         ["git", "config", "user.name", "Dev"].as_slice(),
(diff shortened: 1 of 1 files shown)


naingthet (commented):
Thet's review agent

:white_check_mark: g2g

Parking mid-turn steer on a pending approval (ask + `approval_script`) is the right way to hold the active-turn window without a wall-clock race. LF pin plus newline-normalized restore assertion covers the CRLF mismatch.

**Verdict: g2g**

PR #2308, opened 2026-08-20
fix(self-host): make the server image report its release

naingthet:
## What broke

The server image publish smoke-tests the built binary with `tidebreak --version`. The CLI has no such flag. The binary printed its usage text and exited 2, so the step could never have passed on any build.

It went unnoticed because the first publish attempt died at compile with the dockerignore bug (#2299), and today's dispatch for v0.57.0 was the first build ever to reach the smoke test. It failed there, with both architectures compiled and one push away from done.

## The fix

Add `--version`, and give it something true to say.

The workspace manifest is pinned at `0.0.0` because release numbering lives on the Git tag and the desktop bundle, not in Cargo. Reporting `CARGO_PKG_VERSION` would make the smoke test pass while printing `tidebreak 0.0.0` for every release, which is worse than not answering. So the tag reaches the binary through a `TIDEBREAK_VERSION` build argument that the publish workflow passes and `option_env!` reads at compile time. A build that passes nothing reports `0.0.0-unreleased`.

The smoke test asserts the reported string rather than a zero exit:

```
reported="$(docker run --rm --entrypoint /usr/local/bin/tidebreak "$SERVER_REPOSITORY:$VERSION-$arch" --version)"
[[ "$reported" = "tidebreak $VERSION" ]] || exit 1
```

That is the difference between checking the binary runs and checking it is the release it claims to be. A deployment pins this image by digest; being able to ask a running pod what it is has value beyond the smoke test.

## Test

`workflow-security.test.mjs` gains a case covering the build argument, the Dockerfile's `ARG` default, and the string assertion.

Verified against `cargo` builds:

- unstamped build reports `tidebreak 0.0.0-unreleased`
- `TIDEBREAK_VERSION=0.57.0` build reports `tidebreak 0.57.0`
- `--version extra` exits 2 with the usage text

Verified end to end against the staged Docker context, which is the path that failed today:

```
docker build --build-arg TIDEBREAK_VERSION=0.57.0 ...   # exit 0
docker run --rm --entrypoint /usr/local/bin/tidebreak tidebreak-self-host:verify --version
tidebreak 0.57.0
```

Clippy is clean and 37 workflow tests pass.

## Related

Stacks with #2306, which fixes the trigger that should have started this build in the first place. Both are needed before a release publishes an image without a human dispatching it.

### DIFF
--- a/.github/workflows/publish-server-image.yml
+++ b/.github/workflows/publish-server-image.yml
@@ -259,13 +259,22 @@ jobs:
         run: |
           docker build \
             --file deploy/self-host/Dockerfile \
+            --build-arg "TIDEBREAK_VERSION=$VERSION" \
             --tag "$SERVER_REPOSITORY:$VERSION-${{ matrix.arch }}" \
             .
 
       - name: Smoke test the built binary
         run: |
-          docker run --rm --entrypoint /usr/local/bin/tidebreak \
-            "$SERVER_REPOSITORY:$VERSION-${{ matrix.arch }}" --version
+          # Proves the binary executes on this architecture, and that it
+          # carries the release it claims to. A published image whose version
+          # is a guess is worse than one that cannot answer.
+          reported="$(docker run --rm --entrypoint /usr/local/bin/tidebreak \
+            "$SERVER_REPOSITORY:$VERSION-${{ matrix.arch }}" --version)"
+          echo "$reported"
+          [[ "$reported" = "tidebreak $VERSION" ]] || {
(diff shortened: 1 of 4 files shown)


naingthet (commented):
Thet's review agent

:white_check_mark: g2g

`--version` plus `TIDEBREAK_VERSION` is the right honesty: Cargo stays `0.0.0`, the tag stamps the image, and the smoke test checks the string not just exit 0.

**Verdict: g2g**

PR #2282, opened 2026-08-19
refactor(desktop): unify work/chat transcript cards

naingthet:
## Summary

- standardize ordinary app and output cards on one responsive transcript footprint
- share the boxless expandable tool-row component between Work/Chat and Code mode
- move Work/Chat execution backend and outcome pills into the expanded detail
- reuse the bare command/output presentation while preserving running and failed expansion behavior

## Compatibility and risk

- no wire-format, persistence, routing, or execution behavior changes
- chart outputs and embedded MCP app surfaces retain their content-sized frames
- accessibility labels and Code-mode status, duration, exit-code, and streaming behavior remain covered

## Verification

- `pnpm exec vitest run src/ToolCallCard.test.tsx src/ToolCallCard.dom.test.tsx src/code/CodeTranscript.dom.test.tsx src/AgentActivityTimeline.dom.test.tsx src/MessageList.dom.test.tsx` — 78 tests passed
- `pnpm exec tsc --noEmit`
- `git diff --check`
- visually verified the updated weekly-planner Work/Chat transcript and an existing populated Code workspace in the local app

PR #2309, opened 2026-08-20
fix(desktop): keep the code tool line on one row

naingthet:
## What changed

The card unification (#2282) wrapped every `ToolCardShell` title in an inline span. Code mode passes its title as two elements — the bold verb and the `MiddleTruncate` command — with a flex `titleClassName`, but that class landed on the outer span while the elements sat inside the plain inline wrapper. The block-level `MiddleTruncate` broke onto its own line, its nowrap text stopped the row from shrinking, middle truncation died, and the duration/exit/status meta floated mid-row.

- `ToolCardShell` now renders a non-string title directly inside the flex container, so the host's title layout reaches its children. String titles (work/chat cards, agent timeline) keep the old wrapper.
- `humanizeShellCommand` also strips the leading `cd "<worktree>" && ` for display. Headless engines lead every command with the session's own directory, which buried the real command in the truncated middle. Display only; a bare `cd <path>` is left alone.

## Tests

- Regression test in `CodeTranscript.dom.test.tsx`: the verb and command must be direct children of the flex title row, and the cd prefix must not render.
- New `CodeToolCard.stories.tsx` covering the tool line's states, including the long worktree-prefixed command that would have caught this.
- `pnpm test` on the five suites touching the shell and its consumers (79 pass), `tsc --noEmit` clean, `pnpm storybook:build` clean.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/ToolCardShell.tsx
+++ b/crates/tidebreak-desktop/ui/src/ToolCardShell.tsx
@@ -87,12 +87,16 @@ export function ToolCardShell({
           {icon}
         </span>
         <span className={cn("min-w-0 flex-1", titleClassName)}>
-          <span
-            className={cn(typeof title === "string" && "block truncate")}
-            title={typeof title === "string" ? title : undefined}
-          >
-            {title}
-          </span>
+          {typeof title === "string" ? (
+            <span className="block truncate" title={title}>
+              {title}
+            </span>
+          ) : (
+            // Rendered directly so a `titleClassName` flex layout reaches the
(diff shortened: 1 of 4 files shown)


naingthet (commented):
Thet's review agent

:white_check_mark: g2g

Non-string titles rendering as direct flex children is the right #2282 follow-on — the inline wrapper was what forced `MiddleTruncate` onto its own line. Stripping a leading `cd <worktree> &&` for display (and leaving a bare `cd`) is display-only. The regression test pinning verb and command as siblings of the flex row is the one that would have caught this.

**Verdict: g2g**

Issue #1455, opened 2026-08-04
agent_run_claim grows without bound from empty poll scans


naingthet:
`record_empty_claim_scan_on` (`crates/openwave-core/src/db/ops/agent_run.rs:1446-1461`) inserts a claim row with `agent_run_id = NULL` and a fresh UUID token on **every empty poll**. The background worker idles at a 5s cap, so an idle app accretes ~17k rows/day indefinitely.

The only deletion path (`crates/openwave-core/src/db/ops/conversation.rs:701`) filters `AgentRunId IN (subquery)`, which never matches NULL, so nothing ever prunes these rows.

Fix: either stop persisting empty scans (if the receipt exists only for observability, a counter or log line serves), or add a retention sweep for NULL-run claim rows. Independent of any schema-consolidation decision; worth fixing now since it affects every 0.x database in the wild once background agents are wired up — and the workers already poll today.

PR #2329, opened 2026-08-20
fix(server): stop idle agent polling from starving the SQLite pool

### LINKED ISSUE #2316: Workspace creation stalls behind a starved SQLite connection pool
## Problem

Creating a workspace can take minutes even though the actual work is cheap. `git worktree add` on a real repo takes ~0.3s, but `create_workspace` (`crates/tidebreak-server/src/code/runtime.rs:531-617`) makes four sequential store calls and `create_session_of_kind` adds two more, and each call can wait ~30s (worst observed: 130s) just to acquire a database connection. The New Workspace dialog blocks on the whole chain, so the user sees a multi-minute "creating" state.

## Evidence

A dev-channel `tidebreak.log` covering 2026-08-13 to 2026-08-20 contains 5,628 `sqlx::pool::acquire` slow-acquire warnings (waits cluster at ~30s, worst 130s) and 54 slow-statement warnings. The slow statements concentrate on `agent_run_claim` and `agent_run`, including deletes with `rows_affected=0` taking 1-25s. The database file itself is ~1MB, so this is contention, not data volume.

## Cause

Idle agent-run workers poll every few seconds, and each empty scan writes: a prune delete plus a NULL-run claim insert (`crates/tidebreak-core/src/db/ops/agent_run.rs:1281-1290`, #1455). SQLite has a single writer, so under fleet load these small writes serialize behind real turn writes. Each stalled writer holds a pooled connection while it waits, the pool (SeaORM defaults; `DbStore::connect`, `crates/tidebreak-core/src/db/mod.rs:106-129`) drains, and every request in the process queues — workspace creation is just the most visible victim.

## Outcome

Workspace creation and other interactive requests stay responsive while an agent fleet is running. Idle polling does not consume the write path.

## Levers

- Make the idle worker scan read-only; write a claim receipt only when a run is actually claimed, or batch/round-robin the receipts.
- Set `PRAGMA synchronous=NORMAL` alongside WAL in `DbStore::connect_with_options`. The default (`FULL`) fsyncs every commit, which is expensive on macOS and amplifies write serialization.
- Set an explicit pool size and busy timeout instead of inheriting SeaORM defaults; consider a dedicated writer connection so readers never queue behind the write lock.

naingthet:
## What changed

Creating a workspace could take minutes even though `git worktree add` takes ~0.3s. Every store call in the chain waited ~30s for a database connection. Three things compounded, and this fixes all three.

**Idle polling was a write.** An empty agent-run claim scan performed a prune delete plus a NULL-run receipt insert. Workers poll every few seconds, SQLite has a single writer, and each stalled poll held a pooled connection while it waited.

- A scan that finds no background run in a state it could act on now returns from one read, before it takes the durable claim lock — that lock is an `UPDATE`, so taking it is what made an idle poll a write transaction. The locked scan stays the authority for every decision; a run that appears right after the check is claimed by the next poll, and the worker also wakes on notify.
- When runs exist but nothing is claimable (a saturated fleet), the no-work receipt is written once per five-minute window instead of once per poll, and the retention sweep rides the same write. A receipt that ages past retention still forces one write, so a database that accreted receipts before #1455 converges instead of carrying them forever.
- The table stays bounded and the evidence that scans ran survives, so #1455 stays fixed. What it gives up is exact-retry replay for a token whose scan skipped the write: a retry that finds work claims it under the same token, which is a valid outcome for a request whose response was lost, and every worker draws a fresh token per poll today.

**The pool was one connection.** SeaORM gives a SQLite pool a single connection unless it is told otherwise, so one slow write blocked every other request in the process — including the four sequential store calls that create a workspace. Host processes now open 8 connections with an explicit 30s acquire timeout (the value the starved waits were already hitting, now chosen rather than inherited).

**Every commit fsynced.** SQLite connections now open with `synchronous=NORMAL`, the standard WAL pairing, and a 15s busy timeout. Both are per-connection settings, so they ride the connect options; a `PRAGMA` run once at connect would only reach one pooled connection. `journal_mode=WAL` stays a one-shot `PRAGMA` because it is a persistent file-level setting.

## Tests

- `idle_polling_does_not_touch_the_write_path` — repeated polls with nothing claimable leave the claim table byte-identical, including the sweep.
- `saturated_polling_does_not_accrete_empty_claim_scan_rows` — replaces the #1455 test with the fleet shape it now has to hold: three empty scans write one receipt between them, the stale receipt is swept, and a receipt that later ages past retention is swept too.

Local: `cargo test -p tidebreak-core --lib` (682 passed), `cargo test -p tidebreak-server --lib tests::` (1046 passed, one pre-existing PTY flake that passes alone), `cargo fmt --all --check`, and `cargo clippy -p tidebreak-core -p tidebreak-server --all-targets -- -D warnings` all clean.

## Notes for review

- `sea-orm` moves from a dev-dependency to a dependency of `tidebreak-server` so the pool sizing lives at the host call sites. Same workspace pin; the lockfile is unchanged.
- Raising the desktop pool above one connection means writers can now contend inside SQLite rather than inside the pool. The contended paths take the advisory lock as their first statement, so they write before they read and the busy handler covers them; a read-then-write transaction could still see `SQLITE_BUSY_SNAPSHOT`. The issue's remaining lever — a dedicated writer connection so readers never queue behind the write lock — would settle that and is left for a follow-up.

Fixes #2316

### DIFF
--- a/crates/tidebreak-core/src/db/mod.rs
+++ b/crates/tidebreak-core/src/db/mod.rs
@@ -99,6 +99,43 @@ struct DocumentSummaryRow {
     updated_at: chrono::DateTime<Utc>,
 }
 
+/// How long a SQLite writer waits for the write lock before it gives up.
+///
+/// sqlx defaults to 5s, which a real turn write can exceed while a fleet runs;
+/// waiting is better than surfacing "database is locked" to the caller.
+#[cfg(feature = "sqlite")]
+const SQLITE_BUSY_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(15);
+
+/// Apply the write policy that a `PRAGMA` at connect time cannot.
+///
+/// Constraint: `synchronous` and `busy_timeout` are per-connection settings,
+/// so a pool larger than one connection only honours them if they ride the
+/// connect options. `journal_mode` is the exception — it is a persistent
+/// file-level setting, which is why WAL stays a one-shot `PRAGMA` below.
+///
+/// `synchronous=NORMAL` is the standard WAL pairing: WAL already keeps a
+/// commit durable across a process crash, and the default `FULL` buys
+/// durability across a power loss by fsyncing every commit — expensive on
+/// macOS, and it lengthens every write while SQLite's single writer is the
+/// contended resource (#2316).
+///
+/// SeaORM only applies this hook to the SQLite driver, so it is a no-op for a
+/// Postgres self-host.
+#[cfg(feature = "sqlite")]
+fn with_sqlite_write_policy(mut options: ConnectOptions) -> ConnectOptions {
+    options.map_sqlx_sqlite_opts(|sqlite| {
+        sqlite
+            .synchronous(sea_orm::sqlx::sqlite::SqliteSynchronous::Normal)
+            .busy_timeout(SQLITE_BUSY_TIMEOUT)
+    });
+    options
+}
+
+#[cfg(not(feature = "sqlite"))]
+fn with_sqlite_write_policy(options: ConnectOptions) -> ConnectOptions {
+    options
+}
+
 impl DbStore {
     /// Connect to `url` and run migrations. For a SQLite file that should be
     /// created if missing, include `?mode=rwc` (e.g.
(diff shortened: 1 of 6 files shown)


naingthet (commented):
Thet's review agent

:white_check_mark: g2g

**Verdict: g2g**

PR #2330, opened 2026-08-20
perf(code): warm harness installs and stop re-probing on every create

### LINKED ISSUE #2318: Session create blocks on cold harness install and re-probes on every create
## Problem

The New Workspace dialog blocks on `createCodeSession`, and session create pays two avoidable costs on the request path (`create_session_of_kind`, `crates/tidebreak-server/src/code/runtime.rs:1001-1064`):

- **Cold harness install.** `ensure_pinned_harness` runs `npm install <package>@<pin>` over the network the first time each pinned version is used (`crates/tidebreak-harness/src/pin.rs:119-179`). Installed harnesses are 37-297MB, so a pin bump turns the next session create into a minutes-long silent stall. The dialog has no progress surface for it — just an eventual error on failure.
- **A fresh probe every time.** `runtime.rs:1041` drops the cached probe before every create, so each session create re-runs the harness binary through a shell (~1s of Node CLI startup) even when nothing changed.

## Outcome

Session create is fast when the pinned harness is already installed, and a cold install is visible instead of a silent stall.

## Shape

- Install pins ahead of need: kick off `ensure_installed` for the selected harness when the dialog opens (or at boot after a pin bump), and report progress on the live bus the way clone progress already works.
- Re-probe only when the resolved binary path or pin changed since the last probe, instead of unconditionally invalidating.

naingthet:
Fixes #2318

## What changed

**Warm installs, off the request path.** `POST /code/harnesses/{kind}/install` starts `ensure_installed` for one engine in a detached task and reports progress on the owner's `WS /code/updates`, mirroring how `git clone` progress works. Phases are `installing`, then `ready` or `failed` — npm reports no usable percentage to a pipe, so there is no bar to fake. The route answers immediately in all three cases (already installed, already running, now started) and never starts two installs for one pin. It sits on the deployment plane beside `/code/harnesses/refresh`, because installing a pin writes a binary to the machine.

On success the task also invalidates the stale probe and takes the cold one itself, so create pays for neither the install nor the ~1s engine startup.

**Dialog wiring.** The New Workspace dialog fires the call when it opens and when the engine changes, but only when the doctor says that engine is not installed — an engine already on disk is never asked for, and a member of a shared deployment never hits an admin route for nothing. A new `HarnessInstallNote` renders the phase under the engine picker instead of the previous silent stall, and a finished install re-reads the doctor (`GET /code/harnesses`, not the doctor's own install-everything refresh) so the engine turns selectable.

**One installer per pin.** `tidebreak_harness::ensure_installed` now takes a per-install-directory async lock and re-checks the marker after acquiring it. npm has no guard for two writers in one `node_modules`, and the marker is written only after a successful install, so a torn tree from an interleaved run would have been blessed by the next marker write. Keyed on the directory, so two data directories do not serialize against each other.

**Probe only when it can have changed.** `create_session_of_kind` used to drop the memoized probe unconditionally. It now drops it only when the resolved binary path differs from the one just ensured. The pin version is a path segment (`tools/harnesses/{kind}/{version}/…`), so one comparison covers both the binary and the pin. A probe that found nothing, or no cached probe, is still stale; a failed or empty probe still surfaces `harness_not_found`. The `#[cfg(not(test))]` block around the ensure is unchanged in shape, so test behavior is intact.

## Tests

- `tidebreak-harness`: one pin directory admits one installer at a time, and a second pin is not held up by the first.
- `tidebreak-server`: the probe-freshness contract — same path stays cached, a pin bump stales it, and no-probe / found-nothing are both stale.
- UI: the `harness_install` notice validator (including refusing an unknown engine) and the one-install-per-engine reducer.
- Storybook: `Code/Harness install note` covers installing, failed, ready, and absent.

## Checks

`cargo fmt --all --check`, `cargo clippy -p tidebreak-server -p tidebreak-harness --all-targets -- -D warnings`, `cargo test -p tidebreak-server code::` (130 passed), `cargo test -p tidebreak-harness` (2 pre-existing failures in `child::unix_process_tree_tests`, identical on a clean tree in this sandbox), `pnpm test` (1412 passed), `tsc --noEmit`, `pnpm storybook:build`.

## Left out

Boot-time warming after a pin bump. The dialog is the surface the issue names, and warming every engine at launch spends four installs on engines a launch may never use. Other create surfaces (the CLI, watch sessions) still take the install on the request path, where the existing ensure covers them.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/api/client.ts
+++ b/crates/tidebreak-desktop/ui/src/api/client.ts
@@ -99,6 +99,7 @@ import {
   type CodeWorkspaceSnapshot,
   type CodeCloneDefaults,
   type CodeCloneJobSnapshot,
+  type CodeHarnessInstallSnapshot,
   type CodeSubscriptionUsage,
   type HarnessDoctorReport,
   type HarnessKind,
@@ -123,6 +124,7 @@ import {
   parseCodeApproval,
   parseCodeCloneDefaults,
   parseCodeCloneJob,
+  parseCodeHarnessInstall,
   parseCodeRepo,
   parseCodeSession,
   parseCodeSessionList,
@@ -1800,6 +1802,27 @@ export class ApiClient {
         ),
       ),
       "harness models",
+    );
+  }
+
+  /**
+   * Warm the pinned install of one engine ahead of a session create.
+   *
+   * Answers as soon as the server knows where the install stands; the phases
+   * that follow arrive on `WS /code/updates`. Safe to call repeatedly — an
+   * installed pin answers `ready` and one already running is not restarted.
+   */
+  async startHarnessInstall(
+    kind: HarnessKind,
+  ): Promise<CodeHarnessInstallSnapshot> {
+    return requireParsed(
+      parseCodeHarnessInstall(
+        await this.json(
+          `/code/harnesses/${encodeURIComponent(kind)}/install`,
+          { method: "POST", headers: this.headers() },
+        ),
+      ),
+      "harness install",
     );
   }
 
--- a/crates/tidebreak-desktop/ui/src/api/types.ts
+++ b/crates/tidebreak-desktop/ui/src/api/types.ts
@@ -137,6 +137,7 @@ import {
   type CodeUpdateNotice as WireCodeUpdateNotice,
   type CodeCloneDefaults as WireCodeCloneDefaults,
   type CodeCloneJobSnapshot as WireCodeCloneJobSnapshot,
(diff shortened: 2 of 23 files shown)


naingthet (commented):
Thet's review agent

:white_check_mark: g2g

Warm install off create is the right cut: detached task, phases on the owner's updates socket, one installer per pin directory (lock + re-check after acquire). Create still has its own ensure, so correctness does not depend on the warm path. Dialog only fires when the doctor says not found, so a shared-deployment member does not hit an admin route for an engine already on disk. Ready installs render nothing.

**Verdict: g2g**

PR #2331, opened 2026-08-20
feat(code): move new worktrees to a configurable, user-visible root

### LINKED ISSUE #2317: Move code worktrees out of platform app data to a user-visible, configurable root
## Problem

Code worktrees land under platform app data: `~/Library/Application Support/io.brightwave.tidebreak[.dev|.staging]/code/worktrees/<repoId>-<slug>/<workspaceId>-<slug>/`. The path is hard-coded in `worktree_dir` (`crates/tidebreak-server/src/code/worktree.rs:699-724`); the desktop app takes Tauri's `app_data_dir()` with no override (`crates/tidebreak-desktop/src/lib.rs:519-527`). `TIDEBREAK_DATA_DIR` only affects CLI and self-host.

Worktrees are user work, not app state:

- They hold uncommitted code on real branches. App data is conventionally disposable — uninstall flows and "reset app data" delete it without the user expecting to lose work.
- `~/Library` is hidden in Finder and file dialogs, and the UUID-prefixed segments make paths long and unreadable in terminals and editors.
- The bundle ID keys the path, so each channel grows its own tree (three on a dev machine today).
- Time Machine backs up whatever agents build inside (one current tree is 2.4GB, mostly a `node_modules`).

The database, logs, blobs, and managed tools are correctly placed; only the worktrees don't belong.

## Outcome

Worktrees live under a user-visible, configurable root with human-readable names, and the app-data root keeps only true app state. Existing workspaces keep working (their absolute `gitdir` links are repaired or grandfathered).

## Shape

- Reuse the `code_clone_parent_dir` pattern (`crates/tidebreak-server/src/code/clone.rs`): a per-owner setting for the worktree root, defaulting to something visible such as `~/Tidebreak/workspaces/<repo>/<workspace>` or a `worktrees/` sibling of the clone parent.
- Keep the ID as a uniqueness suffix, not the leading segment.
- Migration: worktree `.git` files and the repo's `.git/worktrees/*` entries store absolute paths, so either run `git worktree repair` per repo on first boot after the change, or apply the new root to new workspaces only.

naingthet:
Fixes #2317.

## What changed

Code worktrees left platform app data. A worktree holds uncommitted work on a
real branch, so it does not belong in a directory that uninstall and "reset app
data" flows treat as disposable — and `~/Library` is hidden, UUID-prefixed, and
duplicated per channel.

New worktrees now land at `<root>/<repo-slug>/<workspace-slug>-<short-id>/`:
the readable name leads, and eight hex digits of the workspace id trail as the
uniqueness suffix. The repo segment is the slug alone; workspace ids are unique
across repos, so two same-named repos share a folder without ever sharing a
worktree.

`<root>` comes from a new deployment-wide `code_worktree_root` setting, built on
the `code_clone_parent_dir` pattern it sits beside: one shared value, read and
written on the admin plane (`GET`/`PUT /code/worktree-root`), with the same
per-owner path segment a multi-user machine needs. Setting a root that does not
exist creates it; a relative path or a path that names a file is refused when
set, not at the first workspace that fails.

## The default, and why

- **Desktop:** `~/Tidebreak/workspaces`, supplied through a new
  `Config::code_worktree_root_default` boot field. All three channels share the
  one root on purpose — the issue counts three trees on a dev machine as part of
  the problem, and app-data keying exists to protect app *state*, not to fork
  the user's work.
- **Headless (CLI, self-host, tests):** unchanged `<data_dir>/code/worktrees`.
  A headless deployment's data directory is already an operator-chosen, visible
  location, and `TIDEBREAK_DATA_DIR` is already the lever. Keying off
  `Profile::Desktop` would have caught the CLI too, so the embedding names the
  default instead of the profile implying it.
- A stored setting overrides either.

## Migration: none, by design

Existing workspaces keep the absolute `worktree_path` on their row and their
checkouts stay exactly where they are. Nothing recomputes a path for a row that
already has one — `restore_worktree` and every other caller read the stored
path, and `worktree_dir` is now reached only from `create_workspace`.

No mover and no `git worktree repair` pass. Git records absolute paths in both
the worktree's `.git` file and the repo's `.git/worktrees/*` entry, so a
boot-time relocation across channels, with sessions possibly mid-turn, is a much
larger risk than the problem it closes. The cost is accepted and stated: a
machine can carry workspaces in two locations until the old ones are archived.

## UI

Settings → Coding harnesses gains a "Workspace folder" section: the folder in
force, a native Browse button, Save, and Reset to default, with the default
named in the hint either way. Stories cover inherited, custom, edited, saving,
no-picker, and headless-default states.

## Checks

- `cargo fmt --all --check` — pass
- `cargo clippy -p tidebreak-server -p tidebreak-core --all-targets -- -D warnings` — pass
- `cargo test -p tidebreak-server --lib code::` — 130 passed
- `cargo test -p tidebreak-server --lib code_clone` / `worktree` / `wire_types` — pass
- `cargo test -p tidebreak-core --lib config` — pass
- `pnpm exec tsc --noEmit` — pass
- `pnpm exec vitest run src/code/parsers.test.ts src/AppShell.dom.test.tsx src/WireFixtures.test.ts` — 67 passed
- `pnpm storybook:build` — pass

New tests pin the two contracts worth pinning: path construction (name first,
id last, unique per workspace, headless root byte-identical to the old one), and
end to end that moving the root moves the next workspace while the previous
workspace's stored path and checkout are untouched. The assertion on the *old*
row is the load-bearing one — an implementation that recomputed paths on read
would pass a test that only checked the new workspace.

### DIFF
--- a/crates/tidebreak-core/src/config.rs
+++ b/crates/tidebreak-core/src/config.rs
@@ -127,6 +127,19 @@ pub struct Config {
     /// [`Config::bind_addr`].
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub listen_addr: Option<SocketAddr>,
+    /// Where new code-mode worktrees land when no `code_worktree_root` setting
+    /// is stored.
+    ///
+    /// Worktrees hold uncommitted work on real branches, so a windowed
+    /// embedding points this at a visible place in the user's home directory —
+    /// the desktop app uses `~/Tidebreak/workspaces` — rather than letting user
+    /// work sit in disposable app data. Embeddings that leave it absent (the
+    /// CLI, self-host, tests) keep worktrees under
+    /// `{data_dir}/code/worktrees`, which is the right answer for a headless
+    /// deployment whose data directory *is* its user-visible location. Either
+    /// way the stored setting wins once an operator sets one.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub code_worktree_root_default: Option<PathBuf>,
 }
 
 /// The bind every profile uses when no address is configured: loopback, and
@@ -191,6 +204,7 @@ impl Config {
             auth_gateway_verifier_url: None,
             public_url: None,
             listen_addr: None,
+            code_worktree_root_default: None,
         }
     }
 
@@ -292,6 +306,7 @@ impl Config {
             auth_gateway_verifier_url,
             public_url,
             listen_addr,
+            code_worktree_root_default: None,
         })
     }
 
--- a/crates/tidebreak-desktop/src/lib.rs
+++ b/crates/tidebreak-desktop/src/lib.rs
@@ -535,6 +535,17 @@ fn home_dir(app: &tauri::AppHandle) -> Result<PathBuf, String> {
         .map_err(|e| format!("resolve home dir: {e}"))
 }
 
+/// Default root for code worktrees: `~/Tidebreak/workspaces`.
+///
+/// Not created here. The server creates each worktree's parents when it adds
+/// one, so an install that never opens code mode leaves no empty folder in the
+/// user's home directory.
+fn worktree_root_default(app: &tauri::AppHandle) -> Result<PathBuf, String> {
+    Ok(home_dir(app)?
+        .join(channel::PRODUCTION_PRODUCT_NAME)
+        .join("workspaces"))
+}
+
 /// Resolve the directory Tauri stages bundle resources into.
 ///
 /// Packaged builds use Tauri's normal resource path (`Contents/Resources` on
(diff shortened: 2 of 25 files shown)


naingthet (commented):
Thet's review agent

:white_check_mark: g2g

**Verdict: g2g**

naingthet (commented):
Thet's review agent
restacked onto main after #2330/#2329. Kept worktree root + harness-install warming.
**Verdict: g2g**

naingthet (commented):
Thet's review agent

:white_check_mark: g2g

**Verdict: g2g**

Issue #2346, opened 2026-08-20
feat(desktop): redesign the in-app browser for shared human and agent control


naingthet:
Parent: #2335
Depends on: host-owned browser sessions.

## Outcome

Turn the current compact browser tab into a polished developer browser whose human/agent ownership and failure states are always legible.

## Deliverable

- Refine navigation, site identity/security, address/search, profile state, and overflow actions.
- Add responsive viewport presets and custom width without compromising native child-view bounds.
- Add an agent status strip showing controller, current action, origin, Stop, and Take over.
- Add inspect mode that highlights the same semantic elements exposed to agents.
- Cover empty, loading, slow load, failure, unsupported frame, consent waiting, agent active, paused/takeover, popup, staged download, compact, and split-pane states.
- Add or update typed Storybook fixtures rather than duplicating ad hoc sample data.
- Keep full console/network/eval diagnostics behind a separate future developer capability.

## Acceptance criteria

- Human and agent ownership is understandable without opening a dialog.
- All app overlays correctly obscure the native browser surface.
- Keyboard focus, screen-reader labels, compact layouts, and split panes remain usable.
- A dedicated browser Storybook suite covers every meaningful state.
- Focused UI tests, typecheck, and storybook:build pass before publishing.

naingthet:
The first shared human/agent browser UI pass is implemented in local branch `thet/agent-browser-driveable-ui` (not published yet).

Implemented:
- refined browser chrome, navigation grouping, address/security treatment, notices, loading, and fallback states;
- native-authority sharing states: `Share with agent`, exact-origin shared, all-local-sites shared, paused cross-origin review, and `Stop sharing`;
- visible agent controller/current-action row with Stop and Take over;
- access-only event handling so consent state updates without a controller transition;
- nonfatal consent/control notices that preserve the visible page;
- Storybook coverage for empty, loading, unshared, shared, paused, controlled, takeover-required, slow, failure, popup, download, compact, and controller states;
- DOM coverage for sharing, revocation, paused navigation, Stop, takeover, workspace-scoped host calls, and stale events.

Validation: full UI suite passes 203 files / 1439 tests; TypeScript, formatting, diff check, and `storybook:build` pass. Real visual QA is still pending because no in-app browser session was attached during this pass. Viewport presets and semantic inspect mode also remain open.

naingthet:
[...]

Issue #2342, opened 2026-08-20
feat(browser): expose a session-scoped browser CLI and MCP bridge to code harnesses


naingthet:
Parent: #2335
Depends on: #2339, #2341, and #2371.

## Outcome

Give every supported code harness the same session-scoped browser capability without creating engine-specific browser implementations.

## Deliverable

- Add `BrowserChannelSpec` to `SessionSpec`; do not route browser authority through generic `extra_env`.
- Mint a capability scoped to local owner, workspace, controller session, and allowed browser ids.
- Write the credential into a session-private file, inject only its path through adapter-owned environment, and revoke/delete it when the harness ends.
- Serve authenticated loopback `/code/browser/*` routes backed by a desktop-supplied runtime contract.
- Add `tidebreak browser list|navigate|snapshot --json` and `tidebreak browser-mcp` as the shared provider-neutral bridge.
- Launch the MCP bridge through the absolute packaged executable carried by the browser channel; never rely on ambient `PATH`.
- Advertise the same bridge to Claude, Codex, and OpenCode; provide Grok a deterministic CLI fallback when its captured interface cannot mount stdio MCP.
- Keep tokens out of argv, provider config, logs, approval previews, and persisted session metadata.

## Staged capability boundary

1. First release: list, navigate, and semantic snapshot only.
2. Follow-up: wait and screenshot after native observation stability and race hardening.
3. Semantic act only after the native engine can truthfully advertise `semantic_actions: true`; it must remain absent while false.

## Acceptance criteria

- Claude, Codex, and OpenCode can discover exactly the same three initial browser MCP tools; Grok can perform the same operations through the documented CLI fallback.
- An expired, copied, cross-workspace, hidden, stopped, stale-instance, stale-document, or ended-session capability is refused.
- A real code session completes the local fixture list → navigate → snapshot workflow through the packaged executable.
- Harness exit and desktop restart revoke or recover capability state conservatively.
- CLI JSON shapes, MCP error categories, bounded I/O, redirect refusal, token redaction, and provider config merging are contract-tested.

## Tracked implementation slices

- #2371 — packaged bridge executable
- #2372 — desktop native runtime adapter
- #2373 — OpenCode and Grok provider coverage
- #2374 — end-to-end shipped-path fixture
- #2400 — wait and screenshot bridge follow-up

naingthet:
[...]

Issue #2341, opened 2026-08-20
feat(browser): add origin consent, audit, stop, and human takeover


naingthet:
Parent: #2335
Depends on: host-owned browser sessions.

## Outcome

Apply Tidebreak's existing capability and native-confirmation model to browser observation and control while preserving fast multi-step use.

## Deliverable

- Add `BrowserObserveOrigin`, `BrowserControlOrigin`, and `BrowserTransferFiles` to the grant vocabulary.
- Grant public sites by normalized origin and support a separate all-loopback-origins-for-this-workspace scope.
- Pause cross-origin navigation and redirects before exposing ungranted content to an agent.
- Keep ordinary actions within a granted origin free of per-click prompts.
- Omit password and one-time-code values from snapshots and require human takeover rather than agent filling.
- Require act-time native confirmation for public form submission, file transfer, and other commit-shaped external effects.
- Add a native Stop latch checked before every action and a visible controller/current-action state.
- Pause queued actions and transfer ownership when the user interacts with the controlled browser.
- Audit controller, browser/tab, origin, action type, semantic target label, confirmation state, timestamp, and outcome without storing entered text or full page content.

## Acceptance criteria

- The renderer and page cannot enumerate or redeem pending approvals.
- Revoked, expired, cross-workspace, and stopped capabilities fail before engine dispatch.
- Human input wins over queued agent input deterministically.
- Page-authored content is treated as untrusted data in prompts and tool results.
- Grant, confirmation, audit-before-action, takeover, and Stop behavior are test-covered.

naingthet:
The native consent and control layer is implemented in local branch `thet/agent-browser-driveable-ui` (not published yet).

Implemented:
- normalized exact-origin grants plus workspace-wide loopback grants;
- capability expiry/revocation and cross-workspace denial before engine dispatch;
- cross-origin redirect pause before destination URL/document exposure;
- native public-site and loopback sharing prompts;
- exact, expiring, single-use confirmation records for consequential actions;
- audit intent persisted before dispatch, with outcomes and no entered text, full URLs, or page content;
- Stop latching before queued dispatch, human takeover, and revocation back to unshared human control;
- password and OTP fields omitted and routed to human takeover;
- current-origin rechecks immediately before confirmation and dispatch.

Adversarial desktop browser tests pass 29/29, including broken-audit-storage Stop/takeover behavior. The issue remains open until trusted native semantic input and real in-app fixture validation land.

naingthet:
[...]

Issue #2340, opened 2026-08-20
feat(browser): expose browser tools to foreground chat


naingthet:
Parent: #2335
Depends on: semantic macOS driver and browser consent/control.

## Outcome

Let foreground chat agents drive the same visible browser the user sees through validated client-executed tools.

## Deliverable

Register bounded named tools for browser list/open/close, navigate, snapshot, screenshot, click, fill, select, check, key press, scroll, and wait. Reads use the read-only client-execution class; navigation and input use the browser-sensitive approval vocabulary.

Calls park in durable state, are claimed and heartbeated by the desktop, are reauthorized against browser/session/origin scope, and resolve through the existing client-execution path. Screenshot results reuse the existing image-result pipeline.

Every action carries browser id, tab id, snapshot id, document epoch, and semantic target ref where applicable. Page content is untrusted result data and never widens host authority.

## Acceptance criteria

- A real foreground turn completes the fixture's local Todo-style workflow and returns a final screenshot.
- Reads remain available where plan-mode policy permits; acting tools follow the established sensitive-tool behavior.
- Lost claims, desktop restart, stale targets, stopped control, and expired grants resolve conservatively without replaying input.
- No renderer command can claim or resolve a browser tool call.

naingthet:
Critical-path implementation is now claimed by the orchestrator. This track will stack behind the trusted native-driver PR from #2339 rather than registering browser tools globally. The design target is authoritative foreground chat/workspace capability derivation, durable client-execution claim/heartbeat/resolve, browser-sensitive approval classes, and screenshot reuse through the existing image pipeline. No edit will begin until the #2339 branch exposes the trusted runtime boundary it needs.

naingthet:
Status update — August 21, 2026

The shared native runtime and code-harness bridge are now landed, but foreground-chat browser tools have not been implemented. This issue remains the owner for authoritative foreground execution/workspace-to-browser capability derivation, durable client-execution claim/heartbeat/resolve, browser approval classes, and screenshot result hydration.

Observation-first list/navigate/snapshot work can reuse the shipped runtime. Acting tools must remain absent until #2339 can truthfully report `semantic_actions: true`.

naingthet:
Implementation update — August 21, 2026

The observation-first foreground slice is now tracked as #2413. It is assigned, in progress, and scoped to exactly list, navigate, snapshot, wait, and screenshot.

[...]

PR #2351, opened 2026-08-21
feat(browser): add agent-ready in-app browser foundation

naingthet:
## Summary

- add engine-neutral browser contracts, the WKWebView-first architecture decision, and a deterministic local browser fixture
- move browser identity, workspace binding, profile metadata, document epochs, controller ownership, grants, confirmations, and audit into the native host
- add bounded semantic snapshots with ephemeral refs, stale-target checks, explicit unsupported cross-origin frames, and sensitive-field omission
- add native origin sharing, cross-origin redirect pause, Stop, takeover, revocation, and consequential-action confirmation
- redesign the browser chrome around native sharing/controller state and add dedicated Storybook coverage

## Trust boundary

Page content is always returned as untrusted data. Password, one-time-code, and file-input values are omitted. Audit records exclude entered text, full URLs, and page content.

`semanticActions` deliberately remains `false`: the page-world action prototype is not advertised as trusted input. Foreground and harness browser tools are also not globally registered until the runtime can derive an authoritative workspace capability. Those follow in #2339, #2340, and #2342.

## Validation

- `cargo test --target-dir /tmp/tidebreak-browser-build -p tidebreak-desktop --lib browser_` — 29 passed
- `cargo test --target-dir /tmp/tidebreak-browser-core-build -p tidebreak-core browser::tests --lib` — 7 passed
- focused browser/workspace Vitest lane — 48 passed
- browser fixture endpoint tests — 4 passed
- `git diff --check`

A real in-app visual/native-driver run remains pending because no attached in-app browser session was available during this pass.

## Tracking

Advances #2336, #2337, #2339, #2341, and #2346 under epic #2335 and the Agent browser milestone. The remaining end-to-end tool bridge stays tracked in #2340 and #2342.


naingthet:
Continuous review finding (non-CI):

- Browser control can become permanently latched after Stop, human takeover, capability expiry, or capability revocation. The dispatch halt / prior controller capability remains authoritative, but the UI exposes no operation that clears it, so a later agent can be rejected forever as either stopped or owned by another controller. Please add an explicit, tested state transition that releases/reset control before this lands.
- “Review & resume” currently approves the destination and clears the paused state, but the navigation callback already returned false and the blocked navigation is not retained or replayed. Approval therefore does not actually resume the redirect/navigation. Please preserve and retry the pending destination, or rename/reframe the action so it does not promise resume.

I’ll re-review the next head.

naingthet:
Addressed both continuous-review findings in `8213c7df0`:\n\n- Human takeover now publishes Stop, drains the active dispatch gate, clears stale controller/pending-navigation state, and then releases the halt latch. The contract test now proves a fresh agent capability can claim afterward.\n- Paused navigation now retains the exact host-validated URL and the native webview replays it after the user approves the destination, so **Review & resume** performs the promised resume.\n\nLocal validation was limited to `git diff --check` by instruction; no Cargo, rustfmt, Rust tests, or Tauri/Rust build was run. CI is the Rust verification lane. Ready for re-review.

naingthet:
Incremental re-review of `8213c7df05` (P1/P2-only, non-CI): **G2G**.

Both prior blockers are actually fixed:

- Human takeover now publishes halt, drains the dispatch gate, clears the stale controller and pending navigation, then releases the latch. A later agent capability can claim. Take over is the UI path after Stop, expiry, or revocation, so control is no longer permanently latched.
- Paused navigation keeps the host-validated URL. After Review & resume, the native webview replays that URL. The renderer never chooses the replay target.

No new P1/P2 in this head. CI is green. Merging.

PR #2354, opened 2026-08-21
fix(code): move editor tab drag onto pointer events

naingthet:
## What changed

Tab dragging moves from native HTML5 drag to pointer events, via dnd-kit.

The strip was the only HTML5 drag source in the app, so there was never any
evidence that path worked here. Two suspects, and no way to separate them:
Tauri's `dragDropEnabled: true`, which `documents.rs:412` needs for Finder
drops, and WebKit refusing to begin an ancestor's drag from the inner
`<button role="tab">` a reader actually presses. dnd-kit's `PointerSensor`
listens for pointer events and never touches the native drag path, so both
suspects stop mattering and `dragDropEnabled` can stay on.

- One `DndContext` in `CodeWorkspacePage.tsx` wraps both strips and the split
  zone. `SortableContext` per region; each tab is a `useSortable`.
- `TabPointerSensor` subclasses `PointerSensor` to ignore presses that start on
  a control marked `data-no-drag`, so pressing close and drifting a few pixels
  still closes the tab. Activation distance is 4px, so a click still selects.
- The split zone is a `useDroppable`, rendered only while a drag is live.
- `DragOverlay` draws the moving tab. Required, not decorative: the strip is
  `overflow-x-auto` and would clip a transformed child.
- `editorDrag.ts` is rewritten rather than deleted. It used to harden a private
  MIME payload so a drop could survive `dataTransfer`; there is no
  `dataTransfer` now, so it holds the dnd-kit id protocol instead — one id per
  tab, per strip, and for the split zone — plus the pure `dropEditorTab` rules
  that turn a finished drag into a layout.
- Reordering within a strip is new. It was not possible before at all.

**Move left / move right on the tab context menu.** Reordering by drag is
pointer-only: `useSortable`'s `attributes` are deliberately left off, because
making each tab wrapper a second focusable stop would collide with the
tablist's own arrow keys. The menu carries the same move for anyone on a
keyboard, alongside the "move to split right" entry that was already there.

Three exact-pinned dependencies, matching the lockfile:
`@dnd-kit/core` 6.3.1, `@dnd-kit/sortable` 10.0.0, `@dnd-kit/utilities` 3.2.2
(`@dnd-kit/accessibility` 3.1.1 arrives transitively). All four are recorded in
the third-party notices.

Atlassian's Pragmatic drag-and-drop was rejected: it is built *on* HTML5 DnD
and would hit the same wall. `@dnd-kit/react` 0.5.0 was rejected as pre-1.0.

## Validation

`pnpm test`, `pnpm exec tsc --noEmit`, and `pnpm storybook:build` pass.

Three new tests in `CodeWorkspacePage.dom.test.tsx` cover the gesture itself,
which is what actually regressed:

- A pointer press **on the inner tab button** plus a move past the activation
  distance starts a drag. This is the exact thing WebKit would not do.
- The same sequence on the close control does not start one.
- Move right in the context menu reorders the strip.

The first two discriminate against each other, so neither is vacuous.

**What this does not prove.** dnd-kit resolves a drop target from element
rects, and jsdom reports every rect as zero, so no automated test here can
assert where a tab lands. That half is covered by the pure `dropEditorTab`
rules in `editorDrag.test.ts`, which are the real contract. Two things still
want a human in the running app: that the pointer path behaves the same inside
WKWebView as it does in jsdom, and that Finder drops into the composer still
work with `dragDropEnabled: true`.

## Compatibility and risk

No wire, schema, or persisted-layout change; the layout algebra in
`code/codeChrome.ts` is untouched. The blast radius is the center tab strip.
Worst case is that dragging is no better than it was, which is to say not
working — the pointer path is additive over a native path that never fired.

## Tracking

Part of a six-PR stack. Independent of the others.

### DIFF
--- a/crates/tidebreak-desktop/ui/package.json
+++ b/crates/tidebreak-desktop/ui/package.json
@@ -15,6 +15,9 @@
     "tauri": "cargo tauri"
   },
   "dependencies": {
+    "@dnd-kit/core": "6.3.1",
+    "@dnd-kit/sortable": "10.0.0",
+    "@dnd-kit/utilities": "3.2.2",
     "@dukelib/sheets-wasm": "0.1.23",
     "@embedpdf/core": "2.15.0",
     "@embedpdf/engines": "2.15.0",
--- a/crates/tidebreak-desktop/ui/pnpm-lock.yaml
+++ b/crates/tidebreak-desktop/ui/pnpm-lock.yaml
@@ -13,6 +13,15 @@ importers:
 
   .:
     dependencies:
+      '@dnd-kit/core':
+        specifier: 6.3.1
+        version: 6.3.1(react-dom@19.2.7(react@19.2.7))(react@19.2.7)
+      '@dnd-kit/sortable':
+        specifier: 10.0.0
+        version: 10.0.0(@dnd-kit/core@6.3.1(react-dom@19.2.7(react@19.2.7))(react@19.2.7))(react@19.2.7)
+      '@dnd-kit/utilities':
+        specifier: 3.2.2
+        version: 3.2.2(react@19.2.7)
       '@dukelib/sheets-wasm':
         specifier: 0.1.23
         version: 0.1.23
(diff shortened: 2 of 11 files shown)


naingthet:
Continuous review finding (non-CI): the implementation only registers the pointer sensor. The PR description says KeyboardSensor makes tab reordering keyboard-reachable, but there is no KeyboardSensor/keyboard coordinate getter and no equivalent reorder command. Keyboard users can select tabs but cannot perform the newly added reorder operation. Please either implement the keyboard reorder path and cover its wiring, or remove the claim and provide an accessible alternative.

Separately, because the original failure is specific to the Tauri/WebKit runtime, the pointer activation/collision path still needs the stated in-app validation before this can be considered g2g. I’ll re-review the next head.

naingthet:
Both points taken.

**KeyboardSensor.** You are right, and the claim was wrong rather than merely unimplemented — it was carried over from the plan. `useSortable`'s `attributes` are deliberately left off, because turning each tab wrapper into a focusable element would put a second stop in the tab order and collide with the tablist's own arrow keys. So a `KeyboardSensor` had nothing to attach to.

Rather than fight that, the reorder is now a **Move left / Move right** pair on the tab context menu, beside the "move to split right" entry that was already there. Keyboard-reachable, no collision with the tablist, and it reuses the same pure `reorderEditorTab` the drag path calls. Covered by a new dom test that opens the menu and asserts the strip order in the URL. The description no longer claims a KeyboardSensor.

**In-app validation.** The author cannot run it, so I have pushed the coverage as close to the gesture as jsdom allows instead of leaving it on trust. Two new tests fire a real pointer sequence — `pointerdown` **on the inner `<button role="tab">`**, then a move past the 4px activation distance — and assert the drag starts. That inner button is precisely what WebKit refused to begin an ancestor drag from. A third fires the same sequence on the close control and asserts no drag starts, so the pair discriminates and neither is vacuous. The Storybook `ActiveDragTarget` play function now presses the tab rather than the wrapper, for the same reason.

What remains unproven, stated in the description rather than assumed away: dnd-kit resolves a drop target from element rects, jsdom reports every rect as zero, so **where** a tab lands cannot be asserted here — that half is the pure `dropEditorTab` rules. And two things genuinely need a human in the app: that the pointer path behaves in WKWebView as it does in jsdom, and that Finder drops into the composer still work with `dragDropEnabled: true`.

Head is `ee568a08c`. The stale third-party notices that were failing CI are fixed in the same push.

naingthet:
Incremental re-review of `ee568a08c962a687f6b986d9165d72ebb5cb6192` (P1/P2-only, non-CI): **G2G from code review.**

The keyboard-access concern is fixed: Move left / Move right are available from the tab context menu in both editor regions and reuse the same reorder algebra. The new pointer tests also cover activation from the inner tab button and exclusion of the close control. No new concrete P1/P2 regression found.

The remaining real-app WKWebView/Finder checks are author validation, not a code-review blocker under this review bar. I’ll re-review if the head changes.

naingthet:
Re-review of `ee568a08c9` after leaving draft (P1/P2-only, non-CI): still **G2G**. Head has not changed since the prior review. Keyboard reorder is on the tab context menu; pointer tests cover the inner tab button and exclude the close control. Merging.

PR #2355, opened 2026-08-21
feat(code): run several agents in one workspace

naingthet:
## What changed

A code workspace held one interactive agent, enforced by a `session_exists`
conflict on create and recorded as a deferral in `docs/deferred.md`.

The rule was never about the agent — it was about the checkout. A workspace is
one git worktree, and two harnesses editing the same files at once overwrite
each other and race for `.git/index.lock` during per-turn checkpoints. Allowing
one session was the cheapest way to make that impossible.

This lifts the cap on conversations and puts the guarantee where it belongs. A
workspace now holds any number of interactive sessions and at most one watch
session, and turns across all of them serialize on a per-workspace async mutex
the session worker holds for the length of a turn. New ADR 0054 states the
rule; the `docs/deferred.md` entry shrinks to point at it.

**This departs from the plan, deliberately.** The plan said the turn path would
reuse the existing watch/interactive serialization rather than adding a second
mechanism. That mechanism turned out not to be one: the check in `watch.rs` is
one-directional and advisory, and `submit_turn` had no cross-session check at
all. The old check stays as a cheap pre-filter; the mutex is the mechanism.

`submit_turn` treats a busy sibling the way ADR 0009 treats a busy session: the
message parks in the session's follow-up slot and the route answers `Queued`, so
a send never holds its connection open waiting for the lock.

Conversation tabs (#2356) and fork (#2357) build on this. This is the server
half and ships no UI.

## Validation

- `cargo test -p tidebreak-server --lib code::` — 147 passed
- `cargo fmt --all --check`

Covers: several interactive sessions create cleanly; concurrent turns across two
sessions in one worktree still serialize; the one-watch-per-worktree guard still
holds.

## Compatibility and risk

No schema or wire format change — the cap was a runtime guard, not a constraint.
The risk is contention: a long turn now blocks a sibling's turn instead of that
sibling being impossible to create. That is the intended trade, and the queue
answer keeps it from looking like a hang.

## Tracking

None.

### DIFF
--- a/crates/tidebreak-server/src/code/recovery.rs
+++ b/crates/tidebreak-server/src/code/recovery.rs
@@ -631,7 +631,13 @@ mod tests {
         })
         .await
         .unwrap();
-        let handle = crate::code::session_worker::spawn_session_worker(session, engine, sink, None);
+        let handle = crate::code::session_worker::spawn_session_worker(
+            session,
+            engine,
+            sink,
+            None,
+            std::sync::Arc::new(tokio::sync::Mutex::new(())),
+        );
         let (reply, _turn) = tokio::sync::oneshot::channel();
         handle
             .commands
--- a/crates/tidebreak-server/src/code/runtime.rs
+++ b/crates/tidebreak-server/src/code/runtime.rs
@@ -747,7 +747,7 @@ impl CodeRuntime {
                 ));
             }
         }
-        self.end_workspace_sessions(owner, id).await?;
+        let workers_stopped = self.end_workspace_sessions(owner, id).await?;
         remove_worktree(std::path::Path::new(&repo.root_path), path)
             .await
             .map_err(map_worktree)?;
@@ -772,13 +772,23 @@ impl CodeRuntime {
         workspace.status = CodeWorkspaceStatus::Archived;
         workspace.archived_at = Some(Utc::now());
         save_workspace(&self.db, &workspace).await?;
-        // The checkout is gone and every session on it has ended, so nothing
-        // can take this workspace's turn lock again. Restore mints a fresh
-        // one on the first session it starts.
-        self.worktree_turns
-            .lock()
-            .expect("worktree turn locks")
-            .remove(&workspace.id);
+        // Drop the turn lock only once every worker confirmed it was gone.
+        // The lock is an `Arc`, so forgetting it here does not disturb a
+        // worker that outlived its shutdown grace — it hands the *next* one a
+        // second lock over the same checkout, which is the one thing this is
+        // supposed to make impossible. Keeping the entry costs a map slot per
+        // archived workspace and keeps restore on the same lock.
+        if workers_stopped {
+            self.worktree_turns
+                .lock()
+                .expect("worktree turn locks")
+                .remove(&workspace.id);
+        } else {
+            tracing::warn!(
+                workspace = %workspace.id,
+                "code-mode: keeping the workspace turn lock; a worker outlived its shutdown"
+            );
+        }
         Ok(workspace)
     }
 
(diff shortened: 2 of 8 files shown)


naingthet:
Continuous review findings (non-CI):

- The worktree mutex is process-local. During recovery, a still-live orphaned harness is fenced, but non-fenced sibling sessions can be attached with a fresh mutex and begin turns against the same checkout. That breaks the single-writer guarantee this PR is meant to establish. Recovery needs a durable/external ownership fence, or it must prevent every sibling from writing until the orphan is gone.
- Two simultaneous first sends can both pass workspace_turn_running before either session is marked Running. One worker gets the mutex; the other blocks before replying, so its HTTP request hangs instead of returning Queued. A timeout/retry can then duplicate intent. Reserving/enqueueing the workspace turn needs to be atomic with submission.
- Archive removes the lock-map entry even when worker shutdown times out. Restore can then mint a second mutex while the old worker is still alive. Do not release the ownership primitive until the old worker is proven gone.
- A worker waiting on the mutex is not draining interrupt/shutdown commands, so cancellation can arrive before the queued turn starts yet only be handled after it starts.

The existing test covers an already-running sibling, not these races/recovery paths. I’ll re-review the next head.

naingthet:
All four are fixed in `e1ef5d7`, with tests that fail against the previous head.

**1. The pre-flight check was not a reservation.** You were right that `workspace_turn_running` answers for a moment that has already passed. It is deleted. `drive_turn` now takes the lock with `try_lock` on the send path, and `submit_turn` converts the `WorktreeBusy` refusal into the same follow-up park a busy session gets. Taking the lock is the reservation, so there is no window to lose.

**2. The loser no longer holds its request open.** Same change. The new `TurnWait` distinguishes a send — which has a connection open and may not wait — from a queued turn, which has already been acknowledged and can. `two_idle_siblings_sending_at_once_get_one_turn_and_one_queue` fires both sends concurrently and asserts one turn reply and one queued reply. Against the old code it fails with two turn replies, the second returned 3.4s after the first — the hang, reproduced.

**3. Archive keeps the lock when a worker survives.** `shut_down_worker` and `end_workspace_sessions` now report whether every worker confirmed it stopped, and `archive_workspace` removes the `worktree_turns` entry only when they did, logging a warning otherwise. This one is a guard without a direct test: reproducing it needs a worker wedged past the 5s grace, and I did not find a way to arrange that in-process that would be worth the fixture. Flagging that rather than implying coverage.

**4. The queued wait was deaf.** New `await_worktree_turn` selects between the lock and the command channel, answering an interrupt with `Ok(())` and dropping the turn before it starts — no engine call, and no turn row, since the row is written after the lock. `interrupting_a_queued_turn_stops_it_before_it_reaches_the_worktree` asserts the stop returns within 1s against a 2s sibling turn and that the queued session ends with zero turns. Isolating just this change back to `lock().await` fails it on the timeout.

**One thing you offered two remedies for, where I took the cheaper one.** For the fence you suggested either a durable cross-process fence or refusing turns workspace-wide. I did the latter: `workspace_fence_reason` refuses a turn anywhere in a workspace holding a fenced session, with kind `workspace_fenced`, and the reap reopens it. A real cross-process fence is a claim this code cannot back — the lock is a process-local mutex — so I would rather be strict than describe a guarantee that is not there. Record 54 says so explicitly now.

Record 54 also had three passages that your findings made false: it described the queued wait as necessarily deaf, kept the pre-flight check as a "cheap pre-filter", and did not mention the fence. All updated, and the pre-flight check now appears under Alternatives Considered as something rejected rather than retained.

naingthet:
Incremental re-review of `0db47ab209` (P1/P2-only, non-CI): **G2G**.

The lock-reservation fix (same change as `e1ef5d7433`, now rebased) addresses the four prior blockers:

- Taking the lock is the reservation. Two idle first sends no longer both pass a pre-flight read; the loser parks as a follow-up.
- A fenced sibling closes the whole workspace to turns until it is reaped, so a live orphan cannot share the checkout with a new writer.
- Archive keeps the turn lock unless every worker confirmed it stopped, so restore cannot mint a second lock over the same files.
- A queued wait still answers control, so an interrupt drops the turn before it reaches the worktree.

No new P1/P2. Holding merge until required CI finishes on this head.

naingthet:
Pushed the review fixes, plus one thing your review surfaced indirectly: the
clippy lane was red on this branch. Adding `wait: TurnWait` put `drive_turn` at
eight arguments, one over clippy's limit.

Rather than an `allow`, the lock and the wait mode now travel as one
`WorktreeTurn<'_>`. Neither says anything alone — the lock names the checkout
being reserved, the wait says what a caller may do when someone else holds it —
so pairing them is what they already meant. `cargo clippy --workspace
--all-targets --locked -- -D warnings` is clean and `cargo test -p
tidebreak-server --lib code::` passes.

The fix is squashed into the review-fix commit rather than stacked, since it is
repairing a line that commit introduced.

PR #2356, opened 2026-08-21
feat(code): give every agent its own tab

naingthet:
> Stacked on #2355. Review that first; this diff is against its branch and
> retargets to `main` once it lands.

## What changed

A workspace can now run several agents (ADR 0054), so the center strip shows one
tab per agent instead of a single fixed "Main agent". Each tab carries its
engine's icon and its own attention dot, so an agent that needs a person is
visible from the tab nobody is looking at.

`CodeUpdatesStore` keeps a digest per session rather than one per workspace,
mirroring the `childrenByWorkspace` map already beside it. List surfaces collapse
the set through `workspaceDigest`, which picks the agent that most wants a
person — a need first, then a running engine, then anything still live — so a
card still speaks with one voice. The workspace page reads all of it to label the
tabs. `liveCodeSession` becomes `liveCodeSessions`, ordered oldest first, so the
first agent keeps the left-hand slot and a new one appends to the right.

Selection lives in the `?task=` search param the watch rail already uses, so a
reload returns to the same agent. "New agent" in the plus menu opens a draft tab
showing the existing start picker; starting it creates a sibling session in the
same worktree.

**One thing the plan asked for is not here.** The plan said each conversation tab
would be closable. Only the unstarted draft tab is. Ending a started agent needs
a server route that does not exist yet, so that control waits for one rather than
shipping as a button that half works.

## Validation

- `pnpm exec vitest run src/code/parsers.test.ts src/code/CodeUpdatesStore.test.ts src/code/CodeWorkspacePage.dom.test.tsx` — 87 passed
- `pnpm exec tsc --noEmit`
- `pnpm storybook:build`

Stories cover three agents in one workspace, one running, one needing you, one
idle, plus the draft tab.

Not verified locally: two agents streaming at once against a real server. Worth
confirming both stream and that their turns serialize.

## Compatibility and risk

No wire format change — the digest notice already carried `session`, and only the
client's indexing of it changes. A reconnect restates a full snapshot, so a
missed notice self-heals.

## Tracking

None.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeCenterTabs.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeCenterTabs.tsx
@@ -8,6 +8,7 @@ import {
 import { CSS } from "@dnd-kit/utilities";
 import {
   ArrowRightFromLine,
+  Bot,
   ChevronLeft,
   ChevronRight,
   CircleX,
@@ -20,7 +21,6 @@ import {
   Globe2,
   SquareTerminal,
   ListX,
-  MessageSquare,
   MoveLeft,
   MoveRight,
   PanelRightClose,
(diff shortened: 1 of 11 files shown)


naingthet:
Continuous review finding (non-CI): stale or ended `?task=` routes no longer fall back safely. The task-selection effect accepts any matching session without checking that it is live, while the fallback effect exits whenever a task param exists. An ended task can therefore remain selected, and an unknown task ID can prevent the intended main-agent fallback. The previous implementation explicitly required `lifecycle !== "ended"`.

Please validate the task param against a live session, clear/replace invalid URL state, and add direct-load coverage for unknown and ended task IDs. I’ll re-review the next head.

naingthet:
Fixed in `3458766`.

You identified both halves correctly, and they compound: the selection effect accepted any id present in the list — an ended session included — while the fallback effect stepped aside whenever a param existed at all. So an ended task stayed selected, an unknown one selected nothing, and in both cases the URL went on naming it.

The param now resolves through a single `namedTask` memo that holds only while the session is still listed and still live. Everything downstream reads that instead of the raw param:

- **Stale params are cleared.** A param that resolves to nothing triggers `openWorkspaceTask(undefined, { replace: true })`. Replace rather than push, so Back does not walk the reader into the same dead link.
- **The fallback reads live conversations, not every session.** `sessions.some(...)` was the other way an ended agent stayed selected — it ends under the reader, and existence alone kept it. `conversations.some(...)` requires a live interactive session.

**One thing worth calling out, since it is a behavior I added rather than restored.** Clearing waits on a new `sessionsLoaded` flag. Before the list arrives, a param that names nothing is indistinguishable from one naming a session the page has not heard of yet — clearing eagerly would drop a *good* link on every reload. The flag resets when the workspace or the reload token changes, so it never carries across workspaces.

Direct-load coverage for both shapes you named, in `CodeWorkspacePage.dom.test.tsx`:

- `falls back to the main agent when ?task names a session that is gone` — unknown id. Against the previous head it fails on the transcript never rendering, since nothing was selected.
- `falls back to the main agent when ?task names an ended session` — the session is still in the list, so it isolates the `lifecycle !== "ended"` check specifically. Against the previous head it fails on `expected { task: 'sess-2' } to not have property "task"`.

Full suite: 1471 tests, `tsc --noEmit`, and `storybook:build` all pass. Rebased onto the updated #2355 head, which now carries the turn-lock fixes from your other review.

naingthet:
Incremental re-review of `3458766fdd` (P1/P2-only, non-CI): **G2G**.

The prior `?task=` fallback bug is fixed. `namedTask` only holds a listed, still-live session. A param that names nothing is replace-cleared after the session list loads, and the fallback selects from live conversations. Unknown and ended ids both fall back to the main agent without leaving a dead `task` in the URL.

This slice is stacked on #2355. Not merging until that base lands.

naingthet:
This PR was **G2G** on `3458766fdd` / `a41c4c7c52`, then closed when #2355 squash-merged and deleted `thet/multi-session-per-workspace`. GitHub will not reopen it because that base branch is gone.

The `?task=` fallback fix itself is still good. Rebase `thet/code-conversation-tabs` onto `main` (drop the already-landed #2355 commits) and open a new PR into `main`.

naingthet:
Re-review of restacked `d262095722` onto `main` (P1/P2-only, non-CI): still **G2G**.

`namedTask` still only holds a listed, live session, and a stale `?task=` is replace-cleared after `sessionsLoaded`. Holding merge until desktop UI finishes on this head.

naingthet:
Reopened rather than replaced — the history and your two reviews are worth
keeping, and GitHub will reopen a PR like this one if you put both refs back the
way it remembers them. Restoring `thet/multi-session-per-workspace` at its old
head cleared the first refusal; restoring this branch to `a41c4c7c52` cleared
the second ("force-pushed or recreated"). The base could then move to `main`,
and the rebased head `d262095722` went back on top. The temporary base branch is
deleted again.

Auto-merge is on, squash, **without** `--delete-branch`. Deleting this branch on
merge is exactly what closed this PR when #2355 landed, and #2357 is stacked on
it. I will retarget #2357 to `main` and delete this branch by hand once this
one lands.

PR #2357, opened 2026-08-21
feat(code): fork an agent into a new tab

naingthet:
> Stacked on #2356, which is stacked on #2355. Review those first; this diff is
> against #2356's branch and retargets as they land.

## What changed

Forking hands one agent's transcript to a fresh one, in a new tab, on whatever
engine the reader picks.

Because the child can be any harness, the handoff has to be something any
harness can read. The server renders the parent's turns to markdown and writes
them into the worktree at `.tidebreak/forks/<session-id>.md`. Pure serialization
of what the journal already holds — no model call and no summary, because a
generated one would be a second thing to be wrong, and the point of the file is
that the child sees what the parent saw.

`.tidebreak/` carries a `.gitignore` holding `*`, so it hides itself and the
ignore file with it. That is what keeps the transcript out of `git status`
without touching the repository's tracked ignore file or the `.git/info/exclude`
that every other worktree of the same repository reads. A long session drops its
oldest turns to fit 512 KiB and the header says so: the child needs recent
context far more than it needs the first turn.

`POST /code/sessions/{id}/fork` writes the file and reports where it landed.
Creating the child is a separate call, so the reader chooses the engine and edits
the framing line before anything is sent.

**The child receives a path, not bytes.** `Composer` gains a workspace-file chip
for exactly this: nothing is uploaded and no document is created, and the surface
names the path after the message it sends. This is a new concept rather than a
reuse of `ComposerFiles`, whose items are `ImportedDocument`s with a
`documentId` a worktree file does not have. Removing the chip removes the
reference, so the framing line stays the reader's to write.

Fork is on the conversation tab's context menu and in the workspace header,
offered for interactive agents only — a watch child is the harness's own run, not
a conversation to continue.

## Validation

- `cargo test -p tidebreak-server --lib code::` — 152 passed, 5 of them new in `code::fork`
- `cargo fmt --all --check`
- `pnpm test` — 206 files, 1473 tests passed
- `pnpm exec tsc --noEmit`
- `pnpm storybook:build`

Server tests cover the self-ignoring directory, per-turn event scoping, subagent
events staying with the task call that ran them, a visible failure, and dropping
the oldest turns to fit. Stories cover a composer seeded with a transcript chip,
a truncated one, and the tab menu.

Not verified locally: an end-to-end fork against a real worktree. Worth
confirming the file exists on disk and that the child's first message names it.

## Compatibility and risk

Adds one route and one wire type (`CodeForkTranscript`); `wire.ts` is
regenerated. The server writes a file inside the worktree, which is new — it is
confined to `.tidebreak/`, ignored, and named by session id, so a re-fork
overwrites its own file and nothing else.

## Tracking

None.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/Composer.tsx
+++ b/crates/tidebreak-desktop/ui/src/Composer.tsx
@@ -11,6 +11,7 @@ import {
 } from "react";
 import {
   ArrowUpRight,
+  FileText,
   FolderOpen,
   Image as ImageIcon,
   Mic,
@@ -196,6 +197,25 @@ export type ComposerFiles = {
   onRemove: (documentId: string) => void;
 };
 
+/**
+ * A file that already sits in the agent's working directory.
+ *
+ * Nothing is uploaded and no document is created: the chip carries a path,
+ * and the surface names that path in the message it sends. Chat has no such
+ * files, so only code mode passes these.
+ */
+export type ComposerWorkspaceFile = {
+  /** Workspace-relative, which is what the engine is told to read. */
+  path: string;
+  /** One short line under the path — what the file holds, or how big it is. */
+  detail?: string;
+};
+
+export type ComposerWorkspaceFiles = {
+  items: readonly ComposerWorkspaceFile[];
+  onRemove: (path: string) => void;
+};
+
 export type ComposerFolders = {
   items: ChatFolderAccess[];
   /**
@@ -306,6 +326,8 @@ export type ComposerProps = {
   slash?: ComposerSlash;
   images?: ComposerImages;
   files?: ComposerFiles;
+  /** Paths the agent can already read, shown as chips and named on send. */
+  workspaceFiles?: ComposerWorkspaceFiles;
   folders?: ComposerFolders;
   /**
    * Workspace-relative paths `@` can insert as plain text. The parent fetches
@@ -352,6 +374,7 @@ export function Composer({
   slash,
   images,
   files,
+  workspaceFiles,
   folders,
   pathMentions,
   voice,
@@ -918,6 +941,20 @@ export function Composer({
           ))}
         </ul>
       )}
+      {workspaceFiles && workspaceFiles.items.length > 0 && (
+        <ul
+          className="m-0 flex list-none flex-wrap gap-2 p-0"
+          aria-label="Attached workspace files"
+        >
+          {workspaceFiles.items.map((file) => (
+            <WorkspaceFileChip
+              key={file.path}
+              file={file}
+              onRemove={() => workspaceFiles.onRemove(file.path)}
+            />
+          ))}
+        </ul>
+      )}
       {folderChips.length > 0 && folders && (
         <ul
           className="m-0 flex list-none flex-wrap gap-2 p-0"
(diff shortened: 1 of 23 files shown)


naingthet:
Continuous review findings (non-CI):

- The 512 KiB transcript cap is not actually enforced. The newest section is always retained when `kept == 0`, even if that one turn is already larger than the cap, and the header is added after budgeting. A single large turn can therefore produce a multi-megabyte file while reporting `truncated: false`. Please bound/truncate within a turn and include header bytes in the budget.
- Re-fork writes truncate and write directly to the final session path. Concurrent forks, or an agent reading while a re-fork occurs, can observe a partial file. Please write to a temporary sibling file and atomically rename, with per-path serialization if concurrent writers remain possible.

I’ll re-review the next head.

naingthet:
Both findings are fixed in `9a0e3bea5`. Thank you — the second one was a real
window, and the test for it fails against the old code.

**The cap now bounds the file.** `render_transcript` reserves the header before
it budgets anything, using the longest header it can produce (the one that
reports every turn dropped), so the header can never push the file past the cap.
Turns are then kept newest-first against what is left. A turn that is over the
budget on its own is cut on a character boundary and marked in place with
`_(the rest of this turn was too large to include)_`, and `truncated` is true
for that case as well as for dropped turns. Three tests cover it: the header
counted against the cap, a single oversized turn, and a multi-byte turn cut on a
character boundary.

The head is what survives a cut, not the tail. The heading and the ask live
there, and without them the section stops being readable markdown — cutting from
the front would keep the engine's last words and lose the question they answer.

**Writes publish by rename.** `write_transcript` stages to a sibling
`<name>.<uuid>.part` and renames onto the final path, which replaces atomically
on POSIX and through `MOVEFILE_REPLACE_EXISTING` on Windows. The staged name
carries a fresh uuid rather than a fixed suffix, which is why per-path
serialization is not needed: two concurrent forks stage to different files and
neither can be caught writing over the other's half-written bytes, so the later
rename simply wins and every reader sees one whole version or another. A failed
write removes its own staged file.

`a_reader_never_catches_a_re_fork_mid_write` is the reproduction: a blocking
reader loops on the path while 40 alternating short and long writes run, and
asserts every read starts with the header. Against the previous head it fails
with `a reader caught 0 bytes mid-write`.

**Not covered.** The cap is on bytes, not on tokens — a 512 KiB transcript can
still be more than a small context window holds, and the child engine is what
truncates then. The UI chip says "most recent N turns" whenever `truncated` is
set, which is slightly loose for the single-clipped-turn case; the in-file cut
marker is the precise signal there, and a second wire field to tell the two
apart did not seem worth it.

naingthet:
Incremental re-review of `9a0e3bea5e` (P1/P2-only, non-CI): **G2G** on this slice.

Both prior blockers are fixed:

- The 512 KiB cap now includes the header and clips a single oversized turn instead of writing past the bound. `truncated` is true when either oldest turns were dropped or a turn was cut.
- Re-fork publishes by staging a unique `.part` file and renaming, so a reader never sees a truncated mid-write.

Not merging: this is still stacked on `thet/code-conversation-tabs` (#2356 closed when #2355 deleted that base). Rebase onto `main` after the tab PR is reopened onto `main`.

naingthet:
Restacked onto the reopened #2356, which now targets `main`: head is
`72fbb4060`. No conflicts and no content change from `9a0e3bea5e` — the fork
commits just moved onto the rebased parent.

I kept the base as `thet/code-conversation-tabs` rather than pointing it at
`main` now, since #2356's commits are not on `main` yet and retargeting early
would fold them into this PR's diff. #2356 has auto-merge on without
`--delete-branch`, so this PR will not be closed out from under it the way that
one was. Once #2356 lands I will rebase this onto `main`, retarget, and delete
the stale branch.

Full suite on this head: 1495 tests, `tsc --noEmit`, `storybook:build`, and
`cargo test -p tidebreak-server --lib code::` (159) all pass.

naingthet:
Re-review of restacked `72fbb4060287` (P1/P2-only, non-CI): **G2G** on the slice, **NO GO to merge**.

The transcript cap and atomic publish fixes are still present. After #2356 landed, this PR was retargeted onto `main` and GitHub reports it **DIRTY** (conflicts). Rebase onto `main` and ping; this loop will merge once the head is conflict-free and CI is green.

naingthet:
Restacked onto `main` now that #2356 has landed: head is `c19a3f8c3`. The two
commits that PR carried are gone from this branch, so the diff is just the fork
slice — server `fork.rs`, the UI action, and the transcript chip.

GitHub retargeted this PR to `main` on its own when the parent branch was
deleted, rather than closing it the way #2356 was closed. The difference is that
#2356's base branch was deleted while it was still the base of an open PR
pointing at a branch that no longer had a merged counterpart; here the parent
merged first, so GitHub had a base to move this to.

Auto-merge is on with `--delete-branch`, which is safe now — nothing is stacked
on this one.

Re-verified on this head: `cargo fmt --all --check`, `cargo test -p
tidebreak-server --lib code::` (159), 1495 UI tests, `tsc --noEmit`, and
`storybook:build` all pass.

naingthet:
Head correction: `8d4184916`, not `c19a3f8c3`. After that comment I rebased once
more onto `9deffde0f` (#2365, the record renumber) so this branch is not a
commit behind `main`. No content change — the fork slice is byte-identical, and
it cites no decision record, so the renumber does not touch it.

Re-verified on `8d4184916`: `cargo fmt --all --check`, `cargo test -p
tidebreak-server --lib code::` (159), 1495 UI tests, `tsc --noEmit`, and
`storybook:build`.

naingthet:
Re-review of rebased `c19a3f8c37` onto `main` (P1/P2-only, non-CI): **G2G**.

Same transcript-cap and atomic-publish fixes as `9a0e3bea5e`. Holding merge until CI on this head finishes.

PR #2359, opened 2026-08-21
feat(web-search): search through the chat's own model provider

naingthet:
## What changed

A chat on GPT-5.6 Sol asked a question that needed the live web and got "Web search needs a provider", even with a ChatGPT subscription that includes search — because the OpenAI adapter refused any request carrying a hosted search (Responses has no wire field for the per-turn `max_uses` budget) and the Gemini adapter only grounds a request carrying no tools of its own. This adds a `WebSearchProvider` backend whose transport is one dedicated, tool-free sub-request to the chat's own model provider, so the host counts the calls and the budget is enforced before egress instead of on the wire; the OpenAI guard narrows to requests that carry *other* tools — the same predicate `declares_search_grounding` already applies — and Gemini needs no adapter change. Settings copy and the attention card are updated to match, including a pre-existing inversion where the panel claimed automatic mode prefers built-in search over the configured engine (it has always preferred the engine).

## Validation

`cargo test -p tidebreak-router --lib` (176), `cargo test -p tidebreak-server --lib` (1074), `pnpm vitest run` (1467 across 205 files), `storybook:build`, plus `cargo clippy` and `cargo fmt --check` clean. **Not verified locally:** the end-to-end dev run — two facts were read from the adapters rather than confirmed against live APIs, namely that the Codex backend accepts the hosted `web_search` tool under subscription auth, and that a tool-free `googleSearch` request behaves as the Gemini adapter expects. Unrelated flakiness exists on `main` and reproduces without this change: `code_terminals::create_write_read_and_two_cursors` and `configuration::api_key_put_configures_it_and_delete_reverts` each failed once under parallel load and pass in isolation.

## Compatibility and risk

Wire change: `WebSearchProviderKind` gains a `model_provider` variant (`wire.ts` regenerated); it is never operator-selectable and is absent from `CREDENTIAL_PROVIDERS` and the settings dropdown. No `supports_vendor_web_search` row moves, so in-turn search on Anthropic and the agent-turn path for OpenAI/Gemini are untouched, and explicit `host` mode still never falls back to a model provider. Known gap carried, not introduced: Gemini's Search Suggestions markup (`attribution_html`) is dropped on this path exactly as it already is on the native one, and Google requires displaying it with grounded results — worth a follow-up that fixes both paths.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/WebSearchProviderRequiredCard.tsx
+++ b/crates/tidebreak-desktop/ui/src/WebSearchProviderRequiredCard.tsx
@@ -11,6 +11,12 @@ import { Button } from "@/components/ui/button";
  * reads the same as a consent prompt or a clarifying question. The tool group
  * above already names and labels the search, so the card carries no icon of its
  * own.
+ *
+ * Reaching this card is narrower than it used to be: Claude, GPT, and Gemini
+ * chats search through their own model provider, so what is left here is a chat
+ * on a model whose provider cannot search — a self-hosted or pass-through
+ * route — or a host pinned to `host` mode with no key. Naming a second remedy
+ * matters, because for most readers switching models is the faster one.
  */
 export function WebSearchProviderRequiredCard() {
   const titleId = useId();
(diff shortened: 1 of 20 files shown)


naingthet:
Re-review of `c4895aea1e` (P1/P2-only, non-CI): **G2G**.

The sub-request path is sound: it is tool-free, never continues, and cannot recurse into host `web_search`. Host mode does not silently fall through to the chat's model provider. The OpenAI mixed hosted-search + host-tools shape stays refused.

The follow-up commit caps `max_results` across every hosted-search item in that one sub-request, so a reasoning model that browses several times cannot return N× the requested rows.

No remaining P1/P2 on this head.

Issue #2339, opened 2026-08-20
feat(browser): implement the macOS semantic browser driver


naingthet:
Parent: #2335
Depends on: host-owned engine-neutral browser sessions.

## Outcome

Make the visible macOS browser reliably observable and driveable through semantic targets instead of guessed coordinates.

## Deliverable

- Implement the selected macOS `BrowserEngine` adapter; default to WKWebView unless the engine spike selects CEF.
- Add bounded semantic snapshots, screenshots, URL/title/load state, and controlled tab/popup lifecycle.
- Emit ephemeral refs scoped to browser, tab, snapshot, and document epoch.
- Re-resolve every target immediately before acting and return `stale_target` after navigation or DOM drift.
- Support click, fill/type, select, check, focus, hover, key press, and scroll.
- Add deterministic waits for URL, load state, text presence/absence, element state, and bounded timeout.
- Report opaque cross-origin frames as `unsupported_frame` on engines that cannot inspect them; never guess inside them.
- Permit control only while the target tab is active and visible in v1.

## Acceptance criteria

- Snapshot -> act -> re-snapshot completes the local fixture's ordinary form and SPA flows.
- Stale refs and changed epochs cause no input synthesis.
- Screenshots and semantic snapshots refer to the same document generation.
- Popup, navigation, hidden-tab, timeout, engine-crash, and unsupported-frame failures use typed results.

naingthet:
A bounded semantic-observation prototype is now implemented on local branch `thet/agent-browser-driveable-ui` (not published yet).

Current coverage:
- bounded visible-content and interactive-node snapshots;
- compact refs scoped to snapshot id and document epoch;
- native registry revalidation of browser, workspace, visibility, epoch, snapshot, and target before dispatch;
- stale-target rejection after navigation or target replacement;
- password, one-time-code, and file-input values omitted with human takeover required;
- same-origin frame traversal and explicit `unsupported_frame` results for opaque cross-origin frames;
- page content labeled as untrusted data.

`semanticActions` deliberately remains `false`: the current action prototype dispatches page-world DOM events and is not trusted native input. This issue remains open for native input synthesis, waits, screenshots, typed engine failures, and a real snapshot -> act -> re-snapshot fixture run.

naingthet:
[...]

PR #2360, opened 2026-08-21
feat(browser): native semantic driver with waits, screenshots, and typed errors

naingthet:
## Summary

Advance #2339 from the bounded snapshot prototype toward a trustworthy WKWebView-first driver. Stacked on #2351 (`thet/agent-browser-driveable-ui`).

### Core contracts (`browser.rs`) — 522 lines added

- **`BrowserAction`**: Click, Focus, Fill, Select, Check, Press, ScrollIntoView with well-formed validation
- **`BrowserActArgs` / `BrowserActResult` / `BrowserActStatus`**: 10 typed statuses — `Ok`, `StaleTarget`, `HumanTakeoverRequired`, `HiddenTab`, `UnsupportedFrame`, `UnsupportedNative`, `InvalidValue`, `TargetObscured`, `EngineFailure`, `Timeout`
- **`BrowserWaitCondition`**: UrlChanged, LoadState, TextPresent, TextAbsent
- **`BrowserWaitArgs` / `BrowserWaitResult` / `BrowserWaitStatus`**: Resolved/TimedOut/Stopped
- **`BrowserScreenshotArgs` / `BrowserScreenshotResult`**: epoch-bound; snapshot_id + document_epoch must match
- Tool specs and validators for all new tools
- Full test coverage for contracts

### Desktop implementation (`browser_semantics.rs`) — 410 lines added

- **Deterministic waits**: bounded poll loop with Stop-latch checking, hard timeout, starting URL snapshot for UrlChanged, text-in-page probing via evaluateJavaScript
- **Epoch-bound screenshots**: WKWebView `takeSnapshotWithConfiguration` → TIFF → PNG via raw `objc_msgSend` (no additional crate dependencies); epoch verified against snapshot before capture
- **`browser_native_act`**: correctly returns `UnsupportedNative` while `semantic_actions` remains `false` — we never claim trusted input until the native adapter is wired
- **`act_status_from_target_error`**: mapping from BrowserTargetError to typed BrowserActStatus

### Desktop registry (`browser_control.rs`) — 10 lines changed

- Extended `BrowserTargetError` with `UnsupportedFrame`, `UnsupportedNative`, `EngineFailure`, `Timeout`
- Added `screenshot_epoch` to `BrowserRecord`; cleared on navigation/hide/lifecycle
- Added `record_screenshot_epoch` and `screenshot_matches_snapshot` methods

### Fixture tests (`server.test.mjs`) — 102 lines added

- 9 new tests: wait scenario surfaces (delays, URL changes, text presence, deterministic reset, cross-origin frame stability, screenshot-readiness, typed error contracts)
- Non-Rust validation: **13/13 pass**

## Not yet implemented

- **Trusted native input synthesis** (the `semantic_actions=false` gap). The action path returns `UnsupportedNative` for every action. This is correct: we must not claim trusted input until the adapter can send input without page-authored DOM events.
- **harness/tool bridge wiring**. Browser tools are not globally registered. This waits for authoritative foreground/harness workspace context (#2339/#2340).

## Validation

- Non-Rust fixture tests: 13 pass (0 fail)
- Rust build/tests intentionally NOT RUN (per instruction)
- `git diff --check`: clean
- CI will handle clippy/fmt/test on the PR

## Review hot spots

1. `capture_browser_image` in `browser_semantics.rs` uses raw `objc_msgSend` with `extern "C"` blocks for TIFFRepresentation. This avoids pulling in `objc2-app-kit` but is delicate. CI on macOS will verify whether the function signature is correct.
2. `BrowserRecord.screenshot_epoch` is cleared in every location that clears `semantic_snapshot`. Reviewers should confirm no lifecycle path was missed.
3. The `page_contains_text` wait helper reuses the existing `evaluate_json` pattern. Reviewer should confirm the quoting of user-supplied text into a JS string is sound.


naingthet:
Continuous review findings (P1, non-CI): **NO GO**.

Native act staying `UnsupportedNative` is fine. The wait/screenshot path is not.

- **Stop does not abort an in-flight wait.** `poll_wait_condition` holds the dispatch gate for the whole poll and only samples halt at the start of each iteration. It does not `select!` on the halt watch during `evaluateJavaScript` (up to 10s) or `sleep`. After a text probe returns, a satisfied condition is reported as `Resolved` without reading halt again. Stop itself is blocked on that same gate, so a wait can succeed after the user has already hit Stop. Subscribe to the halt watch and treat Stop as terminal before returning `Resolved`.
- **Screenshot is not bound to the snapshot or to the captured document.** `browser_screenshot` calls `begin_agent_control` first, which clears the stored semantic snapshot, then never matches `snapshot_id`. Epoch is checked once, then the code awaits `takeSnapshotWithConfiguration` and copies the caller's `snapshot_id` into the result. A navigation during that async snapshot can return pixels from a later document (and origin) tagged with the earlier epoch. Re-check epoch and origin after the image returns, and refuse the capture unless `snapshot_id` still names a live snapshot.
- Related: the capture path encodes `TIFFRepresentation` bytes but labels them `image/png`. Callers that decode PNG will get a garbage image.

Text waits also do not re-check origin after `evaluateJavaScript` returns, so a navigation during the probe can satisfy `TextPresent` on a different origin than the grant. Fix that with the Stop/`select!` change.

Issue #2335, opened 2026-08-20
epic: agent-driven in-app browser


naingthet:
## Outcome

Ship a visible, Tidebreak-managed browser that both foreground chat agents and code-mode harnesses can drive through one engine-neutral capability. The first production adapter is macOS-first and targets localhost, preview URLs, public documentation, and ordinary forms.

The browser keeps its own Tidebreak-managed profile, never imports a personal browser profile, exposes semantic snapshots and stale-safe actions, remains visible and cancellable, and constrains uploads/downloads through existing Tidebreak capabilities.

## Foundations

- #2287 added first-class native browser tabs and is the UI/runtime starting point.
- #1746 recorded the earlier parked product direction; this epic graduates it into owned implementation work without reopening it.
- Decision 13 supplies the trust vocabulary: structured reads before actions, stale-target checks, native consent, consequential-action confirmation, visible control, Stop, and audit.
- #2351 landed the engine-neutral/native-control foundation and selected WKWebView for macOS v1.

## Delivery order and status

- [x] #2337 — engine qualification decision and reusable fixture
- [x] #2336 — finish host-owned profile reset, persistence, migration, and crash recovery
- [x] #2339 — add trusted native semantic actions and enable `semantic_actions` only when truthful
- [x] #2341 — consent, audit, Stop, and takeover layer
- [x] #2340 — add foreground acting tools (observation complete in #2413)
  - [x] #2413 — expose observation tools to foreground chat
- [x] #2342 — session authority, browser routes, shared CLI/MCP bridge, and code-harness integration
- [x] #2371 — package the browser bridge executable with desktop builds
- [x] #2372 — bind session-scoped browser routes to the native desktop registry
- [x] #2373 — complete OpenCode MCP and Grok CLI-fallback coverage
- [x] #2374 — prove the shipped code-session browser contract end to end
- [x] #2400 — expose wait and screenshot through the session bridge
- [x] #2346 — browser UI and split-pane/unsupported-frame story coverage (inspect mode complete in #2421; visual inspection complete in #2345)
- [x] #2344 — bounded upload and download flows
- [x] #2345 — release hardening and deferred-roadmap cleanup

## Current first-release boundary

- On macOS, the WKWebView adapter advertises `semantic_actions: true`. Foreground chats and local code harnesses can use `browser_act` alongside list, navigate, snapshot, and wait. Each action rechecks the live controller, visibility, origin grant, document, and target.
[...]

PR #2362, opened 2026-08-21
feat(browser): implement native semantic driver with waits, screenshots, and typed errors

naingthet:
## Summary

- add native semantic snapshot, wait, screenshot, and typed failure contracts for the shared in-app browser
- keep semantic action synthesis explicitly unsupported until the trusted action path is ready
- extend the deterministic browser fixture for wait, stale-target, frame, and screenshot coverage

## Trust boundary

The native host remains authoritative for workspace scope, document epochs, controller state, grants, and stale targets. Page content is untrusted, sensitive values stay omitted, and native semantic actions are not advertised by this slice.

## Validation

- browser fixture Node tests: 13/13 passed
- `git diff --check`
- no Cargo, rustc, rustfmt, Rust tests, or Tauri/Rust-backed builds were run locally; CI owns Rust validation

## Review status

Draft while the native/macOS and trust/concurrency review loops converge to no P1/P2.

## Tracking

Supersedes #2360, which GitHub automatically closed when the stacked foundation branch was squash-merged and deleted. Advances #2339 under epic #2335 and the Agent browser milestone.

### DIFF
--- a/crates/tidebreak-core/src/browser.rs
+++ b/crates/tidebreak-core/src/browser.rs
@@ -18,14 +18,30 @@ pub const BROWSER_LIST_TOOL: &str = "browser_list";
 pub const BROWSER_NAVIGATE_TOOL: &str = "browser_navigate";
 /// Read one authorized tab as a bounded semantic page snapshot.
 pub const BROWSER_SNAPSHOT_TOOL: &str = "browser_snapshot";
+/// Poll for a bounded deterministic page condition with a hard timeout.
+pub const BROWSER_WAIT_TOOL: &str = "browser_wait";
+/// Capture an epoch-bound screenshot whose generation matches the most
+/// recent semantic snapshot.
+pub const BROWSER_SCREENSHOT_TOOL: &str = "browser_screenshot";
+/// Perform a single trusted semantic action on a re-resolved target.
+///
+/// This tool is registered only when the engine adapter can synthesise
+/// native input behind origin consent and a live Stop latch. Until then
+/// it returns [`BrowserActStatus::UnsupportedNative`] for every action.
+pub const BROWSER_ACT_TOOL: &str = "browser_act";
 
-/// The first browser tool slice intentionally exposes observation and
-/// navigation only. Semantic page input remains unavailable until the native
-/// adapter can send trusted input behind origin consent and a Stop latch.
-pub const BROWSER_TOOLS: [&str; 3] = [
+/// The complete set of browser tools this contract supports.
+///
+/// Wait and screenshot are available on any engine that can inspect a page.
+/// Semantic act requires native input synthesis and is registered only when
+/// the engine adapter reports [`BrowserEngineCapabilities::semantic_actions`]
+/// as true.
+pub const BROWSER_TOOLS: [&str; 5] = [
     BROWSER_LIST_TOOL,
     BROWSER_NAVIGATE_TOOL,
     BROWSER_SNAPSHOT_TOOL,
+    BROWSER_WAIT_TOOL,
+    BROWSER_SCREENSHOT_TOOL,
 ];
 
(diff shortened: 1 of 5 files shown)


naingthet:
Re-review of `eb5df3aa3b` (P1, non-CI): **NO GO**.

Same blockers as closed #2360:

- **Stop does not abort an in-flight wait.** The poll holds the dispatch gate and only samples halt at the start of each iteration. A text probe can still return `Resolved` after the user has already hit Stop. Subscribe to the halt watch and treat Stop as terminal before returning `Resolved`.
- **Screenshot is not bound to the snapshot or the captured document.** `snapshot_id` is not matched to a live snapshot, epoch is checked only before the async `takeSnapshot`, and the bytes are `TIFFRepresentation` labeled `image/png`. Re-check epoch and origin after the image returns, and encode PNG.

Native act staying `UnsupportedNative` is fine.

naingthet:
**P1/P2 review fixes pushed — head `b9ff31ae4`**

Addresses all seven findings from the review plus the CI clippy lint:

| # | Finding | Fix |
|---|---------|-----|
| P1 | Incompatible `Sel::register` API | `Sel::register` is infallible in objc2 0.6.4 — removed `.unwrap_or_else()` |
| P2-2 | Screenshot `max_width`/`max_height` ignored | Build `WKSnapshotConfiguration` with a clamped `rect` when bounds are specified |
| P2-3 | Nil passed for nonnull PNG properties | Use `NSDictionary::dictionary()` (empty dict) instead of `nil` |
| P2-4 | Async results lack completion-time fencing | Re-check document epoch after async image capture; discard image if epoch changed |
| P2-5 | Screenshot trusts arbitrary snapshot IDs | Added `BrowserRegistry::validate_snapshot_id` — validates against the latest stored semantic snapshot |
| P2-6 | Waits can exceed deadline, miss Stop, attribute stale results | Re-check halt latch and epoch before returning `Resolved`; bound sleep to remaining deadline |
| P2-7 | Screenshot dimensions and payload unbounded | Added `MAX_SCREENSHOT_BYTES` (8 MB) ceiling; honor `MAX_BROWSER_SCREENSHOT_DIMENSION` via snapshot config |
| CI | clippy::manual_clamp | Replaced `max(80, min(timeout/20, 500))` with `(timeout/20).clamp(80, 500)` |

**Checks run (no Rust/Tauri builds):**
- `git diff --check` — clean
- Node fixture tests: 13/13 passed
- CI owns Rust compilation and clippy

Native act remains `UnsupportedNative` — this PR adds observation (snapshot, wait, screenshot) only, not trusted input.

naingthet:
Exact-head trust/concurrency review at `49c2145e4` found four P2s (no P1):

1. Semantic snapshot completion can repopulate and return page data after Stop/revocation because completion checks only workspace + document epoch.
2. Semantic snapshot completion lacks an instance-id fence, so an old WebView callback can write into a re-created same-id record when epochs coincide.
3. Screenshot completion splits full authorization, snapshot validation, and epoch recording across separate lock acquisitions, leaving a Stop/revoke/replacement TOCTOU window.
4. Text waits advertise a hard timeout but use the fixed 10-second JavaScript timeout, so a 100 ms request can overshoot by up to 100×.

A repair lane is active on the native registry/semantics files. This PR remains blocked from ready status until those fixes are pushed, CI passes, and a fresh exact-head trust review returns zero P1/P2. No local Rust commands are being run; CI owns compile/format/test validation.

naingthet:
Incremental re-review of `2f1cb972d0` (P1/P2-only, non-CI): **G2G**.

Wait now races halt against probes and refuses `Resolved` after Stop. Screenshot binds `snapshot_id`, encodes PNG, and completes under one fence. Semantic snapshot completion cannot repopulate after Stop or revoke. Native act stays `UnsupportedNative`. Merging.

PR #2365, opened 2026-08-21
docs(decisions): renumber the sessions record to 55

naingthet:
## What

Two decision records landed as `0054`. #2351 accepted the agent-browser
contract and #2355 accepted multiple sessions per workspace, each taking "the
next number" from a `main` that did not yet have the other. The browser record
merged first, so it keeps 54 and the sessions record becomes 55.

## Why it matters

The number is how the codebase cites a record. Eleven comments across
`runtime.rs`, `session_worker.rs`, and `watch.rs` say "record 54", and after the
collision each of them names two different decisions. `docs/code-mode.md` and
`docs/deferred.md` link the file, so their links still resolved — the prose did
not.

## What changed

The file is renamed, its title goes from `# 54.` to `# 55.`, and the citations
follow. The record's text is untouched. `docs/decisions/README.md` needs no
edit: it says the directory is the index and nothing lists the records
elsewhere.

## Not included

Nothing stops this recurring. Two agents branching from the same `main` will
both take the same next number again, and no lane checks for it. A duplicate-
number check would fit the existing `scripts/*.test.mjs` pattern, but it is a
separate change from fixing the collision that already exists.

## Verification

`cargo fmt --all --check` and `cargo check -p tidebreak-server` pass. No
behavior changes — comments, docs, and one filename.

### DIFF
--- a/crates/tidebreak-server/src/code/runtime.rs
+++ b/crates/tidebreak-server/src/code/runtime.rs
@@ -146,7 +146,7 @@ pub(crate) struct CodeRuntime {
     /// overlap: two harnesses editing the same files, and two checkpoints
     /// racing for `.git/index.lock`, is corruption rather than concurrency.
     /// A worker takes the workspace's lock for the length of a turn, so a
-    /// sibling's turn starts after this one ends. See record 54.
+    /// sibling's turn starts after this one ends. See record 55.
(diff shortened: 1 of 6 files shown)


naingthet:
Review of `e0561ecae374` (P1/P2-only, non-CI): **G2G**.

This only renumbers the sessions decision from 54 to 55 so it does not collide with the browser record. Comments, docs links, and the file name match. Merging.

PR #2367, opened 2026-08-21
feat(server): add browser route seam and BrowserRuntime trait

naingthet:
Part of epic #2335. Stacked on #2366 (session browser authority) — the base branch is `thet/browser-session-authority`, so this diff shows only the route-seam slice.

## What this adds

**`BrowserRuntime` trait** (`crates/tidebreak-server/src/code/browser_runtime.rs`)
A server-private `async_trait` seam with `list` / `navigate` / `snapshot` / `wait_for` / `screenshot` / `act`, each taking the `BrowserSubject` derived from the caller's capability token plus already-validated args from `tidebreak-core::browser`. Typed `BrowserRuntimeError` variants map to stable statuses: `UnknownBrowserId` → 404, `SessionEnded` → 403, `StaleTarget` → 409, `Unsupported` → 501, `Failed` → 500. The desktop's engine adapter implements this in the native-driver slice; until then `AppState.browser_runtime` is `None` and every route answers 501.

**Six POST routes** (`crates/tidebreak-server/src/routes/code/browser.rs`)
`/code/browser/{list,navigate,snapshot,wait,screenshot,act}`, registered as a separate router merged into the API *after* `route_layer(require_token)` wrapped the launch-token routes, so they stay under `require_app_origin` + CORS but never see the launch bearer. Each handler resolves its own `Authorization: Bearer` header through `BrowserTokenRegistry::subject_for_token`, then walks a fail-closed ladder: missing/unknown/revoked token → 401; session gone, ended, or fenced, or workspace inactive → 403; session/workspace mismatch or missing workspace → 404 (indistinguishable from a missing target). Owner, workspace, and session identity come only from the token — the deny-unknown-fields bodies make a payload naming `workspace_id`/`session_id`/`owner_id` a 400, and the token never appears in a path, query, or response body.

**Synchronous revocation hook**
`BrowserTokenRegistry::revoke` now fires an optional observer (installed at bind time when a runtime is attached) after dropping in-memory authority, calling the new `BrowserRuntime::revoke_session` — so the native capability is revoked before the session-end call returns, not on a later poll. Fires once per actually-removed entry; repeated revokes stay silent.

**Core contracts** (`tidebreak-core/src/browser.rs`)
Adopts the wait/screenshot/act contracts (`BrowserWaitArgs`/`Result`, `BrowserScreenshotArgs`/`Result`, `BrowserActArgs`/`Result`, `BrowserAction`, statuses, validators, tool specs) byte-for-byte from `thet/browser-native-driver`, so the two branches converge without conflict.

**Plumbing**
`AppState.browser_runtime: Option<Arc<dyn BrowserRuntime>>` mirrors the `LocalVoiceRunner` posture and threads through `bind_inner`; the public bind functions pass `None` until the desktop adapter lands (the trait is `pub(crate)` for now, so it deliberately cannot appear in the public bind signatures yet). New `ServerError::forbidden` / `not_implemented` helpers.

## Tests

`crates/tidebreak-server/src/tests/code_browser.rs` drives the full router over loopback HTTP with a `FakeBrowserRuntime` and directly-seeded workspace/session rows: valid token lists (and the runtime sees exactly the token's subject), missing/unknown/app-launch token → 401, ended session → 403, cross-workspace token → 404, revoked token → 401 with the hook observed synchronously, missing runtime → 501, stale snapshot → 409 `stale_browser_target`, ill-formed URLs → 422 before the runtime, subject identifiers in bodies → 400, and no error body echoes the token.

## Validation

Rust builds and tests are CI-only per repo policy; `git diff --check` is clean locally. Keeping this draft until CI is green and reviews converge.

### DIFF
--- a/crates/tidebreak-core/src/browser.rs
+++ b/crates/tidebreak-core/src/browser.rs
@@ -18,31 +18,14 @@ pub const BROWSER_LIST_TOOL: &str = "browser_list";
 pub const BROWSER_NAVIGATE_TOOL: &str = "browser_navigate";
 /// Read one authorized tab as a bounded semantic page snapshot.
 pub const BROWSER_SNAPSHOT_TOOL: &str = "browser_snapshot";
-/// Poll for a bounded deterministic page condition with a hard timeout.
-pub const BROWSER_WAIT_TOOL: &str = "browser_wait";
-/// Capture an epoch-bound screenshot whose generation matches the most
-/// recent semantic snapshot.
-pub const BROWSER_SCREENSHOT_TOOL: &str = "browser_screenshot";
-/// Perform a single trusted semantic action on a re-resolved target.
-///
-/// This tool is registered only when the engine adapter can synthesise
-/// native input behind origin consent and a live Stop latch. Until then
-/// it returns [`BrowserActStatus::UnsupportedNative`] for every action.
-pub const BROWSER_ACT_TOOL: &str = "browser_act";
 
-/// The complete set of browser tools this contract supports.
-///
-/// Wait and screenshot are available on any engine that can inspect a page.
-/// Semantic act requires native input synthesis and is registered only when
-/// the engine adapter reports [`BrowserEngineCapabilities::semantic_actions`]
-/// as true.
-pub const BROWSER_TOOLS: [&str; 6] = [
+/// The first browser tool slice intentionally exposes observation and
+/// navigation only. Semantic page input remains unavailable until the native
+/// adapter can send trusted input behind origin consent and a Stop latch.
+pub const BROWSER_TOOLS: [&str; 3] = [
     BROWSER_LIST_TOOL,
     BROWSER_NAVIGATE_TOOL,
     BROWSER_SNAPSHOT_TOOL,
-    BROWSER_WAIT_TOOL,
-    BROWSER_SCREENSHOT_TOOL,
-    BROWSER_ACT_TOOL,
 ];
 
 /// Maximum wire length of an opaque browser id.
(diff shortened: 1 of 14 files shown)


naingthet:
CI status: rustfmt and clippy are green as of f18b9038 (the first push had a call-width violation in `bind_configured` and a staged dead-code warning on `BrowserRuntimeError`; both fixed). The remaining red `test` lane is inherited from the base branch — `thet/browser-session-authority` (ede163b1, #2366) fails the same lane with the identical signature (2 annotations, ~6-minute fast-fail) before this PR's commits. Once #2366's branch goes green, a rebase here should bring this PR fully green. Keeping draft until then.

naingthet:
Test-lane update after a full bisection (this sandbox cannot read Actions logs, so this was done with scratch probe PRs, all since closed and deleted):

Every probe below is the green base `thet/browser-session-authority` @ 449a602e plus the listed delta. The `test` lane verdicts:

| delta on top of green base | test lane |
| --- | --- |
| core browser contracts alone (`browser.rs`, core `lib.rs`) | ✅ |
| + `BrowserTokenRegistry` revocation hook | ✅ |
| + `BrowserRuntime` trait | ✅ |
| + `AppState.browser_runtime` field + bind threading | ✅ |
| + `ServerError::forbidden` / `not_implemented` | ✅ |
| + `routes/code/browser.rs` with module doc, imports, and all five helper fns (`authorize`, `bearer_token`, `attached_runtime`, `require_well_formed`, `map_runtime_error`) — handler fns stripped | ✅ |
| + the six handler fns restored — **still unrouted and unreferenced** (no re-export, no router registration) | ❌ |
| every superset containing the six handlers (incl. this PR's head dac3c65b) | ❌ |

So the flip is precisely the six `pub async fn browser_*` handler bodies, in a state where nothing references them. clippy (`--workspace --all-targets -D warnings`) is green on the full head, `desktop test` and `postgres state machine` are green, and the red `test` job completes in ~5–6 min with no rustc-style annotations — a runtime test failure (or an exotic codegen/link effect) whose identity is one glance at any red run's log.

Caveats: the two decisive probes are single runs each (Actions stopped picking up my re-run pushes — likely throttled after ~16 workflow runs), though the 7-green/7-red correlation across all runs without/with the handlers is consistent throughout.

**Ask:** could someone with log access paste the failing test name from this PR's `test` job? With the name, the fix is almost certainly quick. Until then this stays draft.

PR #2368, opened 2026-08-21
feat(harness): advertise browser MCP to Claude Code

naingthet:
## Summary

Replace the hardcoded ambient `tidebreak` command in the Claude Code browser MCP config with the trusted absolute path from `BrowserChannelSpec.bridge_command`.

## Changes

- `browser_mcp_config_entry` now takes `bridge_command: &Path` instead of hardcoding `"tidebreak"`
- `merged_mcp_config_json` passes `spec.bridge_command()` to the entry builder
- Tests updated for two-argument `BrowserChannelSpec::new(capability_file, bridge_command)`
- New test: paths with spaces remain one JSON command value (not split)
- Module doc comment updated to show absolute path example

## Security

- The capability/token travels only through `TIDEBREAK_BROWSER_CAPFILE` in the child environment
- No secrets in argv, config values, logs, or errors
- No `act`, `wait`, or `screenshot` tools advertised
- No browser config emitted when `spec.browser` is `None`

## Stacking

Based on `thet/browser-bridge-command-seam` (#2367). Merge after #2367.

Resolves part of #2342.

### DIFF
--- a/.github/workflows/cache-macos.yml
+++ b/.github/workflows/cache-macos.yml
@@ -79,6 +79,7 @@ jobs:
             target/aarch64-apple-darwin/release/deps
             target/aarch64-apple-darwin/release/tidebreak-desktop
             target/aarch64-apple-darwin/release/tidebreak-host-broker
+            target/aarch64-apple-darwin/release/tidebreak
             target/aarch64-apple-darwin/release/libtidebreak_desktop_lib.*
             target/x86_64-apple-darwin/release/.cargo-lock
             target/x86_64-apple-darwin/release/.fingerprint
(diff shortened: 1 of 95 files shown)


naingthet:
Review of `887be77d9b` (P1/P2-only, non-CI): **G2G**.

Claude Code's browser MCP entry uses `BrowserChannelSpec.bridge_command` (absolute trusted path), not ambient `tidebreak`. The capfile is not leaked to another session. Stacked on the bridge-command seam; not merging until that base lands.

naingthet:
Incremental re-review of `ec015eb692` on `main` (P1/P2-only, non-CI): still **G2G**.

Claude Code still uses `BrowserChannelSpec.bridge_command`, not ambient `tidebreak`. Merging after #2390 if this head stays current.

PR #2369, opened 2026-08-21
feat(harness): advertise browser MCP to Codex

naingthet:
## Summary

Replace the hardcoded ambient `tidebreak` command in the Codex browser MCP `-c` override with the trusted absolute path from `BrowserChannelSpec.bridge_command`.

## Changes

- `compose_app_server_plan` uses `spec.bridge_command()` to format the `-c` override string
- Tests updated for two-argument `BrowserChannelSpec::new(capability_file, bridge_command)`
- New test: bridge command with spaces remains one quoted command value
- Test assertions updated to check for the bridge path instead of hardcoded "tidebreak"

## Security

- Capability/token travels only through `TIDEBREAK_BROWSER_CAPFILE` in the child environment
- No secrets in argv (other than the env var name), config values, logs, or errors
- No `act`, `wait`, or `screenshot` tools advertised
- No browser config emitted when `browser` is `None`

## Stacking

Based on `thet/browser-bridge-command-seam` (#2392). Merge after #2392.

Resolves part of #2342.

### DIFF
--- a/.github/workflows/cache-macos.yml
+++ b/.github/workflows/cache-macos.yml
@@ -79,6 +79,7 @@ jobs:
             target/aarch64-apple-darwin/release/deps
             target/aarch64-apple-darwin/release/tidebreak-desktop
             target/aarch64-apple-darwin/release/tidebreak-host-broker
+            target/aarch64-apple-darwin/release/tidebreak
             target/aarch64-apple-darwin/release/libtidebreak_desktop_lib.*
             target/x86_64-apple-darwin/release/.cargo-lock
             target/x86_64-apple-darwin/release/.fingerprint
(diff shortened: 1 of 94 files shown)


naingthet:
Review of `4ed91db02c` (P1/P2-only, non-CI): **G2G**.

Codex's `-c` override uses `spec.bridge_command()`, not ambient `tidebreak`. Unique diff vs the stacked base is unchanged by the merge/format commits. Stacked; not merging until the seam lands.

naingthet:
Incremental re-review of `494c3c66ab` on `main` (P1/P2-only, non-CI): still **G2G**.

Codex still uses `spec.bridge_command()`, not ambient `tidebreak`. Holding merge until CI is fully settled on this head.

Issue #2373, opened 2026-08-21
feat(harness): expose the browser bridge to OpenCode and Grok


naingthet:
Parent: #2342
Depends on: #2371 and the shared browser CLI/MCP bridge

## Outcome

Complete provider coverage without provider-specific browser implementations: OpenCode receives the shared MCP server, while Grok gets a deterministic CLI fallback when its captured interface cannot mount stdio MCP.

## Deliverable

- Add a session-local OpenCode MCP declaration while preserving and safely merging existing `OPENCODE_CONFIG_CONTENT`.
- Launch the browser bridge with the absolute packaged Tidebreak executable supplied by `BrowserChannelSpec`, never ambient `PATH`.
- Keep the capfile path in the inherited `TIDEBREAK_BROWSER_CAPFILE` environment only; do not place it or the bearer token in argv/config/logs.
- For Grok, advertise exact `tidebreak browser list|navigate|snapshot --json` usage in the session instructions when native MCP mounting is unavailable.
- Expose exactly list, navigate, and snapshot; do not advertise act while `semantic_actions` is false.

## Acceptance criteria

- OpenCode sessions see the same three `tb-browser` MCP tools as Claude and Codex without losing user-supplied OpenCode configuration.
- Grok sessions receive an executable, absolute-path CLI fallback that uses the inherited session capfile.
- Browser configuration is absent when no browser channel exists.
- Contract tests cover malformed/conflicting OpenCode config, path quoting, no-secret serialization, and Grok instructions.

Issue #2372, opened 2026-08-21
feat(desktop): bind code-session browser routes to the native registry


naingthet:
Parent: #2335
Depends on: #2339 and #2342

## Outcome

Make the desktop host the concrete browser runtime behind the session-scoped `/code/browser/*` routes, using the existing native `BrowserRegistry` as the sole authority.

## Deliverable

- Supply a desktop-owned implementation of the server browser runtime contract at bind time, before code-session recovery starts.
- Map the authenticated owner/workspace/session scope into a short-lived native browser capability and controller label.
- Implement only list, navigate, and semantic snapshot in the first slice.
- Authorize listing through `list_for_capability`, and authorize observations/mutations through `begin_agent_control` plus `dispatch_agent`; never use workspace-wide listing as an authorization shortcut.
- Preserve Stop, human takeover, origin grants, visibility, instance, document-epoch, and stale-snapshot fences across async completion.
- Leave `semantic_actions: false` and do not advertise or route act, wait, or screenshot in this slice.
- Keep browser authority native/server-owned; no capability id or route bearer crosses into the renderer or page.

## Acceptance criteria

- A valid code-session capability can list only shared visible browser sessions in its workspace, navigate an authorized session, and receive a semantic snapshot.
- Copied, revoked, cross-workspace, hidden, stopped, stale-instance, and stale-document requests fail closed.
- Standalone server configurations without a desktop browser runtime return a stable configuration-required response.
- Focused contract fixtures cover the scope mapping and lifecycle failure cases; Rust compilation remains CI-only.

naingthet:
Completed by merged PRs #2390 and #2392.

The desktop now supplies the server-owned `BrowserRuntime` from the native `BrowserRegistry`, pairs it with the absolute packaged bridge command before session recovery, maps owner/workspace/session scope into native capabilities, and exposes only list, navigate, and semantic snapshot. The final lifecycle fix preserves live channels across interrupt, revokes only route state on reap/relaunch, and reserves native tombstones for terminal session end.

Exact-head CI and the external P1/P2 review converged green. Closing this implementation slice; wait/screenshot remain in #2400 and semantic actions remain in #2339.

PR #2364, opened 2026-08-21
feat(harness): add scoped browser channel contract

naingthet:
## Summary

Adds the engine-neutral browser channel contract consumed by Tidebreak code harnesses. This is the bounded adapter/env-injection slice of #2342 and supersedes closed PR #2361, which GitHub closed when its stacked foundation branch was deleted.

### What this delivers

- Adds optional `BrowserChannelSpec` to `SessionSpec`, carrying only a session-private capability-file path.
- Reserves `TIDEBREAK_BROWSER_CAPFILE` as an adapter-owned environment key.
- Injects the browser channel through the shared Tokio command path for Claude, Codex, OpenCode, and Grok.
- Filters reserved keys from inherited/settings environment and applies the trusted browser value last, preventing user overrides.
- Preserves existing behavior when no browser channel is present.
- Adds focused contract coverage for presence, absence, settings/snapshot override resistance, and all four adapters.
- Keeps runtime tracking references on #2342; capability mint/write/revoke and CLI routing will arrive as stacked follow-up slices.

## Trust boundary

This PR defines how a trusted, session-scoped capability-file path reaches a harness child. It does not mint tokens, write capability files, expose browser HTTP routes, or accept model-supplied workspace/browser authority.

## Validation

- `git diff --check origin/main...HEAD`
- Verified exactly four adapter call sites use `inject_env_tokio`.
- Verified no adapter calls the legacy injection path.
- Verified changed paths are limited to the harness contract/adapters and required server `SessionSpec` constructors.

Per project instruction, no Cargo, rustc, rustfmt, Rust tests, Tauri commands, or Rust-backed builds were run locally. CI is the Rust compilation authority.

## Tracking

- Epic: #2335
- Runtime/harness outcome: #2342
- Supersedes: #2361

PR #2390, opened 2026-08-21
feat(server): pair browser runtime with trusted bridge command

naingthet:
## Summary

- carry an absolute trusted browser bridge command alongside each session capability file
- require native runtime and bridge command as one fail-closed binding before advertising browser tools
- preserve runtime-only route service for source compatibility without minting agent browser channels
- reject relative bridge commands before capability minting

## Validation

- exact-head API/security review: zero P1/P2 findings
- `node --test crates/tidebreak-desktop/tests/browser-fixture/server.test.mjs` (4/4)
- `git diff --check`
- gitleaks clean
- no local Rust tooling run; CI is the Rust compiler/formatter/test oracle

Stacked on #2367. Tracks the bridge-command seam needed by #2372 and #2373.

### DIFF
--- a/.github/workflows/cache-macos.yml
+++ b/.github/workflows/cache-macos.yml
@@ -79,6 +79,7 @@ jobs:
             target/aarch64-apple-darwin/release/deps
             target/aarch64-apple-darwin/release/tidebreak-desktop
             target/aarch64-apple-darwin/release/tidebreak-host-broker
+            target/aarch64-apple-darwin/release/tidebreak
             target/aarch64-apple-darwin/release/libtidebreak_desktop_lib.*
             target/x86_64-apple-darwin/release/.cargo-lock
             target/x86_64-apple-darwin/release/.fingerprint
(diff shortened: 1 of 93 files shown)


naingthet:
Review of `46b30f3aae` (P1/P2-only, non-CI): **G2G**.

Minting requires both a native runtime and an absolute bridge command. A relative command fails before any capability file is written. Sessions do not swap capfiles or commands. Stacked on #2364; not merging until that base lands.

naingthet:
Incremental re-review of `04fce5717a` on `main` (P1/P2-only, non-CI): still **G2G**.

This head restacks the bridge-command seam onto `main`. Minting still requires both a native runtime and an absolute bridge command. Relative commands still fail before a capfile is written. Holding merge until CI on this head finishes.

naingthet:
Incremental re-review of `5fb8b5e974` on `main` (P1/P2-only, non-CI): still **G2G**.

Restack/merge of `main` plus rustfmt. Minting still requires runtime + absolute bridge command. Merging.

PR #2392, opened 2026-08-21
feat(desktop): browser runtime adapter with bridge-command pairing

naingthet:
## Summary

Add the server-side bridge-command seam that pairs the browser runtime with an absolute trusted bridge executable path. This is the foundation for the desktop adapter (#2372) and all provider adapters.

## Changes

- `BrowserChannelSpec` now requires both `capability_file` and `bridge_command` (absolute path)
- `BrowserChannelBinding` is a pair of native runtime plus absolute `PathBuf`
- Server validates absoluteness at boot and in the token registry (defense in depth)
- `SessionSpec.browser` is minted only when both runtime and command exist
- Old runtime-only binder preserves the runtime for route behavior but provides no command
- Relative bridge command rejection test added

## Security

- Two enforcement layers: boot boundary + token registry defense in depth
- No capfile or token minted when bridge command is missing or relative
- `BrowserTokenRegistry` lock held across path check, capfile write, and map commit (no TOCTOU)

## Review

Parfit review (round 1): zero findings.

## Stacking

Based on `thet/browser-route-seam` (#2367) and `thet/browser-native-driver` (#2362).

Draft until the full desktop adapter is implemented.

Resolves part of #2372.

### DIFF
--- a/.github/workflows/cache-macos.yml
+++ b/.github/workflows/cache-macos.yml
@@ -79,6 +79,7 @@ jobs:
             target/aarch64-apple-darwin/release/deps
             target/aarch64-apple-darwin/release/tidebreak-desktop
             target/aarch64-apple-darwin/release/tidebreak-host-broker
+            target/aarch64-apple-darwin/release/tidebreak
             target/aarch64-apple-darwin/release/libtidebreak_desktop_lib.*
             target/x86_64-apple-darwin/release/.cargo-lock
             target/x86_64-apple-darwin/release/.fingerprint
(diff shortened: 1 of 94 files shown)


naingthet:
Review of `1434ccb372` (P1, non-CI): **NO GO**.

`interrupt` (session continues) and `reap` (same session id, new worker) both call `revoke_browser_session`, which deletes the capfile and plants an enduring native tombstone.

- Stop-turn then leaves the still-running worker injecting a dead capfile.
- Reap mints a new HTTP token that the adapter refuses as `SessionEnded`.

Interrupt may halt in-flight dispatch. Do not delete the capfile or tombstone a live session. End/archive keep today's revoke. Reap/relaunch must be able to mint again. Cover interrupt-then-list and reap-then-list.

naingthet:
Incremental re-review of `645fb5a15b` (P1, non-CI): still **NO GO**.

`interrupt` still calls `revoke_browser_session` first. The session continues; the capfile and native capability do not. Same fix as the prior comment.

naingthet:
Incremental re-review of `3a7392164d` (P1, non-CI): still **NO GO**.

This head is a merge of `main`. `interrupt` still calls `revoke_browser_session` first, so a live session keeps running against a deleted capfile and a native tombstone. Same fix as before: interrupt may halt in-flight dispatch; do not revoke/tombstone a session that continues.

naingthet:
Incremental re-review of `048b7b8381` (P1/P2-only, non-CI): **G2G**.

The prior interrupt/reap tombstone is fixed:

- `interrupt` no longer calls `revoke_browser_session`. The worker and capfile stay live for later turns.
- Reap uses `revoke_browser_channel` (HTTP token/capfile only), then `attach_and_spawn_worker` remints. Native tombstone stays on terminal end paths only.

Holding merge until CI on this head finishes.

naingthet:
Incremental re-review of `7c91c613fe` (P1/P2-only, non-CI): still **G2G**.

The new commit is rustfmt after the lifecycle fix. `interrupt` still does not revoke the capfile. Holding merge until CI on this head finishes.

PR #2394, opened 2026-08-21
feat(harness): advertise browser CLI fallback to Grok

naingthet:
## What

Grok CLI print-mode has no MCP or structured tool channel, so the browser bridge is exposed as shell commands the agent can invoke. When `SessionSpec.browser` is `Some`, concise prompt instructions are appended to the prompt file text exposing exactly three verbs through the trusted absolute `BrowserChannelSpec.bridge_command`:

- `browser list --json`
- `browser navigate --browser-id <id> --url <url> --json`
- `browser snapshot --browser-id <id> [--max-nodes <n>] --json`

## Security

- The bridge executable path is shell-quoted for spaces and platform-shaped paths using POSIX single-quote escaping.
- Page content returned by `snapshot` is explicitly stated as untrusted data, not instructions.
- `act`, `wait`, and `screenshot` are not advertised.
- The token/capfile remains only in the inherited `TIDEBREAK_BROWSER_CAPFILE` environment variable — never in command examples, argv, logs, or prompt text.
- When browser is `None`, existing behavior is preserved: no instructions appended, prompt passes through unchanged.

## Tests

In-file tests cover:
- Presence/absence of browser instructions
- Exact allowed verbs (list, navigate, snapshot)
- No forbidden verbs (act, wait, screenshot)
- No token/capfile leakage in prompt text
- Shell quoting with spaces and embedded single quotes
- Untrusted-content warning present
- Bridge command path inclusion

## Validation

- `rustfmt --edition 2021` clean
- `git diff --check` clean
- No dependencies added
- No Rust builds run (per constraints)

Part of #2373.

### DIFF
--- a/.github/workflows/cache-macos.yml
+++ b/.github/workflows/cache-macos.yml
@@ -79,6 +79,7 @@ jobs:
             target/aarch64-apple-darwin/release/deps
             target/aarch64-apple-darwin/release/tidebreak-desktop
             target/aarch64-apple-darwin/release/tidebreak-host-broker
+            target/aarch64-apple-darwin/release/tidebreak
             target/aarch64-apple-darwin/release/libtidebreak_desktop_lib.*
             target/x86_64-apple-darwin/release/.cargo-lock
             target/x86_64-apple-darwin/release/.fingerprint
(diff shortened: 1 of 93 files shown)


naingthet:
Review of `65b8df1659` (P1/P2-only, non-CI): **G2G**.

Grok's browser fallback uses the trusted absolute bridge path, not ambient `tidebreak`. The three verbs stay on that binary. Formatting-only head change from `3f282344a8`. Stacked; not merging until the seam lands.

naingthet:
Incremental re-review of `789d438445` on `main` (P1/P2-only, non-CI): still **G2G**.

Grok still uses the trusted absolute bridge path. Restack after the seam landed.

PR #2397, opened 2026-08-21
docs(code): point the last record-54 citations at 55

naingthet:
## What

Three comments in the desktop UI still say "record 54" when they mean the
sessions record, which is now 55.

## Why #2365 missed them

#2365 renumbered the record and rewrote its citations. #2356 added these three
while that PR was open. Each branched from a `main` that did not yet have the
other, so neither could see the other's lines — the same collision #2365
existed to fix, one layer down. A stale citation now names the agent-browser
record instead of the one it means.

## What changed

Three comment lines, in `CodeUpdatesStore.ts`, `parsers.ts`, and
`CodeWorkspacePage.tsx`. `git grep "record 54"` and `git grep
"0054-multiple"` are both empty after this.

## Worth noting

This is the second time in one day that two branches took the same view of
`main` and produced an inconsistency no lane caught. #2365 flagged a
duplicate-number check as out of scope; a citation check ("every `record N`
names a file that exists, and only one does") would have caught both this and
the original collision. Still a separate change, but the case for it is
stronger now.

## Verification

`pnpm exec tsc --noEmit` passes. Comments only — no behavior change.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeUpdatesStore.ts
+++ b/crates/tidebreak-desktop/ui/src/code/CodeUpdatesStore.ts
@@ -37,7 +37,7 @@ export type CodeUpdatesState = {
   /**
    * Conversation digests, keyed workspace → session. Never a watch.
    *
-   * A workspace runs several agents (record 54), so it names a set rather
+   * A workspace runs several agents (record 55), so it names a set rather
    * than one digest. List surfaces collapse the set through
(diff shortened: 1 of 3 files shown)


naingthet:
Review of `cfa7790ff5` (P1/P2-only, non-CI): **G2G**.

Comment-only: remaining record-54 citations now point at 55. Merging.

PR #2401, opened 2026-08-21
fix(code): key checkpoint refs on the session, not the workspace

naingthet:
## What

A workspace's checkpoint refs were keyed `refs/tidebreak/checkpoints/<workspace>/<ordinal>`, but `next_turn_ordinal` counts **per session**. Since decision 0055 let a workspace hold several sessions, every session reaches turn 1 — so siblings wrote the same ref and `git update-ref` silently overwrote.

The ref gains a session segment: `refs/tidebreak/checkpoints/<workspace>/<session>/<ordinal>`.

## How it showed up

Two sessions on one workspace, each completing one turn. Both turn rows recorded the same `checkpoint_ref`, only one ref existed, and the other session's checkpoint commit was left dangling and GC-eligible. `code files --ws <id> --turn 1` returned one session's turn id attached to the *other* session's tree, contradicting the turn's own stored diffstat.

`previous_checkpoint_oid` had the same bug one step removed: it resolved `checkpoint_ref(workspace, ordinal - 1)`, so a turn would diff against whichever sibling last wrote the preceding ordinal.

## Why the workspace stays first in the path

`delete_workspace_refs` matches on the `<prefix>/<workspace>/` string and `workspace_id_from_ref` takes the first path component, so both keep working unchanged with an extra segment appended. Archive-time reaping is covered by an assertion in the new test.

## Upgrade behaviour

Refs written under the old scheme are not migrated. Existing turns degrade gracefully: `resolve_diff_range` reads `to` from the turn row's stored `checkpoint_ref` string, and `previous_checkpoint_oid` already falls back to scanning the session's turn rows when the computed ref does not resolve. Old refs are still reaped by prefix when their workspace is archived. Persisted-format changes are in scope pre-v1 per decision 0002.

## Tests

`sibling_sessions_do_not_share_a_turn_one_checkpoint` covers it: two sessions, one workspace, both at turn 1. It asserts distinct refs, that both survive, that the first session's ref still resolves to a tree holding only its own file, and that archiving reaps both.

Verified the test has teeth — reverting `checkpoint_ref` to the workspace-only key fails it with the exact production symptom:

```
assertion `left != right` failed: sibling sessions must not share a ref
  left: "refs/tidebreak/checkpoints/82d2a2a7-.../1"
 right: "refs/tidebreak/checkpoints/82d2a2a7-.../1"
```

`cargo test -p tidebreak-server --lib code::` — 184 passed.

## Known adjacent issue, not fixed here

A turn-1 diffstat still absorbs sibling edits, because ordinal 1 diffs against `merge_base(base_ref)` rather than a per-session baseline. That is a separate change and wants its own PR.

### DIFF
--- a/crates/tidebreak-server/src/code/checkpoint.rs
+++ b/crates/tidebreak-server/src/code/checkpoint.rs
@@ -4,7 +4,12 @@
 //! untracked files — as a synthetic commit created through a temporary index
 //! file. The user's index, `HEAD`, and reflog are untouched. The commit is
 //! referenced only by a hidden ref
-//! `refs/tidebreak/checkpoints/<workspace>/<ordinal>`.
+//! `refs/tidebreak/checkpoints/<workspace>/<session>/<ordinal>`.
+//!
+//! The session segment is load-bearing. A workspace holds several sessions
+//! (decision 0055) and `next_turn_ordinal` counts per session, so every
+//! session reaches turn 1; a workspace-keyed ref let one sibling's snapshot
+//! overwrite another's.
 //!
 //! Diffs are produced here, bounded in bytes and file count, with truncation
 //! marked on the payload. The renderer never runs git.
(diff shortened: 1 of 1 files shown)


naingthet:
Review of `081f81beeb` (P1/P2-only, non-CI): **G2G**.

Checkpoint refs are now `refs/tidebreak/checkpoints/<workspace>/<session>/<ordinal>`, which matches `next_turn_ordinal` (per session). Two sessions on one workspace can both complete turn 1 without `update-ref` silently overwriting the other session's snapshot. `previous_checkpoint_oid` uses the same key, so turn N diffs against that session's N−1, not whichever sibling last wrote the ordinal.

Workspace stays the first path segment, so `delete_workspace_refs` prefix match and `workspace_id_from_ref` still reap every session's refs on archive. Reads keep using the turn row's stored `checkpoint_ref`; the existing turn-row fallback covers old workspace-keyed refs. The sibling-session test locks the overwrite and the archive-reap invariant.

PR #2402, opened 2026-08-21
perf(code): reuse the checkpoint index and diff in one git spawn

naingthet:
Stacked on #2401 — review that one first; this branch contains its commit.

## Two per-turn costs in the checkpoint path

**A cold index every turn.** `snapshot_tree` created a fresh `NamedTempFile` index, so `git add -A` started with no stat cache and re-stat'd and re-hashed the entire worktree on every checkpoint. It now reuses one index per worktree, resolved via `git rev-parse --git-path tidebreak-checkpoint-index` so linked worktrees each get their own and the user's `index` is never touched.

The reused index deliberately skips `read-tree HEAD`. That command overwrites the index wholesale and throws away the stat cache, which is the whole reason for keeping the file; `add -A` reconciles whatever the index holds against the worktree — adding, updating, and staging deletions — so the tree comes out the same. Any failure, including a concurrent snapshot holding `<index>.lock`, logs and falls back to the previous temp-index path, so this cannot fail a turn.

**One git spawn per changed file.** `produce_diff` looped over changed files calling `git diff -- <path>` serially, up to `MAX_DIFF_FILES = 64`. `collect_changes` has already capped `listed.files` at `max_files`, set `truncated` when it did, and ordered them the same way git orders its own output, so passing those paths as a single pathspec preserves both bounds and the ordering while paying one process instead of N.

## Numbers

Measured with git directly, outside the server:

| Operation | Before | After |
|---|---|---|
| `snapshot_tree`, 20,000-file worktree | 0.81–0.93 s | ~0.20 s |
| `snapshot_tree`, 1,600-file worktree | 0.28–0.31 s | ~0.17 s |
| `produce_diff`, 30 changed files | 0.55 s | 0.13 s |

This is per turn, and it gates the session returning to idle, so a follow-up turn waits on it.

## Correctness

The reused index is the risky half, so `reused_index_matches_a_cold_snapshot_through_every_mutation` asserts the warm tree equals a cold-index tree after each of: adding a file, adding a second, modifying a tracked file, deleting a file, and committing so `HEAD` moves out from under the index. It also asserts the index path is not the user's `index`.

The existing `truncation_is_marked_when_caps_are_exceeded` covers the diff bounds and still passes unchanged.

`cargo test -p tidebreak-server --lib code::` — 185 passed. `cargo clippy -p tidebreak-server --all-targets` clean.

### DIFF
--- a/crates/tidebreak-server/src/code/checkpoint.rs
+++ b/crates/tidebreak-server/src/code/checkpoint.rs
@@ -297,35 +297,29 @@ pub(crate) async fn produce_diff(
         });
     }
 
-    let mut body = String::new();
+    // One `git diff` for every included path rather than one spawn per file.
+    // `collect_changes` has already capped `listed.files` at `max_files` and
+    // set `listed.truncated` when it did, so passing those paths as a single
+    // pathspec preserves both bounds and the output order (git sorts by path,
+    // as does `--name-status -z`) while paying one process instead of N.
     let mut truncated = listed.truncated;
-    for (included, entry) in listed.files.iter().enumerate() {
-        if included >= bounds.max_files {
-            truncated = true;
-            break;
-        }
-        let remaining = bounds.max_bytes.saturating_sub(body.len());
-        if remaining == 0 {
-            truncated = true;
-            break;
-        }
-        let raw = git_bytes(
-            worktree,
-            &["diff", "--find-renames", from, to, "--", &entry.path],
-            GIT_SNAPSHOT_TIMEOUT,
-        )
+    if listed.files.is_empty() {
+        return Ok(BoundedDiff {
+            diff: String::new(),
+            truncated,
+            stat: Diffstat {
+                truncated,
+                ..listed.stat
+            },
+        });
+    }
+    let mut args: Vec<&str> = vec!["diff", "--find-renames", from, to, "--"];
+    args.extend(listed.files.iter().map(|entry| entry.path.as_str()));
+    let raw = git_bytes(worktree, &args, GIT_SNAPSHOT_TIMEOUT)
         .await
         .map_err(CheckpointError::internal)?;
-        let (chunk, chunk_truncated) = truncate_bytes(&raw, remaining);
-        if !body.is_empty() && !chunk.is_empty() && !body.ends_with('\n') {
-            body.push('\n');
-        }
-        body.push_str(&chunk);
-        if chunk_truncated {
-            truncated = true;
-            break;
-        }
-    }
+    let (body, body_truncated) = truncate_bytes(&raw, bounds.max_bytes);
+    truncated |= body_truncated;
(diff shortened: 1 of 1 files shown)


naingthet:
Review of `33f4f02b8a` (P1/P2-only, non-CI): **G2G**.

Stacked on the session-keyed checkpoint refs from #2401. The reusable `tidebreak-checkpoint-index` lives in the worktree git dir via `--git-path`, not the user's `index`. A contended or unreadable index falls back to a private temp file so a concurrent snapshot cannot fail the turn. Reuse skips `read-tree HEAD` so the stat cache survives; `add -A` still reconciles adds, edits, and deletions, including after HEAD moves. Diffs of the capped file list go through one `git diff` spawn, then the existing byte bound.

naingthet:
Incremental re-review of `33f4f02b8a` (P1/P2-only, non-CI): **NO GO**.

**P2:** `snapshot_tree` always reuses with `reset_from_head = false` as soon as `rev-parse --git-path` returns a path. That path exists before the index file does, so the first snapshot is `git add -A` on an empty index. `add -A` does not stage files that are in `HEAD` but now match gitignore; `read-tree HEAD` does. A tracked-then-ignored file is in a cold snapshot and missing from the reused one, so later checkpoints, diffs, and restore treat it as deleted.

**Fix:** Resolve `--git-path` against the worktree. If the index file is missing, call `snapshot_tree_with_index(..., true)` (`read-tree HEAD` then `add -A`). If it exists, keep `reset_from_head = false`. Add a tracked-then-gitignored file to `reused_index_matches_a_cold_snapshot_through_every_mutation` so a missing seed fails the test.

naingthet:
Incremental re-review of `305b5d63aa` (P1/P2-only, non-CI): **NO GO**.

The previous P2 is still on this head. `snapshot_tree` still calls `snapshot_tree_with_index(..., false)` as soon as `rev-parse --git-path` returns a path. That path exists before the index file does, so the first snapshot is `git add -A` on an empty index. `add -A` does not stage files that are in HEAD but now match gitignore; `read-tree HEAD` does. Checkpoints, diffs, and restore then treat those files as deleted.

**Fix:** If the resolved index file is missing, seed with `snapshot_tree_with_index(..., true)`. If it exists, keep `false`. Add a tracked-then-gitignored file to `reused_index_matches_a_cold_snapshot_through_every_mutation`.

naingthet:
Incremental re-review of `e5d76df5b8` (P1/P2-only, non-CI): **G2G**.

`--git-path` is joined onto the worktree when relative, so existence is not cwd-relative to the server. A missing index file now seeds with `read-tree HEAD` (`reset_from_head = !index_path.exists()`); a present file keeps the stat cache. `reused_index_matches_a_cold_snapshot_through_every_mutation` covers a tracked-then-gitignored file.

PR #2403, opened 2026-08-21
fix(code): make the browser bridge test fixture absolute on Windows

naingthet:
## Summary

Windows CI on `main` failed in `Code mode Windows lifecycle tests` ([run 32468324417](https://github.com/brightwave-inc/tidebreak/actions/runs/32468324417/job/96729661088)). Not a flake: 17 tests failed identically on all three attempts.

`BrowserTokenRegistry::issue` refuses a bridge command that is not absolute, and `Path::is_absolute` is platform-defined. The test fixture was the Unix literal `/usr/local/bin/tidebreak`, which Windows reads as relative, so every capability-minting test got `bridge command must be an absolute path`. The two `tests::code` failures are the same cause one layer up: session creation could not mint a channel, so the response carried no `id`.

- Give the server one `test_bridge_command()` helper that branches on the host, and point all three fixture sites at it (`code/browser_channel.rs`, `tests/code.rs`, `tests/code_browser.rs`). `tests/code_browser.rs` is not caught by the lane's `code::` filter but carried the same latent bug.
- Add `the_test_bridge_fixture_is_absolute_on_this_host`, so a wrong fixture names itself instead of surfacing as a wall of identical `unwrap` panics.

## Flake sweep

Audited the last 10–15 `main` runs of the Windows and Linux test lanes for retries the wrappers were absorbing.

- Windows lane: no real retries. The 3-attempt wrapper is currently masking nothing.
- Linux lane: one genuine flake, in `committed_cancellation_wakes_container_and_fences_reverse_egress_before_heartbeat` ([run 32425751130](https://github.com/brightwave-inc/tidebreak/actions/runs/32425751130)). It failed with `provider_calls=1`, meaning the fence had held and only the 5s bound blew. That bound is a hang-guard, not a latency contract, and the test already sits inside a 30s bound, so it moves to 20s.

## Test plan

- [x] `cargo test -p tidebreak-server --lib code:: --locked` (the lane's exact filter) — 184 passed
- [x] `cargo test -p tidebreak-server --lib code_browser --locked` — 14 passed
- [x] `cargo test -p tidebreak-server --lib committed_cancellation_wakes_container --locked`
- [x] Inverted the `cfg!(windows)` branch locally to force the Windows literal: reproduced the exact CI failure cluster, and the new guard test fired with `the test bridge fixture must be absolute on this platform, got: C:\Program Files\Tidebreak\tidebreak.exe`
- [x] `cargo fmt --all --check`, `cargo clippy -p tidebreak-server --all-targets --locked`
- [ ] Windows lane green on this PR (`windows-ci` label applied)

## Note

The Windows lane only runs pre-merge when a PR carries `windows-ci`; otherwise it is a post-merge backstop on `main`. That is the deliberate cost choice in the workflow, and it is why this landed. Worth considering whether changes under `crates/tidebreak-server/src/code/browser*` should auto-apply the label, but I have not changed the trigger here.

### DIFF
--- a/crates/tidebreak-server/src/code/browser_channel.rs
+++ b/crates/tidebreak-server/src/code/browser_channel.rs
@@ -423,15 +423,30 @@ fn write_capfile(
     Ok(())
 }
 
+/// An absolute bridge-command path for tests, valid on the host platform.
+///
+/// [`BrowserTokenRegistry::issue`] refuses a bridge command that is not
+/// absolute, and `Path::is_absolute` is platform-defined: `/usr/local/bin`
+/// roots a path on Unix and is merely relative on Windows. Every test that
+/// mints a capability takes its bridge path from here so one definition
+/// stays correct on both hosts. The server never opens this path — the
+/// desktop sibling resolver owns existence and executability — so the
+/// target does not have to exist.
+#[cfg(test)]
+pub(crate) fn test_bridge_command() -> PathBuf {
+    if cfg!(windows) {
+        PathBuf::from(r"C:\Program Files\Tidebreak\tidebreak.exe")
+    } else {
+        PathBuf::from("/usr/local/bin/tidebreak")
+    }
+}
+
 // ── tests ───────────────────────────────────────────────────────────────────
 
 #[cfg(test)]
 mod tests {
     use super::*;
     use std::collections::HashSet;
-
-    /// Fixture bridge command path used by every test issuance.
-    const TEST_BRIDGE: &str = "/usr/local/bin/tidebreak";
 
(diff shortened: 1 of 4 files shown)


naingthet:
Review of `c0ecc89ce1` (P1/P2-only, non-CI): **G2G**.

`BrowserTokenRegistry::issue` refuses a non-absolute bridge command, and `Path::is_absolute` is platform-defined: `/usr/local/bin/...` is relative on Windows. `test_bridge_command` now returns a drive-rooted path on Windows and the Unix path elsewhere, and every minting test takes the fixture from there. A named test fails first if the fixture is wrong on the host instead of a wall of identical unwrap panics.

PR #2405, opened 2026-08-21
fix(code): report codex token usage as a disjoint turn total

naingthet:
## The contract was implicit, and codex broke it twice

`CodeUsage` reuses the four field names the chat side documents on `RendererTurnUsage` as **disjoint per-turn totals**, but carried only a one-line comment. The codex adapter filled them in two ways that both disagree with it.

**It read the last model call, not the turn.** `thread/tokenUsage` reports a running `total` beside the most recent call's `last`, and `usage_from` took `last`. The captured fixtures show the two genuinely diverge on a multi-call turn:

```
"total":{"totalTokens":28912,"inputTokens":28794,"cachedInputTokens":23040,"outputTokens":118}
"last": {"totalTokens":14471,"inputTokens":14466,"cachedInputTokens":14080,"outputTokens":5}
```

So a turn that produced 118 output tokens recorded 5. That is the under-reporting shape you see when a codex turn streams a long answer and the stored row says a couple of hundred tokens.

**It double-counted the cache.** `totalTokens == inputTokens + outputTokens` holds in every captured payload, so `inputTokens` is the whole prompt and `cachedInputTokens` is a subset of it. Filing that subset into `cache_read_input_tokens` while leaving it inside `input_tokens` counts it twice.

Reading `total` and subtracting the cached and written portions leaves the three prompt-side counts disjoint and summing back to `inputTokens`.

## What this changes for readers

Anything comparing harnesses from `code_turn.usage` was reading a codex row that was simultaneously one call's slice and inflated by its cached prefix. Two lanes running identical work on the same fixture reported 256,833 against 47,232 prompt-side tokens; most of that gap was this, not efficiency.

## Scope: codex only, deliberately

While checking the other adapters against the same contract:

- **grok is already correct.** Its final usage event is the cumulative turn total, not a per-call snapshot. From `tool-use.ndjson`: calls of 9050 and 139 input, 384 and 9344 cache-read, 56 and 28 output; the closing event reports 9189 / 9728 / 84, and those sum to its `total_tokens` of 19001. `permission-denied.ndjson` checks out the same way across three calls. Its fields are disjoint already. No change.
- **opencode is unverified.** Every opencode fixture has zero usage lines, so there is nothing captured to establish whether its `tokens` object is per-message or cumulative, or whether `input` excludes `cache.read`. Changing it on a guess would be worse than leaving it. It wants a capture first; I did not touch it.

## Tests

- `token_usage_is_a_disjoint_turn_total` — the real two-call payload: asserts the turn's output (118, not 5), the cache read, and that the three prompt-side counts reconstruct `inputTokens`.
- `token_usage_without_a_total_falls_back_to_the_object_itself` — an object with no `total` still normalizes rather than reporting zeros.
- Five codex replay fixtures regenerated with `UPDATE_HARNESS_FIXTURES=1`. Only `usage` values moved; every event sequence is unchanged.

`cargo test -p tidebreak-harness codex` — 45 passed.

Note: `child::unix_process_tree_tests::{terminating,dropping}_a_process_tree_*` fail on this machine, and they fail identically on a clean checkout of `main` — pre-existing and unrelated to this change.

### DIFF
--- a/crates/tidebreak-core/src/code/event.rs
+++ b/crates/tidebreak-core/src/code/event.rs
@@ -126,13 +126,30 @@ pub struct Diffstat {
     pub truncated: bool,
 }
 
-/// Token accounting as reported by the engine. Missing fields stay zero.
+/// Token accounting for one turn.
+///
+/// The four counts are **disjoint** and they are **turn totals**, summed over
+/// every model call the engine made while servicing the turn. This is the same
+/// contract the chat side states on `RendererTurnUsage`, and the reason to
+/// state it here is that every adapter reports something different natively:
+/// one engine sends a running total beside the last call's slice, another
+/// folds the cached portion back into the prompt count, another overwrites a
+/// snapshot per message. Normalizing belongs in the adapter, so that anything
+/// reading this struct — cost accounting, the CLI's turn list, the desktop's
+/// context indicator — can compare two harnesses without knowing which engine
+/// produced the row.
+///
+/// Concretely: `input_tokens` is the *fresh*, uncached prompt only. It never
+/// includes `cache_read_input_tokens` or `cache_creation_input_tokens`, so
+/// the prompt an engine actually sent is the sum of all three. Missing fields
+/// stay zero, which is not the same as "the engine sent zero" — an engine that
+/// does not surface cache counts reports nothing rather than a real zero.
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Default, TS)]
 pub struct CodeUsage {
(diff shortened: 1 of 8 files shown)


naingthet:
Review of `ba7a3e68c5` (P1/P2-only, non-CI): **G2G**.

`usage_from` reads `thread/tokenUsage.total` (the turn so far), not `last` (the most recent call). Cached and cache-write tokens are subtracted from `inputTokens` so the four `CodeUsage` fields stay disjoint and still reconstruct the engine's prompt. A payload without `total` falls back to the object itself.

PR #2406, opened 2026-08-21
fix(cli): deliver a completed turn's checkpoint before code run exits

naingthet:
## What

`code run`'s stream loop breaks on the turn's terminal frame. The server journals `checkpoint_recorded` just *after* publishing that frame, so the checkpoint was never delivered — across a full exercise run the journal held **20** and the CLI streamed **0**.

Two consequences:

- A `--json` consumer never received the turn's `diffstat`. That event is the only place it appears on the stream.
- A script acting the moment `code run` exited raced a checkpoint that was not durable yet — measured 0.25–1.6 s behind the process.

## How

After a turn completes, drain the stream until the matching `checkpoint_recorded` arrives, bounded by a five-second wait. It ends the instant the frame lands, so the normal cost is the tens of milliseconds the server actually takes.

A checkpoint that fails journals a `harness_notice` instead, so that ends the wait too. A timeout is quiet — a missing checkpoint must not fail a turn that already succeeded. Only a turn this submit owns is drained, so a stream that never bound one does not wait out the window on a sibling's turn.

## Verified against a live server

Same command that produced 0 of 20 before:

```
{"seq":114,"event":{"type":"turn_completed","usage":{...}}}
{"seq":115,"event":{"type":"checkpoint_recorded","turn_id":"227711e1-...","diffstat":{"files":0,...}}}
```

## Tests

`bound_turn_is_only_reported_for_a_turn_we_own` covers the gate accessor the drain depends on: nothing bound before a start, still nothing once the turn is merely named by `on_ran`, ours after its live `TurnStarted`, still ours after the terminal frame so the drain can run, and immediately ours for an attached turn.

`cargo test -p tidebreak-cli` — 92 + 2 + 7 + 16 passed. `cargo clippy -p tidebreak-cli --all-targets` clean.

## Not in this change

The stream frames still carry no timestamp, so measuring turn latency from `--json` output still needs a sidecar stamper. That is a server wire change (`SequencedCodeEventFrame` plus the generated TS types) and wants its own PR.

### DIFF
--- a/crates/tidebreak-cli/src/code.rs
+++ b/crates/tidebreak-cli/src/code.rs
@@ -892,6 +892,11 @@ impl TurnGate {
         self.ours
     }
 
+    /// The turn this submit owns, once bound.
+    fn bound_turn(&self) -> Option<CodeTurnId> {
+        self.ours.then_some(self.expected).flatten()
+    }
+
     fn on_ran(&mut self, id: CodeTurnId) {
         if self.ours && self.expected == Some(id) {
             return;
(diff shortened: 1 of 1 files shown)


naingthet:
Review of `a55fa126a1` (P1/P2-only, non-CI): **G2G**.

After a terminal `TurnCompleted` this submit owns, `drain_checkpoint` waits up to 5s for that turn's `checkpoint_recorded` (or a `harness_notice`) so `--json` gets the diffstat and a script exiting `code run` does not race a checkpoint that is not durable yet. The wait is bound to `TurnGate::bound_turn`, so an attach/submit does not drain someone else's turn. Timeout is quiet: a missing checkpoint must not fail a turn that already succeeded.

PR #2407, opened 2026-08-21
fix(code): surface turn cache cost and dropped engine events

naingthet:
Two things Tidebreak already measures and then hides.

## `code turns` hid the fields that carry the cost

The line printed `in=` and `out=` only. On every Anthropic-routed harness the engine puts almost the whole prompt into the cache fields, so a real turn list looked like this:

```
3  completed  Give me a two-sentence summary of what you chan…  +0 -0 (0 files)  in=2 out=153
```

That turn reads as the cheapest in its session. It wrote ~12k cache tokens, and a cache write bills *above* base input — it was the most expensive per token. `cache-read=` and `cache-write=` now print when non-zero.

## Dropped engine events left no trace

`unrecognized_event_count` rises on the session, but the only record of *which* events were dropped went to `debug!` — and `tidebreak_harness` was the single workspace crate missing from `DEFAULT_DIRECTIVES`, so the crate sat at the global `warn` floor and the debug line was unreachable in a normal run.

Both had to be wrong for the count to be unexplainable, and both were. A run that dropped 34 events across two harnesses left nothing to identify them by.

The event *kind* now logs at `info`: it is what makes a gap findable and it carries no engine payload. The truncated body stays at `debug` for whoever is actually chasing one. Applied to all four adapters.

## The guard

`every_workspace_crate_has_a_log_directive` walks `crates/` and asserts each one appears in `DEFAULT_DIRECTIVES`. `tidebreak_harness` was the only crate missing; with it added, all 13 are covered. This is the omission that made the debug line unreachable, so it is worth a test rather than a second look next time.

## Tests

`cargo test -p tidebreak-server --lib logging` — 5 passed. `cargo test -p tidebreak-harness` — 183 passed. `cargo test -p tidebreak-cli` — 91 + 2 passed. Clippy clean on all three.

`child::unix_process_tree_tests::*` is flaky on this machine and fails the same way on a clean `main` — unrelated.

### DIFF
--- a/crates/tidebreak-cli/src/code.rs
+++ b/crates/tidebreak-cli/src/code.rs
@@ -1302,7 +1302,23 @@ fn print_turn_line(turn: &CodeTurnSnapshot) {
     let usage = turn
         .usage
         .as_ref()
-        .map(|usage| format!("  in={} out={}", usage.input_tokens, usage.output_tokens))
+        .map(|usage| {
+            // Cache reads and writes are most of the prompt on every
+            // Anthropic-routed harness, and a cache write bills above base
+            // input. Printing only `in=` made the most expensive turn in a
+            // session read as the cheapest.
+            let mut line = format!("  in={} out={}", usage.input_tokens, usage.output_tokens);
+            if usage.cache_read_input_tokens > 0 {
+                line.push_str(&format!("  cache-read={}", usage.cache_read_input_tokens));
+            }
+            if usage.cache_creation_input_tokens > 0 {
+                line.push_str(&format!(
+                    "  cache-write={}",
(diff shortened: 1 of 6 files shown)


naingthet:
Review of `bc79b2df66` (P1/P2-only, non-CI): **G2G**.

`code turns` now prints `cache-read` and `cache-write` when they are non-zero, so an Anthropic-routed turn that wrote ~12k cache tokens no longer reads as the cheapest in the session. Unrecognized engine event *kinds* log at info; the truncated payload stays at debug. `tidebreak_harness` is on the default log directive list, with a test that every workspace crate has one.

PR #2408, opened 2026-08-21
fix(desktop): make a failed boot recoverable

naingthet:
## What

The boot screen rendered the caught error and nothing else. This is what a
reader saw this morning after the remote machine briefly stopped answering:

> **Tidebreak**
>
> TypeError: Load failed

Attached to a remote machine, that is a locked door. The address lives in the
shell, the only command that forgets it (`disconnect_remote_machine`) is behind
the Machine settings panel, and that panel is behind the very client that just
failed to reach the machine. Recovery meant quitting the app and editing
`remote-machine.json` by hand.

The screen now carries what the reader needs to act:

- **Which machine.** `Could not reach https://…` when attached remotely, and a
  sentence naming the step otherwise. The raw error stays underneath.
- **Try again.** Boot was a mount-once effect, so a reader who reconnected the
  VPN or woke the other machine had no way to say so.
- **Work on this computer.** The same command the Machine panel runs, offered
  where a reader who cannot open that panel can still reach it.
- **Copy debug info.** A JSON report — app version, stage, attachment, error.

## What is not here

The underlying network failure is not reproducible: the machine and the gateway
both answer fine from this laptop now (`/models` → 401, CORS echoes
`tauri://localhost`), so the outage looks transient. This PR is about the part
that was not transient — a transient outage should not cost you the app.

## Credentials

`ServerInfo` carries the bearer this window would have used, and the point of a
copy control is that the payload is safe to paste in public. `bootDebugReport`
is built from named fields rather than by spreading an existing object, and a
test asserts the serialized output carries neither the token nor an
`authorization` key.

## Tests

- `BootFailure.dom.test.tsx` — headline, the recoveries each state offers, the
  copy, and the report's shape including redaction.
- `AppShell.dom.test.tsx` — an unreachable remote machine detaches into a
  working local shell; a connect failure retries and clears.
- `stories/BootFailure.stories.tsx` — the five states, per `CLAUDE.md`.

Full UI suite green locally (210 files, 1551 tests); `tsc --noEmit` and
`storybook:build` clean.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/AppShell.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/AppShell.dom.test.tsx
@@ -123,6 +123,10 @@ const unmanaged: import("./api").ManagedPolicy = {
 allow_local_mcp_servers: false,
 };
 const getPolicy = vi.fn(async () => unmanaged);
+/** The catalog fetch — the boot step a failing remote machine takes down. */
+const listModels = vi.hoisted(() =>
+  vi.fn(async () => ({ models: [] as unknown[], roles: [] as unknown[] })),
+);
 const getGatewayStatus = vi.fn(async () => ({
   base_url: "https://gateway.example",
   signed_in: false,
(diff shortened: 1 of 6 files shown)


naingthet:
Review of `580bab05f1` (P1/P2-only, non-CI): **G2G**.

A failed boot now names the attached machine, offers Try again (bumps `bootAttempt` and drops the stale client), and — when remote — Work on this computer, which is the same `disconnectRemoteMachine` Settings could not reach. The debug payload is built from named fields (stage, address, error name/message) so the bearer on `ServerInfo` cannot ride a spread onto the clipboard.

Issue #2400, opened 2026-08-21
feat(browser): expose wait and screenshot through the session bridge


naingthet:
Parent: #2342
Depends on: the initial list/navigate/snapshot bridge stack and the native observation path in #2362.

## Outcome

Extend the session-scoped agent browser bridge with bounded wait and screenshot operations, completing the second staged capability boundary from #2342 without advertising semantic actions.

## Deliverable

- Add `wait` and `screenshot` to the desktop `BrowserRuntime` contract and authenticated `/code/browser/*` routes, backed by the native driver's existing observation paths.
- Add `tidebreak browser wait|screenshot --json` and `browser_wait` / `browser_screenshot` to `browser-mcp`.
- Make Claude, Codex, and OpenCode discover the same five browser MCP tools; extend Grok's deterministic absolute-CLI fallback with the same operations.
- Apply the existing session capability, workspace, browser ownership, origin grant, visibility, Stop, instance, document-epoch, and stale-target fences.
- Keep capfile tokens and paths out of argv, provider config, logs, prompts, approval previews, and tool output.
- Reuse the existing bounded image-result path for screenshots and preserve the untrusted-page-content boundary.
- Keep `semantic_actions: false`; do not expose click, fill, select, key, scroll, or any generic act operation in this slice.

## Acceptance criteria

- MCP advertises exactly five browser tools: list, navigate, snapshot, wait, and screenshot; the CLI exposes the same five operations.
- Wait is timeout-bounded and supports only native-driver predicates with deterministic validation and typed failures.
- Screenshot is bound to the authorized browser instance and document generation; a stale or replaced document cannot be returned as current.
- Expired, copied, cross-workspace, hidden, stopped, stale-instance, stale-document, and ended-session authority fails closed before engine dispatch or result publication.
- The dependency-free E2E fixture covers successful wait/screenshot and the negative authorization/staleness matrix through the packaged absolute executable.
- No semantic action tool is advertised.

naingthet:
Status update — August 21, 2026

This follow-up is now unblocked: the native wait/screenshot observation path merged in #2362, and the initial session-scoped list/navigate/snapshot bridge plus all provider adapters merged through #2396.

The next slice remains exactly the five-tool boundary documented here: add wait and screenshot to routes, CLI/MCP, providers, and the deterministic E2E fixture. Keep `semantic_actions: false`; do not include click, fill, select, key, scroll, or a generic act tool.

naingthet:
[...]

PR #2409, opened 2026-08-21
feat(browser): bridge wait and screenshot into code sessions

naingthet:
## Summary

- add authenticated `wait` and `screenshot` operations to the server-owned browser runtime and `/code/browser/*` routes
- preserve the issuing snapshot across observation continuations while rechecking the same live capability, controller, origin grant, visibility, Stop state, instance, and document epoch
- retain typed authorization and stale-target failures for screenshot races instead of collapsing them to HTTP 500
- cover the route contract and native registry fences without exposing semantic action tools

Refs #2400

## Validation

- `git diff --check`
- Rust formatting, compilation, Clippy, and tests intentionally delegated to CI per the no-local-Rust-build instruction

### DIFF
--- a/crates/tidebreak-desktop/src/browser_control.rs
+++ b/crates/tidebreak-desktop/src/browser_control.rs
@@ -858,6 +858,63 @@ impl BrowserRegistry {
         Ok(record.snapshot(browser_id, agent_access_for_record(&state, record)))
     }
 
+    /// Re-check live authorization for observation without clearing the
+    /// stored semantic snapshot.
+    ///
+    /// Unlike [`begin_agent_control`](Self::begin_agent_control), this
+    /// preserves `semantic_snapshot` and `screenshot_epoch` so a wait or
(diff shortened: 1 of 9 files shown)


naingthet:
Review of `c0dcdaa0b1` (P1/P2-only, non-CI): **G2G**.

`/code/browser/wait` and `/screenshot` go through the same `authorize` as snapshot. Wait and screenshot take the caller's last-known `snapshot_id` and `document_epoch` and recheck live capability, controller, origin grant, visibility, Stop, instance, and document epoch before polling or capturing. Typed auth and stale-target failures stay 401/403/409 rather than collapsing to 500.

naingthet:
Incremental re-review of `ba82b30d07` (P1/P2-only, non-CI): **G2G**.

Product wait/screenshot path is unchanged from the previous G2G: same authorize, snapshot/epoch recheck, and typed 401/403/409. The new head only tightens the route tests.

Merging if required checks are green.

PR #2410, opened 2026-08-21
feat(cli): expose browser wait and screenshot

naingthet:
## Summary

- expose `tidebreak browser wait` and `tidebreak browser screenshot` with required browser/snapshot/document identity and bounded options
- advertise exactly five tools from `browser-mcp`: list, navigate, snapshot, wait, and screenshot; semantic actions remain absent
- validate screenshot PNGs and attach pixels through `ImageRef`/`ImageData` without placing base64 in model text, structured tool data, or durable serialization
- give Grok the same five-operation absolute-CLI fallback while Claude, Codex, and OpenCode continue dynamic MCP discovery

Refs #2400

## Validation

- `git diff --check`
- Rust formatting, compilation, Clippy, and tests intentionally delegated to CI per the no-local-Rust-build instruction

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7021,6 +7021,7 @@ dependencies = [
  "base64 0.23.1",
  "chrono",
  "futures",
+ "image",
  "reqwest 0.12.28",
  "serde",
  "serde_json",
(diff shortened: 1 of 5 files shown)


naingthet:
Review of `0c516b8920` (P1/P2-only, non-CI): **G2G**.

`tidebreak browser wait` and `screenshot` require `--browser-id`, `--snapshot-id`, and `--document-epoch`. `browser-mcp` advertises exactly those five tools; semantic actions stay absent. Screenshot pixels attach through `ImageRef`/`ImageData` rather than base64 in model text or durable serialization.

PR #2411, opened 2026-08-21
feat(code): let a session change permission mode

naingthet:
## Plan mode had no exit

`permission_mode` was settable only at session creation. There was no route, no CLI verb, and no desktop control to change it — so the plan→act handoff dead-ended: a user approves the plan, the engine goes on refusing every edit, and the only way forward is a new session that loses the conversation.

Worse, the guidance the user gets is wrong for this surface. The harness answers with its own affordance:

> Switch the session out of plan mode (shift+tab to cycle the permission mode, or re-run without `--permission-mode plan`)

There is no TUI to press shift+tab in, and the user never typed that flag — Tidebreak composes it. In one measured run that turn spent 91,122 tokens producing a refusal, on top of 246,463 tokens of planning that a new session would strand.

## What this adds

`POST /code/sessions/{id}/mode` taking `{"permission_mode": "plan"|"ask"|"auto"|"allow"}`, and `tidebreak code session mode <id> <mode>` in front of it.

## Why it restarts the engine

The mode reaches the engine through `SessionSpec`, which is built once when the worker attaches. A running worker keeps the posture it started with no matter what the row says, so persisting alone would be a lie until something happened to relaunch it. The change stops the worker and reattaches — the same sequence `reap` already performs, minus the fence requirement.

Order matters: the outgoing worker writes its own final state as it stops, so the new mode is persisted *after* the shutdown. Writing first would let the worker clobber it.

## What it refuses

- **While a turn is running** (`409 turn_running`). A turn that began under one posture must not have it changed underneath it — that is the point of the posture.
- **A mode the engine cannot honor** (`422`), by `refuse_unhonored_mode`, the same capability rule session creation applies per decision 0038. Never approximated.
- An ended session (`409 session_ended`).

Setting the mode a session already has returns 200 without relaunching.

## Tests

`a_session_can_leave_plan_mode_and_the_engine_relaunches_under_the_new_one` drives the real route: plan → auto returns 200 with the new mode on the returned session, the change survives a re-read, the session is not left ended by the relaunch, re-setting the same mode is a no-op, and `allow` — which the scripted adapter declares unsupported — is refused 422 without moving the row.

`cargo test -p tidebreak-server --lib code::` — 185 passed. `cargo test -p tidebreak-cli` — 91 + 2 passed. Wire types current. Clippy clean.

## Not in this change

No desktop control yet — this is the API and the CLI. The UI affordance and a structured plan-accepted event (code mode has no plan event at all, where chat has `exit_plan_mode`) are follow-ups.

### DIFF
--- a/crates/tidebreak-cli/src/api/code.rs
+++ b/crates/tidebreak-cli/src/api/code.rs
@@ -430,6 +430,18 @@ impl Client {
         .await
     }
 
+    pub async fn set_session_permission_mode(
+        &self,
+        session: CodeSessionId,
+        mode: CodePermissionMode,
+    ) -> Result<CodeSessionSnapshot> {
+        self.post_json(
+            format!("{}/code/sessions/{session}/mode", self.base_url()),
+            &serde_json::json!({ "permission_mode": mode }),
+        )
+        .await
+    }
+
     pub async fn list_approvals(
         &self,
         session: Option<CodeSessionId>,
--- a/crates/tidebreak-cli/src/code.rs
+++ b/crates/tidebreak-cli/src/code.rs
@@ -50,6 +50,7 @@ usage: tidebreak code doctor [--refresh]
        tidebreak code ws archive <id> [--force]
        tidebreak code session start --ws <id> --harness <kind> [--mode plan|ask|auto|allow]
        tidebreak code session show <id>
+       tidebreak code session mode <id> plan|ask|auto|allow
        tidebreak code session reap <id>
        tidebreak code run (--session <id> | --ws <id>) [<message>]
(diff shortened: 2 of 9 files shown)


naingthet:
Review of `bef7cac1d3` (P1/P2-only, non-CI): **G2G**.

`POST /code/sessions/{id}/mode` is owner-scoped. Same mode is a no-op. A live turn and an ended session refuse. The outgoing worker is shut down before the row is saved so its final write cannot clobber the new posture, then `attach_and_spawn_worker` relaunches under it. A mode the harness cannot honor is refused here (decision 0038), not approximated at the next turn.

PR #2412, opened 2026-08-21
fix(desktop): keep a profile's credentials in one keychain item

naingthet:
## Why

macOS pins a keychain item's access to the binary that created it, so an app
update strands every credential the previous build wrote. `secret_rehome`
already repairs that — but per item, and so is the prompt. `rehome_one` reads,
deletes, stores, and reads back, and the first two hit an item the new binary
does not own yet.

This laptop's profile holds four credentials (`provider.openai.credential`,
`provider.anthropic.credential`, `provider.openai.chatgpt_oauth_v1`,
`gateway.credentials_v1`), so an update costs roughly eight approvals. The log
shows the pass running twice in two days, four credentials each time:

```
00:05:15 INFO secret_rehome: re-homed 4 stored credential(s) …
12:49:29 INFO secret_rehome: re-homed 4 stored credential(s) …
```

`CachingSecretProvider` removes *repeat* reads. It cannot remove the first read
of each distinct item, which is where the prompt is — the prompt count is the
item count, by construction. And `static_secret_keys` names eighteen fixed keys
before per-record MCP and connected-app keys, so the ceiling grows with the
feature set.

## What

One item per profile — `tidebreak.secret_bundle_v1`, a JSON object of key →
value — behind a `SecretProvider` decorator, so nothing above it learns where a
key sits. Boot wraps `Caching(Bundled(Keychain))`.

Three rules make it safe:

- **Reads fall back to a per-key item.** Migration runs at server boot, and the
  shell's remote attachment reads its token before that. Without the fallback
  the launch after an update reads as "not attached" and drops you back onto
  this computer. The fallback makes ordering stop mattering.
- **Writes merge against the store.** The desktop builds two providers over the
  one item, and `tidebreak rehome-secrets` runs without the instance lock by
  design. A write takes a process-wide lock, re-reads, merges, and confirms the
  read-back.
- **Migration stores before it removes.** The value is in the bundle and read
  back before the old item is touched, so the "removed, then could not store it
  again" state the per-item repair can reach is now unreachable.

`desktop.remote-machine.token` joins the swept set — a real stored credential
the enumeration never named, so it had never been re-homed at all.

[ADR 0056](docs/decisions/0056-one-credential-item-per-profile.md) records the
alternatives, including the ACL approach `keyring` does not expose.

## Risks, stated plainly

- **Migration is one-way.** A build older than this one sees no per-key items
  and reads every credential as absent. Relevant to anyone downgrading.
- **A lost-update window across processes.** Narrowed to the write itself, not
  closed. Writes are rare; the daemon holds the instance lock.
- **One item is one failure domain.** An undecodable bundle is every credential
  at once, so the decoder refuses rather than reporting an empty profile — which
  would look like a fresh install and then overwrite the recoverable item.

## Tests

`secret_bundle.rs` (15) asserts the property directly: three credentials in, one
stored item. Plus two providers not clobbering each other, the pre-migration
read fallback, a delete reaching the leftover item, and the undecodable-item
refusal.

`secret_rehome.rs` asserts a migrated profile re-homes *exactly one item*, by
recording the operations — an implementation that swept correctly but still
rewrote each key passes a value assertion and fails this one.

Checked by hand against the real macOS keychain on a scratch service (live
credentials untouched, scratch items cleaned up):

```
before:  gateway.credentials_v1, provider.anthropic.credential
pass 1:  moved provider.anthropic.credential into the credential bundle
         moved gateway.credentials_v1 into the credential bundle
after:   tidebreak.secret_bundle_v1        ← one item, both values intact
pass 2:  re-homed the credential bundle    ← a later update touches only this
```

`cargo fmt` clean; clippy clean on both crates; `tidebreak-core` (694+17),
`tidebreak-server` lib (1130), and `tidebreak-desktop` lib (227) green.
`tests::code_terminals::create_write_read_and_two_cursors` failed once under
parallel load and passes on its own — a PTY timing flake, unrelated.

### DIFF
--- a/crates/tidebreak-cli/src/main.rs
+++ b/crates/tidebreak-cli/src/main.rs
@@ -945,26 +945,39 @@ async fn serve() -> Result<()> {
     server.serve().await
 }
 
-/// Rewrite each stored credential so the item belongs to this binary's code
-/// signature.
+/// Rewrite this profile's stored credentials so their item belongs to this
+/// binary's code signature.
 ///
 /// macOS keeps prompting for credentials an earlier, differently signed build
 /// created — see [`tidebreak_server::secret_rehome`] for why an approval given at
 /// that prompt does not survive the next rebuild. Run this through Cargo
 /// (`cargo run -p tidebreak-cli -- rehome-secrets`) so the dev signing runner
-/// applies; each credential asks for access once more, and then stops asking.
+/// applies; access is asked for once more, and then stops being asked.
+///
+/// Credentials live in one item, so this normally rewrites exactly that one.
+/// A profile last written by a build that predates the bundle also has its
+/// leftover per-key items swept in on the way past.
 async fn rehome_secrets() -> Result<()> {
+    use tidebreak_core::BUNDLE_KEY;
     use tidebreak_server::secret_rehome::RehomeOutcome;
 
     let config = profile_config()?;
     let mut touched = 0usize;
     let mut lost = 0usize;
     for (key, outcome) in tidebreak_server::rehome_configured_secrets(&config).await? {
+        // The bundle is the item itself; every other key is a credential that
+        // used to have one of its own. Saying "re-homed" about both would tell
+        // a reader nothing about which happened.
+        let bundle = key == BUNDLE_KEY;
         match outcome {
             RehomeOutcome::Absent => {}
+            RehomeOutcome::Rehomed if bundle => {
+                touched += 1;
+                println!("tidebreak: re-homed the credential bundle");
+            }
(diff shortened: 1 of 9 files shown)


naingthet:
Review of `f997ab5106` (P1/P2-only, non-CI): **G2G**.

A profile's credentials live in one `tidebreak.secret_bundle_v1` item, so macOS prompts once per binary instead of once per key. Writes take a process-wide lock, re-read, merge, and confirm. Absorb stores into the bundle and reads back before deleting the per-key item; a failed delete leaves both and the bundle wins. Rehome is still delete-then-store of that one item.

PR #2414, opened 2026-08-21
fix(code): fence a session whose turns keep failing

naingthet:
## What happened

A grok session ran two turns fine, then failed the third:

```
Auth recovery succeeded but 4 authenticated inference requests were still
rejected (401); giving up after 3 retries.
```

Turns four and five failed identically — the fifth was a bare `Reply with the single word: ok`. Throughout, the session reported `lifecycle: "idle"` with `fence_reason: null`: indistinguishable from one waiting for work.

`session_worker.rs` already states the invariant this breaks, for the resume-lost case:

> The engine has lost this session: every later turn would fail the same way. Fence it so the user is offered a reap instead of a session that is idle and broken.

Only `HarnessError::ResumeLost` reaches that arm, and only the codex and opencode adapters raise it. `FenceReason` had no variant for a credential that died.

## The approach

Not string-matching provider errors — that is brittle and only ever covers the failures someone already saw. Count consecutive failures instead: three in a row with no success between them is a property of the *session* rather than of any one turn, whatever the cause. An expired key, a revoked token, a provider that cannot be reached, an engine that is simply broken — all read the same from outside, and all deserve the same remedy.

One failure stays ordinary: a bad prompt, a transient hiccup, a tool that blew up. Fencing on that would be worse than the problem.

`FenceReason::RepeatedTurnFailures { count, detail }` carries the last failure's message. The turn row has no error column, so the journal is the only place that message survives — the fence is what makes it reachable without replaying events.

## Desktop

`parseFenceReason` returns `null` for a reason type it does not know, and its caller drops the **whole session** on a null rather than just the badge. So the new variant is parsed there too, with a test — this would otherwise have made fenced sessions vanish from the list.

## Tests

`a_session_whose_turns_keep_failing_is_fenced_rather_than_left_idle` drives a scripted always-failing engine through three turns: after one and two the session must still be `idle` with no fence reason, and after the third it must be `fenced` with `repeated_turn_failures`, `count: 3`, and a detail carrying the `401`.

Verified the test has teeth — raising the threshold to 99 fails it with exactly the production symptom:

```
assertion `left == right` failed: three failures in a row is a property of the session
  left: String("idle")
```

`parseFenceReason` gets its own vitest case, including the malformed shapes it must still reject.

`cargo test -p tidebreak-server --lib code::` — 186 passed. Wire types regenerated for the new variant. Clippy and fmt clean.

### DIFF
--- a/crates/tidebreak-core/src/code/attention.rs
+++ b/crates/tidebreak-core/src/code/attention.rs
@@ -76,6 +76,18 @@ pub enum FenceReason {
         /// Bounded human-readable detail, as the engine reported it.
         detail: String,
     },
+    /// Consecutive turns failed without one succeeding between them.
+    ///
+    /// Whatever broke is not specific to a prompt — an expired credential,
+    /// a revoked key, an engine that cannot reach its provider — so the next
+    /// turn would fail the same way. Fencing offers a reap instead of a
+    /// session that reads idle and is not.
+    RepeatedTurnFailures {
+        /// How many turns failed in a row.
+        count: u32,
+        /// Bounded detail from the last failure, as the engine reported it.
+        detail: String,
+    },
 }
 
 /// Server-computed attention for one unit of supervised work.
--- a/crates/tidebreak-desktop/ui/src/code/parsers.test.ts
+++ b/crates/tidebreak-desktop/ui/src/code/parsers.test.ts
@@ -7,32 +7,33 @@ import {
   parseCodeCloneDefaults,
   parseCodeCloneJob,
   parseCodeCommit,
-  parseCodeEvent,
-  parseCodePush,
-  parseCodeSession,
-  parseCodeSessionList,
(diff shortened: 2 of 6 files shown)


naingthet:
Review of `d424fa2816` (P1/P2-only, non-CI): **G2G**.

After a failed turn, `repeated_failure_fence` counts consecutive `Failed` turns (including the one just closing) and fences at three with `RepeatedTurnFailures`. A success in between resets the streak. The fence detail is the latest `TurnFailed` journal message, so the session no longer reads idle while every next turn would fail the same way.

naingthet:
Incremental re-review of `1dd1f817da` (P1/P2-only, non-CI): **G2G**.

Rebased onto main. The fence is unchanged: three consecutive failed turns, streak resets on success, `RepeatedTurnFailures` keeps the resume ref.

Merging if required checks are green.

Issue #2413, opened 2026-08-21
feat(browser): expose observation tools to foreground chat


naingthet:
Parent: #2340
Depends on: #2400 for the complete five-operation runtime boundary.

## Outcome

Let foreground chat agents use the visible Tidebreak-managed browser through the same engine-neutral observation contract as code sessions, without exposing semantic actions or allowing a chat to reach a code workspace browser.

## Authority boundary

- Derive a dedicated browser workspace scope from the authoritative foreground chat id (`foreground-chat:<chat_id>`); never accept a model- or renderer-supplied workspace id.
- Keep foreground scopes disjoint from code `WorkspaceId` values and fail closed on any scope mismatch.
- Register tools only when a real desktop browser runtime and foreground browser surface are bound.
- Use the durable client-execution claim / heartbeat / resolve path; never replay a call after dispatch may have started.

## Deliverable

- Advertise exactly five validated client-executed tools: list, navigate, snapshot, wait, and screenshot.
- Add a foreground browser executor that derives scope from the persisted chat record, issues bounded native capability, and dispatches through `BrowserRegistry`.
- Reuse the existing screenshot image-result publication path; page pixels or base64 must not enter model text or durable structured data.
- Add the minimal foreground browser surface and origin-sharing flow required for the user to open, see, stop, take over, and authorize the browser associated with that chat.
- Preserve instance, visibility, origin-grant, Stop, document-epoch, and stale-snapshot fences.
- Keep `semantic_actions: false`; do not register click, fill, select, key, scroll, or a generic act tool.

## Acceptance criteria

- A foreground turn can list an explicitly chat-scoped browser, navigate it, take a semantic snapshot, wait on a bounded predicate, and return a screenshot attachment.
- A copied id, different chat, code workspace, hidden/stopped browser, expired capability, replaced instance, or stale document fails before dispatch or result publication.
- Calls recover conservatively across claim loss and desktop restart without replaying uncertain native work.
- Tool registration is absent when the desktop browser runtime/surface is unavailable.
- Focused server/client-execution/native/UI contract tests and browser stories cover loading, empty, success, failure, compact, consent, Stop, and takeover states.
- Semantic actions remain absent until #2339 can truthfully report them.

naingthet:
First slice (fail-closed server plumbing) is up at #2416.

Adds:
- Five observation tools registered as validated client tools when `foreground_browser` is `true`: list, navigate, snapshot, wait, screenshot
[...]

PR #2416, opened 2026-08-21
feat(browser): add fail-closed foreground browser tool-surface plumbing

naingthet:
Adds fail-closed server-side foreground browser tool-surface plumbing. It must be mergeable independently and must not advertise tools in production until a real foreground browser executor/surface explicitly opts in.

## What

- Reuses the canonical core browser `ToolSpec`s and argument validators from `crates/tidebreak-core/src/browser.rs`.
- Registers exactly five client-executed tools when `foreground_browser` is `true`: `browser_list`, `browser_navigate`, `browser_snapshot`, `browser_wait`, `browser_screenshot`.
- Does **not** register `browser_act` — semantic action requires native input synthesis and is gated on a separate capability check.
- Approval classes: list/snapshot/wait/screenshot are `ReadOnly`; navigate is `Sensitive` (it can cross origins and must use the existing sensitive-tool posture until native browser consent reauthorizes it).
- Defaults to `false` in every production binding. Only becomes `true` when an explicit foreground browser executor/surface opts in. The code-mode `BrowserRuntime` is not sufficient to enable this gate.

## Tests

Five focused tests proving:

1. **Absent by default** — all six `BROWSER_TOOLS` names never register when `foreground_browser` is `false`
2. **Exact five tools when enabled** — list, navigate, snapshot, wait, screenshot registered as client-executed; `browser_act` never registers
3. **Approval classes** — reads are `ReadOnly`; navigate is `Sensitive`
4. **Validators reject malformed args** — bad browser ids, non-HTTP URLs, out-of-range dimensions, short timeouts all rejected
5. **Plan mode** — includes the four observation reads, excludes navigate (Sensitive), never contains act

## Production safety

No tools are advertised until `foreground_browser` is explicitly `true`. Every existing caller passes `false`. The code-mode `BrowserRuntime` in `bind_inner` does not trigger this gate — a desktop foreground browser executor must explicitly opt in.

Refs #2413


naingthet:
Review of `3990df5c9a` (P1/P2-only, non-CI): **G2G**.

Foreground browser tools register only when `foreground_browser` is true. Production assembly does not opt in, so chat never advertises them until a desktop executor does. The five observation tools reuse core `ToolSpec`s and validators; list/snapshot/wait/screenshot are `ReadOnly`, navigate is `Sensitive` (plan mode excludes it). `browser_act` is not registered. Tests pin default-off, the five-tool set, approval classes, and malformed-argument rejection.

naingthet:
Incremental re-review of `997fde2bc1` (P1/P2-only, non-CI): **G2G**.

Production `bind_inner` now passes `foreground_browser: false` into `agent_deps_with_cancellation_acceleration`, so chat stays fail-closed until a desktop executor opts in.

naingthet:
Incremental re-review of `20afc33455` (P1/P2-only, non-CI): **G2G**.

Test formatting only. Fail-closed production gate is unchanged.

PR #2417, opened 2026-08-21
ci: run the Windows lane when a change reaches a Windows boundary

naingthet:
Follow-up to #2403.

## Problem

The Windows lane runs on a pull request only when someone applies the `windows-ci` label. Asking authors to remember that has now failed twice, and both times the break reached main:

- #2307 — CRLF mismatch and mid-turn steer races
- #2403 — a Unix-only bridge path fixture, which failed 17 tests on all three retry attempts

Both were changes to paths this lane runs tests for. Neither carried the label.

## Change

Add a `windows` scope to the change detector and let it trigger the lane on its own. The scope covers the crates and modules the lane actually runs tests for, not merely `cargo check`s:

| Path | Windows step it covers |
| --- | --- |
| `crates/tidebreak-code-execution/*`, `crates/tidebreak-harness/*` | Code mode Windows lifecycle tests |
| `crates/tidebreak-host-broker/*` | Host broker Windows tests |
| `crates/tidebreak-server/src/code/*`, `crates/tidebreak-server/src/tests/code*` | Code mode Windows lifecycle tests |
| `crates/tidebreak-server/src/desktop_schema.rs` | SQLite profile open and migrations |
| `crates/tidebreak-core/src/keychain.rs` | Credential Manager round trip |
| `crates/tidebreak-desktop/scripts/prepare-sidecar.mjs` | Prepare target-named sidecar binaries |
| `.github/workflows/ci.yml` | This file decides what every lane runs |

The label stays as the manual override for a change the scope cannot see. Main, scheduled, and manual runs keep the automatic backstop. The `windows` scope must imply the `rust` scope, asserted in the detector the same way `workspace` already is.

## Working within the trusted-policy ratchet

The `release policy` job runs the base branch's copy of `scripts/workflow-security.test.mjs` against this pull request's workflow, so a pull request cannot loosen a policy and rewrite its assertion in one commit.

The gate therefore keeps the label clause ahead of the new scope clause. Main's copy of the test asserts the label gate as a substring, and that substring survives intact:

```
github.event_name != 'pull_request' || contains(..., 'windows-ci') || needs.changes.outputs.windows == 'true'
```

This commit's own copy of the test then asserts the wider policy: the scope clause, the output plumbing, the implication guard, and every boundary path. Dropping a path from the ladder fails that test.

## Test plan

- [x] `node --test scripts/*.test.mjs` — 187 passed
- [x] Base branch's copy of `workflow-security.test.mjs` run against this workflow, as the `release policy` job does — 38 passed
- [x] Extracted the detector from the YAML and drove it against fixture file lists. Both historical breaks now trigger the lane; `sandbox_container_run/tests.rs`, `tests/lifecycle.rs`, `router/src/lib.rs`, `ui/src/App.tsx`, and `README.md` do not.
- [x] Broke the `rust` scope for a Windows path on purpose and confirmed the implication guard exits 1
- [x] This pull request edits `ci.yml`, so the Windows lane runs on it. That is the end-to-end check.

## Deliberately out of scope

`Cargo.lock` and `crates/*/Cargo.toml` do not set the scope. A native dependency bump is a genuine Windows risk, but including them would run an 18-minute lane on every Dependabot pull request. Apply the label for a risky bump instead.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -125,6 +125,7 @@ jobs:
       notices: ${{ steps.scope.outputs.notices }}
       rust: ${{ steps.scope.outputs.rust }}
       ui: ${{ steps.scope.outputs.ui }}
+      windows: ${{ steps.scope.outputs.windows }}
       workspace: ${{ steps.scope.outputs.workspace }}
     steps:
       - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
@@ -148,6 +149,7 @@ jobs:
               echo "notices=true"
               echo "rust=true"
               echo "ui=true"
+              echo "windows=true"
               echo "workspace=true"
             } >> "$GITHUB_OUTPUT"
           }
@@ -197,6 +199,7 @@ jobs:
           notices=false
           rust=false
           ui=false
+          windows=false
           workspace=false
           git diff --no-renames --name-only -z "$diff_range" \
             > "$RUNNER_TEMP/changed-files"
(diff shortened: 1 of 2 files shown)


naingthet:
Review of `7839d5c495` (P1/P2-only, non-CI): **G2G**.

The Windows lane now runs when the change detector sets `windows` — the paths that lane actually tests (`code-execution`, `harness`, `host-broker`, `server/src/code`, `tests/code*`, `desktop_schema`, `keychain`, sidecar prepare) — or when `windows-ci` is still applied. `windows=true` without `rust=true` fails the detector. The workflow-security test pins the gate and those boundaries.

PR #2418, opened 2026-08-21
test(sandbox): give the container-run hang-guards one honest bound

naingthet:
Follow-up to #2403, which found this class while sweeping for flakes.

## Problem

`sandbox_container_run/tests.rs` wraps awaited steps in a `timeout` whose failure path prints a diagnostic snapshot. Those bounds ranged from 5 to 30 seconds with no stated reason, and several sat inside a test bound of 90 or 120 seconds. A 5-second bound inside a 120-second test is not a deliberate deadline. It is an arbitrary number waiting to fire on a busy runner.

One already did. In [run 32425751130](https://github.com/brightwave-inc/tidebreak/actions/runs/32425751130), `committed_cancellation_wakes_container_and_fences_reverse_egress_before_heartbeat` failed with `provider_calls=1`. That count proves the fence held and the second provider call never started. Only the bound blew.

## Change

Name the bound once, at the 30 seconds the `wait_until` helpers in this file already use:

```rust
/// How long an awaited step may take before this suite calls it stuck.
const HANG_GUARD: Duration = Duration::from_secs(30);
```

Twelve guards now use it, across seven tests. Each verified to sit inside a test bound of 60 seconds or more, so the guard still fires first and still prints its snapshot.

`committed_cancellation_wakes_container_and_fences_reverse_egress_before_heartbeat` needed its own bound raised from 30 to 90 seconds to contain the guard. That test runs `signal_woken_config`, whose heartbeat is one hour, so its outer bound proves nothing about timing and widening it costs nothing.

## What this deliberately leaves alone

Six bounds in this file are assertions, not guards. Each runs against a 30-second heartbeat and proves the outcome arrives *without* waiting for that tick. Their messages say so:

- `cancellation_interrupts_held_provision_and_sweeps_its_tagged_side_effect` — "cancellation must not wait for provision or the 30-second heartbeat"
- `cross_process_cancellation_interrupts_held_provision_without_a_local_signal` — "the durable watcher must beat the 30-second heartbeat"
- `cross_process_cancellation_interrupts_blocked_model_resolution`
- `pre_attach_fence_keeps_polling_setup_that_holds_the_claim_lock`
- `exact_provision_admission_refuses_a_remotely_cancelled_claim`
- `failing_provision_intent_reconciles_remote_cancellation`

Raising any of those past 30 seconds would delete the assertion. All six stay at 5 seconds, and the constant's doc comment records why, so the next person to standardize these does not flatten them.

The `dial_container` retry in the Docker end-to-end test also stays: its 2-second bound is a per-attempt dial timeout inside a 60-attempt loop, not a hang-guard.

## Test plan

- [x] `cargo test -p tidebreak-server --lib sandbox_container_run --locked` — 41 passed, run five times, no variation
- [x] Checked programmatically that every `HANG_GUARD` site sits inside a test bound of 60 seconds or more, so no guard is unreachable
- [x] Confirmed by diff that none of the six heartbeat assertions changed
- [x] `cargo fmt --all --check`, `cargo clippy -p tidebreak-server --all-targets --locked`

The suite finishes in about 1.5 seconds, so none of these bounds is near firing in normal operation. That is the point: they only matter when something is stuck.

### DIFF
--- a/crates/tidebreak-server/src/sandbox_container_run/tests.rs
+++ b/crates/tidebreak-server/src/sandbox_container_run/tests.rs
@@ -1339,6 +1339,19 @@ fn signal_woken_config() -> SandboxContainerRunConfig {
     }
 }
 
+/// How long an awaited step may take before this suite calls it stuck.
+///
+/// Each assertion is proven by the outcome its step settles on. This bound
+/// only adds a diagnostic snapshot to the failure, in place of a bare
+/// elapsed-time expect. Set it clear of the slowest legitimate wake path. A
+/// bound tight enough to fire on a loaded runner reports a flake rather than
+/// a bug (#2403).
+///
+/// Not every bound in this file is a hang-guard. Where a failure message
+/// names the 30-second heartbeat, the bound proves the outcome arrives
+/// without waiting for that tick, so it must stay well under 30 seconds.
+const HANG_GUARD: Duration = Duration::from_secs(30);
+
 /// Poll `probe` until it returns `Ok`. The timeout is only a hang-guard: the
 /// proof is the observed transition, and a stuck wait prints the last snapshot
 /// instead of a bare elapsed-time expect.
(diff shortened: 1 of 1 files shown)


naingthet:
**G2G** at `7fc23142560d`.

Test-only. One `HANG_GUARD` (30s) replaces the scattered 5s/10s/20s hang-guards so a loaded runner does not report a flake as a bug (#2403). The outer cancellation-before-heartbeat case moves to 90s so the 30s inner bound can actually fire. Production container-run code is untouched. No remaining P1/P2 that CI would miss.

PR #2419, opened 2026-08-21
feat(ui): foreground-chat browser panel and scope

naingthet:
Reuse CodeBrowserTab as the shared native browser surface for foreground chats.

### What is here

- Extracts usePortalOverlayOpen into a shared lib hook, used by both CodeWorkspacePage and ChatRoute to hide the native WKWebView while app-owned portaled overlays are open.
- Adds foregroundBrowserScope(chatId) helper — returns "foreground-chat:<chat_id>" — with unit tests.
- Teach ChatRoute renderPanel to render restored {type:"browser"} panels with CodeBrowserTab using the foreground scope, overlay obscuration, and title-change callback.
- One default browser tab per chat: the opener focuses an existing browser panel if present, otherwise creates a fresh browser id, seeds the session, and opens it.
- Add Browser row to ChatStatusChip (expanded card and compact popover), visible only when a native host is available. Copy: "Browser" / "Open shared tab".
- 7 focused tests pass (3 unit + 4 DOM). Storybook builds clean.

### What is not here

No semantic actions, foreground tool registration, or client-execution dispatch. Those are separate fail-closed PRs. No Inspect-mode files were modified.

### Changed files

| file | what |
|------|-------|
| lib/usePortalOverlayOpen.ts | new: shared overlay obscuration hook |
| code/browser/foregroundBrowserScope.ts | new: canonical scope helper |
| code/browser/foregroundBrowserScope.test.ts | new: 3 unit tests |
| ChatRoute.tsx | new: browser panel rendering + openBrowser handler + overlay wiring |
| ChatStatusChip.tsx | new: Browser row (expanded + compact, native-host-gated) |
| ChatStatusChip.browser.dom.test.tsx | new: 4 DOM tests for browser row |
| code/CodeWorkspacePage.tsx | refactor: use shared usePortalOverlayOpen, remove local copy |

Refs #2413

### DIFF
--- a/crates/tidebreak-desktop/ui/src/ChatRoute.tsx
+++ b/crates/tidebreak-desktop/ui/src/ChatRoute.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useMemo, useRef, useState } from "react";
+import { lazy, Suspense, useEffect, useMemo, useRef, useState } from "react";
 import { useNavigate } from "@tanstack/react-router";
 import { toast } from "sonner";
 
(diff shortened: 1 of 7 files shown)


naingthet:
**G2G** at `94c265c2610f`.

Foreground chat gets a Browser row on the status chip that opens a native tab scoped as `foreground-chat:<chatId>`. Overlay detection moves to a shared `usePortalOverlayOpen` hook so the native webview still yields to Radix portals. Scope helper is deterministic and disjoint across chats. No remaining P1/P2 that CI would miss.

PR #2420, opened 2026-08-21
fix(code): checkpoint every terminal turn, not only completed ones

naingthet:
## The data loss

A turn that fails or is interrupted recorded no checkpoint at all.
`after_turn_completed` returned early unless the turn's status was
`Completed`, so an interrupted turn that had already rewritten three files
left those edits in the worktree with nothing pointing at them:

- the turn's own diff is empty, because it has no ref to diff against;
- a future restore skips straight past the work;
- the next turn's checkpoint absorbs those edits and attributes them to the
  wrong turn.

An audit of a driven session shows it: failed turns store `checkpoint_ref:
null` and `diffstat: null`.

## What now gets checkpointed

Every terminal turn — `Completed`, `Failed`, and `Interrupted`. A `Running`
turn still records nothing, and the `checkpoint_ref.is_some()` idempotence
guard is unchanged.

- `after_turn_completed` is now `after_turn_ended`, and its guard names the
  three terminal statuses explicitly, so a status added later does not get
  checkpointed by accident.
- The failure posture is unchanged on the new paths: a checkpoint that cannot
  be taken journals a `HarnessNotice` warning and the recorded work stands. It
  never fails the turn.
- `CheckpointRecorded` is journaled the same way for all three, so the CLI and
  the UI see one event regardless of how the turn ended.
- The synthetic commit message is now `checkpoint turn <n> (<status>)`, so
  `git log` on the hidden refs distinguishes a completed turn from an
  interrupted one. Nothing parses that message; the leading shape is unchanged.
- `session_worker` also checkpoints on the harness-error path. That branch
  returns before the old call site, so a failed turn whose stream broke was
  reaching no checkpoint even with the guard widened.

The ordinal chain still holds: a later turn resolves `ordinal - 1` and now
finds the failed turn's ref, so its own diff covers only its own edits.

Decision 0032 describes this as "every completed turn"; the shape it settles
is unchanged, so this widens the implementation rather than opening a new
record.

## Tests

Five tests in `code::checkpoint`, four of which fail on `main`:

- an interrupted turn records a ref whose tree holds its edits, persists it,
  and journals `CheckpointRecorded`;
- a failed turn does the same;
- a running turn records nothing — no ref, no event;
- a checkpoint failure on a failed turn warns and keeps the turn;
- turn 2 parents off a failed turn 1 and excludes turn 1's file from its diff.

`cargo test -p tidebreak-server --lib code::` passes (198 tests), plus
`cargo fmt --check` and clippy on the crate.

### DIFF
--- a/crates/tidebreak-server/src/code/checkpoint.rs
+++ b/crates/tidebreak-server/src/code/checkpoint.rs
@@ -1,10 +1,16 @@
 //! Per-turn worktree checkpoints via a temporary git index.
 //!
-//! A completed turn records the worktree's full state — tracked changes and
-//! untracked files — as a synthetic commit created through a temporary index
-//! file. The user's index, `HEAD`, and reflog are untouched. The commit is
-//! referenced only by a hidden ref
-//! `refs/tidebreak/checkpoints/<workspace>/<session>/<ordinal>`.
+//! A turn that ends — completed, failed, or interrupted — records the
+//! worktree's full state, tracked changes and untracked files alike, as a
+//! synthetic commit created through a temporary index file. The user's index,
+//! `HEAD`, and reflog are untouched. The commit is referenced only by a hidden
+//! ref `refs/tidebreak/checkpoints/<workspace>/<session>/<ordinal>`.
+//!
+//! A failed or interrupted turn is checkpointed for the same reason a
+//! completed one is: the engine may have rewritten files before it died, and
+//! edits outside the chain are edits the per-turn diff and any future restore
+//! cannot see. They would otherwise land in the next turn's checkpoint, under
+//! the wrong turn.
 //!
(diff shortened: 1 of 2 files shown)


naingthet:
**G2G** at `6106b97fbf12`.

The worker now checkpoints on every terminal status, including the harness-error return that used to skip the old call site. `after_turn_ended` still no-ops for `Running` and for a turn that already has a ref, checkpoint failure still warns instead of failing the turn, and the ordinal parent chain still walks the previous ref so a later turn does not absorb a failed turn's edits. Tests cover interrupted, failed, running, checkpoint-failure, and chain-off-failed. Boot recovery still closes a dead running turn as interrupted without a ref — leftover of the same class, not a regression in this change. No remaining P1/P2 that CI would miss.

PR #2421, opened 2026-08-21
feat(browser): add native inspect-mode overlay toggle to the toolbar

naingthet:
Adds a native-owned visual overlay that highlights the same semantic
interactive elements the agent snapshot path resolves. The toggle appears
in the browser toolbar only when the engine truthfully reports semantic
snapshot support and a ready page is open.

What's included:
- Rust BrowserRecord.inspect_enabled and BrowserSnapshot.inspect_enabled
- set_inspect/clear_inspect on BrowserRegistry, with automatic clear on
  page navigation and visibility change
- SetInspect/RemoveInspect variants in the code-browser command enum
- JS overlay script (inject/remove) embedding the same semantic selectors used
  by the snapshot engine, dispatched through the existing evaluate_json bridge
- TypeScript: BrowserSession.inspectEnabled, BrowserHostSnapshot.inspectEnabled,
  BrowserHostAction set_inspect/remove_inspect variants
- BrowserToolbar: ScanEye toggle button with aria-pressed, keyboard
  accessibility, compact-layout support, and backpressure-synced state
- recordRuntime syncs inspectEnabled from host snapshots
- navigation_started clears inspectEnabled in the local session
- Storybook: InspectOff and InspectOn stories
- DOM tests: toggle availability, visibility gating on engine capability,
  compact-toolbar presence

Refs #2346

### DIFF
--- a/crates/tidebreak-desktop/src/browser_control.rs
+++ b/crates/tidebreak-desktop/src/browser_control.rs
@@ -545,11 +545,7 @@ impl BrowserRegistry {
         Ok(())
     }
 
-    pub(crate) fn clear_inspect(
-        &self,
-        browser_id: &str,
-        workspace_id: &str,
-    ) -> Result<(), String> {
+    pub(crate) fn clear_inspect(&self, browser_id: &str, workspace_id: &str) -> Result<(), String> {
         let mut state = self.lock();
         let record = state
             .records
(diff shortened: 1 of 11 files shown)


naingthet:
**G2G** at `a1c23e335071`.

Inspect is a per-session flag on the native browser record, injected as a pointer-events-none overlay of interactive elements, and cleared on navigation, close, and disable. Disable also drops the cached semantic snapshot so a later snapshot is not taken against a highlighted DOM. Workspace-scoped commands; no agent tool path. No remaining P1/P2 that CI would miss.

PR #2422, opened 2026-08-21
feat(desktop): drive code mode from the keyboard

naingthet:
## What changed

Code mode gains a **Ship** group in its keymap so a branch reaches merged without the mouse: `⌘⇧↩` runs whatever the workspace header's primary control says next, and `⌘⇧P/R/W/M/O/G/A` name the individual steps (push and open, rebase, watch, merge, view on GitHub, source control, archive), with `⌘⌥↑/↓` walking the workspace rail. A chord names an intent rather than an action — `⌘⇧P` is a commit on a dirty tree, a push on an unpushed branch, and a create on a pushed one — and a chord with nothing to do says why, because a silent no-op reads the same as a key that never fired.

The second commit takes pull-request state changes off the agent path. Marking a draft ready and merging both publish work to other people, and [decision 42](docs/decisions/0042-user-initiated-pr-merge.md) reserves them for the user — but the header asked an agent to do both, and an agent has its own shell, so the general `gh` runner's refusal of merge argv never applied. Merging now calls `POST /code/workspaces/{id}/pr/merge`, readying calls a new `POST /code/workspaces/{id}/pr/ready`, both behind the confirmation the review sidebar shows and gated by the same `prMergeControls` table it reads; `⌘⇧M` merges a green pull request and arms auto-merge on anything GitHub can still land, naming which in the dialog. `prWorkflowPrompt` loses those two prompt bodies and takes a narrowed `PrPromptAction`, so the type system now refuses to let either be wired back onto the agent path.

## Validation

`tsc --noEmit` clean, 565 UI tests pass, and `pnpm storybook:build` succeeds. `cargo test -p tidebreak-server` passes 1152 tests; the one failure was `tests::code_terminals::create_write_read_and_two_cursors`, a PTY timing test that passes in isolation and shares no code with this change. New coverage pins the shift-disambiguated pairs (`⌘P`/`⌘⇧P`, `⌘W`/`⌘⇧W`, `⌘R`/`⌘⇧R`), the intent-to-action resolution at each workflow stage, that green merges and pending arms auto-merge through the endpoint rather than `submitCodeTurn`, that the ready path runs `gh pr ready` while its runner still refuses merge argv, and — via `@ts-expect-error` — that neither state change can be prompted. Each new assertion was checked to fail when its wiring is removed rather than passing vacuously.

## Compatibility and risk

One new endpoint; no persisted or wire-format changes and no new generated types. Both merge repoints narrow what the agent path can do rather than widening it, and the merge-only runner is untouched. Auto-merge is armed only by an explicit user chord behind a confirmation, never by the watch or any other automation, which decision 42 excludes on purpose.

## Tracking

None. #2431 was the second commit, merged into this branch rather than opened against `main` separately.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/AppShell.tsx
+++ b/crates/tidebreak-desktop/ui/src/AppShell.tsx
@@ -43,11 +43,14 @@ import {
 } from "./code/codeChrome";
 import { CodeDeliveryMonitor } from "./code/CodeDeliveryMonitor";
 import { useCodeUiStore } from "./code/CodeUiStore";
+import { stepRailWorkspace } from "./code/railNavigation";
 import {
   codeRepoIdFromPath,
   codeWorkspaceIdFromPath,
+  isCodeRoute,
   shellShortcutMode,
 } from "./code/routes";
+import type { WorkflowShortcut } from "./code/workspaceWorkflow";
 import { layoutFromSearch, searchFromLayout, type PanelSearch } from "./panel/panelUrl";
 import type { LayoutState } from "./panel/panelTypes";
 import { connectOutputs } from "./deliverables";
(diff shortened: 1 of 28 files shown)


naingthet:
**G2G** at `ae2dc37f70f0`.

The Ship chords are shift-disambiguated from `⌘P` / `⌘W` / `⌘R`, scoped to code mode, and resolved as intents against the same workflow model the header already draws. Merge leaves the agent prompt path: the header button and `⌘⇧M` both call `mergeCodePr` behind the review sidebar's confirmation and `prMergeControls`, which is the decision 42 boundary. Archive still goes through the existing confirm dialog, and the new suites pin the shift pairs, blocked chords, and the green-PR endpoint path rather than `submitCodeTurn`. No remaining P1/P2 that CI would miss.

PR #2424, opened 2026-08-21
fix(code): read the context ring from the resident prompt, not turn spend

naingthet:
## The ring was reading spend

`ContextUsageIndicator` divided the sum of `CodeUsage`'s four counts by the model's context window. Those four are the turn's **spend**, summed across every model call the engine made. A turn that ran six calls re-sent its transcript six times, so the sum is roughly six prompts while the window only ever held one.

Measured on a real lane: the turn summed to **258,738** prompt-side tokens while the widest prompt in it was **44,172** — a 5.9x over-read. The ring clamped to a red `critical` 100% while the window sat at 22% of 200k.

The fixtures show the same shape on every engine. Claude's `tool-use` turn spends 148,401 against a resident 49,603; grok's `permission-denied` spends 54,748 against a resident 18,311.

## The ring shows context, the hover shows spend

`CodeUsage` gains a fifth field, `context_tokens`: the prompt resident on the turn's final model call. It is the only honest numerator for "how full is the window", and every engine already publishes what it takes to compute it.

| Engine | Source of the resident prompt |
| --- | --- |
| claude | Last entry of `result.usage.iterations[]`, summing its three prompt-side counts. A single-call turn omits the array, so the top-level object is that call. |
| codex | `tokenUsage.last.inputTokens + last.cacheWriteInputTokens`, read from the **raw** payload — `usage_from` subtracts the cache figures back out of `input_tokens` to keep the spend counts disjoint, which leaves that field a remainder rather than a prompt. |
| opencode | The `step-finish` / `message.updated` snapshot, which is already one call. The cumulative figure lives on `session.updated`, which this parser does not read. |
| grok | The last per-call `usage` event, tracked separately. Its closing `end` payload overwrites the snapshot with the turn's cumulative total, so reading the snapshot at turn end would have reproduced the same over-read. |

On the UI side, `ContextUsageIndicator` takes an explicit prop shape instead of a `RendererTurnUsage`, so both surfaces map into it deliberately:

```ts
{ contextTokens: number | null; spend: {...}; contextWindow?: number; modelName?: string }
```

- The ring fills from `contextTokens / contextWindow`, with the same clamping and the same `normal` / `warning` / `critical` bands.
- `contextTokens` null or zero renders the unmetered presentation — an empty track, no percent, no false `critical`.
- The hover keeps the In / Out / Cache read / Cache write breakdown but labels it **Turn spend**, and states the context reading on its own line against the window.

## Chat mode is unchanged

Chat passes its existing four-field sum as `contextTokens`, so its behaviour is identical to before this PR. `RendererTurnUsage` has no per-call data, so chat carries the same latent over-read. Fixing it needs the provider paths to publish the last call's prompt; a comment at that call site and in `ChatUsageDialog` says so. Deliberately out of scope here.

## Notes

- **No schema epoch bump.** `context_tokens` is `#[serde(default)]`, so journal rows written before the field existed still deserialize. The pinned journal fixture was regenerated as a change record only; `DESKTOP_SCHEMA_EPOCH` stays at 35. The TypeScript validator accepts a row without the field and reads it as zero, with a test covering that.
- **No fixture has more than one `iterations` entry**, so the claude multi-call case is covered by a unit test over a synthetic three-call `result` alongside the fixture-backed assertions.
- **One inconsistency worth a second opinion:** `usage_from` for codex subtracts `cacheWriteInputTokens` out of `input_tokens` (treating it as inside `inputTokens`), while `context_tokens` adds it back (treating it as outside). Both cannot be right. Every captured codex payload reports `cacheWriteInputTokens: 0`, so nothing observable distinguishes them today, and this PR follows the reading that matches Anthropic's own split. Worth settling when a capture with a non-zero cache write exists.

## Checks

- `cargo test -p tidebreak-harness --lib` — 194 pass, including six new `context_tokens` tests
- `cargo test -p tidebreak-core --lib` — 710 pass
- `cargo test -p tidebreak-server --lib code` — pass
- `cargo fmt --all --check`, `cargo clippy -p tidebreak-core -p tidebreak-harness -p tidebreak-server --all-targets` — clean
- `tsc --noEmit` and the full UI suite — 1585 pass
- `pnpm storybook:build` — succeeds; new `Composer/Context ring` story covers normal, warning, critical, unmetered, and no-reading

### DIFF
--- a/crates/tidebreak-core/fixtures/code-journal-events.json
+++ b/crates/tidebreak-core/fixtures/code-journal-events.json
@@ -73,7 +73,8 @@
       "input_tokens": 11,
       "output_tokens": 22,
       "cache_read_input_tokens": 33,
-      "cache_creation_input_tokens": 44
+      "cache_creation_input_tokens": 44,
+      "context_tokens": 88
     },
     "checkpoint": {
       "checkpoint_ref": "refs/tidebreak/checkpoints/ws/1",
--- a/crates/tidebreak-core/src/code/event.rs
+++ b/crates/tidebreak-core/src/code/event.rs
@@ -144,6 +144,11 @@ pub struct Diffstat {
 /// the prompt an engine actually sent is the sum of all three. Missing fields
 /// stay zero, which is not the same as "the engine sent zero" — an engine that
 /// does not surface cache counts reports nothing rather than a real zero.
+///
+/// None of the four answers "how full is the window". Summing turn totals
+/// counts the same transcript once per model call, so a long turn reads as a
+/// multiple of the prompt that was actually resident. That reading has its own
+/// field: [`CodeUsage::context_tokens`].
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, Default, TS)]
 pub struct CodeUsage {
     /// Fresh, uncached input tokens. Excludes both cache fields.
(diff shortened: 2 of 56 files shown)


naingthet:
**G2G** at `7acba65494d9`.

Code mode fills the ring from the last model call's resident prompt (`CodeUsage.context_tokens`), not the turn's summed spend. Harness parsers for Claude, Codex, Grok, and OpenCode take the last iteration/call; fixtures and unit tests pin last-call < spend. Chat still uses summed totals and says so — that needs a per-call figure on `RendererTurnUsage`, which is out of this change. Zero occupancy is treated as unmeasured rather than an empty window. No remaining P1/P2 that CI would miss.

PR #2425, opened 2026-08-21
fix(code): settle an approval whose tool call already resolved

naingthet:
## The bug

An approval parked on a tool call stayed `pending` forever once the engine resolved that call on its own. Claude Code times a permission prompt out after 60 seconds — that timer is the engine's, not Tidebreak's — fails the tool call, and finishes the turn. Both `code_approval` rows from the driven session that surfaced this are still `state: pending`, `decided_at: NULL`, while their tool calls read `failed` and the turn reads `completed`.

Two harms:

1. **`code approvals --session <id>` lists requests nobody can act on.** The session's attention reads `done_unreviewed`, so nothing surfaces them either.
2. **`code approve <stale-id>` reports success for a command that never ran.** `approval_bridge.rs` accepts a decision when nothing is parked, deliberately, so a decision arriving after a server restart still lands. But the same path let a decision on a dead request flip the row to `approved` and journal an `ApprovalResolved` for a command the engine had already failed and would never run. The user was told they approved something that did not happen. The added test confirms this: without the fix, the decision route answers `200 OK`.

## The fix

Approvals are parked on the engine's `tool_use_id`, and `ToolCompleted` carries the same id — verified through `event_from_prompt_request` (which copies `PermissionPromptRequest.tool_use_id` into `HarnessApprovalRef::call_id`) and the Claude parser, which reads `tool_use_id` off the `tool_result` block. That is the join.

- `CodeApprovalState` gains `Abandoned`: the tool call resolved before anyone decided. Added to the enum, the DB check constraint, and the generated wire types. `DESKTOP_SCHEMA_EPOCH` bumps so disposable local databases rebuild, matching how every earlier code-mode baseline edit landed.
- New `code/approval_sweep.rs` marks a pending approval `Abandoned`, sets `decided_at`, and journals the resolution. It runs from three places: on `ToolCompleted` for the matching call, on every event that closes a turn, and at session end. The turn-end hook sits on the journal path rather than in `drive_turn`, so every route that closes a turn — the sink, the worker's own close, the failure arm — sweeps once.
- Deciding a settled approval now fails with `approval_not_pending` and the message "approval `<id>` is no longer awaiting a decision: it is abandoned", instead of succeeding.

**The restart path is unchanged.** An approval that is still `pending` with no parked waiter is still decidable — that is exactly what the bridge's no-waiter branch is for, and `pending_approval_survives_restart_and_is_decidable` still covers it. The distinction is the row's state, not whether a waiter is parked. Recovery settles an open turn through `append_event` rather than the journal path, so a restart deliberately does not sweep.

## Wire contract

`ApprovalDecisionKind` gains a third variant, `Abandoned`, rather than the journal gaining a new event.

`ApprovalResolved` is the one event that says "this approval reached a terminal state". Every consumer already listens to it to stop showing the request, so reusing it means an existing reader does the right thing with no change. A separate `ApprovalAbandoned` event would let any consumer that only handles `ApprovalResolved` keep rendering the stale pending card this PR exists to remove — it would reintroduce the bug in every un-updated reader. The variant is not a decision and its doc comment says so.

## Attention

`done_unreviewed` stays right. `compute_attention` filters on `pending`, so an abandoned row no longer pins the session to "an approval is waiting", and the turn's own verdict is the honest one.

The user still learns the request went undecided, through three surfaces:

- The journal carries `ApprovalResolved { decision: abandoned }`.
- `code run` prints `approval <id> went undecided; the engine stopped waiting` — previously it printed the "decide with:" prompt and then went silent forever.
- The desktop card drops its buttons, badges "Not decided", and says a decision can no longer reach it.

## Tests

Four new, three of which fail before the change:

| Test | Without the fix |
| --- | --- |
| `an_approval_is_abandoned_when_its_tool_call_resolves_undecided` | row reads `pending` |
| `deciding_an_abandoned_approval_is_refused` | decision returns `200 OK` |
| `a_turn_that_ends_without_a_tool_result_abandons_its_approval` | row reads `pending` |
| `marks an undecided request abandoned rather than denied` (reducer) | the card would read "denied" |

`ScriptedAdapter::with_unattended_approvals` models an engine that asks and then stops waiting, which is what the 60-second timeout does.

The restart case is already covered by `pending_approval_survives_restart_and_is_decidable`, which still passes; a duplicate would not change what we do.

## Checks

`cargo fmt --all --check`, clippy on `tidebreak-core`, `tidebreak-server`, and `tidebreak-cli` with `-D warnings`, the server code-mode suite, `tidebreak-core --lib` (710), the CLI suite, UI typecheck, focused UI tests, and `storybook:build`. Wire types regenerated through `UPDATE_WIRE_TYPES=1 cargo test -p tidebreak-server`, not hand-edited.

`tests::code_terminals::create_write_read_and_two_cursors` fails under parallel load on this machine, and fails the same way on a clean `main` — pre-existing and unrelated.

## Known gap

An abandoned approval's MCP request stays parked in `ApprovalBridge`, because the sink has no handle on the bridge. That leak predates this change and costs one map entry per timed-out approval; unparking it needs the completer plumbed into `LiveSink`, which is more surgery than this fix warrants.

### DIFF
--- a/crates/tidebreak-cli/src/code.rs
+++ b/crates/tidebreak-cli/src/code.rs
@@ -13,9 +13,9 @@ use std::str::FromStr;
 
 use futures::StreamExt as _;
 use tidebreak_core::{
-    AgentError, Attention, AttentionState, CapLevel, CodeApprovalId, CodeApprovalKind, CodeEvent,
-    CodePermissionMode, CodeSessionId, CodeSessionLifecycle, CodeTurnId, HarnessCaps, HarnessKind,
-    RepoId, Result, WorkspaceId,
+    AgentError, ApprovalDecisionKind, Attention, AttentionState, CapLevel, CodeApprovalId,
+    CodeApprovalKind, CodeEvent, CodePermissionMode, CodeSessionId, CodeSessionLifecycle,
+    CodeTurnId, HarnessCaps, HarnessKind, RepoId, Result, WorkspaceId,
 };
 use tokio_tungstenite::tungstenite::Message;
 
@@ -1262,6 +1262,18 @@ fn render_event(event: &CodeEvent, dangling: bool, streamed_text: &mut bool) -> 
         CodeEvent::TurnInterrupted => {
             finish_line(dangling);
             eprintln!("tidebreak: turn interrupted");
+            return false;
+        }
+        // `code run` already told the user to decide this one. Say when the
+        // window closes, or the prompt is the last thing they ever hear.
+        CodeEvent::ApprovalResolved {
+            approval_id,
+            decision: ApprovalDecisionKind::Abandoned,
+        } => {
+            finish_line(dangling);
+            eprintln!(
+                "tidebreak: approval {approval_id} went undecided; the engine stopped waiting"
+            );
             return false;
         }
         CodeEvent::TurnCompleted { .. }
--- a/crates/tidebreak-core/src/code/event.rs
+++ b/crates/tidebreak-core/src/code/event.rs
@@ -192,7 +192,7 @@ pub enum HarnessNoticeLevel {
     Error,
 }
 
-/// Decision recorded on [`CodeEvent::ApprovalResolved`].
+/// Outcome recorded on [`CodeEvent::ApprovalResolved`].
 #[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize, TS)]
 #[serde(tag = "type", rename_all = "snake_case")]
 pub enum ApprovalDecisionKind {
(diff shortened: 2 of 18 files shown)


naingthet:
**G2G** at `74a839ce87eb`.

When the engine times out or drops a parked tool call, the approval row no longer stays `pending` forever. Sweep on tool completion (join via `call_id`), turn end, and session end; journal `ApprovalResolved { Abandoned }` so every existing listener stops showing a live card. `decide_approval` refuses non-pending rows (`approval_not_pending`) so `code approve` cannot report success for a command the engine already failed. Epoch 36 covers the new check constraint. Tests cover timeout, dropped-call, and UI abandoned copy. No remaining P1/P2 that CI would miss.

PR #2427, opened 2026-08-21
test(core): fail a baseline edit that ships without an epoch bump

naingthet:
## Why

The baseline *is* the desktop schema, and nothing noticed when it moved.

`Migrator::up` records one migration name (`m20260814_000001_baseline`), so SeaORM cannot tell that the statements behind that name changed. The reset that repairs an existing database is keyed on `DESKTOP_SCHEMA_EPOCH` — an integer in `tidebreak-server` that a contributor editing `tidebreak-core` has to remember to bump.

[Decision record 2](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0002-pre-v1-schema-and-persisted-format-mutability.md) named this gap in its own validation section:

> Similarly, no test fails if a baseline edit ships without an epoch bump. Reviewers of a change touching `db/migration/baseline/` should confirm `DESKTOP_SCHEMA_EPOCH` moved in the same diff.

This is that test. It retires the review instruction.

The record also named it as a trigger to revisit: *"or earlier if ... a table-level equivalent of the shape fixture is added, which would let the same reasoning cover the tables."*

## What it does

`the_schema_baseline_is_pinned` renders every baseline table, index, and seed row to SQLite DDL and compares against `crates/tidebreak-core/fixtures/schema-baseline.sql`.

It follows the two journal fixtures it sits beside — a change *detector*, not a compatibility contract — and its failure message names the fix:

```
the schema baseline changed; if this is deliberate, bump DESKTOP_SCHEMA_EPOCH
in tidebreak-server so existing databases are discarded, then regenerate with
UPDATE_SCHEMA_FIXTURE=1 cargo test -p tidebreak-core
```

The fixture renders one column per line, so a one-column edit reviews as a one-line diff rather than a rewritten table:

```sql
CREATE TABLE "chat" (
    "id" uuid_text NOT NULL PRIMARY KEY,
    "project_id" uuid_text,
    ...
    FOREIGN KEY ("project_id") REFERENCES "project" ("id") ON DELETE RESTRICT,
    CHECK ("attachment_revision" >= 0)
);
```

## Limits, stated

- **One backend.** It renders SQLite, because the epoch it points at guards the desktop profile. A Postgres-only difference is not something this fixture claims to see.
- **It cannot make the decision.** Regenerating with `UPDATE_SCHEMA_FIXTURE=1` and leaving the epoch alone is green. The fixture makes the change visible; a human still owes it a bump.

Both are written into the record rather than left implicit.

## Docs

- Decision record 2: consequences and validation updated — the gap it accepted is closed, the review instruction is retired, and the remaining Postgres limit is stated.
- `docs/how-tidebreak-works.md`: the epoch paragraph now says the test fails until you bump.
- `docs/releases.md`: 1.0 checklist item 2 now flips *both* fixtures' messages at v1, not just the journal's.

## Verification

Injected a duplicate column into the `chat` baseline:

```
test db::migration::tests::the_schema_baseline_is_pinned ... FAILED
assertion `left == right` failed: the schema baseline changed; if this is
deliberate, bump DESKTOP_SCHEMA_EPOCH in tidebreak-server ...
```

Reverted, and all four `db::migration` tests pass.

### DIFF
--- a/crates/tidebreak-core/fixtures/schema-baseline.sql
+++ b/crates/tidebreak-core/fixtures/schema-baseline.sql
@@ -1,0 +1,1453 @@
+CREATE TABLE "project" (
+    "id" uuid_text NOT NULL PRIMARY KEY,
+    "title" text,
+    "attachment_revision" integer NOT NULL DEFAULT 0,
+    "created_at" timestamp_with_timezone_text NOT NULL,
+    "owner" text NOT NULL DEFAULT 'local',
+    CHECK ("attachment_revision" >= 0),
+    CHECK ("attachment_revision" <= 9007199254740991)
+);
+
+
+CREATE TABLE "setting" (
+    "key" text NOT NULL PRIMARY KEY,
+    "value_json" jsonb_text NOT NULL
+);
+
+
+CREATE TABLE "chat" (
+    "id" uuid_text NOT NULL PRIMARY KEY,
+    "project_id" uuid_text,
+    "title" text,
+    "attachment_revision" integer NOT NULL DEFAULT 0,
+    "created_at" timestamp_with_timezone_text NOT NULL,
+    "model" text,
+    "reasoning_effort" text,
+    "permission_mode" text,
+    "network_policy" text NOT NULL DEFAULT '{"mode":"off"}',
+    "owner" text NOT NULL DEFAULT 'local',
+    FOREIGN KEY ("project_id") REFERENCES "project" ("id") ON DELETE RESTRICT,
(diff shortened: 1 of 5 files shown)


naingthet:
**G2G** at `50a07fe68dbd`.

Pins the rendered SQLite baseline so a table/index/seed edit without `DESKTOP_SCHEMA_EPOCH` is a red test instead of a review-memory item. The remaining hole (regenerate the fixture and skip the bump) is named in the decision record. Docs and the v1 inversion note match. No remaining P1/P2 that CI would miss.

PR #2426, opened 2026-08-21
feat(browser): execute foreground browser tools natively

naingthet:
## Summary

- derive an isolated `foreground-chat:<chat_id>` browser scope from persisted chat authority
- claim, heartbeat, dispatch, resolve, and conservatively recover the five foreground browser tools through the desktop executor
- publish screenshots as chat image attachments without storing pixels or base64 in durable tool results
- advertise the tools only when the desktop executor and browser surface are installed

## Stack

- depends on #2416
- the first parent commit is #2416 head `219ff17b08f360fd971c391cf32d1f9ff0716d66`
- keep this pull request draft until #2416 merges; then merge `main` into this branch, confirm the diff, and mark it ready

Part of #2413.

## Checks

- `git diff 219ff17b0..9cb70a59e --check`
- exact five-tool discovery; no `browser_act` dispatch
- server binding arity checks
- parent and `main` ancestry checks
- no UI files changed
- base64 appears only in ephemeral screenshot decoding and its fixture

Rust builds and tests were not run locally. CI owns Rust validation.

PR #2432, opened 2026-08-21
perf(code): hold one claude child per session

naingthet:
## What

`claude` now runs as **one child per session** instead of one per turn.
`--input-format stream-json` is always passed, stdin stays open, and each turn
is one user line written into a process that is already warm. The turn ends on
the stream's own `result` line.

Decision record [`0057`](docs/decisions/0057-one-claude-child-per-session.md)
carries the reasoning, the alternatives, and the protocol facts this rests on.

## Why, measured

The old model paid process boot on every turn and re-derived the engine's
environment block each time, which rewrote the tail of the cached prompt prefix
whenever the worktree had changed.

I could not measure this end to end against the running dev profile: an "after"
number there would need the desktop app rebuilt and restarted, which this work
deliberately did not do. Instead I drove `claude` 2.1.238 directly, once per
turn and once on one child, with the same four prompts in matched scratch
repositories. Two runs, and the second reproduced the first.

Turn 2 writes a file. Turn 3 is the one that shows the cache cost.

**Time from sending a prompt to the engine's `system/init` — the gap the
journal shows as `turn_started` → `session_started`:**

| turn | per-turn child | one child |
| ---- | -------------- | --------- |
| 1    | 1.601 / 1.264 s | 0.687 / 0.541 s |
| 2    | 0.783 / 0.611 s | 0.032 / 0.026 s |
| 3    | 0.722 / 0.591 s | 0.050 / 0.022 s |
| 4    | 0.601 / 0.622 s | 0.017 / 0.029 s |

**Time from sending a prompt to the first `assistant_delta`:**

| turn | per-turn child | one child |
| ---- | -------------- | --------- |
| 1    | 3.170 / 4.041 s | 2.197 / 2.335 s |
| 2    | 5.142 / 7.059 s | 4.173 / 4.332 s |
| 3    | 3.425 / 3.116 s | 1.505 / 1.903 s |
| 4    | 1.931 / 2.689 s | 1.336 / 1.416 s |

**`cache_creation_input_tokens` on the turn *after* the worktree-changing
turn:**

| turn | per-turn child | one child |
| ---- | -------------- | --------- |
| 3    | **7,869 / 7,867** | **29 / 28** |
| 4    | 52 / 23        | 14 / 19   |

That is the claim in the issue, reproduced: with a persistent child the
follow-up turn reads the prefix back instead of rewriting it.

These absolute latencies are smaller than the 2.855 / 2.165 / 2.714 s reported
on a 1600-file repository, because a 2-file scratch repo gives the engine
almost no directory snapshot or git status to derive. The gap is what scales
with repository size; the shape of the improvement does not.

## Interrupt

A stop is now a `control_request` on stdin, not a signal. Verified live: the
engine answers `control_response` success, writes
`[Request interrupted by user]`, and ends the turn with
`result terminal_reason=aborted_streaming`, which the parser already maps to
`TurnInterrupted`. The child stays up and runs the next turn, so a stop no
longer costs the session its warm process — and the turn now ends on the
engine's own terminal event rather than on a signal, which is strictly more
information than before.

A **second** stop for the same running turn stops the process instead of
waiting on an engine that is not answering. That is not session-ending: the
next turn respawns and resumes. There was no second-interrupt escalation
before, so nothing regressed; this adds the safety valve the old kill-based
path gave for free.

A stop that arrives **between** turns never takes the process. The old adapter
had no child at all there, and a session-long child must not be worse to stop
into.

## Child death

`run_turn` checks whether the child is still running before writing, and a
write that fails on a reused child is retried once on a replacement. Either way
the replacement launches with `--resume <session id>`, so the user's message
lands on the same transcript instead of fencing the session. `child_pid`, the
pid watch, and `spawn_epoch` are untouched; existing recovery holds.

A child that dies *mid*-turn still reports its exit status and stderr as an
incomplete turn, exactly as before — that test is unchanged and still passes.

## Behaviour change to note

`session_started` now fires **once per session** instead of once per turn. The
engine reprints `system/init` for every user message, so the parser's existing
guard is what absorbs it. Checked the consumers: the CLI folds `SessionStarted`
into a no-op branch, and the desktop reducer only records `harnessKind` /
`harnessVersion`, which is idempotent. The server still journals its own
`SessionStarted` at attach.

## What the analysis got slightly wrong

- **`system/init` is not printed until the first user message arrives.** So an
  eager spawn at session attach could not fire `session_started` at session
  open anyway. Combined with boot re-attach restoring every non-ended session
  concurrently with no cap, eager spawn would launch that many `node` processes
  at every app launch. The child is spawned lazily, on the first turn, and the
  record says what would make eager spawn worth revisiting.
- **`--model` is a launch flag**, so a turn asking for a different model retires
  the child and starts one on the requested model, resuming the session. The
  brief did not mention model switching; without this a session would silently
  ignore a model change after its first turn.
- **Approvals are genuinely unaffected**, verified rather than assumed: the
  request parks a oneshot inside the axum handler and the decision returns as
  that same HTTP response. Nothing touches the child's stdin or waits for it to
  exit. The one thing worth stating is that the per-attach approval token cannot
  outlive its child, because worker replacement calls `shutdown`, which stops
  the process.
- **`session_state_changed`** is a system subtype on 2.1.238 that the parser did
  not know, so it was inflating `unrecognized_event_count` twice per turn. It is
  now a recognized no-op.

## Follow-up, not in this PR

`mid_turn_steering` is reachable now that stdin stays open, and stays `Unknown`
for this adapter until it has been captured and fixtured.

## Tests

- New fixture `two-turn.ndjson`: one `SessionStarted` across two
  `TurnCompleted`s, and the second turn's cache creation a fraction of the
  first's. A parser reset per turn would emit two and pass everything else.
- Two turns on one child: same pid answered both, process alive between them,
  exactly one user line per turn reached stdin.
- Interrupt mid-turn: turn ends `Clean` with `TurnInterrupted`, child survives,
  next turn runs on the same pid.
- Second stop: process taken, next turn succeeds on a replacement launched with
  `--resume`.
- Dead child between turns: respawned with `--resume <id>` and completes.
- Stops between turns leave the child alone, and the engine's answers to them do
  not read as the next turn ending.

`cargo test -p tidebreak-harness` passes except for two
`child::unix_process_tree_tests` cases that fail identically on `main` under
full parallel load and pass in isolation. `cargo clippy -p tidebreak-harness
--all-targets` is clean, `cargo fmt` applied, server code-mode tests run.

## Why this is a draft

Everything below the adapter is covered and green, but nothing here has been
exercised end to end against a running build. The protocol facts were verified
by driving the real `claude` 2.1.238 binary directly; the adapter tests drive
shell-script engines. Three paths are therefore reasoned about rather than
observed on a persistent child: an approval round trip, the browser MCP
channel, and an image turn. They share the same stdin and the same HTTP
endpoint as before, and the token cannot outlive its child, but that is an
argument, not a run.

This also lands a new decision record, and merging a record is its acceptance.
Both want a human pass before this goes in, so it is not on auto-merge.


naingthet:
**G2G** at `3fdb84e7f648`.

Claude Code keeps one print-mode child per session with stdin held open; a turn ends on the stream `result`, not process exit. A dead or model-changed child is retired and the next turn respawns with the session resume ref. First interrupt is a `control_request`; a second in-flight stop kills the process. Parser ignores a repeated `system/init` and the new `session_state_changed` frame. Two-turn fixture pins one SessionStarted and a follow-up that does not rewrite the cached prefix. Draft — not merging. No remaining P1/P2 that CI would miss.

naingthet:
**G2G** at `9cfcb70bd9a3`.

Rebase onto main. Unique work is unchanged: one print-mode Claude child per session, turn ends on stream `result`, first interrupt is `control_request`, two-turn fixture pins one SessionStarted and a follow-up that does not rewrite the cached prefix. Draft — not merging. No remaining P1/P2 that CI would miss.

naingthet:
**G2G** at `f7e0fb33be46`.

Merge from main. Unique work unchanged: one print-mode Claude child per session, turn ends on stream `result`, first interrupt is `control_request`. Draft — not merging. No remaining P1/P2 that CI would miss.

naingthet:
**G2G** at `aae2ab62ca53`.

Merge from main after #2426. Unique work unchanged: one print-mode Claude child per session. Draft — not merging. No remaining P1/P2 that CI would miss.

PR #2433, opened 2026-08-21
fix(code): name a tool call while it runs, not after it finishes

naingthet:
Supervision is the premise: a reader watches a tool while it runs. On the
reference-tier adapter they watched a card that named nothing. Measured
across a driven exercise, every `tool_started` on claude_code carried an
empty `cmd`/`path`/`cwd` (9/9, 8/8, 4/4, 2/2 across four sessions), and
opencode was nearly as bad (8/9, 8/8). In one session two `Bash` calls each
ran 60s before timing out, and the reader watched two unlabelled "command"
cards for the whole minute.

Before, from the journal — the call runs nameless and the arguments arrive
only with the result:

```
seq=6 tool_started   {"call_id":"toolu_01QgUb...","detail":{"cmd":"","cwd":"","kind":"command"},"name":"Bash"}
seq=7 tool_completed {"call_id":"toolu_01QgUb...","detail":{"cmd":"ls -R /Users/.../ex-a-claude && cat .../Makefile","cwd":"","kind":"command"},...}
```

After — the same pair, from the regenerated `tool-use` fixture:

```
tool_started   {"call_id":"toolu_019ykd...","name":"Read","detail":{"kind":"file_read","path":"/workspace/README.md"}}
tool_completed {"call_id":"toolu_019ykd...","outcome":"succeeded","preview":"1\tdemo\n2\t\n3\tA tiny fixture repository.\n4\t"}
```

A reader now sees `Read /workspace/README.md` or `Command run ls -R …` the
moment the card appears, for the whole time the tool runs — and the elapsed
timer ticks next to a line that says what is taking the time. The rail's
activity chip reads the same detail, and a `Task` row now names the subagent
it spawned instead of saying "Task" until the subagent finishes.
`tool_completed` no longer repeats a correction nobody needs.

## Approach

Both options in the brief were on the table: move the start to
`content_block_stop`, or keep the early start and publish a correction. The
captured streams settled it. Claude Code repeats each finished block on an
`assistant` line *just before* its `content_block_stop`, so the assembled
arguments are already on the wire one line earlier than the brief assumed —
and a second `ToolStarted` would need the UI reducer to upsert instead of
insert, because it inserts a card per `tool_started`.

So the adapter holds the call the engine opened and starts it at the first
view that carries its arguments, whichever of the two arrives first. That is
one `ToolStarted` per call with the detail already filled, no new
`CodeEvent` variant, and no reducer change. It costs the few hundred
milliseconds the arguments spend streaming — during which the tool has not
run yet.

- **claude_code**: `content_block_start` opens the call with `input: {}` and
  streams the arguments as `input_json_delta`. The parser buffers the
  fragments (capped, since a `Write` streams a whole file through that
  channel) and starts the call at `content_block_stop`, or at the `assistant`
  line if that lands first. A `tool_result` for a call still in flight starts
  it first, so no completion lands on a call the transcript never opened.
- **opencode**: different cause, same symptom. The `pending` tool part
  carries `input: {}` *and* `time: null` — the tool has not run. The parser
  used to start the call there and dedupe every later part, so the `running`
  part's assembled arguments were dropped. It now starts the call on
  `running`, which carries both the arguments and `time.start`. A call that
  resolves without a `running` part still starts on its terminal part.
  Visible in the fixture: with an approval parked, the call sits at `pending`
  until the reply, so its row now appears after the approval it was waiting
  on — the approval card already names the command.
- **codex** and **grok** are untouched. `item/started` and `tool_call`
  already carry the whole input.

## Tests

- One contract test in `lib.rs` replays every captured stream from all four
  adapters: no `tool_started` may carry a detail that names nothing, and no
  `tool_completed` may precede its start. It is the fix for claude_code and
  opencode and the pin for codex and grok.
- The `tool-use` replays for both adapters now assert the assembled subject
  on the start.
- Two claude parser unit tests: a call starts named from the streamed
  arguments alone (no `assistant` line), and a result starts a call whose
  arguments never assembled.
- Expected sequences regenerated with `UPDATE_HARNESS_FIXTURES=1 cargo test
  -p tidebreak-harness`. Codex and grok expectations are byte-identical.

`cargo test -p tidebreak-harness`, `cargo test -p tidebreak-server --lib
code::`, `cargo fmt --all --check`, and `cargo clippy -p tidebreak-harness
--all-targets` are clean. `child::unix_process_tree_tests` flakes on this
machine on `main` as well (1 in 3 runs); it does not touch the parsers.

No UI change: `CodeToolCard` already renders `toolSubject(detail, name)` on
a running row, so the filled detail shows up with no reducer or story work.

### DIFF
--- a/crates/tidebreak-harness/fixtures/claude-code/2.1.233/approval-allow.expected.json
+++ b/crates/tidebreak-harness/fixtures/claude-code/2.1.233/approval-allow.expected.json
@@ -11,18 +11,14 @@
     "name": "Write",
     "detail": {
       "kind": "file_edit",
-      "path": ""
+      "path": "/workspace/probe.txt"
     }
   },
   {
     "type": "tool_completed",
     "call_id": "toolu_01AccXU5wrjo1GDyiNaAGDXM",
     "outcome": "succeeded",
-    "preview": "File created successfully at: /workspace/probe.txt (file state is current in your context — no need to Read it back)",
-    "detail": {
-      "kind": "file_edit",
-      "path": "/workspace/probe.txt"
-    }
+    "preview": "File created successfully at: /workspace/probe.txt (file state is current in your context — no need to Read it back)"
   },
   {
     "type": "assistant_delta",
--- a/crates/tidebreak-harness/fixtures/claude-code/2.1.233/approval-deny-with-feedback.expected.json
+++ b/crates/tidebreak-harness/fixtures/claude-code/2.1.233/approval-deny-with-feedback.expected.json
@@ -11,18 +11,14 @@
     "name": "Write",
     "detail": {
       "kind": "file_edit",
-      "path": ""
+      "path": "/workspace/probe.txt"
     }
   },
   {
     "type": "tool_completed",
     "call_id": "toolu_01Pwo7nMXPQZWnBMyeTuL8ts",
     "outcome": "failed",
-    "preview": "no — use the fixtures directory instead",
-    "detail": {
-      "kind": "file_edit",
-      "path": "/workspace/probe.txt"
-    }
+    "preview": "no — use the fixtures directory instead"
   },
   {
     "type": "assistant_delta",
--- a/crates/tidebreak-harness/fixtures/claude-code/2.1.233/approval-request.expected.json
+++ b/crates/tidebreak-harness/fixtures/claude-code/2.1.233/approval-request.expected.json
@@ -11,7 +11,7 @@
     "name": "Write",
     "detail": {
       "kind": "file_edit",
-      "path": ""
+      "path": "/workspace/probe.txt"
     }
   }
 ]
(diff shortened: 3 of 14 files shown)


naingthet:
**G2G** at `6ba395e9b5e2`.

Claude Code holds `ToolStarted` until `content_block_stop` (or the assistant block) assembles arguments, so a supervising UI names the path/cmd while the tool still runs. OpenCode waits for the `running` part instead of starting on empty `pending` input. Completions no longer need to carry the naming correction. Cross-harness fixture test pins `specificity() > 0` on every start. Overflow of streamed Write JSON is capped at 16 KiB. No remaining P1/P2 that CI would miss.

naingthet:
**G2G** at `ba61afd71734`.

Rebase after the baseline fixture landed. Unique work is unchanged: Claude holds `ToolStarted` until arguments assemble; OpenCode waits for `running`; cross-harness fixture pins named starts. No remaining P1/P2 that CI would miss.

naingthet:
**G2G** at `b5e5a903e4c8`.

Merge from main after #2437/#2438. Unique work unchanged: Claude holds `ToolStarted` until arguments assemble; OpenCode waits for `running`; cross-harness fixture pins named starts. No remaining P1/P2 that CI would miss.

PR #2434, opened 2026-08-21
fix(code): diff a session's first turn from where it started

naingthet:
## The bug

A session's first turn reported a diffstat that included other sessions'
edits. `previous_checkpoint_oid` returns `None` for ordinal 1, so turn 1 diffed
`merge_base(base_ref)` against its own snapshot — the whole worktree against the
base branch. Since record 55 a workspace can hold several sessions over one
worktree, so a sibling's earlier edits were credited to this session's first
turn.

Measured in a two-session workspace: the claude_code session's turn 1 edited
exactly one file, `receipts/parser.py` (+10/-1), and recorded
`{"files": 2, "insertions": 11, "deletions": 1}` — it swallowed the codex
session's line in `receipts/__init__.py`. The codex session escaped only
because it ran first. That turn now reports
`{"files": 1, "insertions": 10, "deletions": 1}`.

## The fix

A session records a **start baseline** when it is created: the same
`snapshot_tree` a checkpoint takes, committed and pointed at by
`refs/tidebreak/checkpoints/<workspace>/<session>/0`. The user's index and
`HEAD` stay untouched, as they already did.

Ordinal 0 is the point of the numbering. `previous_checkpoint_oid` resolves
`ordinal - 1`, so turn 1 now finds the baseline the same way turn 5 finds turn
4 — one comparison changes, and nothing else in the chain does. The commit
parents follow: baseline, then one commit per turn.

The baseline stays inside the workspace-prefixed ref namespace, so
`delete_workspace_refs` and `sweep_orphaned_refs` reap it by prefix along with
the session's turns, with no new cleanup path.

Recording the baseline is best effort. It runs in `create_session_of_kind`
before the worker spawns, so no turn can beat it, and a failure warns rather
than refusing the session.

## What did not change

- **A session with no baseline.** Every session created before this change has
  none, and `merge_base(base_ref)` is still the answer for it.
- **The only session in a fresh workspace.** Its baseline and the merge base are
  the same tree.
- **`code diff --ws <id>` with no `--turn`.** Still the whole workspace against
  the base branch. Only the per-turn diff for ordinal 1 moved.
- **Turns 2 and later.** They chain off the previous turn's checkpoint exactly
  as before, including off a failed turn's checkpoint (#2420).

One behaviour does shift, and deliberately: work already sitting in the worktree
when a session starts — a sibling's edits, or a setup script's — no longer lands
in that session's turn 1. The workspace diff still shows it.

## Tests

Four tests in `code::checkpoint`, two of which fail on `main`:

- two sessions in one workspace: the later session's turn 1 counts only its own
  edit, `(1, 10, 1)` rather than the measured `(2, 11, 1)`, and
  `resolve_diff_range` agrees, so `code diff --turn` shows the same range;
- turn 1 parents off the baseline and turn 2 off turn 1;
- a lone session's turn 1 still matches the base-ref diff exactly;
- a session with no baseline falls back to the base ref and parents off `HEAD`.

`cargo test -p tidebreak-server --lib code::` passes (206 tests), as does
`--lib tests::code` (100), plus `cargo fmt --all --check` and clippy on the
crate.

### DIFF
--- a/crates/tidebreak-server/src/code/checkpoint.rs
+++ b/crates/tidebreak-server/src/code/checkpoint.rs
@@ -17,6 +17,11 @@
 //! session reaches turn 1; a workspace-keyed ref let one sibling's snapshot
 //! overwrite another's.
 //!
+//! Ordinal 0 is the session's start baseline: the worktree as it stood when
+//! that session was created. A first turn diffs against the baseline rather
+//! than against the repo's base ref, so whatever a sibling session already
+//! changed in the shared worktree stays out of this session's turn 1.
+//!
 //! Diffs are produced here, bounded in bytes and file count, with truncation
 //! marked on the payload. The renderer never runs git.
 
@@ -48,6 +53,15 @@ pub(crate) const MAX_DIFF_BYTES: usize = 256 * 1024;
 pub(crate) const MAX_DIFF_FILES: usize = 64;
 
 const REF_PREFIX: &str = "refs/tidebreak/checkpoints";
+
+/// Ordinal of a session's start baseline, one below its first turn.
+///
+/// Numbering the baseline keeps it in the same ref path as the session's
+/// checkpoints, so [`previous_checkpoint_oid`] resolves `ordinal - 1` for
+/// turn 1 the way it does for every later turn, and both
+/// [`delete_workspace_refs`] and [`sweep_orphaned_refs`] reap it by prefix
+/// along with the rest of the workspace's refs.
+const BASELINE_ORDINAL: i64 = 0;
 
 /// Byte and file-count caps for a produced diff or file list.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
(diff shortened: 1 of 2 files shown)


naingthet:
**G2G** at `fd48d7108461`.

Each session records ordinal-0 baseline at create time, so turn 1 diffs against where *this* session started rather than the repo base (which would credit sibling edits in the shared worktree). Missing baseline still falls back to `merge_base(base_ref)`. Later turns still parent off the previous checkpoint. Tests cover sibling isolation, a lone session, pre-baseline sessions, and the chain. No remaining P1/P2 that CI would miss.

naingthet:
**G2G** at `2e36ca3e78ac`.

Rebase after the baseline fixture landed. Unique work is unchanged: ordinal-0 session baseline so turn 1 diffs against this session's start, not sibling edits in the shared worktree. Missing baseline still falls back to `merge_base(base_ref)`. No remaining P1/P2 that CI would miss.

naingthet:
**G2G** at `d96bfe1053f7`.

Merge from main after #2437/#2438. Unique work unchanged: ordinal-0 session baseline so turn 1 diffs against this session's start, not sibling edits. Missing baseline still falls back to `merge_base(base_ref)`. No remaining P1/P2 that CI would miss.

PR #2435, opened 2026-08-21
fix(core): refresh schema baseline for abandoned approvals

naingthet:
## Why

#2425 deliberately added the `abandoned` approval state and bumped `DESKTOP_SCHEMA_EPOCH` from 35 to 36. #2427 landed its pinned schema fixture immediately before that merge, so `main` now fails `db::migration::tests::the_schema_baseline_is_pinned` with the old three-state constraint.

The same deterministic failure appears in the `main` CI run for `5055f5b56` and in #2426 after merging that exact `main` head.

## Change

Refresh the pinned `code_approval.state` constraint to include `abandoned`. No second epoch bump is needed because #2425 already supplied the required bump for this schema change.

## Checks

- `git diff --check`
- No Rust command ran locally; CI validates the generated fixture against the migration.


naingthet:
**G2G** at `d5f04d9a79f7`.

Regenerates `schema-baseline.sql` so the `code_approval` state check includes `abandoned`, matching the epoch-36 constraint from #2425. That is exactly what `the_schema_baseline_is_pinned` requires. No remaining P1/P2 that CI would miss.

PR #2437, opened 2026-08-21
fix(core): match the schema fixture to the abandoned approval state

naingthet:
## What

Regenerates `crates/tidebreak-core/fixtures/schema-baseline.sql` so it matches the baseline on `main`.

## Why

`main` is currently red on `the_schema_baseline_is_pinned`.

#2425 added `abandoned` to the `code_approval.state` check constraint and bumped `DESKTOP_SCHEMA_EPOCH` to 36. It merged within seconds of #2427, which introduced the fixture, so #2425 never ran against the tripwire and the fixture never learned about the new state.

This is the failure mode `CLAUDE.md` names: a conflicting PR ran nothing, so green was not evidence.

## Scope

One line. The epoch already moved in #2425, so nothing is owed here beyond the regeneration.

Verified: the test fails on `main` and passes with this change.


naingthet:
**G2G** at `bf2078fcd739`.

Same one-line baseline fixture: `code_approval.state` check includes `abandoned`. No remaining P1/P2 that CI would miss.

PR #2438, opened 2026-08-21
perf(core): keep idle SQLite polls read-only

naingthet:
## What changed

Idle turn and blob-retirement polls now check for claimable work before taking SQLite's single-writer lock, while the locked scans remain authoritative for claims and exact retries. Regression tests hold an unrelated write transaction and verify that empty polls return without waiting for it.

## Validation

- `cargo test -p tidebreak-core --lib blob_retirement_ -- --nocapture`
- `cargo test -p tidebreak-core --lib turn_claim -- --nocapture`
- `cargo clippy -p tidebreak-core --all-targets -- -D warnings`
- `cargo fmt --all --check`

## Compatibility and risk

No persisted or wire-format changes; behavior changes only when a poll has no work.

### DIFF
--- a/crates/tidebreak-core/src/db/ops/blob.rs
+++ b/crates/tidebreak-core/src/db/ops/blob.rs
@@ -229,20 +229,22 @@ pub(in crate::db) async fn claim(
             "blob retirement lease expiry must be after claim time".into(),
         ));
     }
+
+    // Idle fast path. The locked scan below is still authoritative, but a scan
+    // with no claimable retirement has nothing it can decide. Taking the no-op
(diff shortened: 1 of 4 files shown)


naingthet:
**G2G** at `8cd19f2601cd`.

Idle blob-retirement and turn-run claims check for claimable work on a read before taking SQLite's write lock. The locked scan stays authoritative when a candidate exists. A row that appears after the hint is picked up by the worker wake or the next poll, which is named in the comments. Tests pin that an empty store does not take the write lock. No remaining P1/P2 that CI would miss.

PR #2439, opened 2026-08-21
fix(code): keep a removed repository's history reachable

naingthet:
## The bug

Deleting a repository registration took its transcripts with it, and did so differently on each backend. Verified against the real schema:

```
SQLite (desktop, foreign keys off by default):
  repos left: 0
  orphaned workspaces: 1        <- history survives, unreachable forever

PostgreSQL (self-host, foreign keys enforced):
  FOREIGN KEY constraint failed (19)
  repos left: 1                 <- removal is impossible
```

`code_workspace.repo_id` has no `ON DELETE`, and nothing sets `PRAGMA foreign_keys=ON`. So on the desktop the delete succeeded and stranded the workspace, session, turn, and event rows with no reachable repository to find them through. On a self-host deployment the same delete failed against exactly the archived workspaces the path requires you to have first, which made repository removal impossible there.

## The change

Removal becomes soft. `code_repo` gains `removed_at`:

- Removal stamps it and drops the registration from `list_repos`.
- `get_repo` still resolves it, so archived workspaces and their history stay readable.
- Creating a workspace or restoring one refuses a removed repository with `repo_removed`.

## Deliberately not included

Reclaiming the bytes on disk. Removing a clone means recursively deleting a directory the user may have registered themselves, so it needs its own confirmation rather than riding on this one. That lands with the wider reclaim work.

## Schema

Baseline edit plus `DESKTOP_SCHEMA_EPOCH` 36 -> 37, per decision record 2. The pinned schema fixture is regenerated.

## Tests

Three new cases in `crates/tidebreak-core/src/db/tests/code.rs`:

- a removed repo keeps its workspace, session, and turn reachable, and drops out of the repo list
- removing twice reports no change rather than restamping
- another owner cannot remove the repository

The first is the regression test for the bug above. `cargo test -p tidebreak-server --lib tests::code::` passes (63 tests).

### DIFF
--- a/crates/tidebreak-core/fixtures/schema-baseline.sql
+++ b/crates/tidebreak-core/fixtures/schema-baseline.sql
@@ -1309,7 +1309,7 @@ CREATE TABLE "code_repo" (
     "removed_at" timestamp_with_timezone_text
 );
 
-CREATE UNIQUE INDEX "idx_code_repo_owner_root_path" ON "code_repo" ("owner", "root_path");
+CREATE UNIQUE INDEX "idx_code_repo_owner_root_path" ON "code_repo" ("owner", "root_path") WHERE "removed_at" IS NULL;
 
 CREATE TABLE "code_workspace" (
     "id" uuid_text NOT NULL PRIMARY KEY,
--- a/crates/tidebreak-core/fixtures/schema-baseline.sql
+++ b/crates/tidebreak-core/fixtures/schema-baseline.sql
@@ -1305,7 +1305,8 @@ CREATE TABLE "code_repo" (
     "setup_script" text,
     "archive_script" text,
     "quick_actions" jsonb_text NOT NULL,
-    "created_at" timestamp_with_timezone_text NOT NULL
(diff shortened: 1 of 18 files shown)


naingthet:
**NO GO** at `527d95956381`.

**P1** — `idx_code_repo_owner_root_path` is still unique on `(owner, root_path)` with no `removed_at IS NULL` predicate. Soft-removed rows keep the path, so `insert_repo` for the same checkout fails. The API error and docs tell the user to register it again. CI would miss this: the new test never re-inserts the same path after `mark_repo_removed`.

Fix: partial unique index on live rows only, or revive the existing row on re-register instead of inserting.

naingthet:
**G2G** at `ec518ec9e5c6`.

The unique index is now live rows only (`WHERE removed_at IS NULL`). `get_repo_by_root_path` ignores removed registrations, so the same checkout can be registered again with a new id while archived history stays on the old row. That closes the previous P1. No remaining P1/P2 that CI would miss.

PR #2442, opened 2026-08-21
fix(code): stop a failure-fenced session from closing its whole workspace

naingthet:
## What happens now

A sibling session fenced for **repeated turn failures** blocks every other session in the workspace with `409 workspace_fenced` until someone reaps it.

I hit this while verifying the persistent-child branch against a live profile. One session tripped a provider content refusal, failed three turns, and fenced. A second, healthy session in the same workspace then refused every turn — including a bare `Say only: ALIVE`:

```
tidebreak: request failed (409 Conflict): workspace_fenced: another session
in this workspace is fenced until it is reaped (a924ab2f-…)
```

## Why it is wrong

The workspace gate was written for fences that imply an **engine process nobody is holding a lock on** — `OrphanAlive`, `ProbeAmbiguous`, `ResumeLost`. For those, blocking is correct and record 55 says so: a process from a previous boot may be writing to the shared worktree.

`RepeatedTurnFailures` does not imply that. The engine answered every time. The failures were an expired credential, a refused prompt, a provider outage — none of which put the worktree at risk. The gate has treated it like an orphan since it was added.

## The change

`FenceReason::blocks_workspace()` names the distinction; `workspace_fence_reason` consults it. A fenced row with no recorded reason still blocks, so the conservative default holds for anything written earlier.

## Tests

- New: a sibling fenced for `RepeatedTurnFailures` does **not** close the workspace. Fails before this change with the exact 409 above.
- Existing `a_fenced_session_closes_its_whole_workspace_to_turns` uses `OrphanAlive` and stays green — that is the regression guard for the half that should still block.

`code::` 207 passed, `tidebreak-core --lib` 717 passed, fmt and clippy clean. `code_terminals::create_write_read_and_two_cursors` fails under full parallel load and passes in isolation on this branch and on clean `main` alike — pre-existing PTY timing flake.

### DIFF
--- a/crates/tidebreak-core/src/code/attention.rs
+++ b/crates/tidebreak-core/src/code/attention.rs
@@ -90,6 +90,28 @@ pub enum FenceReason {
     },
 }
 
+impl FenceReason {
+    /// Whether this fence must also stop every other session in the workspace.
+    ///
+    /// True when an engine process may be writing to the shared worktree
+    /// outside every lock this process holds — an orphan from a previous
+    /// boot, a pid the probe could not identify, or a session the engine no
+    /// longer recognizes. Nothing in the workspace may write until a reap
+    /// settles it (decision 0055).
+    ///
+    /// False for [`Self::RepeatedTurnFailures`]. That fence says the engine
+    /// answered and the turns failed — an expired credential, a refused
+    /// prompt, a provider outage. The process is accounted for and the
+    /// worktree is not at risk, so a healthy sibling session keeps working.
+    #[must_use]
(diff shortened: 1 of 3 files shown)


naingthet:
**G2G** at `ebba718aba77`.

Only fences that imply an unaccounted engine (`OrphanAlive`, `ProbeAmbiguous`, `ResumeLost`) still close the workspace. `RepeatedTurnFailures` does not: the engine answered, so a healthy sibling keeps working. A fenced row with no reason still fail-closes. Test covers the sibling turn. No remaining P1/P2 that CI would miss.

Issue #2440, opened 2026-08-21
feat(code): search conversation history across archived workspaces


naingthet:
`search_workspace` greps files in the worktree, so searchability dies with the very thing archive removes. To archive aggressively and still find past work, search has to cover the transcript.

Search `code_event` and `code_turn` text across a repository's workspaces, including archived and removed ones, and return enough to reach the session that produced a hit.

This is what makes the reclaim tiers safe to use: you can free the checkout because you can still find the conversation.

Split out of the workspace reclaim work so it can land on its own.

naingthet:
Unclaimed. The in-progress label was applied on 2026-08-25 with no branch or pull request. Search across archived workspace transcripts is still the outcome; reclaim it when work starts.

PR #2443, opened 2026-08-21
feat(code): reclaim a workspace's branch without losing the work

### LINKED ISSUE #2441: feat(code): reclaim tiers for workspaces, so archiving frees real disk
Archive already frees the worktree — the gigabytes of build output and dependencies — and keeps the branch so `restore_workspace` can rebuild the checkout. What it never frees is the branch, its git objects, and the clone.

The goal is to archive aggressively and reclaim real disk while keeping every conversation searchable and the work rebuildable.

## Tiers

| Tier | On disk | Rebuild |
| --- | --- | --- |
| Active | worktree | — |
| Archived | branch only | `git worktree add` (today) |
| Released | nothing | unbundle, then `git worktree add` |

History is never deleted by any tier.

Released bundles the branch's unique commits with `git bundle create base..branch` before dropping the branch. That is typically kilobytes against a worktree measured in gigabytes, so the tier reclaims nearly everything and stays exactly rebuildable. Store the bundle as a blob and ask before storing an unusually large one.

Each tier down confirms what it makes unrecoverable: uncommitted work, unpushed commits, and at Released, a branch that is merged nowhere. `archive_blockers` already computes the first two.

## Landed

- Released tier — #2443
- Clone reclaim for checkouts Tidebreak cloned — #2449. Provenance is recorded on `code_repo`; a user-registered directory is not deleted.

## Remaining

- A cleanup surface showing reclaimable bytes per repo and workspace, so the tiers are actionable rather than guesswork. Orca's Resource Manager is the reference.
- Environment for the hook scripts. `run_hook_script` still passes none, so an archive script cannot tell what it is archiving. Orca passes the repo root and workspace name.

Related: #2429, #2440. Schema posture is migrations (#2428).

naingthet:
Closes #2441 (first slice).

## Why

Archive frees the checkout — the gigabytes of build output and dependencies — and keeps the branch, so `restore` is `git worktree add`. What it never frees is the branch and the objects it holds alive. Running many parallel agents, that residue accumulates; the machine this was built on holds ~80 worktrees.

The measurement that shapes the design, taken on a real branch in this repo:

```
bundle (base..branch):   12K
worktree on disk:       1.1G
```

The bytes are in the checkout, not the commits. So the branch can be dropped and still restored exactly, for storage that rounds to nothing.

## What

A third reclaim tier:

| Tier | On disk | Rebuild |
| --- | --- | --- |
| Active | worktree | — |
| Archived | branch only | `git worktree add` |
| **Released** | nothing | unbundle, then `git worktree add` |

- `POST /code/workspaces/{id}/release` bundles `base..branch` into `<data_dir>/code/bundles/<id>.bundle` and drops the ref.
- The bundle is written and measured **before** the ref is dropped, so any earlier failure leaves the workspace exactly where it was.
- Release refuses a branch with commits the base lacks (`branch_unmerged`) unless forced. The bundle makes it recoverable either way — the prompt is for deliberateness, not correctness.
- `restore` fetches the branch back from the bundle, then follows the archived path unchanged. It verifies the bundle first, so a corrupt file fails with the branch still absent rather than half-populating the object store.

**No tier deletes a session, turn, or event.** The row and its journal outlive the bytes. That is what makes reclaiming safe to do routinely.

## UI

A new status is exactly the kind of thing surfaces forget. Every "is this still live?" check now goes through one `isPutAway` helper instead of comparing to `"archived"`, so released workspaces do not read as live work on the rail. Added a `Released` story and fixture.

## Deliberately not included

Clone reclaim (recursively deletes a directory the user may own — needs its own confirmation), a large-bundle size guard (wants a threshold real use should set), and transcript search across put-away workspaces (#2440). All tracked on #2441.

## Schema

Baseline edit plus `DESKTOP_SCHEMA_EPOCH` 37 → 38, per decision record 2. Fixture regenerated. `code_workspace` gains `released_at`, `released_tip`, `bundle_bytes`, and the `released` status.

Decision record: `docs/decisions/0056-workspace-reclaim-tiers.md`.

## Tests

- **Round trip** (`release_frees_the_branch_and_restore_rebuilds_it_from_the_bundle`) — end to end over HTTP: commit work, archive, release, confirm the branch is gone, restore, confirm the file is back. This is the claim; a release that bundles nothing fails here.
- **Efficiency** (`a_bundle_carries_only_the_branch_commits`) — asserts the bundle stays under 100 KB for a one-line commit on a base carrying a 400 KB blob. If a future change bundles full history, the tier stops being worth its step and this fails.
- **Ordering** — a refused release leaves the branch in place.
- **Corrupt bundle** — refused, with no branch created.
- Plus merged/unmerged detection and the git-level round trip.

`cargo test -p tidebreak-server --lib code::` passes (211), `tidebreak-core` (717), UI (531). Clippy and rustfmt clean, `tsc --noEmit` clean.

### DIFF
--- a/crates/tidebreak-core/fixtures/schema-baseline.sql
+++ b/crates/tidebreak-core/fixtures/schema-baseline.sql
@@ -1323,8 +1323,11 @@ CREATE TABLE "code_workspace" (
     "pr" jsonb_text,
     "created_at" timestamp_with_timezone_text NOT NULL,
     "archived_at" timestamp_with_timezone_text,
+    "released_at" timestamp_with_timezone_text,
+    "released_tip" text,
+    "bundle_bytes" integer,
     FOREIGN KEY ("repo_id") REFERENCES "code_repo" ("id"),
-    CHECK ("status" IN ('creating', 'setup_failed', 'active', 'archived'))
+    CHECK ("status" IN ('creating', 'setup_failed', 'active', 'archived', 'released'))
 );
 
 CREATE UNIQUE INDEX "idx_code_workspace_repo_branch" ON "code_workspace" ("repo_id", "branch_name");
--- a/crates/tidebreak-core/src/code/mod.rs
+++ b/crates/tidebreak-core/src/code/mod.rs
@@ -253,8 +253,15 @@ pub enum CodeWorkspaceStatus {
     SetupFailed,
     /// Ready for sessions.
     Active,
-    /// Archived; worktree removed.
+    /// Archived; worktree removed, branch kept.
     Archived,
+    /// Released; worktree and branch both gone, commits kept as a bundle.
+    ///
+    /// The deepest reclaim tier. A worktree is gigabytes of checkout and build
+    /// output; the branch's own commits are usually kilobytes, so bundling
+    /// them and dropping the branch frees nearly everything and still restores
+    /// exactly. The transcript is untouched at every tier.
+    Released,
 }
 
 impl CodeWorkspaceStatus {
@@ -266,6 +273,7 @@ impl CodeWorkspaceStatus {
             Self::SetupFailed => "setup_failed",
             Self::Active => "active",
             Self::Archived => "archived",
+            Self::Released => "released",
         }
     }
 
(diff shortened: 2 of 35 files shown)


naingthet:
**G2G** at `62d2cba54ca8`.

Release bundles `base..branch`, then deletes the branch (`-D`) only after the bundle is on disk. Restore verifies then fetches. Epoch 38 covers `Released` plus tip/size columns. UI treats archived and released as reclaim tiers. Delta since last look is a test-only line-ending compare. No remaining P1/P2 that CI would miss.

PR #2446, opened 2026-08-21
perf(code): stop journaling assistant deltas and bound journal replay

naingthet:
## What changed

`code_event` had no pruning and most of it said nothing. Assistant deltas
were 63–93% of the table by lane, and the text they carry is repeated byte
for byte by the `assistant_message` that follows — 38 of 38 messages across
the eight driven sessions, no mismatches.

- **Deltas are live-only.** `assistant_delta` is published on the session bus
  and never journaled. Everything else keeps the existing discipline:
  journaled first, published second, under the spawn-epoch fence, with
  `(session_id, seq)` monotonic and dense.
- **The frame says which kind it is.** `SequencedCodeEventFrame` gains
  `transient`. A transient frame carries the journal cursor it streamed
  behind, so a client that resumes from the last `seq` it saw loses nothing.
  This is the chat socket's existing shape for out-of-band frames, not a
  second streaming idiom.
- **Mid-turn reconnect stays correct.** The bus keeps the text streamed so
  far, bounded by `MAX_EVENT_TEXT_CHARS`, and retires it on any journaled
  event that ends the run. A connecting socket takes the tail and subscribes
  under one lock, and forwards it only when its replay did not read past the
  position the tail was captured at.
- **A turn that ends mid-sentence gets the message the engine owed it.**
  Synthesizing the message is safe where replaying deltas would not be: both
  the renderer and the CLI treat a message as a replacement for text they
  streamed, so a client that already has it shows it once.
- **Replay is bounded.** `list_events` takes a limit, keeps the newest events
  (`MAX_REPLAY_EVENTS = 2000`), and the first frame of a capped window carries
  `truncated`. The renderer turns that into a transcript line.
- **The per-delta `get_session` is gone.** `note_activity` reads a pessimistic
  in-memory hint that every row read corrects, so the common path costs a
  mutex rather than a query.

`reasoning_delta` is deliberately unchanged. There is no `reasoning_message`,
so those rows are the only record of reasoning there is — dropping them would
empty the thinking accordion on reload rather than deduplicate anything.

Measured on a scripted 40-delta turn: **45 journal rows before, 5 after.**

## Decision record

Dropping durable deltas diverges from record 0030's
journal-before-live-publication rule for one event kind, so this adds
`docs/decisions/0057-assistant-deltas-are-live-only.md`. The divergence is
narrow on purpose: the table shape, sequencing rule, fence, and reconnect
contract are unchanged, so a future unified journal is still a table merge.

## Validation

- `cargo test -p tidebreak-server --lib -- tests::code routes::code::session_events` (106 passed)
- `cargo test -p tidebreak-core --lib -- db::tests::code::` (19 passed)
- `cargo test -p tidebreak-cli --bins -- code` (23 passed)
- `pnpm --dir crates/tidebreak-desktop/ui exec vitest run src/code` (533 passed)
- `cargo clippy -p tidebreak-core -p tidebreak-server -p tidebreak-cli --all-targets -- -D warnings`
- `cargo fmt --all --check`

New tests: a turn journals its message and none of its deltas while the
deltas still arrive on the socket; a socket connecting mid-answer assembles
the same text as one connected throughout; text streamed before an interrupt
survives in the journal; a capped replay flags truncation on its first frame
only and keeps the newest events; an exact-fit window reports no truncation;
a delta the replay overtook is not delivered twice. The existing durable
replay test — which appends delta rows directly — covers journals written
before this change.

## Compatibility and risk

No migration. Delta rows already stored stay readable and still replay. The
wire gains two optional fields; the CLI's decoder ignores unknown fields, so
an older client reading a newer server renders deltas exactly as before.

The bus stops being a pure fan-out — it now owns per-session live state, which
is per-process. Code mode is single-process today; this is a constraint on
that changing.

### DIFF
--- a/crates/tidebreak-cli/src/api/code.rs
+++ b/crates/tidebreak-cli/src/api/code.rs
@@ -113,13 +113,24 @@ pub enum SubmitTurnResponse {
     Queued(QueuedCodeTurn),
 }
 
-/// One journaled event on the per-session WebSocket.
+/// One event on the per-session WebSocket.
 #[derive(Debug, Clone, PartialEq, Serialize, Deserialize)]
 pub struct SequencedCodeEventFrame {
+    /// Journal position, or — on a `transient` frame — the cursor the event
+    /// streamed behind. Resuming from it is correct either way.
     pub seq: i64,
     pub event: CodeEvent,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub replayed: Option<bool>,
+    /// Set on a live-only event no journal row holds: assistant deltas, and
+    /// the catch-up delta a mid-turn reader gets on connect. Render it and
+    /// move on — it will not come back on reconnect.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub transient: Option<bool>,
+    /// Set on the first replayed frame of a capped window: history older than
+    /// this frame was dropped and is not coming.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub truncated: Option<bool>,
 }
 
 /// Doctor report for every registered engine adapter.
@@ -690,6 +701,25 @@ mod tests {
         assert_eq!(encoded["event"]["type"], "assistant_delta");
         assert_eq!(encoded["event"]["text"], "hello");
         assert_eq!(encoded["replayed"], true);
+    }
+
+    #[test]
+    fn a_transient_delta_frame_keeps_the_cursor_it_streamed_behind() {
+        // Deltas are live-only (record 57). `code run --json` prints them the
+        // same as any other frame, and the `seq` they carry is the journal
+        // position to resume from — a reconnect using it loses nothing,
+        // because no row holds the delta.
+        let frame = decode_event_frame(
+            r#"{"seq":12,"event":{"type":"assistant_delta","text":"half"},"transient":true}"#,
+        )
+        .unwrap();
+        assert_eq!(frame.seq, 12);
+        assert_eq!(frame.transient, Some(true));
+        assert_eq!(frame.replayed, None);
+        assert!(matches!(
+            frame.event,
+            CodeEvent::AssistantDelta { ref text } if text == "half"
+        ));
     }
 
     #[test]
--- a/crates/tidebreak-cli/src/api/code.rs
+++ b/crates/tidebreak-cli/src/api/code.rs
@@ -123,10 +123,15 @@ pub struct SequencedCodeEventFrame {
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub replayed: Option<bool>,
     /// Set on a live-only event no journal row holds: assistant deltas, and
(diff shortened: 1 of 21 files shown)


naingthet:
**NO GO** at `51ff828029fd`.

**P1** — A reconnecting client that already applied live deltas gets the full in-memory tail again as another `assistant_delta`.

`send_live_tail` in `crates/tidebreak-server/src/routes/code/session_events.rs` fires whenever `last_seq <= tail.cursor`. Transient frames keep `seq` at that journal cursor, so both the desktop (`getAfter` → `lastSeq`) and the CLI (`CodeStream.last_seq = frame.seq`) reconnect from the same point and receive the whole streamed string.

The reducer concatenates (`assistantBuffer + event.text`). The CLI writes it. The final `assistant_message` later replaces the pane, but CLI output is already wrong, and until that message the desktop keeps appending onto doubled text.

The new tests cover a second reader with an empty buffer (`connecting_mid_answer_replays_the_text_that_already_streamed`), not the socket that just dropped. CI would miss this: no reconnect-with-applied-buffer case.

Fix: replay the tail only when the subscriber has not already seen that cursor, or send a replace-not-append frame.

naingthet:
**NO GO** at `7e186bf9827b`.

**P1** — `send_live_tail` in `crates/tidebreak-server/src/routes/code/session_events.rs` drops tokens that streamed while the socket was down.

`requested_after >= tail.cursor` skips the tail on every reconnect from the live cursor. Transient frames do not move that cursor, so desktop `getAfter` and CLI `CodeStream.last_seq` both resume there. The tail is the only copy of those bytes. The post-attach subscription does not have them.

The prior same-cursor concat is gone when the gap is empty. A disconnect that misses deltas still loses them. The desktop stays wrong until `assistant_message` replaces the buffer. The CLI sets `streamed_text` and then skips that message, so the gap is permanent.

naingthet:
**NO GO** at `cf0dfde23783`.

**P1** — `send_live_tail` in `crates/tidebreak-server/src/routes/code/session_events.rs` still drops tokens that streamed while the socket was down.

`requested_after >= tail.cursor` skips the tail on every reconnect from the live cursor. Transient frames do not move that cursor, so desktop `getAfter` (`lastSeq`) and CLI `CodeStream.last_seq` both resume there. The tail is the only copy of those bytes. The post-attach subscription does not have them.

The first P1 is fixed: a same-cursor reconnect no longer concatenates the full tail. The new test publishes `"third"` only after the resumed socket is already subscribed, so it never covers the gap. A disconnect that misses deltas still loses them. The desktop stays wrong until `assistant_message` replaces the buffer. The CLI sets `streamed_text` and then skips that message, so the gap is permanent.

naingthet:
**G2G** at `b90a0bd12a58`.

`send_live_tail` now forwards the captured tail as a `replacement` whenever replay has not read past `tail.cursor`, including a same-cursor reconnect. The desktop replaces the assistant buffer, and the CLI prints only the unprinted suffix, so already-applied deltas are not concatenated and tokens that streamed while the socket was down are not dropped.

PR #2447, opened 2026-08-21
chore(ui): add a formatter and linter for the desktop TypeScript

### LINKED ISSUE #2444: chore(ui): add a formatter and linter for the desktop TypeScript
The desktop UI has no formatter and no linter. `crates/tidebreak-desktop/ui/package.json` carries no `prettier`, `eslint`, `biome`, or `stylelint` dependency, and there is no config file for any of them. TypeScript is checked by `tsc` and tested by `vitest`; nothing enforces style or catches lint-class mistakes.

The practical cost: an agent or a contributor editing TSX matches the surrounding style by eye. That works until it does not, and it produces diffs where formatting noise hides the change. Rust already has `cargo fmt --check` and `clippy`; the UI has no equivalent.

## Scope

- Pick one tool for format and lint. Biome does both in one dependency and one config, which suits a repo that currently has neither. Prettier plus ESLint is the more conventional pair and has broader plugin coverage for React and a11y rules. Decide deliberately rather than by default.
- Format the existing `src/` in one commit, separate from any behavior change, so the reformat is reviewable as a no-op.
- Add a check lane to CI alongside the existing TypeScript job.
- Add the matching scripts to `package.json` so `pnpm --dir crates/tidebreak-desktop/ui format` and `... lint` work locally.
- Pin exact versions, per the repository's dependency rule.

## Notes

Keep the initial rule set small. A linter that lands with a hundred pre-existing violations gets suppressed rather than fixed; start with the rules that catch real defects and grow the set once it is green.

Storybook and the a11y addon are already installed, so an accessibility rule set has somewhere to land if the chosen tool supports one.

Follow-up from #2443, where hand-matched formatting was the only option available.

naingthet:
## What changed

The desktop UI now formats and lints with Biome. One pinned dependency covers both jobs, which is the right fit for a tree that had neither. Prettier plus ESLint would have added a plugin graph this first pass does not need.

`pnpm --dir crates/tidebreak-desktop/ui format` rewrites `src/`. `pnpm --dir crates/tidebreak-desktop/ui lint` reports defects. The desktop UI CI lane runs both, after install and before tests. Generated wire types stay unformatted so the server bindings test can keep byte-comparing them.

The first lint set is the defect rules that already pass: unreachable code, invalid use-before-declaration, hook call sites, missing JSX keys, `dangerouslySetInnerHTML`, debugger, and the duplicate-binding checks. Recommended, a11y, and hook-dependency rules stay off until they can land green instead of suppressed.

The second commit is a mechanical reformat of `src/` with no behavior change.

## Validation

- `pnpm --dir crates/tidebreak-desktop/ui lint`
- `pnpm --dir crates/tidebreak-desktop/ui exec biome format src`

## Compatibility and risk

None. DevDependency and CI only. Generated `src/generated/` is excluded.

## Tracking

Closes #2444


naingthet:
**G2G** at `36301c1d1e8c`.

Biome format check (no `--write`) and a small defect-only lint set. Generated `src/generated/` stays excluded so wire bindings still byte-compare. The bulk commit is wrapping/quotes/commas; JSX spaces that wrapping would drop are kept with `{" "}`. No remaining P1/P2 that CI would miss.

Issue #2428, opened 2026-08-21
feat: switch the schema posture from epoch resets to migrations


naingthet:
Pre-v1 schema changes are a baseline edit plus a `DESKTOP_SCHEMA_EPOCH` bump, and the bump deletes the local database ([decision 2](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0002-pre-v1-schema-and-persisted-format-mutability.md)). That window should close before 1.0 rather than at it.

**The current rate.** Eleven epoch bumps since 2026-08-12, eight since the baseline squash on 2026-08-14 — roughly one wipe a day. Eight of the eleven are code-mode work (#2123, #2168, #2175, #2209, #2224, #2250, #2291, #2305).

**What has to land first.** Every remaining step of [decision 48](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0048-one-interaction-model.md), in its order — #2315, #2314, #2312, then #2313. The record already says step 5 after 1.0 "would cost migrations this plan never budgets for," and #2313 merges `chat` / `message` / `event` / `turn_run` into the code-shaped tables. Doing that under migrations means rewriting every user's history in place.

Also inside the window, because both change storage layout:

- Object storage for blobs (#2322).
- Any durable rows the agent browser epic adds (#2335).

**Then the switch.** Item 2 of the [1.0 checklist](https://github.com/brightwave-inc/tidebreak/blob/main/docs/releases.md#preparing-and-shipping-100) already describes it: re-squash the baseline, invert decision 2, retire `reset_pre_v1_state`. This issue is the decision to do that part *before* 1.0 rather than as part of it.

**Two things to settle in the same work:**

1. **The self-host chain is implicit.** `Baseline::up` already branches on `if manager.has_table("app")` and calls `ensure_app_owner` / `ensure_code_owner`. That is a migration chain wearing a baseline's clothes, and it encodes ordering as "check whether the column exists," which does not compose. Make it an ordered chain rather than growing more `ensure_*` functions.
2. **The tripwire is in place but partial.** `the_schema_baseline_is_pinned` (#2427) fails on any baseline edit and names the epoch. It renders SQLite only, so a Postgres-only difference is still unguarded.

Do not squash and flip the lifecycle until #2313 is merged.

naingthet:
[...]

PR #2448, opened 2026-08-21
test(core): pin the baseline's PostgreSQL rendering too

naingthet:
Part of #2428.

## Why

`the_schema_baseline_is_pinned` (#2427) rendered SQLite only, and its own doc comment said a Postgres-only difference "is not something this fixture claims to see."

That gap is real. SQLite's type affinity collapses distinctions that are load-bearing on the durable self-host store:

```
SQLite                              PostgreSQL
"agent_run_id" uuid_text            "agent_run_id" uuid
"answer_..._ids" jsonb_text         "answer_..._ids" jsonb
"admitted_at" timestamp_..._text    "admitted_at" timestamp with time zone
"allow_free_form" boolean           "allow_free_form" bool
```

A baseline edit that changes only a Postgres type renders identically under SQLite and ships unguarded. Desktop is disposable and an epoch bump repairs it; the self-host store is neither.

## What

Render both backends into two fixtures. One helper, one loop, same failure message.

The gap closes ahead of the work in #2428, where the migration chain that replaces the epoch gets written against PostgreSQL first — exactly the backend that had no tripwire.

## Verification

Hand-edited `NOT NULL` to `NULL` on one column in the Postgres fixture; the test fails. Reverted; it passes. So the second fixture is enforced, not merely generated.

`cargo test -p tidebreak-core` passes (717). Also updated decision record 2's validation section, which described the single-backend limitation as accepted.

### DIFF
--- a/crates/tidebreak-core/fixtures/schema-baseline.postgres.sql
+++ b/crates/tidebreak-core/fixtures/schema-baseline.postgres.sql
@@ -1,0 +1,1367 @@
+CREATE TABLE "project" (
+    "id" uuid NOT NULL PRIMARY KEY,
+    "title" text,
+    "attachment_revision" bigint NOT NULL DEFAULT 0,
+    "created_at" timestamp with time zone NOT NULL,
+    "owner" text NOT NULL DEFAULT 'local',
+    CHECK ("attachment_revision" >= 0),
+    CHECK ("attachment_revision" <= 9007199254740991)
+);
+
+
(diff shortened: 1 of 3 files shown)


naingthet:
**G2G** at `12f717e99e55`.

Pins the rendered PostgreSQL baseline the same way SQLite is already pinned, so a Postgres-only table/index/seed edit without an epoch bump is a red test. Docs match. No remaining P1/P2 that CI would miss.

PR #2449, opened 2026-08-21
feat(code): reclaim only the checkouts Tidebreak cloned

naingthet:
Part of #2441.

## The problem

Removing a repository left its checkout on disk, and there was no safe way to add reclaim — because nothing recorded who made the directory.

A Tidebreak-performed clone registers through the *same* path a manual registration uses (`clone.rs` passed `RepoRegistration::default()`), so `code_repo` could not distinguish "Tidebreak cloned this" from "the user pointed at their repo."

Any reclaim built on that would have to infer provenance from the path — testing against `code_clone_parent_dir`, a setting that moves, and that a user's own repo may sit under. Inferring wrong means `rm -rf` on work Tidebreak never created.

So this was never a policy question about how aggressive to be. It was a missing fact.

## The change

`code_repo` gains `cloned_from`, set only by the clone path. The manual registration route hardcodes `None`, so a client cannot claim it.

`DELETE /code/repos/{id}?reclaim_checkout=true` then needs no judgment:

- `cloned_from` set → delete the directory, after re-validating it is still the git repository that was cloned (a row written long ago must not become a recursive delete of whatever occupies that path now).
- `cloned_from` null → `checkout_not_reclaimable`. The registration is still removed; that was always the whole action for a directory the user brought.

Removal stays soft either way, so archived workspaces and transcripts remain reachable (#2439).

## Credential redaction

A clone URL can embed a credential — `https://ghp_token@github.com/org/repo.git`. That string is persisted and shown back, so it is redacted to its origin before storage, and the reclaim log line no longer includes it.

## Schema

Baseline edit plus `DESKTOP_SCHEMA_EPOCH` 38 → 39, per decision record 2. Both fixtures regenerated.

## Tests

- `reclaim_refuses_a_checkout_tidebreak_did_not_clone` — over HTTP: a registered repo refuses reclaim with `checkout_not_reclaimable`, its `.git` survives, and plain removal still drops the registration while leaving the directory. This is the guard; without it the route deletes user work.
- `a_cloned_repo_records_where_it_came_from` — provenance round-trips, and a registered repo records nothing.
- `a_clone_url_keeps_its_origin_and_loses_its_credential` — covers `token@host`, `user:token@host`, no-credential, and scp-style.

`tidebreak-core` 718 pass, `tidebreak-server --lib code::` 214 pass, clippy and rustfmt clean.

### DIFF
--- a/crates/tidebreak-core/fixtures/schema-baseline.postgres.sql
+++ b/crates/tidebreak-core/fixtures/schema-baseline.postgres.sql
@@ -1225,7 +1225,8 @@ CREATE TABLE "code_repo" (
     "archive_script" text,
     "quick_actions" jsonb NOT NULL,
     "created_at" timestamp with time zone NOT NULL,
-    "removed_at" timestamp with time zone
+    "removed_at" timestamp with time zone,
+    "cloned_from" text
 );
 
 CREATE UNIQUE INDEX "idx_code_repo_owner_root_path" ON "code_repo" ("owner", "root_path") WHERE "removed_at" IS NULL;
--- a/crates/tidebreak-core/fixtures/schema-baseline.sql
+++ b/crates/tidebreak-core/fixtures/schema-baseline.sql
@@ -1306,7 +1306,8 @@ CREATE TABLE "code_repo" (
     "archive_script" text,
     "quick_actions" jsonb_text NOT NULL,
     "created_at" timestamp_with_timezone_text NOT NULL,
-    "removed_at" timestamp_with_timezone_text
+    "removed_at" timestamp_with_timezone_text,
+    "cloned_from" text
 );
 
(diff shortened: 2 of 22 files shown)


naingthet:
**G2G** at `67c67f88a96a`.

`cloned_from` is set only on the clone path (credentials stripped). Register always stores `None`. Reclaim delete is honored only when that origin is present and the path still validates as the same checkout. User-owned directories stay. Epoch bump covers the new column. No remaining P1/P2 that CI would miss.

naingthet:
**G2G** at `67c67f88a96a`. Clone is the only path that writes `cloned_from`, registration hardcodes `None`, and `?reclaim_checkout=true` refuses anything Tidebreak did not create. The delete re-validates that the stored path is still that git toplevel, and a credential in the clone URL is stripped before the string is persisted.

Issue #2313, opened 2026-08-20
feat: one conversation with engines behind the adapter


naingthet:
[Decision 48](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0048-one-interaction-model.md) step 5.

Start only after the remote hosting wire has been exercised (decision 47). Merge chat and code entities into the code-shaped structures: session/turn, sequenced journal plus updates digest, attention, verbatim approvals.

The internal agent loop implements the [decision 31](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0031-harness-adapter-boundary.md) adapter contract as one engine among several. A conversation bound to a repo-backed workspace selects a harness; one without selects the internal engine.

Chat's waiting-for-client continuations, queued turns, and user-questions cards must become expressible through that adapter. Nothing may reach the internal engine through a side door.

Approvals merge structurally but keep their semantic split as capability: external harnesses keep verbatim payloads and no standing grants; the internal engine keeps grant ladders.

Do not start this step as a translation layer or a discriminator column on the chat `event` table.

PR #2450, opened 2026-08-21
refactor(core): one permission vocabulary for chat and code

### LINKED ISSUE #2315: feat: one permission vocabulary for chat and code
[Decision 48](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0048-one-interaction-model.md) step 2.

Replace the parallel `PermissionMode` and `CodePermissionMode` enums with one vocabulary. Use code mode's lifecycle as the default: the mode is chosen at conversation creation. Mid-conversation changes are a capability an engine declares (the internal engine declares it; harnesses mostly do not), under the honesty rules in decisions 38 and 39.

Do not approximate a mode the engine cannot honor.

naingthet:
Closes #2315. [Decision 48](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0048-one-interaction-model.md) step 2, and the first step of the sequence #2428 depends on.

## What

`PermissionMode` and `CodePermissionMode` had identical variants (`Plan | Ask | Auto | Allow`), identical wire tokens, and near-identical `ALL` / `as_str` / `from_str`. Merged into one type at `tidebreak_core::permission` — a home neither surface owns.

The merged doc comment keeps the parts of each that were genuinely saying different things: chat's, about the server-side approval gate and the one client-executed write that consults the mode itself; code's, about adapters selecting the engine's own posture rather than approximating one.

`CodePermissionMode` is deleted rather than aliased. Decision 48 asks for a merge, and two names for one type is the translation layer it explicitly rejects.

## No schema change

Both surfaces already stored the same four tokens, so this is a type merge and not a data change — no baseline edit, no epoch bump. A test pins the tokens so a later edit cannot quietly orphan one side's persisted modes.

## Lifecycle

Decision 48 wants code's lifecycle as the default, and code already has it: the mode is chosen at session creation, changing it relaunches the engine, and `refuse_unhonored_mode` rejects a mode the engine cannot honor instead of approximating it.

Chat still allows a free PATCH. Unifying that needs chat behind the decision-31 adapter contract, which is #2313 — doing it here would mean inventing a capability seam for an engine that has no adapter yet.

I started to add a `permission_mode_change` capability to `HarnessCaps` and removed it: code mode already changes mode by relaunching, so a flag claiming engines cannot would have been both unused and false.

## Client-visible

`CodePermissionMode` disappears from `wire.ts`; the UI now imports `PermissionMode`. `api/types.ts` had aliased both wire types to the same name, so one alias is deleted.

## Tests

`tidebreak-core` 718 pass, `tidebreak-server --lib code::` 214 pass, UI 531 pass, clippy and rustfmt clean, `tsc --noEmit` clean.

One unrelated failure seen locally: `child::unix_process_tree_tests::dropping_a_process_tree_stops_its_descendant_and_closes_its_pipe`. It reproduces identically on unmodified `main` — passes on the first run after a build, then fails every subsequent run — so it is pre-existing. Filed separately.


naingthet:
**G2G** at `2a85c7f254ce`.

The merge keeps the same four wire tokens (`plan`/`ask`/`auto`/`allow`) and the same variant order, so stored chat and code rows still parse. Chat still treats a missing mode as Ask, adapters still refuse an unhonored mode, and nothing at this head still names `CodePermissionMode`.

naingthet:
**G2G** at `06a11a7858b3`.

The new commit only rewraps three UI call sites after the rename. `permission.rs` is untouched: the four wire tokens stay `plan`/`ask`/`auto`/`allow`, Ask is still the documented default, and stored rows still parse the same way.

naingthet:
**G2G** at `5e3cbe170170`.

`PermissionMode` lives in `tidebreak_core::permission` with wire tokens `plan` / `ask` / `auto` / `allow`, `DEFAULT` is `Ask`, and `CodePermissionMode` is gone rather than aliased. The pin test would catch a token drift that would orphan persisted modes. No P1/P2 beyond what CI already covers on a type merge.

PR #2453, opened 2026-08-21
feat(code): pick a reasoning effort and switch it mid-conversation

naingthet:
## What changed

Code mode's reasoning-effort picker only ever wrote to a map in the renderer, so it forgot on reload and never reached the engine; the level is now stored on the session, threaded onto every turn, and mapped onto what each engine accepts (Claude Code `--effort low..max`, grok `--reasoning-effort low..xhigh`, Codex's per-model ladder from `model/list`, and nothing for opencode, which says so). `Ultra` is a new rung above `Max` for the engines whose own pickers offer one — Codex spells it `ultra`, Claude Code's is ultracode, which composes `--effort xhigh` plus the keyword because 2.1.234 exposes no flag — and the picker gives the top of whatever ladder is on offer a violet treatment, so grok's `xhigh` gets it too and no level name is hard-coded. Changing permission mode now asks the live engine first (Claude Code's `set_permission_mode` control request, Codex's per-turn `approvalPolicy`/`sandboxPolicy`) and relaunches only when it refuses, which is still the case for Allow because its posture is a launch flag record 33 forbids composing by default. Two bugs fell out of the wiring and are fixed here: the session worker persists its own copy of the session row every turn, so a change made between turns was written back to its old value by the next one, and Claude's `ensure_channel` read an accepted live switch as a stale child and respawned the very thing the switch avoided.

## Validation

Full `cargo test --workspace` and all 1,624 desktop UI tests pass, alongside `cargo fmt`, biome format/lint, `tsc --noEmit`, and `storybook:build`; the flags were checked against the pinned binaries themselves rather than from docs (`claude --effort` under `-p` stream-json, Codex's generated app-server JSON schema, `grok --reasoning-effort`). The violet Ultra treatment was reviewed as rendered screenshots of the new Storybook story in both light and dark themes, not just read in the diff. One pre-existing flake in `tidebreak-harness` (`terminating_a_process_tree_stops_its_descendant_and_closes_its_pipe`) fails under full-suite parallelism on unmodified `main` too, so it is not from this change.

## Compatibility and risk

`code_session` gains a nullable `reasoning_effort` column and `DESKTOP_SCHEMA_EPOCH` moves to 39, so local desktop databases are rebuilt; `ReasoningEffort` gains an `ultra` token, and `CodeSessionSnapshot`, `HarnessModel`, and `HarnessModelList` each gain a field, all additive. The permission-mode boundary is unchanged: a session that did not choose Allow never has the bypass-enabling flag composed for it, and moving to or from Allow still takes a fresh child.


naingthet:
**G2G** at `e25c0fd31c2b`.

Effort is stored on the session row, written and read through the session id, and refused while a turn is running. Claude and grok clamp to the ladder the pin actually takes (`ultra` becomes `--effort xhigh` plus the ultracode keyword, grok never emits `max`/`ultra`). Codex sends the catalog token as-is. `as_str` / `from_str` / the snapshot parser all keep `ultra`, so a reload does not fall back to a costlier level.

naingthet:
**G2G** at `6894378d7a87`.

The new commit only folds permission mode, model, and effort into `NewSessionSettings` so create stays under the argument ceiling. Effort still lives on the session row, `set_reasoning_effort` still refuses a running turn, and watch still starts with `reasoning_effort: None`. Claude still clamps Ultra to `--effort xhigh` plus the ultracode keyword, grok still clamps to its own ladder, and Codex still sends the catalog token as-is.

Issue #2451, opened 2026-08-21
fix(harness): a process-tree test leaks state and fails on every run after the first


naingthet:
`child::unix_process_tree_tests::dropping_a_process_tree_stops_its_descendant_and_closes_its_pipe` passes on the first run after a build and fails on every run after that, timing out in `assert_pipe_reaches_eof`:

```
panicked at crates/tidebreak-harness/src/child.rs:740:14:
descendant released the stdout pipe: Elapsed(())
```

Reproduces on unmodified `main`:

```
run 1: ok      (0.01s)
run 2: FAILED  (5.01s)
run 3: FAILED  (5.01s)
run 4: FAILED  (5.01s)
```

The pattern is what makes this worth filing rather than dismissing as flake. A test that fails *only* after a prior run has left something behind is describing accumulated state, and the assertion that breaks is the one about a descendant releasing its stdout pipe — so the candidate explanation is that a descendant from an earlier run survives and holds the pipe open. If that is right, the leak is in the code under test, not the test, and CI hides it by starting clean every time.

Worth confirming before assuming it is only a test bug: check whether a descendant process or its pid file outlives the run, and whether the fixture reuses a fixed path across runs.

naingthet:
Fixed in #2454 (`58b73eeff`). The leak was in `ProcessTreeChild`, not the test: a descendant that forks while `kill(2)` walks the process group escapes the signal and keeps the inherited stdout pipe open. The group kill now stops the group first and re-walks it.

PR #2454, opened 2026-08-21
fix(harness): kill a process group until nothing survives in it

naingthet:
## What

`kill(2)` walks a process group once. A descendant that forks while that walk is in flight never receives the signal, and when its killed parent goes away the new process is reparented to init — still in the group, still holding the stdout it inherited. The pipe never reaches EOF, so a caller draining engine output hangs and a live process leaks.

`ProcessTreeChild` sends exactly one `kill(-pgid, SIGKILL)` from `Drop`, `terminate`, `interrupt`, and the interrupt drop guard, so every one of those paths can leak.

## How

`kill_process_group` stops the group before killing it, so a member that is already stopped cannot start another fork, then re-walks the group: the one fork that was in flight when the stop landed is an ordinary member by the next round. The stop is best effort — only the kill decides the result. It exits early once `kill(2)` stops finding the group, and is bounded to three rounds and about two milliseconds, because it also runs from `Drop`, where there is no runtime to await on.

## Why this is the product, not the test

Issue #2451 reported `dropping_a_process_tree_stops_its_descendant_and_closes_its_pipe` timing out in `assert_pipe_reaches_eof`, and asked whether the leak was in the code under test. It is. Every failing run left a `sleep 60` behind with `ppid 1` and a `pgid` whose leader was already gone — a group member the kill missed. Passing runs left nothing. The reported "passes once, then always fails" pattern was scheduling luck; on this machine the existing test failed about one run in three.

## Tests

`dropping_a_process_tree_stops_a_descendant_that_is_still_forking` makes the window reliable instead of one-in-three: the descendant forks a burst of pipe-holding children right after it publishes its pid, so the kill almost always lands mid-fork. Against the single-walk kill it fails 7 runs out of 8; with the fix, 20 consecutive runs of the whole `unix_process_tree_tests` module pass and leak nothing.

The burst is bounded and its children are `sleep 60`, so a future regression leaves at most a handful of processes that exit on their own — the current test leaks a `while :; do sleep 60; done` that never does.

`spawn_shell_tree` now delegates to `spawn_tree_from`, which both tests share.

### DIFF
--- a/crates/tidebreak-harness/src/child.rs
+++ b/crates/tidebreak-harness/src/child.rs
@@ -279,7 +279,7 @@ impl ProcessTreeChild {
             let Some(process_group) = self.process_group else {
                 return Ok(());
             };
-            signal_process_group(process_group, libc::SIGKILL)?;
+            kill_process_group(process_group)?;
             self.process_group = None;
             Ok(())
         }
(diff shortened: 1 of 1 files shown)


naingthet:
**G2G** at `5b3cdfe558b1`.

`kill_process_group` is the only Unix teardown: `Drop` and `terminate_tree`, `ProcessGroupGuard::kill`, and the interrupt drop guard all go through it. `SIGINT` still single-shots, which is the grace signal, not the leak path.

The group id is the root pid after `process_group(0)`, and every `kill` uses `-process_group` with `process_group > 0`. That cannot target the host session.

Re-walk is three rounds and a 1ms settle, then it returns. `process_group_is_gone` is only an early exit; an unreaped leader does not loop forever.

SIGSTOP is best effort before SIGKILL. A member that appears on the last stop walk and misses the last kill walk can stay stopped, but the new burst test holds the pipe on those children, so that leftover fails CI instead of hanging a drain later. Residual last-round luck is the bound they chose for `Drop`.

Issue #2429, opened 2026-08-21
fix: an epoch reset orphans every code worktree on disk


naingthet:
`reset_pre_v1_state` in `crates/tidebreak-server/src/desktop_schema.rs` deletes the SQLite files and the host-broker durable state, and its own comment says "nothing else in the data directory."

Code worktrees are not in the data directory. [Decision 53](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0053-code-worktrees-live-in-a-user-visible-root.md) moved them to `~/Tidebreak/workspaces`, deliberately, because `worktree_root.rs` puts it plainly:

> A worktree is user work — uncommitted code on a real branch — so it does not belong in the disposable app-data directory the database and logs live in.

So every epoch bump drops the `code_workspace` rows and leaves the trees. Eight bumps since 2026-08-14. What is left behind after each one:

- Directories under `~/Tidebreak/workspaces/<repo-slug>/<workspace-slug>-<short-id>/`, with whatever uncommitted work was in them.
- Entries in each source repo's `.git/worktrees/*`, so `git worktree list` keeps reporting them.
- Branches the workspace created, on the user's real repo.

Nothing sweeps them. `plugin_mcp.rs` has `prune_orphan_plugin_data` for plugin data; there is no equivalent here. `scripts/clean-worktree-artifacts.sh` is about Cargo target directories in development checkouts, not these.

**Deleting them is not the fix.** The record says they are user work, and the reset cannot even read `code_worktree_root` or the `worktree_path` rows to know which trees were ours — the database it would read is the thing being deleted.

**Options, roughly in order of cost:**

1. Read the old database before removing it — raw SQL, tolerant of a schema that no longer matches — and write the orphaned paths to a sidecar file next to the schema marker. Log a warning naming the file. Non-destructive, and the reset already keeps sidecars for exactly this kind of state (the marker itself, `gateway-policy.json`).
2. The same read, plus a first-run surface that lists what was orphaned and offers to archive or delete each tree.
3. The same read, plus re-adoption: recreate `code_repo` and `code_workspace` rows from the sidecar so the trees come back as real workspaces.

Option 1 is the floor and makes the rest optional. Whichever lands, `reset_pre_v1_state`'s comment needs to stop implying the data directory is the whole story.

Worth doing before #2428 closes the epoch window, since after that the resets stop and the orphans left by the previous eight are still on disk.

naingthet:
[...]

PR #2458, opened 2026-08-21
fix: record the code worktrees an epoch reset orphans

naingthet:
## What

`reset_pre_v1_state` deletes the SQLite files and the host-broker durable state, and its comment said "nothing else in the data directory". Code worktrees are not in the data directory. [Decision 53](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0053-code-worktrees-live-in-a-user-visible-root.md) moved them to `~/Tidebreak/workspaces` deliberately, because a worktree is user work — uncommitted code on a real branch.

So every epoch bump drops the `code_workspace` rows and leaves behind the trees, the entries in each source repository's `.git/worktrees`, and the branches the workspaces created on the user's real repository. Eight bumps have gone by and nothing sweeps them.

## How

This is option 1 from #2429, the non-destructive floor.

Right before the database goes, `code::worktree_orphans` reads the rows that name the trees and records every tree still on disk in `orphaned-code-worktrees.json`, beside the schema marker — a sidecar because sidecars survive a reset by design, which is already how the marker and `gateway-policy.json` work. A warning names the file and says plainly that nothing removes the trees, their `.git/worktrees` entries, or their branches.

Deliberate choices:

- **Best effort, never blocking.** The scan reads a database whose schema this binary no longer matches, on a path that has to end with the app booting. It opens read-only so it can never create or migrate the file it is about to delete, `SELECT *`s rather than naming columns that may not exist, falls back to an unjoined query for databases from before code repositories, and closes before the caller deletes the file so Windows can. Every failure is logged and the reset continues.
- **Merges rather than replaces.** Epoch bumps come in runs. A later reset adds to the record and keeps the first sighting of each tree, which is the honest timestamp: it says when the tree stopped being a workspace.
- **A tree already gone is not an orphan.** Only paths still on disk are recorded.
- **An unparseable sidecar is moved to `orphaned-code-worktrees.previous.json`, not overwritten.** It is still the only pointer to somebody's uncommitted work.

Options 2 (a first-run surface) and 3 (re-adoption) stay open on top of this and need no rework — the record is the input either would read.

`prepare` and `reset_pre_v1_state` become `async` to do the read; two sync tests become `#[tokio::test]`. `replace_file` is now `pub(crate)` so the sidecar publishes with the same guarantee as the marker.

## Tests

Three, in `desktop_schema` where the reset contract already lives:

- A reset records the orphaned tree with its path, branch, source repository, and title, does not record a workspace whose tree is already gone, and leaves the tree's contents untouched.
- A second reset adds to the record instead of replacing it.
- A reset with no code worktrees writes no file at all.

## Note on timing

Worth landing before #2428 closes the epoch window. After that the resets stop, but the orphans from the previous eight are still on disk — this records new ones, and does not go back for those.

### DIFF
--- a/crates/tidebreak-server/src/code/mod.rs
+++ b/crates/tidebreak-server/src/code/mod.rs
@@ -25,6 +25,7 @@ pub(crate) mod terminal;
 pub(crate) mod titling;
 pub(crate) mod watch;
 pub(crate) mod worktree;
+pub(crate) mod worktree_orphans;
 pub(crate) mod worktree_root;
 
 pub(crate) use runtime::CodeRuntime;
--- a/crates/tidebreak-server/src/code/worktree_orphans.rs
+++ b/crates/tidebreak-server/src/code/worktree_orphans.rs
@@ -1,0 +1,215 @@
+//! What a pre-v1 epoch reset is about to lose track of.
+//!
+//! Code worktrees do not live in the data directory. Decision 53 moved them to
+//! a user-visible root on purpose, because a worktree is user work —
+//! uncommitted code on a real branch. The reset in [`crate::desktop_schema`]
+//! deletes the database and nothing outside the data directory, so every
+//! epoch bump drops the `code_workspace` rows and leaves the trees behind:
+//! directories full of possibly-uncommitted work, entries in each source
+//! repository's `.git/worktrees`, and branches on the user's real repository.
+//!
+//! Deleting them is not the answer, and neither is silence. So the reset
+//! writes them down. Right before the database goes, read the rows that name
+//! the trees and record them in a sidecar file beside the schema marker, which
+//! survives a reset by design. The trees are still orphaned afterwards, but
+//! they are now recoverable by hand, and adoptable by a first-run surface
+//! later.
+//!
+//! Everything here is best effort. It runs against a database whose schema
+//! this binary no longer matches, on a path that has to end with the app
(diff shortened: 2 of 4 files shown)


naingthet:
**G2G** at `1e372810e95c`.

`reset_pre_v1_state` records first, then deletes only SQLite and host-broker files. The scan never removes trees, `.git/worktrees` entries, or branches.

The sidecar path is a fixed name under `data_dir`. Temporary and previous files use `with_file_name` on that same path, so a workspace `worktree_path` cannot steer the write outside the data directory.

`scan` only reads `worktree_path` from `code_workspace` and keeps rows whose path is still a directory. It does not walk checkouts or follow a user repo outside Tidebreak into the orphan list.

naingthet:
**G2G** at `2652bd36e835`.

The new commit is docs only: `how-tidebreak-works.md` names `orphaned-code-worktrees.json` and states that a reset records the trees and leaves the trees, `.git/worktrees` entries, and branches alone.

`reset_pre_v1_state` still records first (`record_orphaned_worktrees`) then deletes SQLite and host-broker durable files. `worktree_orphans` is read-only scan + sidecar merge. It never removes a worktree directory, a git worktree entry, or a branch.

Prior G2G at `1e372810e95c` still holds: record-then-wipe, no tree or branch delete.

Issue #2314, opened 2026-08-20
feat: one attention surface for chat and code


naingthet:
[Decision 48](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0048-one-interaction-model.md) step 3.

Give every conversation, chat included, the code-mode `AttentionState` so a supervising client watches one queue with one vocabulary. This is the first place the chat and code id spaces meet, and it is what a later phone client needs.

Do not keep a second inbox keyed only on `ChatId`.

PR #2460, opened 2026-08-21
refactor(core): move attention out of code mode

naingthet:
First of a stack for #2314 ([decision 48](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0048-one-interaction-model.md) step 3).

## What

`attention` moves from `crates/tidebreak-core/src/code/` to `crates/tidebreak-core/src/attention.rs` — a home neither surface owns, the same move `PermissionMode` made in #2450.

Pure relocation. No type, variant, or wire name changes. The generated TypeScript is byte-identical, so no client can tell.

## Why now

[Decision 30](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0030-code-mode-separate-surface.md) defined this vocabulary over "a unit of supervised work" rather than over a coding session, and said so in the module's own doc comment, explicitly so chat could adopt it later without renaming. This is that adoption starting.

Splitting the move from the behavior keeps the next PR — chat conversations deriving an `Attention` of their own — reviewable as behavior rather than as several hundred lines of import churn.

## One thing the module now states

A shared vocabulary does not oblige every engine to produce every state. `AttentionState::Fenced` exists for supervising a process the server does not control; the internal engine has nothing to fence. Worth writing down before chat adopts the type, since the alternative reading — that chat must now grow a fencing concept — is wrong.

## Tests

`tidebreak-core` 720 pass, `tidebreak-server --lib code::` 223 pass, clippy and rustfmt clean.

### DIFF
--- a/crates/tidebreak-core/src/attention.rs
+++ b/crates/tidebreak-core/src/attention.rs
@@ -1,7 +1,15 @@
 //! Attention for a unit of supervised work.
 //!
 //! The vocabulary is defined over "a unit of supervised work", not over a
-//! coding session specifically, so chat can adopt it later without renaming.
+//! coding session specifically. Decision 30 wrote it that way on purpose so
+//! chat could adopt it without renaming, and decision 48 step 3 is where that
+//! happens: this module moved out of `code` to a home neither surface owns,
+//! ahead of chat conversations carrying an [`Attention`] of their own.
+//!
+//! Not every engine produces every state. [`AttentionState::Fenced`] belongs
+//! to supervising a process the server does not control, and the internal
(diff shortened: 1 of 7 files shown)


naingthet:
**G2G** at `9b4662bdf7f3`.

Pure relocation: `code/attention.rs` becomes `core/attention.rs`, crate-root re-exports the same names, and `code` only imports what it still uses. Serde tags stay `snake_case`; `as_str` / `from_str` tokens are unchanged. No leftover `CodeAttention` type. The `Manual` + automatic-source bump to `User` in `Attention::new` is the same constructor it had before the move, not a silent upgrade of stored attention.

Nothing here that CI would miss on wire or default behavior.

Issue #2457, opened 2026-08-21
test(core): a Postgres acceptance test only passes against a database no run has touched


naingthet:
`postgres_turn_acceptance_claims_and_receipts_are_atomic` passes against a fresh database and fails on every run after that, against the same database:

```
run 1 (fresh database): ok
run 2 (same database):  FAILED
```

Reproduces on unmodified `main` with the CI image (`postgres:17-alpine`). It fails two different ways depending on how far it gets:

```
assertion `left == right` failed
  left: ChatId(acdb555b-…)
 right: ChatId(c9cacadc-…)
crates/tidebreak-core/tests/postgres_turn_state.rs:2724
```

```
assertion `left == right` failed
  left: Some(TurnRun { id: TurnId(40b0e920-…), … })
 right: None
crates/tidebreak-core/tests/postgres_turn_state.rs:2720
```

## Why

`claim_turn_run` claims *any* runnable turn in the store — it is a work-queue poll, not a per-chat lookup. The test spawns eight racing claims and asserts exactly one wins and that the winner is its own turn:

```rust
let (token, claimed) = winner.unwrap();
assert_eq!(claimed.id, turn_id);
```

That holds only when the queue contains nothing but this test's turn. Every other test in the file leaves queued turns behind, and so does every previous run of this one. On a reused database the claim picks up a stale turn from an earlier run, and the identity assertions compare against the wrong row.

So this is a test-isolation bug, not a product one: `claim_turn_run` is behaving exactly as a global queue claim should. CI hides it because the `postgres` job starts a fresh service container every time, which is also why it is invisible to anyone who does not run the suite twice locally.

## Worth deciding

Fixing it is a design call rather than a one-line change, since the assertions are about global claim behavior and the fixture is shared:

1. Give each test its own database (or schema), so a global claim really is global over one test's rows.
2. Drain or truncate the queue in the fixture before the racing claims.
3. Keep the shared database and weaken the assertions to "some turn was claimed exactly once", which loses the identity coverage that makes the test worth having.

Related in shape to #2451: a test that only passes from a clean slate is describing state that outlives a run, and CI's clean slate keeps it quiet.

PR #2461, opened 2026-08-21
test(core): give the Postgres claim test a queue it owns

naingthet:
## What

`postgres_turn_acceptance_claims_and_receipts_are_atomic` passes against a fresh database and fails on every run after that against the same one, two different ways depending on how far it gets:

```
assertion `left == right` failed
  left: ChatId(acdb555b-…)     right: ChatId(c9cacadc-…)

assertion `left == right` failed
  left: Some(TurnRun { id: TurnId(40b0e920-…), … })     right: None
```

`claim_turn_run` is a work-queue poll across the whole store, not a per-chat lookup — it takes the globally oldest due or expired turn. The test spawns eight racing claims and asserts the single winner is its own turn, which holds only when the queue contains nothing else. Every other test in this file leaves queued turns behind, and so does every earlier run of this one, so the claim picks up a stale turn with an older `available_at` and the identity assertions compare against the wrong row.

Nothing is wrong with `claim_turn_run` — it behaves exactly as a global queue claim should. CI hid this by starting a fresh Postgres service container every run, which is also why it is invisible to anyone who does not run the suite twice.

## How

Option 2 from #2457: empty the store before the racing claims, so a global claim really is global over one turn. That keeps the identity coverage, which option 3 would have thrown away, and does not pay a fresh migration per test the way option 1 would.

The table list comes from `pg_tables` rather than a constant, so a new table is covered without anyone remembering to add it. Two are held back:

- `seaql_migrations`, because dropping it re-runs every migration.
- `advisory_lock`, whose seeded rows are what serializes the claim paths. The baseline inserts them once (`SEED_STATEMENTS`), and every claim fails with `durable advisory lock turn_claim is missing` without them — which is what the first version of this change did.

Only this test resets, because only this test asserts on a scan across the whole store. The helper's doc says when to reach for it.

## Tests

Run against `postgres:17-alpine`, the CI image, five times in a row against a single database that is never recreated: 29 passed every time. On `main` the same loop fails from the second run on. The sibling `postgres_code_owner` suite still passes against the same reused database.

### DIFF
--- a/crates/tidebreak-core/tests/postgres_turn_state.rs
+++ b/crates/tidebreak-core/tests/postgres_turn_state.rs
@@ -118,6 +118,40 @@ async fn expire_postgres_turn_admission(url: &str, turn_id: TurnId) {
     assert_eq!(updated.rows_affected(), 1);
 }
 
+/// Delete every row the product owns, leaving the migrated schema in place.
+///
+/// The tests in this file share one database and serialize on
+/// [`POSTGRES_TEST_LOCK`], which keeps them from running at once but not from
+/// reading each other's leftovers. A test that asserts on a scan across the
+/// whole store has to start from an empty one, or it passes only against a
+/// database no run has touched — which is exactly what CI hands it, and why
+/// running the suite twice locally used to fail.
+///
+/// The table list comes from the catalog rather than a constant, so a new table
+/// is covered without anyone remembering to add it. Two tables are held back:
+/// SeaORM's bookkeeping, because dropping it re-runs every migration, and
+/// `advisory_lock`, whose seeded rows are what serializes the claim paths — the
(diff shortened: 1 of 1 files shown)


naingthet:
**G2G** at `ff62ae734209`.

`claim_turn_run` still scans the whole store, and this test still asserts the winner is its own turn plus the terminal receipt. The new helper empties every product table except `seaql_migrations` and `advisory_lock` before that race, so leftovers from earlier tests or earlier runs cannot win the poll. `expire_postgres_turn_admission` still keys on the test's `turn_id`. Tests stay serialized on `POSTGRES_TEST_LOCK`, so the truncate cannot wipe a sibling mid-flight.

The identity and receipt assertions are intact. CI would not catch a leftover-queue flake on a fresh service container; this does.

PR #2462, opened 2026-08-21
docs: fix the duplicate 56 record and the stale permission-mode entry

naingthet:
Two documentation records went stale on the same day. Both corrections are
prerequisites for writing the next decision record, because the next free
number depends on the first one.

## Renumber the reclaim record to 59

`0056-one-credential-item-per-profile.md` (#2412, 11:11) and
`0056-workspace-reclaim-tiers.md` (#2443, 15:12) both landed as 56. The
reclaim record arrived second, so it takes the next free number, following
the precedent of #2365 and #2397. Nothing in the repository cites either
record, so the rename and its title line are the whole change. The next free
decision-record number is now 60.

## Say that a live session can change its permission mode

`docs/deferred.md` still said the mode is fixed at session creation and that
no adapter can renegotiate it on an attached session. #2411 shipped the
change: the session asks its engine on the engine's own channel — Claude
Code's `set_permission_mode` control request, Codex's per-turn policy fields
— and only an engine that fixes its posture at launch pays for a stop and
re-attach.

The entry is rewritten rather than removed, because `docs/deferred.md` is the
canonical record of parked scope and a silent deletion loses the reasoning.
What survives as parked is moving the posture underneath a running turn,
which `CodeRuntime::set_permission_mode` refuses with `turn_running`, so the
entry now names that instead.

## Checks

Documentation only. No code, schema, or interface change.

### DIFF
--- a/docs/decisions/0059-workspace-reclaim-tiers.md
+++ b/docs/decisions/0059-workspace-reclaim-tiers.md
@@ -1,4 +1,4 @@
-# 56. Workspace Reclaim Tiers
+# 59. Workspace Reclaim Tiers
 
 - Status: Accepted
 - Date: 2026-08-21
--- a/docs/deferred.md
+++ b/docs/deferred.md
@@ -228,17 +228,20 @@ purpose:
   worktree per session, which splits the branch, the diff, and the pull
   request the workspace is keyed to. Reconsider only with a reason to give
   one workspace several trees.
-- **Changing a session's permission mode after it starts.** The mode is
-  chosen at session creation, refused per harness capability there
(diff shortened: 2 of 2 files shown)


naingthet:
**G2G** at `44e0baf80642`.

The reclaim record is `docs/decisions/0059-workspace-reclaim-tiers.md` with title 59. The tree at this SHA has only `0056-one-credential-item-per-profile.md` under 56; there is no leftover `0056-workspace-reclaim` path. Related links in the 59 record resolve. `docs/deferred.md` now treats a live-session mode change as shipped and parks only a change underneath a running turn. The 33 / 38 / 39 citations still point at the right ADRs.

PR #2463, opened 2026-08-21
fix(code): a settled session is idle, not working

naingthet:
Second of the stack for #2314. Follows #2460, which moved the attention vocabulary out of code mode.

## The bug

`AttentionState` had no resting state — `Working | NeedsYou | Stalled | DoneUnreviewed | Fenced | Manual`. So three paths reported a session as `Working` while no engine was running:

1. **`compute_attention`'s final arm.** Reached by a completed-and-reviewed session, and by one that has never taken a turn.
2. **`mark_viewed`.** Wrote `Attention::working(...)` outright — looking at a finished session made it claim to be busy.
3. **`reap_session`.** Sets `lifecycle = Idle`, resolves the fence, then set attention to `Working`.

## The evidence it was already known

`attention_follows_approval_completion_and_view` asserted that viewing a session makes it `Working`, with the message "viewing the session must clear DoneUnreviewed." That expectation was the bug written down: `Idle` was not available to assert instead, so the test pinned the wrong state. Correcting it is the regression test.

## Why now rather than as its own bug

Code mode survives this. A session is short-lived and usually being watched, so a wrong resting state is not on screen long enough to confuse anyone.

Decision 48 step 3 changes the stakes: chat conversations are about to carry the same vocabulary, and a chat idle for a week claiming an engine is busy is not a rounding error. Fixing the vocabulary before chat adopts it means chat inherits a state machine that can express rest.

## Not a schema change

`attention_state` is a `json_binary` column with no CHECK constraint, so a new variant needs no baseline edit and no epoch bump. Existing rows hold the old variants and still parse.

The stall-clearing path also writes `Working`, and that one is correct — it fires only when the session is `Running`. Left alone.

## Tests

`tidebreak-server --lib code::` 223 pass, `tidebreak-core` 720 pass, UI 539 pass, clippy, rustfmt, `tsc`, and `biome check` clean. The CLI and both UI switches over `AttentionState` are exhaustive, so the new variant was a compile break at each — all three now render "Idle".

### DIFF
--- a/crates/tidebreak-cli/src/code.rs
+++ b/crates/tidebreak-cli/src/code.rs
@@ -1529,6 +1529,7 @@ fn attention_label(attention: &Attention) -> String {
         AttentionState::NeedsYou { prompt, .. } => format!("needs_you: {prompt}"),
         AttentionState::Stalled { idle_secs } => format!("stalled {idle_secs}s"),
         AttentionState::DoneUnreviewed => "done_unreviewed".to_owned(),
+        AttentionState::Idle => "idle".to_owned(),
         AttentionState::Fenced { .. } => "fenced".to_owned(),
         AttentionState::Manual { note } => format!("manual: {note}"),
     }
--- a/crates/tidebreak-core/src/attention.rs
+++ b/crates/tidebreak-core/src/attention.rs
@@ -140,6 +140,13 @@ pub enum AttentionState {
     },
     /// Finished work the user has not looked at.
     DoneUnreviewed,
+    /// Nothing is running and nothing is waiting.
+    ///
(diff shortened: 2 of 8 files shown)


naingthet:
**G2G** at `e9652f6951c3`.

`compute_attention` only reaches `Idle` after the `Running` arm has already returned `Working` or `Stalled`, so a live engine is not reclassified. `mark_viewed` writes `Idle` once `DoneUnreviewed` is gone, and `reap_session` matches. The wire token is internally tagged `"idle"`; existing rows keep their old variants and still parse.

The stall-clear path still writes `Working` only while `lifecycle == Running`. That is the right write.

PR #2464, opened 2026-08-21
fix(ui): render the foreground browser panel instead of its tab title

naingthet:
## What changed

`renderPanel` matched `case "browser"` twice, so the foreground browser panel rendered its tab title string instead of mounting `CodeBrowserTab` — both cases landed together in #2419, so the panel has never worked. Biome's `noDuplicateCase` is now on to catch the next one, since the UI lint sets `recommended: false` and TypeScript accepts a string as a valid `ReactNode`. The desktop binary also allows `linker_messages` on macOS debug builds, where ~18MB of `__eh_frame` overflows the 16MB a Mach-O compact unwind table can address and makes `ld` warn on every dev link.

## Validation

`npx tsc --noEmit` and `pnpm --dir crates/tidebreak-desktop/ui lint` both pass, and a probe file confirmed `noDuplicateCase` fires on a duplicate. I did not relink the desktop binary locally because this workspace's target directory is cold, so I checked the lint name with `rustc -W help` and confirmed the attribute compiles clean standalone; CI covers the real link.

## Compatibility and risk

Nothing persisted or on the wire changes, and the `allow` is scoped to macOS debug builds so release builds and every other platform keep their linker warnings.

### DIFF
--- a/crates/tidebreak-desktop/src/main.rs
+++ b/crates/tidebreak-desktop/src/main.rs
@@ -3,6 +3,12 @@
 //! Boots the in-process HTTP/WebSocket surface and hosts the chat UI webview.
 
 #![cfg_attr(not(debug_assertions), windows_subsystem = "windows")]
+// A debug build of this binary emits ~18MB of `__eh_frame`, past the 16MB the
+// Mach-O compact unwind table can address, so `ld` warns on every dev link.
+// Only unwinding a panic pays for that, and nothing here is on a hot path, so
+// the warning is noise. Scoped tight: release builds and every other platform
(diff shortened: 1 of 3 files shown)


naingthet:
**G2G** at `82bce49c040f`.

`renderPanel` now has one `case "browser"` and it mounts `CodeBrowserTab` (with the existing `Suspense`/`Skeleton` wrapper). The title string lives only in `tabLabel`, which is the right place. `noDuplicateCase` is on in the UI biome config, so the same switch cannot land again. The `linker_messages` allow is scoped to macOS debug only.

PR #2466, opened 2026-08-21
feat(core): a chat carries attention, derived from its own rows

naingthet:
Third of the stack for #2314, after #2460 and #2463.

## What

Every conversation now has an attention value in the vocabulary code mode introduced, so one supervising client can watch one queue. `Store::chat_attention_scoped` returns one value per chat that has something to say; absent means `Idle`.

## Derived, not stored

Code keeps attention on `code_session` because a session *is* the unit — its lifecycle lives on the same row. Chat has no equivalent row. A conversation's liveness is spread across `turn_run`, and its waiting work is exactly the set the inbox already projects.

So attention columns on `chat` would duplicate both, and would need a write at every turn transition to stay true. That is the shape #1462 spent a wave removing, and the inbox next door already states the principle in its own doc comment: a projection of the rows that carry the state, "deliberately not a store of its own, so resolving an item through its existing route is what removes it from here."

The test asserts exactly that property — an approval parked, then resolved through its normal route, with no attention-specific write anywhere in between.

## One definition of "busy"

Live turn statuses become `TurnRunStatus::LIVE`, shared with `count_active_work`, which had the list inline. Busy has to mean the same thing to the host's quiescence check and to the reader's badge; a new non-terminal status added to one and not the other would make a live conversation look settled.

## What chat does not produce

Neither `DoneUnreviewed` nor `Fenced`. Reviewed-ness rides an attached events socket chat does not have, and there is no foreign process to fence. This is the principle #2460 wrote into the module: adopting a shared vocabulary is not a promise to use all of it.

## Not a schema change

No baseline edit, no epoch bump.

## Tests

`a_chat_carries_attention_derived_from_its_parked_work` covers needs-you with a kind-only prompt, absent-means-idle, and the hand-back to `Working` when an approval is granted — the turn resumes, so asserting the chat goes quiet there would be asserting that approving a call does nothing.

`tidebreak-core` 720 pass, `tidebreak-server --lib` 1199 pass. One unrelated failure under parallel load, `code_terminals::create_write_read_and_two_cursors`, which passes in isolation; filed as #2467.

### DIFF
--- a/crates/tidebreak-core/src/db/mod.rs
+++ b/crates/tidebreak-core/src/db/mod.rs
@@ -2591,6 +2591,14 @@ impl Store for DbStore {
         ops::inbox::list_inbox_items(self, owner).await
     }
 
+    async fn chat_attention_scoped(
+        &self,
+        owner: &OwnerId,
+        items: &[crate::InboxItem],
+    ) -> Result<std::collections::HashMap<crate::ChatId, crate::Attention>> {
+        ops::chat_attention::chat_attention(self, owner, items).await
+    }
+
     async fn answer_user_questions(
         &self,
         request: &crate::AnswerUserQuestionsRequest,
--- a/crates/tidebreak-core/src/db/ops/active_work.rs
+++ b/crates/tidebreak-core/src/db/ops/active_work.rs
@@ -15,16 +15,10 @@ use super::super::{entities, store_err, DbStore};
 
 pub(in crate::db) async fn count_active_work(store: &DbStore) -> Result<ActiveWorkSnapshot> {
     let active_turns = entities::turn_run::Entity::find()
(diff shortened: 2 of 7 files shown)


naingthet:
**G2G** at `608c7d376a3f`.

Attention is derived, not stored. `TurnRunStatus::LIVE` is the full non-terminal set, and `count_active_work` uses the same list, so a live turn cannot classify as Idle while the host still counts it busy. Scoped reads load every live `turn_run` then keep only chats in the owner set; inbox items are already owner-scoped, so another user cannot appear. A waiting inbox row overwrites Working, and the oldest NeedsYou wins, so parked work is not dropped. After approve the turn is live again, so Working is the right answer, not a settled leak.

PR #2348, opened 2026-08-20
feat(code): add delivery center, archive, and notifications

naingthet:
## Summary

- add a cross-repository Delivery center for pull requests, GitHub Actions runs, and deployments, including filters, saved views, detail drawers, workspace links, and guarded actions
- move archived workspaces into a searchable, filterable Archive page and add Delivery, Archive, and Notifications links above Usage
- add configurable local delivery notifications, background polling, GitHub server routes, strict wire parsing, tests, and Storybook coverage
- widen the workspace rail settings popover and permanently keep archived workspaces out of the live rail

## Validation

- `pnpm --dir crates/tidebreak-desktop/ui exec tsc --noEmit --pretty false`
- focused UI suite: 176 tests across 7 files
- full UI suite before rebase: 1,450 tests across 205 files
- `pnpm --dir crates/tidebreak-desktop/ui storybook:build`
- `cargo test -p tidebreak-server code::delivery --lib`
- `UPDATE_WIRE_TYPES=1 cargo test -p tidebreak-server wire_types::tests::the_generated_bindings_are_current --lib`
- `cargo fmt --check`
- `git diff --check`

## Notes

A controllable browser was unavailable for screenshot-level visual inspection; Storybook and production builds completed successfully.

PR #2468, opened 2026-08-21
docs(decisions): propose triggers as durable rules on pull-request facts

naingthet:
Record 60. Status is `Proposed` — merging is the acceptance, so this is the
argument to push back on before any of the substrate gets built.

## The problem it settles

An agent working a pull request finds out its checks failed by asking. Three
pieces of the answer already exist and none of them know about each other:

- **Watch-and-fix** ([record 50](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0050-watch-and-fix-is-a-durable-task.md))
  already does "CI fails, an agent acts" as a durable row on a try-based sweep.
  Its condition set, its action, and its dedicated `kind = watch` session are
  all fixed.
- **Notification rules** in `CodeDeliveryStore.ts` (#2348) are already
  user-defined and repository-scoped — and live in `localStorage` with no
  record behind them. The durable/ephemeral split is backwards.
- **Mid-turn delivery** is reachable but not uniform. `CodeRuntime::steer`
  refuses unless the harness declares `mid_turn_steering: Supported`, and only
  Codex does.

## What it decides

A trigger is a durable row modelled on `code_watch`, driven by a sweep that
reads rows rather than subscribing to the lossy `broadcast` bus. Conditions are
enum-shaped edge transitions of `PullRequestDigest` fingerprinted against
`head_sha` — no user scripting. Actions in v1 are exactly two: deliver a
bounded message to a session, and raise a notification. Merge, auto-merge and
mark-ready stay with the user under [record 42](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0042-user-initiated-pr-merge.md).

Delivery is steer where the harness declares it and a sweep retry otherwise.
The capture in #2465 is why that split is real: Claude Code reads an injected
line at a tool-call boundary, a turn without one never reaches it, and the miss
is silent. Code sessions have no `queued_turn` table, so the fallback is the
watch's own idiom — hold the fire, submit on a later tick when no session in
the workspace is running a turn.

Three positions the research left open are decided here rather than deferred:
triggers bind **per repository**, they target the workspace's **most recently
active interactive session** named in the interface when armed, and the
client-side notification rules **move into this substrate**. That last one is
the only choice with real cost — a one-way `localStorage` migration — and the
record says why one rule engine is worth it.

## Cost

A baseline schema edit, so `DESKTOP_SCHEMA_EPOCH` goes 39 → 40. Two automation
engines coexist until watch-and-fix is re-expressed. Claude Code — the
reference tier — delivers on a sweep retry rather than mid-turn until its
adapter can bound a missed steer.

Documentation only; no code in this PR.

### DIFF
--- a/docs/decisions/0060-triggers-are-durable-rules-on-pull-request-facts.md
+++ b/docs/decisions/0060-triggers-are-durable-rules-on-pull-request-facts.md
@@ -1,0 +1,225 @@
+# 60. Triggers Are Durable Rules On Pull-Request Facts
+
+- Status: Proposed
+- Date: 2026-08-21
+- Owners: code mode, harness integration
+- Related: [`0009-queued-turns.md`](0009-queued-turns.md),
+  [`0031-harness-adapter-boundary.md`](0031-harness-adapter-boundary.md),
+  [`0042-user-initiated-pr-merge.md`](0042-user-initiated-pr-merge.md),
+  [`0048-one-interaction-model.md`](0048-one-interaction-model.md),
+  [`0050-watch-and-fix-is-a-durable-task.md`](0050-watch-and-fix-is-a-durable-task.md),
+  [`0055-multiple-sessions-per-workspace.md`](0055-multiple-sessions-per-workspace.md),
+  [`0057-one-claude-child-per-session.md`](0057-one-claude-child-per-session.md)
+
+## Context
+
+An agent working a pull request learns its checks failed by asking. Polling
+costs a tool call and a slice of context per look, and it is the agent's own
+loop, so nothing outside the turn can make it look sooner. What the user wants
+is the opposite arrangement: declare once that a failing check should reach the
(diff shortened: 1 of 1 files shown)


naingthet:
**G2G** at `13157080dbb2`.

Record 60 stays inside accepted invariants. Merge, auto-merge, and mark-ready stay user-only (42). Watch stays its own durable loop and is never a delivery target; an overlapping fire is suppressed rather than racing the same digest (50). Queue remains the default path; steer is only where the harness already declares it (9, 31, 57). Binding is owner-scoped and repo-shaped, not a cross-principal rule on another user's PR facts (48). Claude Code is not granted `mid_turn_steering` from the unacked stdin capture. Conditions are digest edge transitions fingerprinted on `head_sha`, not a script surface. Nothing here would commit an unsafe or contradictory product rule if accepted as written.

PR #2470, opened 2026-08-21
feat(core): triggers are durable rows on pull-request facts

naingthet:
First slice of [record 60](https://github.com/brightwave-inc/tidebreak/pull/2468):
the row a trigger lives in, the fingerprint that makes it fire on an edge, and
the classifier that turns a digest into a condition. Nothing sweeps yet — this
is the substrate the sweep and the delivery path land on next.

## What is here

**`code_trigger`** binds per repository, not per workspace. A trigger is a
standing preference about how the user wants to be reached, not a decision
about one pull request the way a watch is. Unique on
`(owner, repo_id, condition)`, so arming a condition twice is an edit —
`save_trigger` upserts on that key rather than erroring on a switch the user
just flipped.

**`code_trigger_fire`** is the fingerprint. Its primary key is
`(trigger_id, workspace_id, head_sha)`, so the insert *is* the claim:
`insert_trigger_fire` returns whether this call was the one that got it, and
two sweeps racing the same edge produce one fire. Fires cascade on trigger
delete, so a re-armed trigger cannot inherit the old one's suppression.

**`classify_trigger_condition`** generalizes the watch's `assess` over the same
`PullRequestDigest` and the same host tokens, and keeps its precedence.

## Judgement calls worth checking

- **Conflicts outrank failing checks.** A conflicting branch reports conflicts,
  not the tests the conflict broke. Waking an agent to fix tests that pass once
  the rebase lands is the failure this avoids.
- **A draft fires nothing**, however bad its checks look — the watch parks on
  drafts for the same reason.
- **Pending checks are not ready.** Reporting `ready_to_merge` while checks are
  still running would fire on every pull request seconds after it opened.
- **A digest with no `head_sha` cannot fire.** Without it the fire cannot be
  bounded, and re-firing every sweep is worse than not firing.

## Tests

Six new. The two that matter are the ones a plausible wrong implementation
still passes half of: a trigger fires once per head **and again on the next**
(keying only on the trigger id fires once and then goes silent forever), and
deleting a trigger clears its fires. `conflicts_outrank_failing_checks` and
`pending_checks_are_not_ready` pin the precedence above.

The repo's own guards caught two real mistakes while building this —
`code_trigger_fire` was missing its `owner` column, and three ops took no
owner. Both are fixed rather than exempted.

## Cost

Baseline edit, so `DESKTOP_SCHEMA_EPOCH` goes 39 → 40 and the pinned schema
fixtures are regenerated.

Ran locally: `cargo test -p tidebreak-core --lib` (728 pass),
`cargo clippy -p tidebreak-core --all-targets -- -D warnings`,
`cargo fmt --all --check`.


naingthet:
**G2G** at `1dec45bf7331`.

`uq_code_trigger_rule` is `(owner, repo_id, condition)`, so two principals can arm the same condition on the same repository without colliding. `save_trigger` conflicts on that triple and only updates `action`, `enabled`, and `updated_at` — it cannot retarget `repo_id` or steal another owner's row. List and delete filter on the acting owner; fires denormalize the same owner and claim on `(trigger_id, workspace_id, head_sha)`.

The row stores a condition and a `deliver`/`notify` action, not a session id. Delivery into a workspace session is not a stored destination on this slice, so there is no column that would fire into someone else's conversation. Nothing sweeps yet, so the remaining owner/workspace pairing lives in the next slice.

PR #2471, opened 2026-08-21
feat(ui): fold code-mode thinking and group its tool calls

naingthet:
## What changed

A running code session filled the viewport with open reasoning blocks and one row per tool call, pushing the answer off the bottom: reasoning now arrives folded, and `CodeSessionReducer` settles a block when the next call or the answer starts rather than only at turn end, so a turn no longer shows four blocks all pulsing and all claiming to be the live one. A contiguous run of tool calls collapses behind one line naming its newest call plus a count, and opens to the same rows — a run of one stays a plain row, a run holding a failure opens itself, and a call an approval is parked on is held out so the card still hangs off a visible command. `MiddleTruncate` also dropped the space when its split landed on one, rendering `cargo test -p tidebreak-server code::recovery` as a single word.

## Validation

`pnpm --dir crates/tidebreak-desktop/ui test` (217 files, 1634 tests), `biome format`/`lint`, `tsc --noEmit`, and `storybook:build`; the new `Code/Activity group` story was rendered and reviewed in its live, settled, failed, and two-call states.

## Compatibility and risk

None — presentation only. No wire, persisted-format, or platform changes.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/ThinkingAccordion.tsx
+++ b/crates/tidebreak-desktop/ui/src/ThinkingAccordion.tsx
@@ -25,12 +25,23 @@ import { cn } from "@/lib/utils";
 export function ThinkingAccordion({
   text,
   streaming,
+  expandWhileStreaming = true,
 }: {
   text: string;
   /** Reasoning is still arriving and no answer has started. */
   streaming: boolean;
+  /**
+   * Whether live reasoning opens itself.
+   *
+   * A surface that pauses to think several times in one turn — code mode does,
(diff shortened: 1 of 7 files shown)


naingthet:
**G2G** at `d06f88bd63b2`.

A parked approval pops the last call out of the run before the rest collapse, so the card still hangs off a visible command. A group that holds a failed or denied call opens itself and stays open until you close it. Code mode keeps thinking folded (`expandWhileStreaming={false}`), and `assistant_delta` / parent `tool_started` settle the live reasoning block so it does not keep pulsing over the answer.

PR #2472, opened 2026-08-21
feat(code): open terminals as center tabs, many at a time

naingthet:
The `+` menu offered "Terminal" beside entries that open a center tab, but it opened a bottom drawer instead. A workspace could show one shell however many the server would run, and the drawer attached to whichever shell it found first.

## What changes

The menu leads with the three things a workspace can hold many of — **New agent**, **New terminal**, **New browser** — then a separator, then the views onto the worktree: Open file…, All changes, Source control, Pull request. The split is a rule rather than an ordering.

Terminals are center tabs. One drags, reorders, moves into the split, and answers the tab chords exactly like a file tab, because it is now an editor tab in `codeChrome.ts` rather than a type carved out of the tab model.

Each tab names its own shell, so several run side by side, numbered `Terminal 1`, `Terminal 2`… to stay tellable apart.

## Two consequences worth reading

**Closing a terminal tab ends its shell.** The server caps a workspace at eight, and a tab is the only handle the reader has on one — a close that only hid the tab would spend that budget with nothing left to reclaim it. Navigating away does *not* end it: a build keeps running, and the tab's address names the shell, so coming back re-attaches with its output intact.

**Cmd+J jumps rather than toggles.** It moves focus to the terminal, creating one if there is none, and the next press returns focus where it was — the flick there and back the drawer was good for, without a second kind of surface. Starting a shell is a server call, so the chord raises the ask through `CodeUiStore` and the workspace page answers it, the way the quick-open chord already works.

`TerminalDrawer` and its per-workspace height store are deleted. A link naming a terminal still opens one: the pane adopts a shell and the address rewrites itself to name it.

## Verification

`tsc`, biome format and lint, and the full UI suite pass (1628 tests). Storybook builds. Two new stories cover several shells at once and the menu open; the strip and the menu were rendered and reviewed.


naingthet:
**NO GO** at `b90680c0fcf0`.

A nameless terminal tab (`panel.terminalId` unset — the old `?tabs=terminal` link) does not share `panelKey` with a named shell. `TerminalPane` then `adoptOrCreate`s the first live PTY, and `adoptCodeTerminalId` writes that id onto the nameless tab without checking whether another tab already owns it. Primary and split can then both mount `TerminalPane` on the same id, so keystrokes go out twice.

`openCodeEditor` / URL dedupe only collapse matching keys, so they never catch this. Close-id diffs, `closeFocusedCodeTab`, and the deleted drawer look fine.

Fix: if the adopted id is already in the layout, focus that tab (or mint a new shell) instead of cloning the identity.

Issue #2473, opened 2026-08-22
test(server): the inbox route test races its own turn claim


naingthet:
`tests::lifecycle::the_inbox_lists_parked_work_until_its_own_route_resolves_it` fails most runs locally, on unmodified `main`:

```
main, six consecutive runs:  FAILED FAILED FAILED ok FAILED FAILED
```

It fails in setup, not in the assertions. Each `park_*_for_route_test` helper calls `accept_turn` and then immediately `claim_turn_run`, and the claim comes back with no turn:

```
called `Option::unwrap()` on a `None` value
  crates/tidebreak-server/src/tests/lifecycle.rs  (the `.turn.unwrap()` in a park helper)
```

Two things point at a race rather than a logic error. The helper that fails moves between runs — sometimes the question helper, sometimes the plan one. And adding an `eprintln!` between the claim and the unwrap changed which helper failed, which only happens when the outcome depends on timing.

The likely mechanism: `accept_turn` sets `available_at` from the wall clock, `claim_turn_run` filters on `due_turn_candidate_condition(now)` with its own `now`, and both go through `canonical_db_timestamp`. A claim landing in the same tick as the accept can see the turn as not yet due.

CI is greener than local here, so this has been surviving on timing. Worth fixing as a condition rather than a coincidence — have the helpers claim the turn they just accepted, or make `available_at` explicit rather than implicit in `now`.

Third timing-sensitive test found today, after #2452 and #2467.

Issue #2467, opened 2026-08-21
test(server): the code terminal cursor test fails under parallel load


naingthet:
`tests::code_terminals::create_write_read_and_two_cursors` fails intermittently when the full `tidebreak-server` lib suite runs, and passes every time in isolation:

```
cargo test -p tidebreak-server --lib
  test result: FAILED. 1199 passed; 1 failed

cargo test -p tidebreak-server --lib code_terminals::create_write_read_and_two_cursors
  test result: ok. 1 passed
```

Seen twice on unrelated branches (#2450, #2466), so it is not tied to a change.

A PTY-backed test that only fails when the machine is busy is usually asserting on output within a window rather than on a condition, which makes the threshold a function of load rather than of behavior. Worth checking whether the read waits for the expected bytes or for a fixed duration before comparing cursors.

Lower priority than #2451 and #2452 — this one looks like a genuine timing assumption in the test, where those two describe leaked state and a lease race.

PR #2474, opened 2026-08-22
feat: one inbox over chat and code

### LINKED ISSUE #2314: feat: one attention surface for chat and code

naingthet:
Closes #2314. Last of the stack, after #2460, #2463, and #2466.

## What

The queue becomes conversations rather than parked calls keyed on `ChatId`. `GET /inbox` returns one entry per conversation that wants the reader:

```
{ conversation: {surface: "chat", chat_id}    | {surface: "code", session_id, workspace_id},
  title, attention, items[], waiting_since }
```

- **Chat entries** derive attention from the rows the inbox already projects, and carry the parked calls behind it — which is what a deep link needs to reopen the transcript where it stopped.
- **Code entries** read attention off the session, and are listed only when the state asks for a person. `Working` and `Idle` stay out of the reader's way.

`InboxItemKind` and the resolution routes are unchanged: an item is still answered through the card that owns it, so resolving stays first-responder by construction.

## The tag, and why it is there

The conversation reference is tagged because chat and code ids are still separate spaces — `chat_and_code_entities_do_not_cross_reference` still holds, and retiring it is step 5.

I said earlier I would stop and say so if step 3 turned out to need the id merge. It doesn't, but it does need *something* to name which space an id belongs to, and this tag is it. It is deliberately the only place either surface's shape shows through, and it is what step 5 collapses.

## A bug this surfaced

The route first took `ScopedCode`, whose extractor rejects the whole request when code mode is unconfigured. That would have made a chat-only deployment answer 500 for its own inbox. It now reads the runtime from state, so absent code mode means the queue is chats only.

## Client

`listInbox` returns entries; the store keys on conversation identity rather than call id; the view renders one row per conversation, showing `+N more` when several calls are parked behind it, and navigates to the chat or the code workspace. Parsers reject an unknown surface, and still reject smuggled detail at both the entry and item level.

## Tests

`tidebreak-core` 720, `tidebreak-server --lib` 1198, UI 1628. Clippy, rustfmt, `tsc`, `biome check` clean.

Two known-flaky tests fail locally under load and are unrelated: `code_terminals::create_write_read_and_two_cursors` (#2467) and `configuration::api_key_put_configures_it_and_delete_reverts`, which passes 4/4 in isolation.

**One caveat worth reviewing:** `the_inbox_lists_parked_work_until_its_own_route_resolves_it` is itself flaky on `main` — six consecutive runs there gave `FAILED FAILED FAILED ok FAILED FAILED`. My updated version passes at the same rate. It races its own turn claim in setup, before reaching any assertion; filed as #2473. If CI goes red on that test, that is why — but it is worth confirming rather than assuming, since I updated its assertions in this PR.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/AppShell.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/AppShell.dom.test.tsx
@@ -107,15 +107,29 @@ const desktopUpdateState = vi.hoisted(() => ({
 const restartDesktop = vi.hoisted(() => vi.fn(async () => {}));
 
 /** One waiting question, as the shell's cross-chat read returns it. */
+/** One chat-surface inbox entry holding a single parked question. */
 function parked(chatId: string, callId: string) {
   return {
-    chatId,
-    chatTitle: null,
-    turnId: "turn-1",
-    callId,
-    kind: "question" as const,
-    action: null,
-    requestedAt: "2026-08-04T00:00:00Z",
+    conversation: { surface: "chat" as const, chatId },
+    title: null,
+    attention: {
+      state: {
+        type: "needs_you" as const,
+        prompt: "waiting",
+        source: "structured" as const,
+      },
+      source: "structured" as const,
+    },
+    items: [
+      {
+        turnId: "turn-1",
+        callId,
+        kind: "question" as const,
+        action: null,
+        requestedAt: "2026-08-04T00:00:00Z",
+      },
+    ],
+    waitingSince: "2026-08-04T00:00:00Z",
   };
 }
 const postMessage = vi.fn(async () => {});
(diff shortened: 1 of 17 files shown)


naingthet:
**G2G** at `50b133089028`.

Inbox chat rows come from `list_inbox_items_scoped` plus `chat_attention_scoped`, so another owner never appears. Code rows go through `ScopedCode::for_owner` and `list_sessions` for that owner; `Working` and `Idle` stay out via `wants_the_reader`, and `NeedsYou` / `Stalled` / `Fenced` / `DoneUnreviewed` / `Manual` stay in. The client rejects an unknown surface or a bad item, and `listInbox` fails the whole payload rather than keeping a half-parsed queue. `idle` is not a listed attention, so it does not wipe a valid queue.

PR #2475, opened 2026-08-22
fix(code): a terminal reports ended only once its output has landed

### LINKED ISSUE #2467: test(server): the code terminal cursor test fails under parallel load

naingthet:
## What

`tests::code_terminals::create_write_read_and_two_cursors` fails intermittently under the full `tidebreak-server` lib suite and passes every time in isolation (#2467). Under full CPU load on this machine it fails 13 runs out of 15, always at `shell did not exit after the marker`.

Two threads own a terminal's end. `start_reaper` waits on the child and flips `ended` the moment the shell exits; `start_reader` drains the PTY and flips it again at EOF. The PTY still holds whatever the shell wrote on its way out, so the reaper wins that race whenever the reader is slow to be scheduled — which is what a busy machine makes common.

Losing that race costs bytes. The reader checked `ended` *after* its read returned and broke out without writing the buffer, so the tail of the client's own command went in the bin. It also made `ended` a lie on the read route: `ended` is what tells a client to stop polling, and it arrived while the cursor could still advance.

## How

Split the two facts the one flag was carrying.

- `ended` still means the shell is gone. That is what a snapshot reports and what `write` and `resize` refuse on, so a dead shell rejects keystrokes the moment it dies.
- `producing` says a PTY reader thread is still feeding the ring. A **read** reports `ended` only once both have settled, so a client that stops polling there has seen every byte, and its cursor is final.

The reader now writes what it read before deciding whether to stop, and stops early only on an explicit close — where the ring is being dropped anyway, and where the alternative is filling one no reader can reach. The reaper leaves `producing` to the reader and says why.

Deliberately *not* changed: the snapshot keeps reporting the shell's own state rather than the drain. A background job that outlives the shell holds the PTY open, and gating the snapshot too would leave that terminal showing as live forever while every write 409s.

The test's own deadline was the other half of the flake. It bounds a loop that already exits on the condition it waits for, and three seconds is not enough to start the developer's `$SHELL` — which sources their rc files — alongside a thousand other tests. Thirty costs nothing on a healthy machine.

## Tests

Two unit tests force the interleaving that only load produced, driving `start_reader` from a scripted source instead of a PTY:

- `reader_keeps_output_the_shell_wrote_before_it_was_reaped` — the reaper flips `ended`, then the last line arrives. Without the fix the read returns `[]` where the tail should be.
- `a_terminal_is_not_finished_while_its_reader_is_still_draining` — a read does not claim `ended` with bytes in flight, while the snapshot already reports the shell gone.

Both fail on `main` and pass here.

End to end, with every core saturated: `tests::code_terminals` fails 13/15 on `main` and 0/15 with this change, plus 0/12 on a re-run after narrowing the snapshot back. Full `cargo test -p tidebreak-server --lib`: 1202 passed.

Closes #2467

### DIFF
--- a/crates/tidebreak-server/src/code/terminal.rs
+++ b/crates/tidebreak-server/src/code/terminal.rs
@@ -63,7 +63,15 @@ struct LiveTerminal {
     ring: ByteRing,
     cols: u16,
     rows: u16,
+    /// The shell is gone. Nothing can be typed at it any more, which is what
+    /// a snapshot reports and what `write` and `resize` refuse on.
     ended: bool,
+    /// A PTY reader thread is still feeding this ring, so bytes can still
+    /// arrive after `ended`. The reaper flips `ended` the moment the child
+    /// exits, which is before the reader has drained what the shell wrote on
+    /// its way out. A read reports `ended` only once both have settled, so a
+    /// client that stops polling there has seen every byte.
+    producing: bool,
     created_at: DateTime<Utc>,
     writer: Option<Box<dyn Write + Send>>,
     master: Option<Box<dyn MasterPty + Send>>,
@@ -162,6 +170,7 @@ impl TerminalHub {
             cols,
             rows,
             ended: false,
+            producing: true,
             created_at: Utc::now(),
             writer: Some(spawned.writer),
             master: Some(spawned.master),
@@ -197,6 +206,7 @@ impl TerminalHub {
             cols,
             rows,
             ended: false,
+            producing: false,
             created_at: Utc::now(),
             writer: None,
             master: None,
@@ -258,7 +268,10 @@ impl TerminalHub {
             next_cursor,
             overflow,
             truncated,
-            ended: live.ended,
+            // `ended` is what stops a client polling, so it has to mean the
+            // cursor is final — not just that the shell is gone while its
+            // last line is still on its way out of the PTY.
+            ended: live.ended && !live.producing,
         }
     }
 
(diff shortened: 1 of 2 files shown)


naingthet:
**G2G** at `c6115380f0e7`.

Read `ended` is now `live.ended && !live.producing`, so a client that stops polling there has seen the ring after the reader wrote its last chunk. The reaper still flips `ended` on `wait` and leaves `producing` alone; the reader writes the buffer before it looks at `producing`, and only `close` clears that flag early because the hub is about to drop the ring.

A snapshot still reports the shell, not the drain, so a write 409s as soon as the child is gone. A background job that keeps the slave open will leave `/read` unended until EOF — that is the same PTY lifetime as before, not a new hang, and it is the trade this change names.

The scripted reader tests force the reaper-first interleaving the full suite only hit under load. The e2e deadline is only a bound on a loop that already exits on marker plus `ended`.

PR #2482, opened 2026-08-22
refactor(ui): put one workspaces header on the code rail

naingthet:
## What changed

The code rail carried a full-width "New workspace" button stacked over a repo
switcher, and by-repo group headers linked to a repo page that only relisted the
workspaces the rail already draws. All of it collapses into one header —
"Workspaces", then list settings, add repo, and new workspace as icon buttons —
and `/code/r/$repoId` is gone along with `CodeRepoPage` and `RepoSwitcherPopover`.
The four callers that pointed at that route now go somewhere real: registering or
cloning a repo opens the new-workspace dialog with that repo picked, a repo row on
`/code` does the same, archiving the workspace you are viewing returns to `/code`,
and Cmd+N no longer reads a repo out of the path.

## Validation

`pnpm exec tsc --noEmit`, `biome format`/`biome lint`, and `vitest run src/code`
(536 tests) plus `src/AppShell.dom.test.tsx` all pass. I also rendered the rail
from the Code/Workspace page story in light and dark at 1440px to check the new
header.

## Compatibility and risk

None — no persisted or wire-format change. A bookmarked `/code/r/<id>` URL no
longer resolves, which only affects hash history inside the desktop app.


naingthet:
**G2G** at `d6a408d6c706abd532ac1b5ed86419c51eaa4f2a`

Callers that used `/code/r/$repoId` now open the new-workspace dialog or `/code`. Removing the repo page and switcher matches that, and nothing in the remaining path still depends on a repo id from the URL.

naingthet:
**G2G** at `5366b2bb5a533d674a94337996b2fd4ccd5732ed`

Rebase looks consistent: `/code/r/$repoId`, `CodeRepoPage`, and `RepoSwitcherPopover` are gone, and the four former callers (add/clone repo, `/code` repo row, archive-while-viewing, Cmd+N) now open a workspace or `/code` instead. No leftover navigate-to-removed-route that would fail at runtime without CI catching it.

Issue #2479, opened 2026-08-22
feat(server): deliver a trigger fire into a session


naingthet:
Part of #2477. Depends on the sweep slice.

Consume fires. A claimed fire reaches the agent, or raises a notification, and
the transcript says why.

## Rules from record 0060

- **Target** is the workspace's most recently active interactive session. Watch
  sessions are never a target — a watch is already acting on the same facts.
- **Delivery is steer where the harness declares the capability, and a sweep
  retry otherwise.** Code sessions have no `queued_turn` table, so the fallback
  is the watch's idiom: hold the fire and submit on a later tick when no session
  in the workspace is running a turn (`watch.rs` has the contention guard).
  Claude Code takes that path today — see #2465.
- **Every fire writes a journal event**, and the delivered message names the
  trigger that fired and the fact that fired it. It must never read as the user
  speaking.
- **Message content follows `fix_turn_instruction`'s discipline**: check names,
  buckets, and URLs, never raw logs.
- The notification action raises attention rather than submitting anything.

## Done when

A fire against a session on a steering-capable harness arrives mid-turn; a fire
against a Claude Code session arrives as a turn on a later tick once the
workspace is quiet; both leave a journal event naming the trigger and the fact;
and neither is submitted while another session in the workspace holds the
worktree.

Issue #2477, opened 2026-08-22
epic: user-defined triggers on pull-request facts


naingthet:
Let a user declare that when something happens on a pull request — checks fail,
review is requested, conflicts appear — Tidebreak delivers a message to the
agent working that workspace, instead of the agent polling GitHub for it.

Design is settled in
[decision record 0060](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0060-triggers-are-durable-rules-on-pull-request-facts.md).
A trigger is a durable row modelled on `code_watch`, its conditions are
enum-shaped edge transitions of `PullRequestDigest` fingerprinted against
`head_sha`, and v1 has exactly two actions: deliver a bounded message to a
session, and raise a notification.

## Landed

- #2468 — the decision record.
- #2470 — `code_trigger` and `code_trigger_fire` rows plus
  `classify_trigger_condition`. Storage and classification only; nothing fires.

## Slices

Order matters: the sweep produces fires, delivery consumes them, and the
interface arms the rules the sweep reads.

- [x] #2478 — sweep pull-request facts and fire triggers
- [x] #2479 — deliver a trigger fire into a session
- [x] #2480 — arm and manage triggers
  - [x] server routes (#2486, corrected by #2493)
  - [x] types on the wire (#2495)
  - [x] the rules component and its stories (#2496)
  - [x] #2503 — render it on the notifications page
- [x] #2481 — move the delivery notification rules onto the substrate
  - [x] the mapping and its tests (#2502)
  - [x] #2504 — run it once, then retire the client rules

## Out of scope

Shell commands and webhooks as actions, scheduled and recurring triggers,
triggers on chat conversations, rerunning a failed check before waking an
agent, and re-expressing watch-and-fix as a built-in trigger. Each is named in
record 0060 with the condition that would start it.

Claude Code mid-turn steering is a separate concern: the capture in #2465 shows
it works only at a tool-call boundary and that a miss desynchronizes the
session, so Claude sessions take the sweep-retry delivery path until an adapter
can bound the miss.

PR #2483, opened 2026-08-22
feat(server): sweep pull-request facts and fire triggers

### LINKED ISSUE #2478: feat(server): sweep pull-request facts and fire triggers
Part of #2477. Depends on #2470, which landed the rows and the classifier.

Produce fires. A try-based sweep reads enabled triggers from the table each
tick, classifies the pull-request facts for the workspaces they cover, and
claims a fire once per `head_sha` per trigger.

## Shape

Structure it on `sweep_watches` (`crates/tidebreak-server/src/code/watch.rs`),
armed from an `ensure_trigger_sweep` beside `ensure_watch_sweep` in
`runtime.rs`. One failing trigger never stops the others.

**Do not copy the watch's data source.** The watch calls
`refresh_workspace_pr` per watched workspace, which invalidates the 20-second
`PR_CACHE_TTL` entry and forces a `gh` read. Triggers bind per repository
(record 0060), so doing that per workspace would spawn one `gh pr view` per
workspace per tick — a rate-limit problem on a machine carrying many
worktrees, and the opposite of the shared-cache position the record takes.

Use `query_pull_requests` (`crates/tidebreak-server/src/code/delivery.rs`)
instead: bulk per repository behind a 30-second list cache.
`CodeDeliveryPullRequestSummary` carries every field
`classify_trigger_condition` reads — `state`, `draft`, `head_sha`,
`review_decision`, `mergeable`, `merge_state_status`, `checks`, `merged_at` —
and its `workspace_links` gives the pull-request-to-workspace mapping without a
second query. The glue is a `From<&CodeDeliveryPullRequestSummary>` for
`PullRequestDigest`, which keeps the record's "edge transitions of
`PullRequestDigest`" wording honest.

## Rules

- Fire on the transition into a condition, once per `head_sha`, not on every
  sweep that still finds it true.
- A trigger firing against an unchanged fact parks rather than retrying,
  following the watch's `last_fix_head`.
- Suppress a fire that overlaps an active watch on the same workspace, or two
  drivers chase one failure.
- Read rows on the sweep; never subscribe. The event bus is lossy `broadcast`.

## Done when

A trigger armed against a repository writes exactly one `code_trigger_fire`
row when a covered pull request enters its condition, writes none on the next
sweep while the fact is unchanged, and writes another once `head_sha` moves and
the condition recurs.

naingthet:
Closes #2478. Part of #2477.

#2470 landed the trigger rows and the classifier with nothing driving them.
This adds the sweep that turns pull-request facts into claimed fires.

## The sweep

Try-based, on its own guard beside the watch's, armed from
`ensure_trigger_sweep`. It reads its work list from the table every tick rather
than subscribing: the event bus is a lossy `broadcast`, and a fact this misses
is a message an agent never gets. One failing repository never stops the
others.

The interval is 53 s against the watch's 47 s. Both sweeps read GitHub, and
landing them on the same tick would double the burst a rate limit sees.

## Why it does not copy the watch's data source

The watch calls `refresh_workspace_pr` per watched workspace, which invalidates
the 20-second `PR_CACHE_TTL` entry and forces a `gh` read. That is right for a
watch, which is a decision about one pull request. Triggers bind per repository
(record 0060), so copying it would spawn one `gh pr view` per workspace per
tick — a rate-limit problem on a machine carrying many worktrees, and the
opposite of the shared-cache position the record takes.

So this groups triggers by repository and takes one bulk `query_pull_requests`
per repository, behind the delivery list cache, never setting `refresh`.
`CodeDeliveryPullRequestSummary` already carries every field
`classify_trigger_condition` reads, and its `workspace_links` give the
pull-request-to-workspace mapping without a second query. `digest_from` is the
only glue, and it needs no case handling because both paths lowercase their
host tokens already.

## Guards

- **No head SHA, no fire.** The fire could not be fingerprinted, and one that
  cannot be bounded would repeat every tick.
- **No fuzzy links.** A fuzzy link is a branch-name guess; firing on one would
  wake an agent about somebody else's pull request.
- **Workspaces under an active watch are skipped wholesale**, or two drivers
  chase one failure. A sweep that cannot read the watches does nothing rather
  than firing beside an unknown one.
- Only `Active` workspaces are targets.

## Scope

Claiming only — `insert_trigger_fire` is the claim, and two sweeps racing one
edge both call it while exactly one gets `true`. Acting on a fire is #2479.

## Tests

Unit tests on the two functions where a bug is silent and reaches a real agent:
the summary-to-digest conversion (including that a merged pull request reads as
`Merged` rather than `Closed`, which `state` alone cannot settle), and the
link filter. The claim itself is already covered by the store test #2470 added.

### DIFF
--- a/crates/tidebreak-server/src/code/delivery.rs
+++ b/crates/tidebreak-server/src/code/delivery.rs
@@ -859,7 +859,7 @@ async fn require_authenticated(runtime: &CodeRuntime) -> Result<GhObservation, S
     Ok(observation)
 }
 
-async fn repository_target_from_local(
+pub(crate) async fn repository_target_from_local(
     repo: &CodeRepo,
 ) -> Result<CodeGitHubRepositoryTarget, String> {
     let remote = git_read(Path::new(&repo.root_path), &["remote", "get-url", "origin"])
--- a/crates/tidebreak-server/src/code/mod.rs
+++ b/crates/tidebreak-server/src/code/mod.rs
@@ -23,6 +23,7 @@ pub(crate) mod session_worker;
 pub(crate) mod setup_script;
 pub(crate) mod terminal;
 pub(crate) mod titling;
+pub(crate) mod trigger;
 pub(crate) mod watch;
 pub(crate) mod worktree;
 pub(crate) mod worktree_orphans;
--- a/crates/tidebreak-server/src/code/runtime.rs
+++ b/crates/tidebreak-server/src/code/runtime.rs
@@ -176,6 +176,8 @@ pub(crate) struct CodeRuntime {
     stall_started: AtomicBool,
     watch_sweep: Mutex<Option<super::watch::WatchSweepGuard>>,
     watch_started: AtomicBool,
+    trigger_sweep: Mutex<Option<super::trigger::TriggerSweepGuard>>,
+    trigger_started: AtomicBool,
     /// Workspaces with a background naming call in flight.
     ///
     /// One call per workspace at a time; a second trigger is dropped rather
@@ -245,6 +247,8 @@ impl CodeRuntime {
             stall_started: AtomicBool::new(false),
             watch_sweep: Mutex::new(None),
             watch_started: AtomicBool::new(false),
+            trigger_sweep: Mutex::new(None),
+            trigger_started: AtomicBool::new(false),
             titling_in_flight: Mutex::new(std::collections::HashSet::new()),
         }
     }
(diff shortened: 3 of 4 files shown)


naingthet:
**NO GO** at `af0c428f83394cbed09cd32ab7f52fd2640110e4`

P1: `sweep_repo` calls `query_pull_requests` with `limit: None` and never follows `next_cursor`. That helper still paginates: `limit.unwrap_or(50)` and a hard cap of 100. A repository with more than fifty Tidebreak-linked pull requests (sorted by `updated_at`) never has the rest classified or claimed. The sweep is the only driver, so those edges stay silent until the PR ages into the first page. Walk the cursor, or take the cached aggregate without going through the paged filter, so one tick sees every linked PR for that repo.

PR #2484, opened 2026-08-22
feat(server): deliver a trigger fire into a session

### LINKED ISSUE #2479: feat(server): deliver a trigger fire into a session

naingthet:
Closes #2479. Part of #2477. **Stacked on #2483** — the first commit here is
that PR's; the diff collapses once it merges.

The sweep claimed fires with nothing consuming them. This delivers one.

## Delivery is settled before the fire is claimed

This is the part worth reading twice. `code_trigger_fire` has no column
marking a fire drained, and the claim is fingerprinted against `head_sha`. So
claiming on a tick that could not deliver would burn the edge and the message
would never arrive — the condition staying true on the next tick is not a
second edge, and `insert_trigger_fire` would return `false` forever after.

Holding the claim back until delivery is settled is what makes record 0060's
"retry on a later tick" work without a schema change. The sweep re-derives the
condition from current facts every tick, so a fire deferred for a busy
workspace is not lost — it is simply claimed later.

## Who receives it

The workspace's most recently active interactive session. Recency is the last
turn the session ran, falling back to when it was created, because a session
row carries no activity timestamp of its own. Watch sessions are never a
target: a watch is already acting on the same facts, and delivering to it
would put two drivers on one loop.

## How it arrives

- **Quiet workspace** → a turn.
- **Busy workspace** → steering, and only where the adapter declares
  `mid_turn_steering` as `Supported`. Otherwise the fire stays unclaimed and a
  later tick delivers it. Claude Code sessions take that path today, per the
  capture in #2465.
- **Notify** → attention, regardless of whether the workspace is busy, because
  attention never touches the checkout.

## Saying where it came from

Every fire journals a `HarnessNotice` naming the trigger and the fact — a
notice rather than a variant of its own, following `note_permission_mode`,
which the journal already uses for lines no harness produced. The delivered
message says outright that nobody typed it, and carries the pull request,
its URL, and failing check names. Content discipline follows
`fix_turn_instruction`: check names, buckets, and URLs, never raw logs.

It also repeats that merging, auto-merge, and draft or review state stay the
user's, per record 42.

### DIFF
--- a/crates/tidebreak-server/src/code/trigger.rs
+++ b/crates/tidebreak-server/src/code/trigger.rs
@@ -18,11 +18,14 @@ use std::time::Duration;
 
 use chrono::Utc;
 use tidebreak_core::db::code::{
-    get_repo, insert_trigger_fire, list_active_watches_all_owners, list_enabled_triggers_all_owners,
+    get_open_turn, get_repo, get_session, insert_trigger_fire, latest_turn,
+    list_active_watches_all_owners, list_enabled_triggers_all_owners, list_sessions_for_workspace,
 };
 use tidebreak_core::{
-    classify_trigger_condition, CodeTrigger, CodeTriggerFire, CodeWorkspaceStatus, OwnerId,
-    PullRequestCheck, PullRequestDigest, RepoId, WorkspaceId,
+    classify_trigger_condition, Attention, AttentionSource, CapLevel, CodeEvent, CodeSession,
+    CodeSessionId, CodeSessionKind, CodeSessionLifecycle, CodeTrigger, CodeTriggerAction,
+    CodeTriggerCondition, CodeTriggerFire, CodeTurnId, CodeWorkspaceStatus, HarnessNoticeLevel,
+    OwnerId, PullRequestCheck, PullRequestDigest, RepoId, WorkspaceId,
 };
 use tracing::{debug, warn};
 
@@ -432,9 +435,7 @@ async fn note_fire(
     trigger: &CodeTrigger,
     digest: &PullRequestDigest,
 ) {
-    let Ok(Some(session)) =
-        tidebreak_core::db::code::get_session(&runtime.db, owner, session_id).await
-    else {
+    let Ok(Some(session)) = get_session(&runtime.db, owner, session_id).await else {
         return;
     };
     let _ = journal_event(
--- a/crates/tidebreak-server/src/code/trigger.rs
+++ b/crates/tidebreak-server/src/code/trigger.rs
@@ -6,7 +6,11 @@
 //! event bus is a lossy `broadcast`, and a fact this misses is a message an
 //! agent never gets.
 //!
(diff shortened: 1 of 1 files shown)


naingthet:
**NO GO** at `a85be425ecd24427c7d9cfdd645bd6a7b648676a`

`fire_one` claims the fire (`insert_trigger_fire`) after `plan_delivery` succeeds, then calls `steer` / `submit_turn`. Those can still fail. The claim is fingerprinted on `head_sha` and there is no drained column, so a failed act burns the edge: the condition staying true is not a second insert, and the message never arrives.

That is the failure the module comment and record 0060 say this PR is written to avoid. Planning is not delivery. Either act first and claim only on success, or treat a post-claim act failure as something that must still be delivered (a drain flag, or an unclaim). As written, a transient `steer`/`submit_turn` error permanently drops the fire.

`Notify` swallowing `apply_attention` after a successful claim has the same shape, quieter: attention never retries for that head.

naingthet:
**G2G** at `d1cb880f9322a7ff7faccdf6e81f11b8a1f52257`

`plan_delivery` now returns `None` before `insert_trigger_fire` when the workspace is busy and mid-turn steering is unsupported, the target is not running, or there is no open turn. That is the earlier burn: a tick that cannot deliver no longer fingerprints `head_sha`, so a later tick can still claim.

The remaining claim-then-`steer`/`submit_turn` window is after delivery is settled, which matches the module contract.

naingthet:
**G2G** at `1653974521f2bf446e329bb4bd33f0a81d0279b2`.

Re-reviewed after the rebase. The new commit only widens the `trigger.rs` imports so delivery types and store helpers resolve. Claim-after-plan, interactive-only targeting, and busy-workspace deferral are unchanged.

PR #2485, opened 2026-08-22
fix(code): the create dialog, the model picker, and image pastes

naingthet:
Four things a code session would not let you do, reported from one opencode session.

## Plain Enter created a workspace

Title and Base ref are single-line inputs inside the New workspace dialog's form, so Enter on either submitted it. A keystroke meant to end typing cut a worktree, and the create it started left Cmd+Enter looking dead for the rest of the dialog's life — which is how this was reported. Enter is field-local now, the way the dialog's own docstring already claimed. Cmd+Enter was never broken: React routes it here even from a picker's portal.

## The model picker showed one row

It opened on the current model's vendor block and left every other vendor behind the unlabelled icon rail. Claude Code lists one vendor and never noticed; opencode lists a mixed catalog, so the menu looked like a model that could not be changed. A mixed catalog now opens on a new All entry with every row in it, the rail still narrows to one vendor, and the row the session is on is scrolled into view.

## A pasted image left a file URL in the draft

Only Claude Code's adapter puts `TurnInput::images` on the wire; codex, opencode, and grok drop them. The composer gates its image path on `caps.image_input`, so on those three a paste was not claimed and the clipboard's text flavour — a `file://` URL — landed in the draft, reading as an attachment that had worked. The paste is claimed and answered instead: "Tidebreak can't send images to opencode yet."

Making it actually work is adapter-side and gated on a captured image round-trip per engine (`no_adapter_declares_image_input_without_an_image_fixture`). Not in this PR.

## A mode change opencode never took

opencode takes its agent and permission ruleset on `POST /session`; resuming is a plain `GET /session/{id}`, and 1.18.18 captured nothing that re-applies either. It has no in-place mode switch, so the runtime relaunched — onto the stored resume ref, which came back running the posture the session was created with while the row, the picker, and the journal all said the new mode had taken. Dial down to Plan and the engine stayed in Allow.

`HarnessAdapter` now states whether relaunching composes a mode changed after the session started. It defaults true, because every other engine takes its posture from the launch plan; opencode says false. The runtime refuses the change once the session has a resume ref rather than recording one the engine never took. A session that has not opened yet still moves, since its relaunch creates rather than resumes.

## Allow all is the create default from the CLI too

`tidebreak code`'s `default_create_permission_mode` walked Ask → Auto → Plan and its doc comment claimed to be the desktop default. It never picked up the [decision 0039 amendment](docs/decisions/0039-allow-is-a-first-class-code-permission-mode.md#amendment-2026-08-18), which made the default the most autonomous posture the engine honors. Every engine declares `allow_mode: supported`, so CLI sessions opened in Ask while desktop sessions opened in Allow all. It walks Allow → Auto → Ask → Plan now, and `resolve_start_mode` already had the sentence that states it.

## Tests

Each fix has a test that fails without it: the dialog keeping Enter out of create, the mixed catalog opening on every vendor, the paste being answered, the refused mode change not moving the row, and the CLI default. New `HarnessModelMenu` story covers single-vendor, mixed, field, loading, and set-at-session-start.

### DIFF
--- a/crates/tidebreak-cli/src/code.rs
+++ b/crates/tidebreak-cli/src/code.rs
@@ -779,13 +779,14 @@ async fn resolve_run_session(
         .ok_or_else(|| AgentError::msg(format!("workspace {workspace} has no active session")))
 }
 
-/// Desktop create default: Ask when the engine can honor it, otherwise the
-/// most autonomous non-bypass mode. Allow is an explicit `--mode`.
+/// Desktop create default: the most autonomous posture the engine honors,
+/// walking Allow -> Auto -> Ask -> Plan (decision 0039, amended 2026-08-18).
+/// `resolve_start_mode` states whichever posture that is before the turn runs.
 fn default_create_permission_mode(caps: Option<&HarnessCaps>) -> PermissionMode {
     match caps {
-        Some(caps) if caps.structured_approvals == CapLevel::Supported => PermissionMode::Ask,
+        Some(caps) if caps.allow_mode == CapLevel::Supported => PermissionMode::Allow,
         Some(caps) if caps.auto_mode == CapLevel::Supported => PermissionMode::Auto,
-        Some(caps) if caps.plan_mode == CapLevel::Supported => PermissionMode::Plan,
+        Some(caps) if caps.structured_approvals == CapLevel::Supported => PermissionMode::Ask,
         _ => PermissionMode::Plan,
     }
 }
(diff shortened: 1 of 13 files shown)


naingthet:
**G2G** at `fa030a6b9dfd6f0144f6684611582bf81231945e`

The four fixes match the reported failures: Enter no longer submits the create form, a mixed catalog opens on every vendor, an image paste on a no-image engine is claimed instead of becoming a file URL, and a mode change that a relaunch cannot carry is refused without moving the row. Tests cover each of those. Nothing here is a high-impact miss CI would not catch.

Issue #2480, opened 2026-08-22
feat: arm and manage triggers


naingthet:
Part of #2477. Depends on #2478 and #2479.

Routes and the rules component have landed (#2486, #2493, #2495, #2496). Remaining work is #2503: mount that surface on the notifications page so a user can arm a trigger outside a test.

## Shape

- Owner-scoped routes to list, create, enable, disable, and delete a trigger,
  following the existing code-mode route shapes in
  `crates/tidebreak-server/src/routes/code/`.
- Triggers bind per repository and apply to workspaces that have a pull
  request, so the arming surface is repository-shaped, not workspace-shaped.
- **Name the target session when the trigger is armed.** Workspaces hold
  several sessions (record 55), so the rule needs to be visible before it
  fires.
- **Say which delivery path a given session is on** rather than implying every
  trigger interrupts. A Claude Code session takes the sweep-retry path; a Codex
  session is steered.
- Storybook stories under `crates/tidebreak-desktop/ui/src/stories/` for the
  states that matter: no triggers, armed, disabled, a repository with no
  workspace carrying a pull request, and the two delivery paths.

## Done when

A user can arm a trigger on a repository from the notifications page, see which session it will reach and how, disable it without deleting it, and watch it fire. That last mount is #2503.

naingthet:
The last slice, #2503, landed as PR #2922. Routes, the rules component, and the notifications-page mount are all on main, so a user can arm a trigger on a repository, see its target session and delivery path, and disable it without deleting it. Closing.

PR #2486, opened 2026-08-22
feat(server): routes to arm and manage triggers

naingthet:
Part of #2480 and #2477. Now rebased onto `main`; #2483 and #2484 have landed,
so this is a single commit.

Until now a trigger row could only be created in a test. These are the routes
behind it.

| Route | Does |
| --- | --- |
| `GET /code/repos/{id}/triggers` | list what is armed |
| `POST /code/repos/{id}/triggers` | arm one |
| `PATCH /code/repos/{id}/triggers/{trigger_id}` | enable or disable |
| `DELETE /code/repos/{id}/triggers/{trigger_id}` | remove |

Owner-scoped through `ScopedCode` like the rest of `/code/*`, and
repository-scoped rather than workspace-scoped: a trigger is a standing
preference about how the user wants to be reached, and record 0060 binds it per
repository.

## Behaviour worth knowing

- **Arming the same rule twice returns the row already there.** One row per
  `(repository, condition, action)`, or a double-armed rule would deliver the
  same message twice.
- **Disabling keeps the row**, so turning a rule back on does not mean
  rebuilding its scoping. That is what the `enabled` column is for.
- **Deleting keeps the recorded fires.** They are what the transcript already
  explained; dropping them would leave journal lines with nothing behind them.
- Listing refuses an unknown repository rather than returning an empty list, so
  a stale id reads as an error instead of "none armed".

## Scope

The server half. The desktop surface is still to come, and it owes two things
record 0060 asks for that only an interface can deliver: naming the target
session when the trigger is armed, and saying which delivery path a given
session is on rather than implying every trigger interrupts.

`CodeTriggerSnapshot`, `CreateCodeTriggerBody`, and `UpdateCodeTriggerBody`
derive `TS` but are not added to the `collect_from` list in `wire_types.rs`, so
`wire.ts` is unchanged. That regeneration belongs with the UI change that first
consumes them.

### DIFF
--- a/crates/tidebreak-server/src/code/runtime.rs
+++ b/crates/tidebreak-server/src/code/runtime.rs
@@ -10,18 +10,19 @@ use chrono::Utc;
 use tokio::sync::oneshot;
 
 use tidebreak_core::db::code::{
-    delete_workspace, get_approval, get_open_turn, get_repo, get_repo_by_root_path, get_session,
-    get_workspace, insert_repo, insert_session, insert_workspace, list_approvals, list_events,
-    list_repos, list_repos_all_owners, list_sessions, list_sessions_all_owners,
-    list_sessions_for_workspace, list_turns, list_workspaces, mark_repo_removed, save_approval,
-    save_repo, save_session, save_workspace, MAX_REPLAY_EVENTS,
+    delete_trigger, delete_workspace, get_approval, get_open_turn, get_repo, get_repo_by_root_path,
+    get_session, get_workspace, insert_repo, insert_session, insert_workspace, list_approvals,
+    list_events, list_repos, list_repos_all_owners, list_sessions, list_sessions_all_owners,
+    list_sessions_for_workspace, list_triggers_for_repo, list_turns, list_workspaces,
+    mark_repo_removed, save_approval, save_repo, save_session, save_trigger, save_workspace,
+    MAX_REPLAY_EVENTS,
 };
 use tidebreak_core::{
     ApprovalDecisionKind, Attention, AttentionSource, CapLevel, CodeApproval, CodeApprovalId,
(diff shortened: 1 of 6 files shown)


naingthet:
**G2G** at `590b995eeeee87eb9e1730385f8f711b38fb2a64`

Owner-scoped list/create/update check the repository first, and the idempotent arm path matches the unique rule the PR describes. Nothing here would change how I ship this slice.

naingthet:
**NO GO** at `d27b131789484c9b60c403b6ff14676e8d1bfc19`

`create_trigger` treats uniqueness as `(repository, condition, action)` and, on a miss, builds a new `CodeTriggerId` then calls `save_trigger`. The write path is unique on `(owner, repo, condition)` and on conflict updates `action`/`enabled` in place. Arming the same condition with a different action therefore returns a snapshot whose `id` was never stored; the existing row keeps its old id and its action is rewritten. PATCH/DELETE against the returned id 404s, and a later GET shows a different id than the create response.

Match the unique key in the read-before-write (condition only), or do not mint a new id when the upsert will hit an existing row. Return that row (and decide whether changing `action` is allowed at all).

naingthet:
**NO GO** at `44827e6e6c16e82ade3b9d55387915a79a4805fa`

P1: `create_trigger` still returns a phantom id when `save_trigger` upserts.

`save_trigger` conflicts on `(owner, repo_id, condition)` and updates `action`/`enabled` without inserting the new `id`. `create_trigger` only short-circuits when **both** `condition` and `action` match. Arming the same condition with a different action (or two concurrent arms of the same rule) still allocates `CodeTriggerId::new()`, calls `save_trigger`, and returns that new id. The row that actually exists keeps the old id.

A 201 then names a trigger that `GET`/`PATCH`/`DELETE` will not find. Match the unique key: look up by condition only, or insert and then reload the stored row instead of returning the unsaved id.

Issue #2321, opened 2026-08-20
feat: enable code mode on a shared hosted machine


naingthet:
[Decision 47](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0047-gateway-linked-hosting.md) gates code mode on a shared machine until `code_*` rows are owner-scoped. The columns and owner-scoped `/code/*` queries are in the schema.

Turn that into the product gate: a second user on the same hosted machine must see neither another owner's workspaces nor their `/code/updates` events. Enable the surface only when that holds.

Do not share repositories, worktrees, or session journals across owners.

naingthet:
PR #2250 completed the shared-machine code-mode isolation gate. It owner-scopes every `code_*` row and `/code/*` route, partitions `/code/updates`, and tests the required two-user behavior on SQLite and PostgreSQL. PR #2487 closed the remaining owner-blind image-attachment path. Closing as completed.

PR #2487, opened 2026-08-22
fix(code): a blob id is not permission to attach an image

### LINKED ISSUE #2312: feat: one turn-submission contract for chat and code
[Decision 48](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0048-one-interaction-model.md) step 4, executing the revisit clause in [decision 46](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0046-rich-turn-input-rides-harness-capabilities.md).

Merge code turn input into the chat attachment model rather than growing a second one. Capability gates decide what an engine accepts (images, slash-command discovery, `@` worktree completion).

Text-only clients stay valid. Do not journal attachment bytes.

naingthet:
Closes #2312 ([decision 48](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0048-one-interaction-model.md) step 4), and closes a missing authority check found while doing it.

## The gap

Code-mode turn attachments resolved on the strength of the bytes existing:

```rust
let Some(meta) = self.blobs.metadata(*blob_id).await? else { ... };
```

The blob store is content-addressed and owner-blind, so that answers *"do these bytes exist anywhere"*, not *"may this session use them"*. A caller who learned a blob id could name it on a turn in a session of their own, and then read the pixels back through that session's `GET /attachments/images/{blob_id}` route.

`resolve_turn_attachments` never even received the session or the owner — the signature is why the check was absent rather than wrong.

Chat has never had this hole. `chat_image_publication` records that a chat reserved an image, and its resolution consults the reservation instead of the blob.

## The fix

`code_session_image` mirrors `chat_image_publication`. Upload writes the reservation; resolution requires it, refusing an unreserved id as **not-found** so a refusal cannot confirm the blob exists elsewhere. Resolution also re-derives the descriptor from the bytes, so a reservation whose content drifted no longer resolves.

## Not reachable today

The desktop profile has a single owner, and code mode on a shared machine is gated behind #2321 — which is precisely the gate that would have opened this. Fixing it before that lands rather than after.

## Why this is also step 4

`CodeTurnAttachment` was `ImageRef` minus `width`/`height`. Those dimensions were computed at upload and thrown away, because resolution rebuilt from blob metadata that does not carry them. Publication carries them, so the merge falls out of the fix: `CodeTurnAttachment` is now `ImageRef`, and `code_turn_attachment` stores the dimensions a renderer needs without reading the bytes back.

## Verification

`a_session_cannot_attach_an_image_published_to_another_session` publishes to one session and attaches from a sibling.

I checked the test actually bites, because it initially passed against a deliberately broken guard for an unrelated reason. With the publication check removed the sibling gets **202 with the image bound**; with it, 400 naming authority. The test asserts on that message, not just the status.

`tidebreak-core` 722, `tidebreak-server --lib code::` pass, UI 1644 pass, clippy, rustfmt, `tsc`, `biome check` clean.

Note for reviewers: my machine currently fails ~12 `tidebreak-core` tests on **unmodified main** (varying per run, resource exhaustion after a long session). CI is the real signal for those.

## Schema

Baseline edit plus `DESKTOP_SCHEMA_EPOCH` 39 → 40, per decision record 2. Both fixtures regenerated. `every_code_table_carries_an_owner_column` caught the new table, which is what it is for.


naingthet:
**G2G** at `372c823b1a783ed853139bea16f31ee438711c99`

Publication is now the capability for a code-mode attach, not blob presence. Resolution is scoped to owner plus session, unpublished and drifted ids share one refusal, and GET still only serves an image already bound on a turn. The sibling-session test is the one that would have passed on the old path. Nothing here is a P1/P2 CI would miss.

naingthet:
**G2G** at `bff6b4bd3970fb9b645906fe1e0f04822fe4cbc9`.

The follow-up only retargets the two existing attachment tests onto the publish path with a real one-pixel PNG. That is the client path the reservation now requires, so they no longer put a blob in the store and name it. The capability test still refuses for the capability, on an image that is actually reserved. No new high-impact issue.

naingthet:
This adds `code_session_image` to the baseline and regenerates both fixtures, but `DESKTOP_SCHEMA_EPOCH` is still 40 — the same value that is on `main`.

An unchanged epoch means `prepare` takes the `saved == SchemaMarker::current()` arm and does not reset, so an existing desktop profile never runs `Baseline::up` again and never gets the table. `reserve_session_image` is on the path this PR makes *required* for resolution, so the first image attachment on an existing profile hits a missing table rather than a permission refusal.

A fresh profile is fine, which is why the suite is green — every test builds its own database.

Fix is the one line, 40 → 41, in `crates/tidebreak-server/src/desktop_schema.rs`.

This is the failure mode decision record 2 names: the fixture makes a baseline edit visible, it cannot make the decision. Regenerating it is what a green run looks like either way.

Not pushing it myself — another session has this branch checked out with uncommitted changes in `crates/tidebreak-server/src/tests/code.rs`, and I would rather not collide.

PR #2489, opened 2026-08-22
refactor(core): make the schema chain ordered instead of guard-driven

naingthet:
The first half of #2428: give the schema an ordered chain, so a change that has
to reach a database the baseline already ran against has somewhere to go.

## What was there

`Baseline::up` branched on `has_table("app")`. On an existing self-host
database it skipped the snapshot entirely and ran `ensure_app_owner` and
`ensure_code_owner` instead — two functions whose ordering lived in their own
`if !has_column(...)` guards.

That is a migration chain wearing a baseline's clothes. It works for one
upgrade. It composes badly for the next one, because nothing records what runs
when, and nothing compares the two paths.

## What this does

`AppOwner` and `CodeOwner` are now appended migrations. The bodies are
unchanged and still idempotent:

- **Fresh database** — the baseline already declares both owner columns and
  both indexes, so each migration finds its work done and does nothing.
- **Pre-squash self-host database** — the baseline still returns early, and the
  two migrations patch exactly what they patched before.

`a_stepwise_upgrade_lands_on_the_fresh_schema` is the check the guards were
standing in for: a database that stopped at the baseline and later took the
rest of the chain must describe the same schema as one built in a single pass.
It reads `sqlite_master`, which stores each `CREATE TABLE` verbatim and appends
what `ALTER TABLE` adds, so it sees a column an appended migration declares
differently — not only a table it forgot to create.

It runs on SQLite while the store it protects is PostgreSQL, and that is
enough: both sides are the same sea-query statements, so a baseline and a
migration that disagree disagree on either backend. What differs between
backends is the rendering, and `the_schema_baseline_is_pinned` already pins
that for both.

## No schema change

Neither fixture moves and the epoch does not bump. Nothing about a fresh
database's schema changes — only which migration name creates it.

## What this does not fix

`Baseline::up` returning early means a pre-squash self-host database never
gets a table added after its own baseline ran. Unfolding the guards does not
change that; it creates the place the fix goes. Closing it needs the schema
pin from #2428, which is the next PR.

### DIFF
--- a/crates/tidebreak-core/src/db/migration.rs
+++ b/crates/tidebreak-core/src/db/migration.rs
@@ -1,12 +1,26 @@
-//! The database schema.
+//! The database schema: a baseline, then an ordered chain.
 //!
-//! Fresh pre-v1 databases are described by a single [`Baseline`]. Desktop
-//! SQLite changes bump `DESKTOP_SCHEMA_EPOCH` so disposable local data is
-//! rebuilt. Self-host databases are durable: a renamed baseline must not
-//! recreate tables that already exist, and any later in-place baseline edit
-//! that must reach an already-recorded schema also gets an ordered upgrade
-//! migration in this module. Squash again before `1.0.0` so that release's
-//! first migration is a clean snapshot, not this public-opening chain.
+//! [`Baseline`] describes a fresh database. Anything that has to reach a
+//! database the baseline already ran against is an appended migration in
+//! [`Migrator::migrations`], in order.
+//!
+//! The ordering is the point. SeaORM records one name per migration and cannot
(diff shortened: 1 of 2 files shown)


naingthet:
**G2G** at `e6d216ba743cdcf139c460913c871c775dd2616b`

The early return on an existing `app` table plus the two appended owner migrations preserve the old durable path: a pre-squash database still skips `CREATE TABLE` and still patches owner columns and indexes under the same `has_column` / `if_not_exists` guards. A database that already recorded this baseline name will treat `AppOwner` and `CodeOwner` as new work and no-op them. The stepwise `sqlite_master` check is the right contract for later appends. No schema pin or epoch change in this slice, as stated.

Issue #2490, opened 2026-08-22
fix(core): a self-host database created before the migration pin has no way back


naingthet:
[Decision 61](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0061-schema-changes-are-migrations.md) makes every schema change an appended migration, so from the pin onward a durable PostgreSQL store stays current. It does nothing for a database created *before* the pin, and that gap is worth writing down.

## What is broken

`Baseline::up` returns early when it finds an `app` table:

```rust
if manager.has_table("app").await? {
    return Ok(());
}
```

That branch exists so `CREATE TABLE` does not re-run against a database that recorded the pre-squash chain. The cost is that such a database never gets a table the baseline gained afterwards. Between 2026-08-14 and 2026-08-21 the baseline gained `code_watch`, `code_trigger`, `code_trigger_fire`, and `code_session_image` — a self-host store from before each one silently lacks it, and finds out at the first query.

SeaORM cannot see this. It records one name per migration, so a database that recorded `m20260814_000001_baseline` never re-runs it no matter how much the statements behind that name changed. The desktop profile was covered by the epoch reset. PostgreSQL has never had an equivalent.

## Why the sidecar does not generalize

`tidebreak-schema.json` sits beside the SQLite file. A server deployment has no such place — the database *is* the state, and the operator may run several against one binary. So the pin has to live inside the database.

## Shape

A `schema_pin` row, written by the baseline and read before the chain runs:

- Pin present and current → run the chain, normal path.
- Pin absent, tables present → a pre-pin database. Refuse to open, and say to recreate it. A loud failure beats a store that is silently missing tables.
- Pin absent, no tables → fresh. Write the pin, run the chain.

Refusing is the honest floor: there is no way to determine what a pre-pin database holds, so there is nothing to migrate it *from*. If that turns out to be too blunt, the alternative is a repair migration that creates any missing baseline table `IF NOT EXISTS` — which works for added tables and not for changed columns, so it would need its own scope.

Adding the row is a baseline edit, which decision 61 freezes. So this rides the next squash — either #2313 or the `1.0.0` one — rather than being appended.

## Not urgent

Self-host is not in production (#2322), so today this is a development-machine problem. It stops being one the moment a deployment is real, and #2321 puts code mode on shared machines, which is where a missing `code_session_image` becomes a user-visible failure rather than a contributor's.

PR #2491, opened 2026-08-22
feat: schema changes are migrations, not database resets

### LINKED ISSUE #2428: feat: switch the schema posture from epoch resets to migrations

naingthet:
Closes #2428. Decision record 61 supersedes record 2.

A schema change stops deleting your database. `Baseline` is frozen, every
change after it is an appended migration, and the epoch becomes a pin that does
not move.

## Why now

The window has been running at about one wipe a day — eleven epoch bumps
between Aug 12 and Aug 21, eight of them since the squash. And it never covered
self-host: the epoch is a file beside SQLite, PostgreSQL has neither the file
nor a reset, so the durable backend has had no schema lifecycle at all.

#2428 originally gated this on #2313. That order came from treating the squash
as one-shot; it isn't, while the product major is `0`. The real gate was churn,
and the three steps that churned the schema hardest — #2450, #2474, #2487 —
have landed.

## What happens to an existing profile

Once, on the next boot, and then never again:

| Marker | What happens |
|---|---|
| At the pin (`pre_v1`, 41) | Re-stamped. Chats, sessions and turns kept. |
| Below the pin | One last reset — its schema is a baseline revision nothing recorded. |
| No marker | Reset, as before. |
| Anything else | Refuses to open. Never destroys. |

The convergence arm is the point, and it is the one path that runs once per
profile and can never be re-tested afterwards, so
`a_pre_v1_profile_at_the_pin_keeps_its_data_and_converges` asserts both halves:
the chat is still there, and the marker moved.

## What the audit turned up

Every baseline-touching commit since the Aug 14 squash, checked against the
epoch in the same commit:

```
#2289  code_session.kind              epoch stayed 33
#2453  code_session.reasoning_effort  epoch stayed 39
#2487  code_session_image             epoch stayed 40   (fixed by #2494)
```

Each added a real column or table and regenerated the fixture without bumping.
The first two were repaired by luck — the next change that did bump rebuilt the
profile from the current baseline. The third reached `main` and broke image
attachments on every existing profile until #2494. In between, a contributor is
booting into a schema missing a column.

That is the failure mode decision record 2 named and accepted: the fixture makes
a baseline edit visible, it cannot decide whether the epoch owes a bump.
Freezing the baseline removes the judgment call instead of reinforcing it —
afterwards a fixture diff means "you edited the wrong file", not "decide
something".

## The tripwires flip with it

- `the_schema_baseline_is_pinned` says the baseline is frozen and to append a
  migration, because an edit there now reaches a fresh database and no existing
  one.
- Both journal fixtures stop being change detectors and become compatibility
  contracts again. Rows written by an older binary survive now, so a payload
  change needs a `#[serde(alias)]` or a migration for the rows already on disk.
  `list_events` collects into `Result<Vec<_>>`, so one unreadable row still
  fails a whole chat's history.

## Precondition, now satisfied

Convergence is only safe if `main`'s last baseline edit is the one that set the
pin. #2487 broke that — it added `code_session_image` to the baseline and left
the epoch at 40, so a converged profile would have been missing that table
permanently. #2494 bumped to 41 after it, so `main`'s baseline is again exactly
what the pin rebuilds to, and this branch pins 41.

That is the last time anyone has to check. Afterwards the baseline is frozen and
the tripwire refuses the edit outright.

## Deliberately not in scope

A self-host database created before the pin still holds whatever the baseline
said the day it first ran. Nothing here repairs that — the fix needs a pin row
inside the database, which SQLite got as a sidecar and PostgreSQL never got.
Filed as #2490, and it rides the next squash since adding the row is itself a
baseline edit.

### DIFF
--- a/crates/tidebreak-core/src/code/event.rs
+++ b/crates/tidebreak-core/src/code/event.rs
@@ -578,11 +578,11 @@ mod tests {
         let existing = std::fs::read_to_string(&path).unwrap_or_default();
         assert_eq!(
             existing, rendered,
-            "the code journal event shape changed, and rows in this shape \
-             already exist in databases nothing deletes; make the change \
-             readable from the old shape with #[serde(alias)], or migrate the \
-             rows, then regenerate with UPDATE_CODE_JOURNAL_FIXTURE=1 cargo \
-             test -p tidebreak-core"
+            "the code journal event shape changed; rows written in the old shape \
+             now survive a schema change, so make the new shape readable from \
+             the old one with #[serde(alias)], or migrate the rows. Then \
+             regenerate with UPDATE_CODE_JOURNAL_FIXTURE=1 cargo test -p \
+             tidebreak-core"
         );
     }
 }
--- a/crates/tidebreak-core/src/code/event.rs
+++ b/crates/tidebreak-core/src/code/event.rs
@@ -557,8 +557,12 @@ mod tests {
     /// `CodeEvent` is written into `code_event.event` and read back, so its
     /// field names are a storage format. Rows in the old shape survive a
     /// schema change now, so a diff here needs a `#[serde(alias)]` or a
-    /// migration, not a refreshed fixture on its own — see the same test on
-    /// `AgentEvent` in [`crate::event`].
+    /// migration, not a refreshed fixture on its own. The chat journal's own
+    /// copy of this test says the same thing.
+    ///
+    /// Do not name the chat payload type here, even in prose:
+    /// `chat_and_code_entities_do_not_cross_reference` scans this file as
+    /// text, and it cannot tell a doc comment from a use.
     ///
     /// Regenerate with
     /// `UPDATE_CODE_JOURNAL_FIXTURE=1 cargo test -p tidebreak-core`.
--- a/crates/tidebreak-core/src/code/event.rs
+++ b/crates/tidebreak-core/src/code/event.rs
@@ -555,9 +555,10 @@ mod tests {
     const JOURNAL_FIXTURE: &str = "fixtures/code-journal-events.json";
 
     /// `CodeEvent` is written into `code_event.event` and read back, so its
-    /// field names are a storage format. Pre-v1 this test is a change
-    /// detector: a payload change that ships without a
-    /// `DESKTOP_SCHEMA_EPOCH` bump would make old rows unreadable.
+    /// field names are a storage format. Rows in the old shape survive a
+    /// schema change now, so a diff here needs a `#[serde(alias)]` or a
+    /// migration, not a refreshed fixture on its own — see the same test on
+    /// `AgentEvent` in [`crate::event`].
     ///
     /// Regenerate with
     /// `UPDATE_CODE_JOURNAL_FIXTURE=1 cargo test -p tidebreak-core`.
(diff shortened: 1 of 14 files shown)


naingthet:
**G2G** at `0c80c468b744d054b65ce039189b8ede28ac1a46`

Convergence, last-reset, and fail-closed arms match the decision: `pre_v1` at epoch 40 is re-stamped and kept, anything below the pin still resets once, and an unknown or newer marker refuses without deleting. The pin is frozen in `LAST_RESET_EPOCH` and the baseline tripwire now tells you to append a migration. The one-shot keep path is covered by `a_pre_v1_profile_at_the_pin_keeps_its_data_and_converges`.

Hold merge until #2487 is either a migration or a bump, so a profile already at 40 is not missing `code_session_image` forever. After that, the tripwire owns the rest.

naingthet:
**G2G** at `4cd3325fbaa378fab24730caf0df3c485c4bfe68`

Re-reviewed the new head. The two commits after `0c80c468b744` rewrap local-data docs and make both journal fixture messages say the same compatibility obligation. Lifecycle, pin, fail-closed, and reset paths are unchanged. No new P1/P2.

naingthet:
**G2G** at `03329339f895f4a0e076d3bb79dcd68f935aff4b`

The new head only keeps the chat payload type out of the code journal fixture prose so `chat_and_code_entities_do_not_cross_reference` does not treat a comment as a use. Lifecycle still converges a `pre_v1` pin-40 profile without a wipe, resets only below the pin, and fails closed on anything else. Nothing here is a high-impact miss CI would not see.

naingthet:
**G2G** at `4947431857f84297cff9e4bd8a9523dfb6347ad3`

Re-reviewed after the conflict rebase. The pin is now `LAST_RESET_EPOCH` 41, matching the `main` baseline that includes `code_session_image`. A `pre_v1` profile at 41 still converges without a wipe, anything below the pin still resets once, and an unknown or newer marker still refuses without deleting. No new P1/P2.

PR #2492, opened 2026-08-22
feat(code): attachments reach every engine as files in the worktree

naingthet:
Follow-up to #2485, which made an image paste on opencode say "Tidebreak can't send images to opencode yet" instead of dropping a file URL in the draft. This removes the limit that message described.

## The capability was doing two jobs

Only Claude Code's adapter reads `TurnInput::images` — `encode_turn_stdin` base64s them onto the stream-json user line. codex, opencode, and grok never touch the field. So `caps.image_input` gated the whole feature: `POST /code/sessions/{id}/turns` refused attachments with `unsupported_attachment` on those engines, and the composer hid the control.

Only half of that was ever a property of the engine. Every engine reads files off disk with the tools it already has — which is exactly how a fork hands a child its parent's transcript, and `messageWithWorkspaceFiles` says so: "The engine reads them from disk, so the prompt carries paths and nothing else."

So the capability now decides only which wire an attachment takes:

- **States image input** → bytes ride the protocol, unchanged.
- **Does not** → the worker writes the file under `.tidebreak/attachments/<blob-id>.<ext>` in the checkout and names the path in the prompt.

The route no longer gates on the capability, and the composer offers attachment on every session. No adapter changes, no new fixtures — the fixture rule behind `image_input: Supported` stays exactly as strict, now governing the narrower claim it always measured.

## Where the files go

`.tidebreak/attachments/`, a sibling of the existing `.tidebreak/forks/`. The directory already hides itself: a `.gitignore` holding `*` covers the tree and the marker with it, without touching the repo's tracked ignore file or the `.git/info/exclude` every other worktree of the same repository reads. It is a plain directory in the checkout, so an agent can write there too.

`publish` (stage-then-rename) and that marker move out of `fork.rs` into `code/scratch.rs`. A reader can be mid-read of an attachment for the same reason it can be mid-read of a transcript, so it wants the same guarantee: one whole version or another, never the middle of one.

Names are the blob id, which is content-addressed — the same image attached twice is one file, and a re-send after a failed turn overwrites itself rather than growing the directory.

## The transcript keeps what was typed

`CodeTurn::user_input` is the message as written; only `TurnInput::text` carries the appended paths. That split already existed — the turn row and the engine input are assembled separately.

## Tests

- An engine with no image protocol: the file lands under `.tidebreak/attachments/`, holds the published bytes, the prompt names the path, no bytes ride the protocol, `user_input` is unchanged, and `.tidebreak/.gitignore` holds `*`.
- An engine that states image input: bytes ride the protocol, nothing is appended to the prompt, and no file is written.
- `attachments_are_rejected_unless_the_adapter_declares_image_input` is deleted — it asserted the gate this removes, and the pair above covers the replacement contract.
- `ScriptedAdapter` gains `turn_inputs()`, recording what each turn actually handed the engine rather than only that it was accepted.

Decision 0046 is amended: it anticipated this under "revisit if an engine ships attachments only over a non-machine-readable path".

## Next

Attaching any file type, not just images, is the follow-up: `CodeTurnAttachment` needs a filename and a media type outside the image enum, plus an upload route that skips image decoding. The worker-side write needs no change — it already takes bytes, a name, and an extension.

## Unrelated failure on main

`tidebreak-core`'s `hydration_is_bounded_so_a_long_chat_cannot_grow_the_outbound_body` fails on `main` at ab724b1 (3/3 with this branch stashed) — chat image-hydration eviction keeps 2 where it expects 7. Not touched here.

### DIFF
--- a/crates/tidebreak-core/src/image.rs
+++ b/crates/tidebreak-core/src/image.rs
@@ -100,6 +100,22 @@ impl ImageMediaType {
         }
     }
 
+    /// The file extension to write these bytes under, without the dot.
+    ///
+    /// Used where an attachment is materialized as a file rather than sent
+    /// over a protocol: an engine that reads the image off disk needs the
+    /// name to say what it is, both for its own sniffing and for a person
+    /// reading the path in the prompt.
+    #[must_use]
+    pub const fn extension(self) -> &'static str {
+        match self {
+            Self::Png => "png",
+            Self::Jpeg => "jpg",
+            Self::Webp => "webp",
+            Self::Gif => "gif",
+        }
+    }
+
     /// Identify a supported raster format from its file signature.
     ///
     /// This deliberately recognizes only the provider-safe formats represented
--- a/crates/tidebreak-desktop/ui/src/Composer.tsx
+++ b/crates/tidebreak-desktop/ui/src/Composer.tsx
@@ -329,13 +329,6 @@ export type ComposerProps = {
   plugins?: ComposerPlugins;
   slash?: ComposerSlash;
   images?: ComposerImages;
-  /**
-   * Why this surface has no image path, when it has none. A paste or drop
-   * carrying an image is claimed and answered with this sentence instead of
-   * falling through to the textarea, where the clipboard's text flavour — a
-   * file URL — lands in the draft and reads as an attachment that worked.
-   */
-  imagesUnavailable?: string;
   files?: ComposerFiles;
   /** Paths the agent can already read, shown as chips and named on send. */
   workspaceFiles?: ComposerWorkspaceFiles;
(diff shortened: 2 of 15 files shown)


naingthet:
**G2G** at `f01a218e272e225d18fcb98b1954de13910022a1`

The capability now only picks the wire, the route accepts attachments on every engine, and the file path is content-addressed under `.tidebreak/attachments/` with the same stage-then-rename publish forks already use. The transcript keeps the typed message. No P1/P2 beyond what the new tests already pin.

naingthet:
**G2G** at `1963e6fcfab43a46c0c31fcafd53a983a1b7725e`

Rebase onto main is mergeable. The capability still only picks the wire: protocol bytes when the adapter states image input, otherwise a published file under `.tidebreak/attachments/` and a path in `TurnInput::text`. `user_input` stays as typed. No new high-impact issue on this head.

PR #2493, opened 2026-08-22
fix(server): arming a trigger returns the row that exists

naingthet:
Follow-up to the NO GO on #2486. That review was right and the PR merged over
it — my watcher gated on CI checks only and used `--admin`, so it never read
the review. This fixes it forward.

## The bug

`create_trigger` treated uniqueness as `(repository, condition, action)`. The
store does not: `save_trigger` upserts on `(owner, repo_id, condition)` and
updates `action` and `enabled` **without touching the stored id**.

So re-arming one condition with a different action fell through the
short-circuit, minted a fresh `CodeTriggerId`, wrote nothing under it, and
answered `201` naming a trigger that `GET`, `PATCH` and `DELETE` could not
find.

## The fix

Two parts, and both are load-bearing.

1. **The lookup matches the store's key** — condition only. Re-arming a
   condition now updates the row already there.
2. **The row is read back after the write** rather than returning what was just
   built. This covers the case the lookup cannot: two arms of one condition
   racing each other both see no row and both mint an id, and only one is
   stored.

## Test

A route-level regression test that arms `checks_failed` as `deliver`, re-arms
it as `notify`, and asserts the second response carries the same id, that
listing returns one row rather than two, and that `PATCH` can reach the id it
was handed.

Verified both ways rather than assumed: against the previous implementation it
fails with the two ids differing, which is the phantom exactly. Against this
one it passes. Reverting only the lookup half still passes, which is why the
read-back is there.

### DIFF
--- a/crates/tidebreak-server/src/code/runtime.rs
+++ b/crates/tidebreak-server/src/code/runtime.rs
@@ -2132,26 +2132,45 @@ impl CodeRuntime {
         action: CodeTriggerAction,
     ) -> Result<CodeTrigger, ServerError> {
         self.get_repo(owner, repo_id).await?;
-        let existing = list_triggers_for_repo(&self.db, owner, repo_id).await?;
-        if let Some(found) = existing
+        // Match the store's unique key, which is (owner, repo, condition) —
+        // not the condition and action together. `save_trigger` upserts on
+        // that key and updates action and enabled without touching the stored
+        // id, so re-arming one condition with a different action keeps the row
+        // that is already there.
+        let existing = list_triggers_for_repo(&self.db, owner, repo_id)
+            .await?
             .into_iter()
-            .find(|t| t.condition == condition && t.action == action)
-        {
-            return Ok(found);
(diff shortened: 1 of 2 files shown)


naingthet:
**G2G** at `8eed725a0e5d2dd4504b1be4f6d980ae218efc26`

`create_trigger` now looks up on condition only, which is the key `save_trigger` upserts on, so re-arming with a different action updates the existing row instead of minting an id that is never stored. The post-write read-back is the right second half: two racing arms of the same condition can both miss the lookup, and the loser still answers with the row that actually landed. The route test covers the #2486 phantom (same id, one listed row, PATCH reaches that id).

naingthet:
**G2G** at `94398c311cf85f19f5660e65352577823d4d8915`

Lookup now matches the store key `(owner, repo, condition)`, and the response is the row after `save_trigger` rather than the minted id. Re-arming a condition keeps a reachable id. No new high-impact issue on this head.

Issue #2481, opened 2026-08-22
feat: move the delivery notification rules onto the trigger substrate


naingthet:
Part of #2477. Depends on #2478, #2479, and #2480.

One rule engine, server-side, owning both actions.

`CodeDeliveryStore.ts` already holds user-defined notification rules —
`pull_request_attention`, `pull_request_ready`, `run_failure` — but they are
client-only, in `localStorage` under `tidebreak.code-delivery`, with no record
behind them. They landed in #2348. Record 0060 puts the durable/ephemeral split
back the right way round.

The mapping and its tests landed in #2502. Remaining work is #2504: run the
migration once on first read, then retire the client rules.

## Shape

- Map each existing rule onto a trigger condition and a notification action.
- Migrate `localStorage` state once on first read; the server is authoritative
  afterwards.
- The migration is one-way and runs once. Removing the key without migrating
  loses a user's rules silently, so the read path has to be the thing that
  migrates.

## Done when

A user carrying existing notification rules keeps them, sees them in the
trigger surface, and the client no longer evaluates rules of its own. That
migration is #2504.

naingthet:
The last slice, #2504, landed as PR #2923. The migration runs once on first read and the client no longer evaluates rules of its own. Closing.

PR #2495, opened 2026-08-22
feat(desktop): carry the trigger types to the renderer

naingthet:
Part of #2480. Unblocks the trigger surface and #2481's migration, both of
which need these names in TypeScript.

The routes landed in #2486 but the desktop app has no way to name a trigger.
`wire.ts` is generated from an explicit `collect_from` list in
`wire_types.rs`, and the trigger types were never on it.

This re-exports `CodeTriggerSnapshot`, `CreateCodeTriggerBody` and
`UpdateCodeTriggerBody` from `routes::code`, registers all three for
generation, and commits the regenerated file. The registration and the
regenerated output have to land together — `the_generated_bindings_are_current`
fails otherwise.

Purely additive: 38 generated lines, no existing type changed. Generated with
`UPDATE_WIRE_TYPES=1 cargo test -p tidebreak-server`; all 12 wire tests pass.

Nothing consumes these yet. The client methods, validators, and the surface
itself follow.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/generated/wire.ts
+++ b/crates/tidebreak-desktop/ui/src/generated/wire.ts
@@ -1338,6 +1338,33 @@ export type CodeTerminalRead = { id: CodeTerminalId, workspace_id: WorkspaceId, 
 export type CodeTerminalSnapshot = { id: CodeTerminalId, workspace_id: WorkspaceId, cols: number, rows: number, ended: boolean, created_at: string, };
 
 /**
+ * What a trigger does when its condition fires.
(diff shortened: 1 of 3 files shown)


naingthet:
**G2G** at `6889b846cb9c85ac479f662ca25ab36e395cba26`

The three trigger types are re-exported from `routes::code` and registered in `collect_from` together with the regenerated `wire.ts`. The snapshot, create body, and update body match what the follow-on surface needs. Purely additive; nothing here is consumed yet, so no contract break.

PR #2496, opened 2026-08-22
feat(ui): a surface for arming triggers

naingthet:
Part of #2480. Follows #2495, which carried the types.

Validators, four client methods, and a rules list with one row per condition.

Each row says what firing does — tell the agent, or just notify — and the
section says how a fire would actually reach the session. A harness that
declares mid-turn steering is named as interrupting; one that does not is
described as waiting for the workspace to go quiet and then taking a turn.
That distinction is what record 0060 asks the interface to make visible, and
Claude Code sits on the second path today.

The component is presentational, so the states worth reviewing are stories:
nothing armed, armed, waiting for quiet, no session available to receive a
fire, busy, and armed-but-switched-off.

`parseCodeTriggers` fails the whole read on one bad row rather than dropping
it, since a partially parsed rule set would show fewer triggers than are armed.

Checks: `tsc --noEmit` clean, `biome lint` clean, 1644 UI tests pass.

Still to come: wiring this into the notifications page against a real
repository, and #2481's migration of the existing `localStorage` rules.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/api/client.ts
+++ b/crates/tidebreak-desktop/ui/src/api/client.ts
@@ -93,6 +93,9 @@ import {
   type CodeWorkspaceBlob,
   type CodeWorkspaceTree,
   type CodeWorkspacePrSnapshot,
+  type CodeTriggerAction,
+  type CodeTriggerCondition,
+  type CodeTriggerSnapshot,
   type CodeWatchSnapshot,
   type CodePrCommentsSnapshot,
   type CodePrMergeMethod,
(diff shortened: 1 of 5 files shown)


naingthet:
**NO GO** at `f75c38fa8fd9cb6a367eade95a691e9bc3517ae2`

The switch cannot express the disable path this PR already documents. `setCodeTriggerEnabled` and the Disabled story keep the rule (`enabled: false`). The row switch does not: turning an armed rule off always calls `onDisarm`, and there is no enable callback. Re-enabling a disabled snapshot calls `onArm` (create) instead of flipping `enabled`. Wiring this to delete on off, or inventing meaning for `onDisarm`, loses the rule the client comment says must survive. Add something like `onSetEnabled` and call it when a snapshot already exists.

PR #2499, opened 2026-08-22
test(ci): let a lane stop a superseded main push by skipping it

naingthet:
First of a pair. This changes no workflow — it only widens two assertions so
the change that follows can land.

The `release policy` job runs main's copy of
`scripts/workflow-security.test.mjs` against a pull request's workflows, on
purpose: otherwise a pull request could introduce a violation and delete the
assertion that forbids it in the same commit. The cost is that a workflow
change the current assertions forbid needs the assertion relaxed in an earlier
commit.

## What it allows

A main push that a newer commit has already replaced must not keep an
expensive lane busy. Cancelling it through a concurrency group does stop it,
but the job then concludes `cancelled`, and GitHub rolls a cancelled check up
as a red X on a commit whose own checks all passed — so main's commit list
reads as broken when it is not.

A guard step that reads main's tip and skips the run reports success instead.
Both shapes now satisfy the Windows lane's assertion. The guard shape only
counts when every step after the guard is gated on it: one ungated expensive
step and a superseded run does its whole job anyway.

Staging may likewise build main on a poll of its tip rather than on every
relevant push. It still may never build a pull request's code.

## Verification

`node --test scripts/*.test.mjs` — 187 pass against the workflows as they
stand today, and against the workflows in the follow-up.

### DIFF
--- a/scripts/workflow-security.test.mjs
+++ b/scripts/workflow-security.test.mjs
@@ -533,6 +533,42 @@ test("merge queue groups re-run required CI", () => {
   assert.match(prTitle, /github\.event_name == 'merge_group'/);
 });
 
+// A main push that a newer commit has already replaced must not keep an
+// expensive lane busy. Two shapes stop it. A concurrency group cancels the
+// older run. A guard step reads main's tip and skips the run instead, which is
+// the shape that also keeps the lane off main's commit list: GitHub rolls a
+// cancelled check up as a red X on a commit whose own checks all passed.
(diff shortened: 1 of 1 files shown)


naingthet:
**G2G** at `b4336dbb996f9a92a376a1fd3d108e53c8a1c2c9`

This PR only widens two security-test assertions so a follow-up can skip superseded main work without a cancelled check. Current workflows still satisfy the old shapes, and nothing here changes product or release behavior.

PR #2500, opened 2026-08-22
ci: stop cancelling superseded runs on main

naingthet:
Stacked on #2499, which relaxes the policy assertions this change needs.
Retarget to `main` once that merges.

## The problem

Over the last 25 commits on main, 20 carried a cancelled staging check and 11
a cancelled Windows check. Neither is a required check, so nothing was ever
blocked — but GitHub rolls a cancelled check up as a red X, so main's commit
list read as broken on 22 of 25 commits. Two genuine `desktop UI` failures sat
in that column unnoticed.

## Staging polls main instead of running on every push

A staging build takes about 45 minutes and every build serializes on
`tidebreak-desktop-staging-build`. A per-push trigger could therefore only
queue merges behind each other, and GitHub keeps at most one pending run per
concurrency group: each merge evicted the run queued behind it. Those 20
cancelled jobs had run zero steps.

**Publish staging desktop** now runs hourly on a `schedule`, plus
`workflow_dispatch`. Each poll reads the commit recorded in the hosted staging
manifest, diffs it against main's tip over the path list the push trigger
filtered on, and builds only when one of those paths moved. A manual run takes
a `force` input to build a tip the poll would skip. When the poll finds
nothing, `publish` is skipped and the run is green.

The concurrency rules are unchanged: the caller still uses a per-run group,
and the publish workflow still serializes with `cancel-in-progress: false` so
an in-flight notarization finishes.

## The Windows lane skips instead of cancelling

`windows-check` drops its cancelling concurrency group for a guard step that
reads main's tip and skips a run the branch has already moved past. A
superseded run reports success rather than `cancelled`. The lane also gets a
backstop that finishes: it was being superseded so often that the post-merge
run it exists for almost never completed.

The check runs inside the job, not ahead of it, because a run is almost always
main's tip when it is created and goes stale while it waits for a runner or
while it works. It fails open — an unreadable tip runs the lane.

Both shapes are pinned by the policy test, which no longer accepts a lane that
cancels itself on a main push.

## Also

- `docs/releases.md` describes the poll and the `force` input.
- Decision record 16 gets a dated amendment. Its identity, signing, and
  concurrency decisions stand; only the "When" column changes.

## Verification

`node --test scripts/*.test.mjs` — 187 pass. Also ran #2499's copy of
`workflow-security.test.mjs` against this branch's workflows the way the
`release policy` job will, and every new shell step through `bash -n`.

## Not fixed here

Staging has not published since `f2e32cf6` at 10:28 today. `publish_staging`
advances `latest.json` only if the built commit is still main's tip, and a
45-minute build rarely is. Polling does not change that.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -530,12 +530,6 @@ jobs:
     #
     if: ${{ needs.changes.outputs.rust == 'true' && (github.event_name != 'pull_request' || contains(github.event.pull_request.labels.*.name, 'windows-ci') || needs.changes.outputs.windows == 'true') }}
     runs-on: windows-latest
-    # Every Rust-scoped main push contains the prior main state, so the newest
-    # Windows run proves everything an older in-progress one would have proven.
-    # PRs, schedules, and manual dispatches remain isolated by run id.
-    concurrency:
-      group: ${{ github.event_name == 'push' && 'windows-check-main' || format('windows-check-run-{0}', github.run_id) }}
-      cancel-in-progress: ${{ github.event_name == 'push' }}
     # The persistence steps below compile tidebreak-server's test target, which
     # is the heaviest codegen this lane pays; leave headroom over the
     # check-plus-broker baseline.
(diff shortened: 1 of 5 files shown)


naingthet:
**G2G** at `ff862ecf8b255443f3c3d3ac7f7bf63065c2a20b`

Hourly poll plus skip-instead-of-cancel keeps main from showing cancelled checks. Staging still only advances `latest.json` when the built commit is tip, which this PR already calls out as unchanged.

naingthet:
**G2G** at `ff884690ca5c3bc892ace9665177b2c1dc51ce84`

Re-reviewed the new head. Same shape as the prior G2G: Windows skip-on-stale-tip instead of cancel, staging on a poll plus force dispatch, policy test pins both. No P1/P2 outside what CI already covers.

PR #2502, opened 2026-08-22
feat(ui): map the delivery notification rules onto triggers

naingthet:
Part of #2481. Follows #2496.

Record 0060 folds the client-side notification rules into the trigger
substrate so there is one rule engine rather than two. This is the mapping,
kept as a pure function so it can be tested without a server.

## Decisions worth reviewing

- **Everything migrates as `notify`, never `deliver`.** These rules raised a
  notification and never sent the agent anything. Migrating them to `deliver`
  would start interrupting work the user never agreed to interrupt.
- **`pull_request_attention` maps to two conditions**, `changes_requested` and
  `conflicts`. It covered more than one fact, and dropping either would
  silently narrow what someone already asked to hear about.
- **An empty scope means every repository**, matching what the rule already
  meant. A repository with no `tidebreak_repo_id` is not something a trigger
  can bind to, so it is skipped rather than guessed at.
- **Results are keyed on `(repository, condition)`** — the server's own unique
  key — so two rules reaching one condition arm it once.

## Scope

The mapping and its tests. Running it once on first read, and then dropping the
client-side evaluation, is the next commit; the `localStorage` write is
one-way, so it is worth landing the mapping under test before anything reads it
for real.

Checks: 7 new tests pass, `tsc --noEmit` clean, `biome lint` clean.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeTriggerMigration.test.ts
+++ b/crates/tidebreak-desktop/ui/src/code/CodeTriggerMigration.test.ts
@@ -1,0 +1,110 @@
+import { describe, expect, it } from "vitest";
+
+import type { CodeGitHubRepositoryRef } from "@/api/types";
+import type { CodeDeliveryNotificationRule } from "@/code/CodeDeliveryStore";
+import { triggersForNotificationRules } from "@/code/CodeTriggerMigration";
+
+function repository(
+  name: string,
+  tidebreakRepoId?: string,
+): CodeGitHubRepositoryRef {
+  return {
+    host: "github.com",
+    owner: "brightwave-inc",
+    name,
+    name_with_owner: `brightwave-inc/${name}`,
+    url: `https://github.com/brightwave-inc/${name}`,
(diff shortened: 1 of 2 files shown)


naingthet:
**NO GO** at `7a1172d6bc253637593e3209e3c46012c69fcf3b`

Scoped rules will silently miss repositories. Persisted `repositoryKeys` come from `codeDeliveryRepositoryKey`, which lowercases `host/owner/name`. `triggersForNotificationRules` builds its own key without that fold and then uses `includes`. A scoped rule for `github.com/BrightWave-Inc/orca` therefore never matches the stored key, so the trigger is never armed. Empty-scope rules still work. Reuse `codeDeliveryRepositoryKey` for the compare.

naingthet:
**G2G** at `018ae83a24e10757f28ef79c68cab82751d3c777`.

Scoped matching now uses `codeDeliveryRepositoryKey`, so a stored lowercase key still binds a host/owner/name that is not already lowercase. The mixed-case scope test covers that path. No other high-impact defect in the mapping.

PR #2505, opened 2026-08-22
fix: restore CI and repair platform regressions

### LINKED ISSUE #2473: test(server): the inbox route test races its own turn claim

### LINKED ISSUE #2490: fix(core): a self-host database created before the migration pin has no way back

### LINKED ISSUE #2497: docs(managed-policy): the sqlite developer toggle no longer matches the code
[`docs/managed-policy.md`](https://github.com/brightwave-inc/tidebreak/blob/main/docs/managed-policy.md#developer-flow--the-sqlite-policy-toggle) tells a developer to write a `managed_policy_v1` row to turn a profile managed, and to delete it to go back. Neither half is true since the provisioned policy moved to the `gateway-policy.json` sidecar.

`import_legacy_setting` (`crates/tidebreak-server/src/managed_policy.rs`) is called once, at boot, from `lib.rs:1630`, and returns early when the sidecar already exists:

```rust
match provisioned.read() {
    Ok(Some(_)) => return Ok(()),
    ...
}
```

So:

- **Writing the row** takes effect on the next *boot*, not "on the next `/policy` read" as the doc says — and only while no sidecar exists.
- **Deleting the row** does not return to the open profile. The sidecar holds the policy and is what every read consults. You have to delete `gateway-policy.json` as well.

## The row now lingers

The comment on `LEGACY_SETTING_KEY` used to say epoch resets remove the row naturally. [Decision 61](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0061-schema-changes-are-migrations.md) ends those resets for any profile at or above the pin, so on those the row stays indefinitely. It is inert while the sidecar exists, which is the case for every profile that has booted since the move — but it means a developer who deletes the sidecar to unpair gets the old policy imported back on the next boot, silently.

I corrected the comment in #2491 to say the row is inert rather than swept. The doc and the unpair path are the parts still worth fixing.

## Shape

Rewrite the developer flow around the sidecar, which is what the code reads: write or delete `gateway-policy.json` directly, and say a restart is needed. Then decide whether `import_legacy_setting` should delete the row after a successful import, which would make unpairing behave the way the doc already claims.

### LINKED ISSUE #2498: test(core): the chat/code cross-reference guard scans prose, and misses a re-export
`chat_and_code_entities_do_not_cross_reference` (`crates/tidebreak-core/src/db/tests/code.rs`) walks `src/code` and `src/db/ops/code` and asserts no file *contains* the strings `ChatId`, `MessageId`, or `AgentEvent`. It reads each file as text, with no idea what part of it is code.

## It fires on prose

A doc comment that merely names a chat type fails the test. That happened in #2491: a retargeted comment on the code journal's shape fixture pointed at the chat journal's copy of the same test by naming its payload type, and the guard rejected it. The fix was to reword the sentence, which is the wrong thing to be spending review on — nothing about the code changed.

## And it misses the case it exists for

The same bluntness cuts the other way. The guard matches spelled-out names, so a glob import plus inference, a type alias, or a re-export under a different name all pass it. It is a spell-checker standing in for a dependency rule.

The `TurnId` arm already shows the strain — it carries a hand-maintained exception:

```rust
if line.contains("TurnId")
    && !line.contains("CodeTurnId")
    && !line.contains("code_turn_attachment")
```

## Shape

Strip comments before scanning. That removes the false positive for a few lines and keeps everything else as it is. Do not reach for `syn` here: [decision 48](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0048-one-interaction-model.md) step 5 (#2313) merges the two id spaces and retires this guard outright, so an AST-accurate version would be built to be deleted.

If #2313 lands first, close this instead.

naingthet:
## What changed

Cargo uses the runner's Git client for the pinned Symphonia fetch, and the core migration chain repairs baseline tables missing from pre-pin self-host databases. Managed policy retires the legacy SQLite row after the sidecar holds a valid policy, the developer guide uses `gateway-policy.json`, and route and boundary tests no longer depend on clock timing or scan comments as code.

## Validation

`cargo fmt --all -- --check`, `node --test scripts/workflow-security.test.mjs`, `cargo metadata --locked --no-deps --format-version 1`, and `git diff --check` pass. Rust compilation and tests were not run locally under the repository's Cargo policy; CI provides that coverage.

## Compatibility and risk

This adds an ordered repair migration and deletes only an obsolete managed-policy row after a valid sidecar exists; wire formats do not change.

## Tracking

Closes #2498
Closes #2497
Closes #2490
Closes #2473

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -21,6 +21,11 @@ on:
 permissions:
   contents: read
 
+env:
+  # Cargo's libgit2 transport intermittently fails TLS handshakes against the
+  # pinned Symphonia revision. The runner's Git client handles the fetch.
+  CARGO_NET_GIT_FETCH_WITH_CLI: "true"
+
 concurrency:
   # On a PR ref only the newest commit matters, so supersede in-progress runs.
   # Every other event gets a unique group (keyed by run id) so nothing cancels
--- a/crates/tidebreak-core/src/db/migration.rs
+++ b/crates/tidebreak-core/src/db/migration.rs
@@ -37,7 +37,12 @@ pub struct Migrator;
 #[async_trait::async_trait]
 impl MigratorTrait for Migrator {
     fn migrations() -> Vec<Box<dyn MigrationTrait>> {
-        vec![Box::new(Baseline), Box::new(AppOwner), Box::new(CodeOwner)]
+        vec![
+            Box::new(Baseline),
+            Box::new(AppOwner),
+            Box::new(CodeOwner),
+            Box::new(BaselineRepair),
+        ]
     }
 }
 
@@ -213,6 +218,54 @@ impl MigrationTrait for CodeOwner {
     }
 }
 
+/// Repair the known gap for a self-host database that predates the frozen
+/// baseline. Such a database makes [`Baseline`] return when it sees `app`, so
+/// tables added to the baseline later never reached it. Create every missing
+/// baseline table and index in dependency order. This repairs additive schema
+/// changes; changed columns still require their own ordered migration.
+struct BaselineRepair;
+
+impl MigrationName for BaselineRepair {
+    fn name(&self) -> &str {
+        "m20260822_000004_baseline_repair"
+    }
+}
+
+#[async_trait::async_trait]
+impl MigrationTrait for BaselineRepair {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
+        for entry in baseline::tables() {
+            let name = entry
+                .table
+                .get_table_name()
+                .expect("every baseline table statement names its table")
+                .sea_orm_table()
+                .to_string();
+            if manager.has_table(&name).await? {
+                continue;
+            }
+            let mut table = entry.table;
+            table.if_not_exists();
+            manager.create_table(table).await?;
+            for mut index in entry.indexes {
+                index.if_not_exists();
+                manager.create_index(index).await?;
+            }
+        }
+        for seed in baseline::SEED_STATEMENTS {
+            manager.get_connection().execute_unprepared(seed).await?;
+        }
+        Ok(())
+    }
+
+    async fn down(&self, _manager: &SchemaManager) -> Result<(), DbErr> {
+        // The repair only creates baseline objects that were missing. Removing
+        // them would also remove objects owned by the baseline on a fresh
+        // database.
+        Ok(())
+    }
+}
+
 #[cfg(test)]
 mod tests {
     use sea_orm::{ConnectionTrait, Database, DbBackend, Statement};
@@ -263,6 +316,7 @@ mod tests {
                 "m20260814_000001_baseline",
                 "m20260814_000002_app_owner",
                 "m20260814_000003_code_owner",
+                "m20260822_000004_baseline_repair",
             ]
         );
         assert!(db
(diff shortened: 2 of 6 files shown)


naingthet:
**G2G** at `8f3d2a1194f26b774a29bb0fb2945df5628c3d56`

The baseline repair only fills missing tables, then seeds with `ON CONFLICT DO NOTHING`, so existing self-host rows stay put. Managed policy deletes the SQLite row only after the sidecar already holds a valid copy, which matches the documented upgrade path. Route helpers claim with `available_at + 1s` instead of wall-clock now, so they no longer flake on due-time races.

naingthet:
**G2G** at `fd338b76cbaa965fabc983c5d0038fe47b74964a`

The new commit keeps route-test park timestamps on the same instant as the claim instead of `Utc::now()`. That is a fixture-only change; no product path or schema moves. Nothing high-impact that CI would miss.

PR #2507, opened 2026-08-22
fix(desktop): stop Cmd+W from quitting the app

naingthet:
## What changed

On macOS a menu accelerator is claimed before the key reaches the webview, so Tauri's default Close Window item always won Cmd+W and the shell's own close-tab binding never ran — and since Tidebreak has one window, the chord quit the app. The new `menu.rs` builds the menu bar explicitly instead of patching Tauri's default: Cmd+W is a Close Tab item the renderer answers, Close Window sits under it with no accelerator (Cmd+Shift+W already starts a pull-request watch in code mode), and everything the default carried — Services, Hide, the Edit clipboard commands, Full Screen, Minimize, Zoom, plus the existing Check for Updates… and Reload items — is preserved. The renderer runs one close for both the menu event and the keydown path, so the chord means the same thing in the packaged app and in a browser dev build, and with no tab to close it does nothing. The menu code moved out of `updater.rs`, which is now only about updates.

## Validation

`tsc --noEmit`, `biome format` and `biome lint` over `crates/tidebreak-desktop/ui/src`, `cargo fmt --check`, and the AppShell, ShellShortcuts, and codeChrome test files (64 passing). Rust was not compiled locally; CI covers it.

## Compatibility and risk

macOS only — no menu bar ships on Windows or Linux. Cmd+W no longer closes the window: use the red button, the File > Close Window item, or Cmd+Q. No persisted or wire-format changes.

### DIFF
--- a/crates/tidebreak-desktop/src/lib.rs
+++ b/crates/tidebreak-desktop/src/lib.rs
@@ -42,6 +42,7 @@ mod documents;
 mod host_access;
 mod host_authority;
 mod image_attachments;
+mod menu;
 mod node_install;
 mod office_install;
 mod office_pdf;
@@ -840,7 +841,7 @@ pub fn run() {
             updater::check_for_update,
             updater::restart_for_update
         ])
-        .on_menu_event(updater::handle_menu_event)
+        .on_menu_event(menu::handle_menu_event)
         .setup(move |app| {
             // The dev server's origin is remote as far as the IPC is
             // concerned, so `tauri dev` needs it granted explicitly. Added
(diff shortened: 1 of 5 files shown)


naingthet:
**G2G** at `96c766aa8f6dfd550c9bc66a9e0b84e2890ebe0a`

Cmd+W is claimed by Close Tab and routed through the same close as the keydown path, so the chord no longer closes the only window. Close Window has no accelerator, which keeps Cmd+Shift+W for PR watch. Nothing here would fail CI and still ship a wrong close or quit.

Issue #2513, opened 2026-08-22
Fold Machine settings into Model Gateway and autofill the address from the gateway


naingthet:
Fold the Machine settings into the Model Gateway section and fill the address in from the gateway, so connecting to your organization's machine is one deliberate click rather than a URL you have to know.

## Autofill

A gateway that hosts a machine should say so. Add `tidebreak_machine_url` to `GET /api/v1/meta`, beside the `tidebreak_download_url` already there, then read it into `GatewayMeta` (`crates/tidebreak-server/src/connectors/gateway.rs:147`) and prefill the address field. Additive both ways: serde ignores unknown fields, so a new client reads an old gateway as absent and an old client ignores a new gateway's field.

Meta rather than the provision deep link: the link fires once at pair time, so a profile paired earlier never sees the value and a machine that moves leaves a stale one behind. Meta is re-read every boot.

Treat the value as a hint for a text field, never as authorization. The connect path does not change — the machine still has to answer `/auth/discovery`, name the same gateway this profile is provisioned to, and echo the resource the desktop derives locally from the URL. `discover_gateway` enforces all three, so a wrong value can only produce an address that fails to connect.

The gateway-side field is a separate change in the gateway repository; build the client half to degrade when it is absent.

## Folding the section

- Remove the Machine section from `ui/src/settings/sections.tsx:296` and move its content into `ui/src/settings/GatewayPanel.tsx`.
- Drop `unmanagedHidden` from the gateway section. It is there because an unmanaged profile has no gateway to configure, but the standalone token path is exactly what an unmanaged profile needs, and hiding the section would strip the only desktop route to a machine that is not behind a gateway.
- Keep the standalone token behind an Advanced disclosure, collapsed. It is the only client path to a deployment running on `TIDEBREAK_AUTH_TOKENS_FILE` (`docs/self-hosting.md:29`); removing it would push those operators to the CLI.
- Do **not** attach automatically on sign-in. Attaching moves a reader's work to another machine and takes away connected folders, computer use, and native export; it stays a deliberate act.

Add Storybook stories for the panel's states — unmanaged, managed and signed out, managed with a machine offered, attached — per the repository's UI workflow.

Issue #2512, opened 2026-08-22
Gate the surfaces that reach this computer when the window works on another machine


naingthet:
Surfaces that reach this computer still offer themselves while the window works on another machine, then fail after the reader has committed to them.

`useAttachedRemotely()` and the module-level `attachedRemotely()` now exist (`ui/src/AppContext.tsx`, `ui/src/host.ts`), so each of these is a gate away from correct. Before this, every one used `hasNativeHost()`, which is `isTauri()` and stays true while attached.

Ordered by how much they cost the reader:

1. **Pasted and dropped images fail after the fact.** `ui/src/useImageAttachments.ts:170` takes the host branch on `hasNativeHost()`, and the handler resolves the conversation against the local store, so every paste fails "Conversation not found". Should take the browser branch and post to the attached machine — a correctness fix, not a gate.
2. **File attach, same shape.** `ui/src/ChatRoute.tsx:754`, `ui/src/HomeRoute.tsx:450`, `ui/src/DocumentDropTarget.tsx:43`. The picker opens, the reader chooses files, then the import fails.
3. **Browser tabs open on the wrong computer.** `ui/src/ChatRoute.tsx:1001`, `ui/src/code/CodeWorkspacePage.tsx:1013` and `:1175`. "Share with agent" shares a browser on this laptop with an agent that is not on it. Making `available` false in `ui/src/code/browser/browserHost.ts:130` while attached fixes nine call sites at once.
4. **Debug bundle.** `ui/src/ChatHeaderTitle.tsx:116`. Built from the local event journal, so it cannot be built for a remote conversation at all — hide it rather than refuse it.
5. **Document download** leaks the raw `native_export_authority_unavailable` code to the reader (`ui/src/document-detail/DocumentDetailRoot.tsx:245`); the sibling outputs panel already routes through `hostErrorMessage`.
6. **Harness worktree Browse.** `ui/src/settings/CodingHarnessesPanel.tsx:105`.
7. **Copy that names the wrong computer.** `ui/src/settings/VoiceTranscriptionPanel.tsx:101` promises recordings never leave this device — false while attached, the audio travels to the machine. Also `ui/src/ExecDisclosure.ts:4`, `ui/src/settings/ExecPanel.tsx:32`, and the local-stdio wording in `ui/src/settings/McpPanel.tsx`.

## Done when

[...]

Issue #2508, opened 2026-08-22
Self-host image ships no Node runtime, git, or gh, so code mode cannot run on a hosted machine


naingthet:
The self-host image cannot run a coding agent or clone a repository, so a machine deployed from it is not usable for code mode.

`deploy/self-host/Dockerfile:70` installs only `ca-certificates` and `curl`. Two consequences, both confirmed against a live deployment:

- Every harness install fails with "the verified managed Node runtime is not installed in this Tidebreak data directory", so Claude Code, Codex, opencode, and Grok all report Not found and code mode stays gated at `CodeHome.tsx:59`.
- There is no `git` and no `gh`, so even a public clone dies at `crates/tidebreak-server/src/code/clone.rs:373` with "failed to spawn git".

## Done when

- The image carries `git`, and the managed Node runtime the harness installer expects, pinned to exact versions.
- A container built from it installs at least the primary pinned harness and clones a public repository end to end.
- `docs/self-hosting.md` states what the image provides, so an operator can tell whether their own base image is sufficient.

Decide deliberately whether `gh` belongs in the image or whether the GitHub path should stop depending on the CLI — see the per-user GitHub identity issue, which may remove the need.

PR #2514, opened 2026-08-22
fix(desktop): a machine that refuses you no longer traps the window

naingthet:
Attaching to a machine points every read through it, including the managed gate's `/policy`. When that machine stops accepting the session — access ended, gateway down — the gate reads the HTTP error as an unreadable policy and blocks the whole app with "Managed policy unavailable. This device's managed policy is misconfigured. Contact your administrator."

Both sentences are wrong. The policy on this computer is fine; the machine refused the token. And the only control on screen is a Retry that retries the same refusal, while the Machine settings that would detach sit behind the gate. The only way out is to edit the address file by hand, which is how this was found.

The blocked screen now names what happened and offers **Work on this computer**. Detaching is a host command, so it works no matter what the machine says.

## Three things behind it

**Surfaces could not tell they were attached.** The only way to ask was an async Tauri call that three places made and kept to themselves, so everything else used `hasNativeHost()` — which is `isTauri()`, and stays true while attached. `attachment` now rides on the app context with `useAttachedRemotely`, and `host.ts` carries the module-level twin for the three callers outside React.

**Saving MCP servers wrote to the wrong machine.** The panel read the attached machine's servers over HTTP but saved through the native command, which writes to the server embedded in this app. A reader edited the machine's config, saved it to their laptop, and watched the edit disappear on the next read. It takes the HTTP branch while attached; the OS confirmation it skips exists to guard stdio commands that would run on this computer, which these would not.

**A provision link into an already-managed profile did nothing visible.** One log line, no dialog, no gate nudge — and that is exactly the state someone reaches for the link from, so it read as the app ignoring them. It now nudges the gate, which re-reads policy and presents its sign-in.

## Not in scope

The remaining remote-attachment gaps are filed as #2508 through #2513. This PR adds the signal the UI gating work (#2512) needs; it does not do that gating.

## Verification

`tsc --noEmit` clean. Full renderer suite: 216 files, 1652 tests, all passing. `biome format` and `lint` clean. `cargo fmt --check` clean; Rust compilation left to CI.

### DIFF
--- a/crates/tidebreak-desktop/src/deep_link.rs
+++ b/crates/tidebreak-desktop/src/deep_link.rs
@@ -279,6 +279,13 @@ fn spawn_pairing(app: tauri::AppHandle, link: ProvisionLink) {
             }
             Ok(PendingRegistration::AlreadyManaged) => {
                 log_pairing(&app, &format!("{origin} already manages this device"));
+                // Nothing to provision, but the click still means "connect me
+                // to this gateway" — and the profile may be managed by it
+                // with no session behind that, which is exactly when someone
+                // reaches for the link again. Nudging the gate makes it
+                // re-read policy now and present its sign-in, instead of the
+                // link reading as the app ignoring them until the next poll.
+                notify_pairing_changed(&app);
             }
             Err(PairingError::Conflict {
                 provisioned_url,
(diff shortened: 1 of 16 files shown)


naingthet:
**G2G** at `dbc7d444d07eb09a55b7cbdc06388ce7a643fe0e`

The blocked path now keys off the host attachment record rather than the client, and detach is a host command, so a refused machine can no longer trap the window behind Retry. MCP save takes the HTTP branch while attached, so it writes the config the panel just read. Nothing here looks like a ship-blocker CI would miss.

PR #2515, opened 2026-08-22
feat(code): arm Claude and Codex fast modes from the composer

naingthet:
Fast mode runs the same model faster and charges more per token. Both engines expose it and neither was reachable from Tidebreak.

## How each engine spells it

**Claude** uses a `fastMode` settings key, which rides `--settings` as inline JSON. It joins model and effort as a launch flag, so `ensure_channel` respawns the child on a toggle exactly as it already does for those two. Anthropic serves the tier on part of the Opus line only, so `model_serves_fast_mode` matches the ids it actually serves rather than inferring from the family. A wrong yes would arm a control whose turns quietly run at standard speed while reading as fast, which is worse than not offering it.

**Codex** uses a service tier and validates the id per model. Its catalog advertises the tier two ways — `additionalSpeedTiers` lists `fast`, `serviceTiers` carries the id a request sends — so the parser reads either, in both camelCase and snake_case. The request sends `priority`, which is what Codex's own picker labels "Fast — 1.5x speed, increased usage". It rides `turn/start` beside the existing model and effort overrides, so there is no relaunch.

## Shape

The session row carries the choice and a turn reads it the way it reads effort. `POST /code/sessions/{id}/fast-mode` mirrors the effort route: refused mid-turn, refused after the session ends, and refused when the session's model does not serve the tier. Only arming is gated — disarming returns to a default every model honors.

Capability is per model, not per engine, which is the same reason the effort ladder is. It rides `ListedHarnessModel` off the engine's own listing, so a model that gains or loses the tier upstream needs no change here.

The toggle sits between the model and effort controls: all three say how the next turn runs, and those two are the pair that costs money to change. It renders nothing when the selected model cannot serve the tier, matching how an empty effort ladder hides the effort control.

## Note on cost reporting

Model Gateway now prices Anthropic fast mode correctly, but still reads OpenAI `priority` as standard. Codex turns armed here will under-report cost until that follow-up lands.

## Checks

`tsc --noEmit` clean, `biome format`/`lint` clean, 138 focused UI tests pass, `cargo fmt` applied. Rust compiles on CI.


naingthet:
**NO GO** at `f3f1e39ff53e0007fd8775969001b6337e7ea119`

P1: `set_fast_mode` refuses arming when the selected model cannot serve the tier, but a later model switch never clears `session.fast_mode`, and Claude still composes `--settings {"fastMode":true}` from `spec.fast_mode` on every launch. The PR says Anthropic rejects `speed` off the Opus ids this matcher allows. So: arm on Opus, pick Sonnet (toggle hides), next turn can fail or still claim a premium the model will not run. `create_session` also accepts `fast_mode: true` with no capability check.

Clear the bit when the new model cannot serve it, or drop the flag at launch unless `model_serves_fast_mode` is still true for the resolved id. Same gate on create.

P2: `FastModeToggle` keeps `useState(fastMode)` and never syncs from the prop. A failed POST rolls `settings.fastMode` back in the page, but the composer stays on the optimistic value, so the spend switch can lie until remount.

naingthet:
**NO GO** at `3303ba4b7c517870b7a72c8a7462485598eff033`

The last review asked that arming not be refused after the user already picked a model that serves the tier. That is still what happens.

`FastModeToggle` uses the composer's selected row (`selectedOption?.fast_mode`), so the control appears as soon as you pick a supporting model. `POST /code/sessions/{id}/fast-mode` still calls `model_serves_fast_mode` against `session.model` on the row. Model changes in the composer only update local state (`onModelChange` → `setModel`); they are not written to the session before the toggle fires.

Pick a model that advertises fast mode, then arm: the request is `fast_mode_unsupported` because the persisted id is still the old one. Disarm still works. Drop the persist-time gate (or key it on the model the next turn will actually send) so a choice the UI already offered can stick.

naingthet:
Both findings addressed at `0cb3cbb5b`, plus the repeat.

**Repeat finding (persist-time gate).** You're right that I fixed the wrong half last time. The route no longer checks the tier at all — it stores intent and nothing else. Arming a model the picker is showing as selected now sticks, because nothing consults `session.model` on that path.

**P1 (stale bit).** Moved the decision to where the model is actually known rather than clearing it in one place:
- Claude composes the `fastMode` settings flag only when the *resolved* model serves the tier (`fast_mode_flags` now takes the model). Armed on Opus, switched to Sonnet, the turn degrades to standard speed instead of sending `speed` to a model that rejects it. `EngineChannel` records the resolved value, so `ensure_channel` respawns when it changes rather than reusing a child composed for the other model.
- A turn that switches model clears the arming when the new model can't serve the tier, which covers Codex, whose validation is server-side. Checked only on a real switch, since `list_models` costs a process there.
- `create_session` needs no gate under this shape: a stored bit for a model that can't serve the tier is inert.

**P2 (optimistic value).** `FastModeToggle`'s value now follows the prop, keyed on `fastMode` and `sessionId`, matching the effort control beside it. A refused POST rolls the composer back with the page.

Also fixed two CI failures that were mine: the migration had no `down`, so `Migrator::down` failed the multi-agent-wait rollback test, and an `api.test.ts` fixture omitted `fast_mode` — its expectation now documents that a payload without the field reads as off.

PR #2522, opened 2026-08-22
feat(code): a machine says which repo sources it can serve

### LINKED ISSUE #2511: Add-repo hard-codes three sources; a machine should report which ones it supports
Add-repo offers three sources unconditionally, and two of them cannot work on a hosted machine.

`ui/src/code/AddRepoPalette.tsx:34` hard-codes Local folder, Git URL, and GitHub. Selecting Local folder auto-opens this computer's directory picker (`:209`), and both clone stages ask for a destination path with a Browse button (`:428`, `:631`) — all filling fields the server resolves on its own filesystem. On a machine that is not this computer every one of those paths is meaningless.

The palette also fetches `getCodeCloneDefaults()` on open with no `.catch` (`:107`). That route is admin-plane (`crates/tidebreak-server/src/lib.rs:527`), so a non-admin member on a shared machine gets an unhandled rejection, `parentDir` stays empty, and both clone forms stay disabled with nothing said.

## The axis

Do not branch on the client's attachment. Attachment is a property of the window: a headless self-host reached from a browser tab reports `local` (`ui/src/boot.ts:60`), and a laptop-hosted machine with no `git` is as unable to clone as a container is. The machine has to report what it supports, on the member plane. Decision 17 sets the convention — probe, don't pin — and decision 47 extends it to the client-to-machine wire.

The harness doctor is the precedent worth copying: the machine reports what it can do and the client renders remediation instead of a broken control (`ui/src/code/CodeHome.tsx:59`).

## Done when

- A member-plane read reports which repo sources this machine supports, each with an availability flag and remediation text.
- `AddRepoPalette` filters its source list from that read rather than hard-coding three rows, and ignores source keys it does not know, so an older client degrades instead of breaking.
- The clone-defaults read is reachable by a member, or the palette tolerates its absence.
- Browse appears only where the picker can name a path the machine will resolve.

naingthet:
Closes #2511.

Add-repo offered Local folder, Git URL, and GitHub unconditionally, and two of them could not work on a machine that is not this computer. The local picker filled a path field the machine resolves on its own disk, and both clone stages asked for a destination directory the caller had no way to name.

A member could not even open the dialog cleanly. It fetched `clone-defaults` with no `.catch`, that route is administrator plane, and a 403 left the destination empty with both clone forms disabled and nothing said.

## What the machine now answers

`GET /code/repos/sources` is a member-plane read of what this machine can serve: whether it can spawn `git`, whether it has a GitHub credential, and whether it places clones under a destination its operator already configured. The palette filters its source list from that answer and shows the machine's own remediation for whatever it drops, so a missing source reads as a fact about the machine rather than a broken dialog.

Unknown source kinds are ignored rather than rendered, so the set can grow without a client release — decision 17's rule, applied to the client-to-machine wire as decision 47 extends it.

## Clones no longer demand a path

`parent_dir` is optional on the clone request. A machine that remembers a destination places checkouts itself, which is what lets someone who cannot see its filesystem clone at all, and keeps a shared machine's layout to itself. A desktop working on its own machine still names its own path; that field is unchanged.

A field nobody was shown is never sent. The palette drops the remembered value rather than posting it — a test caught that, and without it a hidden field would have put checkouts somewhere the reader never chose.

## The axis

Availability is the machine's answer, not the window's. Whether a picker can name a path the machine would resolve is a separate question, and the client answers that one with `hasLocalHostAuthority()`. Branching on attachment alone would be wrong in both directions: a browser tab against a headless deployment reports local, and a laptop-hosted machine without `git` is as unable to clone as a container is.

## Verification

`tsc --noEmit` clean. Full renderer suite: 216 files, 1655 tests. New coverage for a source the machine cannot serve, a machine that places clones itself, and the administrator-only defaults read being refused. `biome format` and `lint` clean, `cargo fmt --check` clean.

The generated wire types are hand-written to match `ts-rs` output rather than regenerated locally; the `the_generated_bindings_are_current` test is the check on that.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/api/client.ts
+++ b/crates/tidebreak-desktop/ui/src/api/client.ts
@@ -103,6 +103,7 @@ import {
   type CodePrMergeMethod,
   type CodeWorkspaceSnapshot,
   type CodeCloneDefaults,
+  type CodeRepoSources,
   type CodeWorktreeRoot,
   type CodeForkTranscript,
   type CodeCloneJobSnapshot,
@@ -142,6 +143,7 @@ import {
 import {
   parseCodeApproval,
   parseCodeCloneDefaults,
+  parseCodeRepoSources,
   parseCodeWorktreeRoot,
   parseCodeForkTranscript,
   parseCodeCloneJob,
@@ -1919,6 +1921,22 @@ export class ApiClient {
     });
   }
 
+  /**
+   * What this machine can add a repository from.
+   *
+   * Member-plane, unlike `getCodeCloneDefaults`, which is administrator-only:
+   * on a shared machine the person adding a repository is usually not an
+   * administrator, and a dialog that cannot read its own options is no dialog.
+   */
+  async getCodeRepoSources(): Promise<CodeRepoSources> {
+    return requireParsed(
+      parseCodeRepoSources(
+        await this.json("/code/repos/sources", { headers: this.headers() }),
+      ),
+      "repo sources",
+    );
+  }
+
   async getCodeCloneDefaults(): Promise<CodeCloneDefaults> {
     return requireParsed(
       parseCodeCloneDefaults(
@@ -1930,10 +1948,14 @@ export class ApiClient {
     );
   }
 
+  /**
+   * Start a clone. `parent_dir` is omitted when the machine places clones
+   * itself — see `CodeRepoSources.chooses_destination`.
+   */
   async startCodeClone(body: {
     url?: string;
     github?: string;
-    parent_dir: string;
+    parent_dir?: string;
     name?: string;
   }): Promise<CodeCloneJobSnapshot> {
     return requireParsed(
--- a/crates/tidebreak-desktop/ui/src/api/types.ts
+++ b/crates/tidebreak-desktop/ui/src/api/types.ts
@@ -167,6 +167,8 @@ import {
   type CodeSessionDigest as WireCodeSessionDigest,
   type CodeUpdateNotice as WireCodeUpdateNotice,
   type CodeCloneDefaults as WireCodeCloneDefaults,
+  type CodeRepoSource as WireCodeRepoSource,
+  type CodeRepoSources as WireCodeRepoSources,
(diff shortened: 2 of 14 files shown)


naingthet:
**NO GO** at `e830de10376cac14e1de4504f7d778ccdfb15383`

1. `repo_sources` marks `github` available only when `git` is present **and** `gh` is found **and** authenticated. The clone path already resolves `owner/repo` to HTTPS when `gh` is missing (`resolve_github_clone_url`, and the old GitHub stage said you can still clone over HTTPS). A machine with git and no GitHub CLI credential can still serve that form for public repos; this probe hides it and leaves only the raw URL field. Gate `github` on `git` (same as `git_url`), and keep `gh` auth as remediation/hint, not as availability.

naingthet:
Fixed at `16d575c8d`.

You are right, and the old GitHub stage said so itself — its hint ended "You can still clone over HTTPS." `resolve_github_clone_url` falls back to the public URL at `gh.rs:168` when `gh` is absent, so hiding the form took away something that works on any machine with `git`.

`github` now needs exactly what `git_url` needs. The missing `gh` credential rides along as `remediation` on an available source, meaning "public repositories only" rather than "not offered".

One thing that fix exposed: the GitHub stage read that hint from `getCodeCloneDefaults`, which is administrator plane. A member on a shared machine — the person this whole change is for — was told nothing about why `owner/repo` would not reach their private repositories. The hint now comes from the member-plane probe, with the defaults read kept as a fallback.

Also in this push: the route never reached the runtime. Handlers take `ScopedCode` and it had no `repo_sources` to forward, which is what failed all five Rust lanes.

naingthet:
**G2G** at `16d575c8deddf5a60456a4f4d4bd1eb7be9c11d3`.

The prior NO GO is fixed. `repo_sources` now offers `github` whenever `git` is available. Missing `gh` is a hint on an available source (public HTTPS fallback), not unavailability. Clone resolution still uses that fallback, so the palette no longer hides a path that works.

No other P1/P2.

PR #2528, opened 2026-08-22
feat(server): a hosted machine offers each caller their own gateway models

### LINKED ISSUE #2517: A hosted machine's model picker shows the curated list, not the caller's gateway entitlements
A caller on a gateway-authenticated hosted machine now sees models, but they are the curated Anthropic list, not what their own Model Gateway account entitles them to. A deployment whose gateway serves GPT or Gemini offers none of it, and a caller entitled to less than the curated list is offered models their next turn will be refused for.

#2509 fixed the half that made the machine unusable: policy names the deployment's gateway (`hosted_gateway_url`), the sign-in gate stays off, and `provider_is_usable` reports the provider the caller's own credential serves. The catalog itself is still `model_registry::models_for(Anthropic)`, hard-coded in `collect_routes` (`crates/tidebreak-server/src/providers.rs`).

## Why the deployment-wide snapshot cannot be reused

`gateway_snapshot_for_policy` reads one snapshot per deployment, written only by `sync_models`, and only while a gateway session exists. A hosted machine holds no session, and entitlements are per user — so a snapshot in deployment-wide state would be one caller's entitlements served to everybody. Any cache here has to be keyed by owner.

## The shape to build

The machine already holds each caller's live machine-bound token (`OboInference::record_caller`, decision 51). Fetch `GET /api/v1/me/catalog` per caller with that credential, cache it in process memory keyed by `OwnerId` beside the exchanged inference token, and let the catalog surfaces read it.

The plumbing is the bulk of the work, because the surfaces that decide what a caller may select take no owner today:

- `catalog_models`, `provider_is_usable`, `model_is_usable`, `list_providers`, `resolve_model_policy` — all `(store, secrets, policy)`.
- `collect_routes` builds the gateway route from the snapshot; on a hosted machine it would build it from the caller's catalog, with the same frozen-identity and protocol handling.
- `model_roles::resolve` has no caller at all. Background roles (chat titling, the approval judge) therefore resolve to nothing on a hosted machine today. Per-caller entitlements do not obviously fix that — decide deliberately whether a background role runs as the caller who triggered it or not at all.

## The open contract question

Settle this before building. `/api/v1/me/catalog` is read with a `control`-audience token (`RESOURCE_CONTROL`). A hosted machine holds two credentials for a caller and neither is `control`: the machine-bound `tidebreak:<sha256>` token, and the `llm`-audience token decision 51 exchanges for. So one of these has to be true, and which one changes the design:

1. The gateway accepts a machine-bound token at `/api/v1/me/catalog`. Nothing new is needed.
2. The exchange can mint a `control` audience for the same subject. That widens what decision 51 calls an inference-only exchange, so it needs a decision record and a gateway-side change.
3. Neither. Then the catalog has to come from somewhere else — the attached desktop's own session is the obvious candidate, but the server validates every model selection, so a client-side-only catalog does not work as it stands.

## Done when

- The picker on a hosted machine lists the models the caller's own gateway account is entitled to, and a caller entitled to nothing is offered nothing.
- Two callers on one machine see different catalogs, and no per-caller catalog is written to deployment-wide state.
- A turn selects and runs a model from that catalog.
- A test pins each of those.

naingthet:
Closes #2517.

A caller on a gateway-authenticated hosted machine saw the curated Anthropic list, not what their own Model Gateway account entitles them to. Models their next turn would be refused for were offered, models the deployment's gateway does serve were missing, and background roles (chat titling, the approval judge) resolved to nothing because they walked stored provider rows a hosted machine never has.

## The contract question, settled

Decision 62 (in this PR) answers the issue's open question with the gateway's new capability: the machine exchanges the caller's live machine-bound token for a `catalog` capability (model-gateway ADR 0082, merged as brightwave-inc/model-gateway#1279) and reads `/api/v1/me/catalog` as that caller. Option 1 (accepting machine-bound tokens at the catalog) and a client-supplied catalog are rejected in the record.

## What changed

- `obo_inference.rs` becomes `obo_gateway.rs`: the same per-caller exchange machinery now mints two capabilities — inference (decision 51, unchanged) and the entitlement snapshot. Snapshots are cached per owner in process memory: fresh 300 s, ETag revalidation, a transport-only stale grace of 3600 s, and refusals always propagate, so a revoked session stops the caller.
- The caller's routes are real gateway routes: `collect_routes` builds them with the same per-protocol, frozen-identity builder the managed profile uses (factored into one `gateway_snapshot_routes`), from the caller's snapshot. Nothing impersonates the curated Anthropic route anymore, so a stored BYOK configuration coexists instead of outranking — decision 51's rule 3 retires with the impersonation that forced it.
- Catalog surfaces resolve per caller: `catalog_models`, `list_providers`, `provider_is_usable`, `model_is_usable`, `resolve_model_policy`, `effective_chat_policy`, selection validation, the turn-accept freeze, the turn worker's re-resolution, and the image-capability gate all take the requesting caller's snapshot.
- Background roles run as the caller who triggered them: the turn worker, chat titler, workspace titler, and approval judge resolve the utility role over the triggering owner's entitlements and drive it on their credential (`resolve_for`). No caller means the work is skipped.
- Named gaps, left open in the record: the web-search sub-request and the sandbox model paths still resolve without a caller.

## Tests

- `obo_gateway`: snapshot fetch shape and conversion, memory freshness, ETag revalidation on 304, refusal fails closed with no grace, per-caller isolation, unknown caller gets nothing — against a fake gateway that asserts the catalog read presents a catalog capability.
- `tests/configuration`: per-protocol routes with frozen identities from the caller snapshot; a caller entitled to nothing offered nothing; two callers see two catalogs and nothing reaches deployment-wide state; a hosted selection freezes through the caller's snapshot and a foreign snapshot refuses it; background roles walk the caller's entitlements and skip without one; the providers list names the caller's gateway; stored BYOK and the caller path coexist; per-caller provider cache isolation (seeded snapshots).

Depends on the gateway serving audience `catalog` (already on model-gateway main; deployments predating it get the explicit member-catalog refusal, and callers are offered nothing rather than a guess).

### DIFF
--- a/crates/tidebreak-server/src/approval_judge.rs
+++ b/crates/tidebreak-server/src/approval_judge.rs
@@ -298,6 +298,9 @@ fn user_prompt(action: &str, context: &[JudgeContextMessage]) -> String {
 
 /// Polls for judge-owned approvals and lands one verdict per call.
 pub(crate) struct ApprovalJudgeWorker {
+    /// Per-caller gateway capabilities on a hosted machine (decisions 51 and
+    /// 62): the judge runs as the caller whose approval it reads.
+    on_behalf_of: Option<Arc<crate::obo_gateway::OboGateway>>,
     store: Arc<dyn Store>,
     resolver: Arc<dyn ProviderResolver>,
     secrets: Arc<dyn tidebreak_core::SecretProvider>,
@@ -317,6 +320,7 @@ impl ApprovalJudgeWorker {
         approvals: Arc<ApprovalBroker>,
     ) -> Self {
         Self {
+            on_behalf_of: None,
             store,
             resolver,
             secrets,
@@ -325,6 +329,14 @@ impl ApprovalJudgeWorker {
             approvals,
             poll: Duration::from_millis(750),
         }
+    }
+
+    pub(crate) fn with_on_behalf_of_gateway(
+        mut self,
+        gateway: Option<Arc<crate::obo_gateway::OboGateway>>,
+    ) -> Self {
+        self.on_behalf_of = gateway;
+        self
     }
 
     pub(crate) async fn run(self) {
@@ -377,19 +389,32 @@ impl ApprovalJudgeWorker {
         let Some(action) = action_description(preview) else {
             return Ok(false);
         };
-        // No utility model configured means no judge, not a cheaper gate.
+        // The judge runs as the caller whose approval it reads: their own
+        // entitlements pick the utility model, and their credential drives it
+        // (decision 62). No utility model configured means no judge, not a
+        // cheaper gate.
+        let owner = self
+            .store
+            .chat_owner(approval.chat_id)
+            .await
+            .unwrap_or_default();
+        let caller_gateway = match (self.on_behalf_of.as_ref(), owner.as_ref()) {
+            (Some(gateway), Some(owner)) => gateway.snapshot_for(owner).await.ok().flatten(),
+            _ => None,
+        };
         let Some(utility) = crate::model_roles::resolve_utility_model(
             &*self.store,
             &*self.secrets,
             &*self.provisioned_policy,
             &*self.os_policy,
+            caller_gateway.as_ref(),
         )
         .await?
         else {
             return Ok(false);
         };
         let context = conversation_digest(&self.store.list_messages(approval.chat_id).await?);
-        let provider = self.resolver.resolve().await;
+        let provider = self.resolver.resolve_for(owner.as_ref()).await;
         let verdict = request_verdict(
             provider.as_ref(),
             &utility,
--- a/crates/tidebreak-server/src/auth.rs
+++ b/crates/tidebreak-server/src/auth.rs
@@ -146,11 +146,12 @@ pub async fn require_token(
                 let bearer: std::sync::Arc<str> = presented
                     .expect("a resolved principal had a presented token")
                     .into();
-                // Remember the live caller token so this caller's turns can be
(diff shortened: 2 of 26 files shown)


naingthet:
**G2G** at `cac2e38eaa0fc7fa381dfca8314a715568d6b2d8`

Per-caller catalog exchange, cache (fresh/ETag/stale grace), and fail-closed refusals match decision 62. Routes are built from the caller snapshot with the same per-protocol builder as the managed path, so BYOK no longer outranks the gateway impersonation that used to stand in for entitlements. Surfaces and background roles take the triggering owner; the named web-search and sandbox gaps are left open on purpose. Isolation and freeze tests cover the mix-up that would actually leak models. Nothing here is a P1/P2 CI would miss.

naingthet:
Clippy on `cac2e38ea`: the route builder made `inference_base_url` dead (both compat URLs are derived from the gateway base now). `428df62ea` removes the field and its accessor; the three URL tests assert the base instead, and the subpath behavior they guarded lives on in the derived form. Nothing else changed.

naingthet:
**G2G** at `428df62ea349a4b6a492c7e4071981430034b94e`

The head after `cac2e38e` only drops the unused precomputed Anthropic inference base. Routes still derive `/compat/anthropic` from the gateway base, and the URL tests now assert that base. Per-caller catalog exchange, cache, fail-closed refusals, and owner-scoped surfaces are unchanged. No new P1/P2.

PR #2530, opened 2026-08-22
feat(code): reconcile pull-request facts and surface the workspace's set

naingthet:
Slice B of #2523 (closes #2525). Builds on the fact substrate from #2529.

The delivery read path now persists what it observes. `query_pull_requests` and `pull_request_detail` upsert fact snapshots for tracked pull requests — exact-linked to a workspace, or already holding a fact row — and mint `contributed` attribution for the exact tiers only (`discovered_via: reconcile`). The branch-name guess stays display-time and never mints. Stored attribution folds back into `workspace_links` before caching, so links survive workspace archive/release, carry an optional `relation`, and keep the established ordering (the notifications store routes to the first link).

A 61-second reconcile sweep (coprime with the 47s watch and 53s trigger sweeps) drives that read per owner over every live registered repository plus every identity already holding facts — which is how a cross-repo pull request the post-turn detector observed keeps itself fresh without a local checkout. The sweep also populates the `code_repo` origin identity columns added in #2529 and refreshes them when origin changes.

Surfaces:
- `GET /code/workspaces/{id}/pull-requests` — the workspace's attributed set from the durable store, open first then newest, no host read.
- `pr_count` on session digests (absent when zero), the cheap client signal to re-read the list. `pr_state` stays single-valued.
- The inspector's Review tab lists the attributed set above the single-PR panel once there is more than one; selecting a non-primary row shows its stored snapshot, the primary row returns to the live resource.
- The workspace card's PR chip gains the attributed count ("3 PRs") while keeping its primary pull request.

Tests: server integration (`code_reconcile.rs`, shimmed gh) — exact tier mints once and is idempotent, branch tier never mints or persists, origin columns populate, a delivery read serves durable links with the relation; UI vitest for the new parsers (unknown relation/state rejected, `pr_count` through digest notices), the chip helper, and the inspector's empty-list fake. Stories: inspector `ReviewPullRequestSet` (multi-PR list with a cross-repo entry), workspace card `SeveralPullRequests`.

Local validation: `UPDATE_WIRE_TYPES=1 cargo test -p tidebreak-server` regen (committed), incremental `cargo check --all-targets` clean, `pnpm test` (1705), `biome lint`/`format`, `pnpm build`, `storybook:build`. Rust test suites run in CI.

Next: stack lanes (#2526), then `pr_opened`/`pr_updated` trigger conditions (#2527).

PR #2529, opened 2026-08-22
feat(code): pull-request facts and workspace attribution

naingthet:
Slice A of #2523 (closes #2524).

A workspace produces pull requests over its whole life, sometimes in other repositories than its checkout, and nothing durable records them: `code_workspace.pr` is one digest with no repo identity, and the delivery surface's PR-to-workspace link is a per-request heuristic over a 30-second cache.

This adds the substrate:

- `code_pull_request` — a durable, confirmed observation of one pull request, keyed `(owner, host, repo_owner, repo_name, number)`, carrying the snapshot stack derivation and trigger edges will need (state, head/base branch, head SHA, host timestamps, `first_seen_at`/`last_seen_at`). Checks and review state stay live-only in delivery.
- `code_pull_request_attribution` — one row per `(pull_request, workspace)` holding the strongest claim. Only two acts mint: `gh pr create` (`authored`) and `git push` whose branch is a PR head (`contributed`). Views, checkouts, comments, closes, and merges never mint, so review/triage agents stay out of the attributed set. Rows carry the session and the subagent `Task` span (`parent_call_id`) when one acted.
- A post-turn detector: reads the closed turn's journaled `ToolDetail::Command` events, recognizes the two acts via `tidebreak_shell_policy::simple_command_argvs` (fail-closed), resolves the repository from `--repo`, the tool output's PR URL, or the cwd's origin, and confirms with one repository-qualified `gh` read before writing anything. Bounded per turn (400-event tail, 200 parses, 4 confirms); never fails or delays the turn. Failed commands are never confirmed.
- The user-initiated create/push routes mint through the same confirm-then-record helper.
- `code_repo` gains nullable `origin_host`/`origin_owner`/`origin_name` (populated by the slice-B reconcile sweep).
- Migration `m20260822_000009_code_pull_request_facts` per ADR 0061 (no baseline edit, no epoch change).
- ADR 0062 records the substrate, the two-act attribution rule, and the exclusions (no journal backfill, no webhooks, no author-search net, checks live-only).

Tests: core ops (identity upsert preserves `first_seen_at`, claim-once, contributed→authored promotion, owner isolation) and detector integration against a shimmed `gh` (create mints authored; push mints contributed; view/comment/checkout/close and failed creates mint nothing and never reach the host; signed-out gh mints nothing; re-running the sweep is idempotent).

Local validation per repository policy: rustfmt on both crates; Rust compilation and tests run in CI. No wire-type changes in this slice.

[...]

Issue #2527, opened 2026-08-22
PR facts: pr_opened/pr_updated trigger conditions and stack-aware classification


naingthet:
Slice D of #2523. `pr_opened` and `pr_updated` trigger conditions sourced from fact edges with a PR-keyed fire table; the trigger CHECK constraint rebuild; Behind/ReviewRequired suppression and watch parking for stacked children waiting on a parent. Deferred: fires for PRs with no attributed workspace; triggers on unregistered repos.

Issue #2523, opened 2026-08-22
Durable cross-repo PR tracking for agent workspaces


naingthet:
Agent workspaces produce many pull requests, sometimes in other repositories than the checkout, and Tidebreak cannot show every PR a workspace or its subagents worked on. `code_workspace.pr` holds one digest with no repo identity, the delivery view is a 30-second cache, and the PR-to-workspace link is a per-request heuristic.

This epic adds a durable `code_pull_request` fact substrate with workspace/session/subagent attribution, fed by a post-turn command detector (journaled `gh pr create` / `git push` acts, confirmed against GitHub) and a reconcile sweep; joins it into the delivery and workspace surfaces; derives stacked-PR chains; and adds `pr_opened`/`pr_updated` trigger conditions.

Attribution rule: only `gh pr create` (authored) and `git push` to a PR head (contributed) mint attribution. Views, checkouts, comments, closes, and merges never do, so review and triage agents stay out of the bird's-eye view.

Slices, in landing order:

- [x] #2524
- [x] #2525
- [x] #2526
- [x] #2527

PR #2531, opened 2026-08-23
feat(code): derive stacked pull requests and render them as lanes

### LINKED ISSUE #2526: PR facts: stacked pull request chains in delivery and workspace surfaces
Slice C of #2523. Server derives stack parents from fact rows (base branch = another open PR's head); `stack_parent_number` on delivery summaries; indented stack lanes in the delivery page; workspace card shows its stack parent from `base_ref`.

naingthet:
Slice C of #2523 (closes #2526). Builds on the fact substrate (#2529) and reconcile path (#2530).

Stacks are derived, never declared. A pull request is stacked on another when its base branch is that pull request's head branch in the same repository, and only open parents count — a merged parent's branch is history, not a base.

Server: the delivery read path annotates each summary with `stack_parent_number` from the durable fact set (the per-repo fact list the fact-persistence seam already loads, kept current with the pass's own upserts). Deriving from facts rather than page items means a parent outside the current page or filter still resolves. An untracked child stacked on a tracked parent gets the annotation too.

Delivery list: children nest directly under their parent, indented per hop with `data-depth` marked, and roots keep the incoming order. A child whose parent is not among the loaded rows stays at depth 0 with a "Stacked on #N" chip instead of a hidden edge. Cycles break at the first revisited node and the remainder render as orphans, never dropped. Chains cap at depth 10. Cross-repo number collisions never link (row keys are full identity).

Rail: the local pre-PR signal reads the same way — a workspace whose `base_ref` names a sibling workspace's branch (with `origin/` stripped, matching how PR creation resolves the base) shows a "Stacked on <title>" child row that opens that sibling. Stacks have been creatable this way all along; this is the read side.

Tests: pure lane arrangement (chains, root order, off-page parent, cross-repo isolation, cycle break, depth cap), `workspaceStackParent` (match, `origin/` strip, self/repo/default-base guards), a delivery dom test asserting `data-depth` `["0","1","2","0"]` plus the orphan chip, the parser round-trip for the new optional field, and the server integration extended with a stacked child (annotated via its tracked parent's fact while holding no fact row itself; the parent on main annotates nothing). Stories: `PullRequestStacks` scenario with a play asserting depths, `StackedOnSibling` workspace card.

Local validation: wire regen committed; incremental `cargo check --all-targets` clean; `pnpm test` (1714), lint, format, `pnpm build`, `storybook:build`. Rust suites run in CI.

Next: `pr_opened`/`pr_updated` trigger conditions + stack-aware watch/trigger classification (#2527).


naingthet:
**G2G** cd6d8b1e0a522160c0a2f8a1e425a3e0ba61564e

Stack parent is derived from durable open heads, lanes nest on the page, and off-page or cyclic parents stay visible as orphans. Workspace sibling matching strips `origin/` the same way PR creation does. No high-impact defect in the read path or the render contract.

PR #2532, opened 2026-08-23
feat(code): fire pr_opened and pr_updated from fact edges, hold stacked children

### LINKED ISSUE #2527: PR facts: pr_opened/pr_updated trigger conditions and stack-aware classification

naingthet:
Slice D of #2523 (closes #2527) — the last slice. Builds on the fact substrate (#2529), reconcile path (#2530), and stack derivation (#2531).

Two new trigger conditions are edges of the durable fact store, not digest states — `classify_trigger_condition` deliberately never returns them:

- `pr_opened` fires once per pull request whose host `created_at` and local `first_seen_at` both postdate the trigger's arming, so arming over existing history stays silent. The fire's fingerprint is the sentinel `opened` instead of a head SHA — the edge is existence, not a head.
- `pr_updated` fires once per distinct head. The first observed head lands as a settled baseline row (`insert_settled_trigger_fire`: state `delivered`, no payload — the state-shape CHECK permits it, the outbox sweep never picks it up, and the conflict identity still dedupes), so arming or first sight never notifies; later heads insert pending fires.

The fact pass runs inside `sweep_owner` before the remote-read gate (which requires a persisted `workspace.pr` digest and would drop fact-only repositories). It reads only the local store — repo identity from the `origin_*` columns the reconcile sweep maintains, facts, attributions — and delivers to eligible attributed workspaces through the existing outbox, so leases, bounded retries, and at-most-once receipts apply unchanged.

Schema: both 8-token CHECKs (`code_trigger.condition`, table-level from the frozen baseline, and `code_trigger_fire.delivery_condition` from the outbox rebuild) widen via `m20260823_000010_trigger_fact_conditions` — a paired rebuild ordered so nothing is dropped while referenced (PostgreSQL enforces the cascade FK), reusing the outbox table builder now parameterized by trigger-table name and token list (the shipped migration's rendered DDL is unchanged; the stepwise-vs-fresh schema test guards that).

Stack awareness: `claim_fires` holds `Behind`/`ReviewRequired` when the summary carries `stack_parent_number` — the child is behind *because of its parent*. To make that signal real on the sweep's path, `query_pull_requests_by_number` now persists facts and annotates stacks like the list read. The watch's `Actionable(Behind)` arm parks with "waiting on its parent pull request #N" (resolved from the fact set via the workspace's stored origin) before the rebase loop can start.

Arming surface: the two conditions join `CodeTriggerRules` with copy that states the baseline rule ("The first observed head never notifies").

Tests: a seeded-data migration test (rows survive the paired rebuild, `pr_opened` inserts on both rebuilt CHECKs, a garbage token still fails); integration against shimmed `gh` (`code_trigger_facts.rs`): `pr_opened` fires exactly once across two sweeps and only for attributed workspaces; `pr_updated` baselines silently then fires exactly the moved head; a BEHIND stacked child produces zero fires while the watch-side parent lookup resolves `Some(412)` from the same facts and `None` on a default base. The condition round-trip test covers the new tokens via `all()`.

Deferred, per ADR 0062: triggers on repositories with no local registration, and fires for pull requests with no attributed workspace.

Local validation: incremental `cargo check --all-targets` clean on both crates; wire regen committed; `pnpm test` (1714), lint, format, `pnpm build`, `storybook:build`. Rust suites run in CI.

### DIFF
--- a/crates/tidebreak-core/src/code/mod.rs
+++ b/crates/tidebreak-core/src/code/mod.rs
@@ -1138,6 +1138,14 @@ pub enum CodeTriggerCondition {
     Merged,
     /// The pull request closed without merging.
     Closed,
+    /// A pull request came into existence (decision 62). Edge-sourced from
+    /// the durable fact store's `first_seen_at`, never from
+    /// [`classify_trigger_condition`].
+    PrOpened,
+    /// A tracked pull request's head moved (decision 62). Edge-sourced from
+    /// the fact store, once per distinct head; the first observed head is a
+    /// silent baseline, never a notification.
+    PrUpdated,
 }
 
 impl CodeTriggerCondition {
@@ -1153,6 +1161,8 @@ impl CodeTriggerCondition {
             Self::ReadyToMerge => "ready_to_merge",
             Self::Merged => "merged",
             Self::Closed => "closed",
+            Self::PrOpened => "pr_opened",
+            Self::PrUpdated => "pr_updated",
         }
     }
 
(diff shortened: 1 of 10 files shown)


naingthet:
**G2G** 6ba48b33a0a7bf6f81f9f3d3ac5c609d844a9960

Fact-edge firing, silent first-head baseline, stacked-child hold, and the paired CHECK rebuild match the stated contracts. No high-impact defect that CI would miss.

PR #2533, opened 2026-08-23
feat(code): hand Fix CI the failing job logs

naingthet:
## What changed

Fix CI used to name the failing checks and tell the agent to go and find their logs, which cost two or three `gh` calls before it saw an error line and then pulled a whole multi-megabyte job log into context in one read. The action now downloads them first: a new `POST /code/workspaces/{id}/pr/check-logs` reads the job id out of each failing check's URL, fetches that job's log through `gh api`, and writes it into the workspace's private root — beside fork transcripts, outside anything Git indexes, and already on the session's read roots — and the prompt names the paths and opens by telling the agent to read them. Each log is capped at 512 KB, tail-kept, and headed with the check, the job URL, the head SHA, and how much was dropped; every press replaces the previous set, so the directory only ever describes one head. The fetch is best-effort, so a GitHub outage, a signed-out `gh`, or a check from a provider with no downloadable log still sends the turn, with the old find-them-yourself wording and a toast saying the files are missing.

## Validation

Ran the full desktop UI suite (1716 tests, including new coverage for the prompt, the fetch fallback, the parser, and the button's busy state), `tsc --noEmit`, Biome lint and format, and the Storybook build. Rust was not compiled locally: `rustfmt --check` is clean on every changed file, and the hand-written `wire.ts` block was verified line-by-line against the Rust doc comments, so CI owns compilation, the `ci_logs` unit tests, and the wire-type regeneration check.

## Compatibility and risk

Additive wire types (`CodeCheckLog`, `CodeCheckLogError`, `CodeCheckLogsSnapshot`) and one new route; no persisted schema change. Logs land in the existing per-workspace private root rather than the worktree, so Git cannot index them and no new read boundary is opened. The watch task's fix turn still runs the old instruction — the materializer is shaped so it can adopt this later.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/api/client.ts
+++ b/crates/tidebreak-desktop/ui/src/api/client.ts
@@ -106,6 +106,7 @@ import {
   type CodeCloneDefaults,
   type CodeRepoSources,
   type CodeWorktreeRoot,
+  type CodeCheckLogsSnapshot,
   type CodeForkTranscript,
   type CodeCloneJobSnapshot,
   type CodeHarnessInstallSnapshot,
@@ -146,6 +147,7 @@ import {
   parseCodeCloneDefaults,
   parseCodeRepoSources,
   parseCodeWorktreeRoot,
+  parseCodeCheckLogsSnapshot,
   parseCodeForkTranscript,
   parseCodeCloneJob,
   parseCodeHarnessInstall,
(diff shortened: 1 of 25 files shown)


naingthet:
**G2G** `afc61e5aa2115e32015cd0a09f2fa9f777d41b68`

The check-log fetch is bounded, written only under the workspace private root, and best-effort so a missing job still starts the fix turn. Job ids come from Actions URLs, filenames are slugged, and stale `.log` files are pruned per head. I do not see a concrete high-impact defect CI would miss.

naingthet:
**G2G** `7fcc839c3e3d8017acd356bbeaab1c15131350a1`

The prior verdict still holds on this head. Check-log fetch stays bounded (six jobs, 512 KB tail, 20s `gh` deadline), files land only under the workspace private root, and a missing or failed download still starts the fix turn with the old find-them-yourself wording and a toast.

No new high-impact defect in the delta.

naingthet:
**G2G** 967e9faf744f1ce0bc27ab5a8d31ad2fa03ab29c

Delta from 7fcc839c3e3d8017acd356bbeaab1c15131350a1 owns the check name before the stream instead of carrying `&PullRequestCheck`. Fetch bounds, private-root writes, and best-effort fallback are unchanged. No P1/P2.

PR #2534, opened 2026-08-23
feat(code): a hosted machine clones and pushes with the gateway's GitHub App

### LINKED ISSUE #2510: Hosted machines have no GitHub identity, so a team cannot link private repositories
A team cannot link its GitHub repositories on a hosted machine, because Tidebreak has no GitHub identity of its own.

The only GitHub authentication is observing the server's `gh` CLI — `crates/tidebreak-server/src/code/gh.rs:888` runs `gh auth status` and reads no token, per decision 34 ("credentials are observed, never brokered"). There is no App, no OAuth, no PAT storage, and `gh.rs` takes no `OwnerId` anywhere. Clone and push deliberately disable prompting (`clone.rs:377`, `gh.rs:1488`), so there is no injection point either.

On a shared machine that means every member acts as whatever single GitHub account the server's OS user is logged into — or, far more likely, as nobody at all.

Meanwhile a gateway deployment can already hold a GitHub App: connected apps of kind `git_forge` with `credential_mode: github_app` mint installation tokens today. That is the credential this should use, so the machine never stores a GitHub secret of its own.

## Done when

- A hosted machine can clone a private repository the gateway's GitHub App is installed on, with no local involvement and no `gh` on the machine.
- The credential is minted per request and never written to the store.
- Commit and PR authorship are legible: state plainly in the UI when work is attributed to the App rather than the person, since `personal_credential_policy: installation_only` deployments cannot attribute it to the person.
- A deployment with no git forge configured says so rather than offering the option.

Needs design agreement with the gateway on the token-minting surface before implementation. Depends on the repo-source capability probe for the UI half.

naingthet:
Closes #2510.

## Problem

A hosted machine has no GitHub identity. The only GitHub authentication is observing the server's `gh` CLI (decision 34), and a hosted machine has no `gh` and no person at the keyboard — so a team cannot clone its private repositories, and every member acts as nobody.

## What this does (decision 63)

The deployment's gateway already holds the right identity: a git-forge connected app in `github_app_installation` mode mints hour-long, repository-scoped installation tokens (gateway ADR 0083, model-gateway#1281 + model-gateway#1294 for the no-mint probe).

- **Per-operation borrow.** A GitHub clone and a workspace push each present the caller's live machine-bound token to `POST /api/v1/tidebreak/git-credential` and receive a token scoped to the one repository the operation touches. Minted per request, held in memory for the operation, never written to the store, the checkout's config, or the persisted clone URL.
- **Injection without storage.** The credential rides the environment into a one-shot `credential.helper`; the helper list is emptied first, so no configured helper answers ahead of it — or stores the dying token when git offers it back. A real `git credential fill` test pins the helper.
- **Per-caller repo sources.** On hosted machines the `github` source stops consulting `gh` and reflects the gateway's probe for the requesting caller: offered with the attribution sentence, or hidden with the gateway's reason (`no_git_forge` → "not offered", connect-mode → "register an installation-mode forge", etc.). Local machines keep the existing observation path bit for bit.
- **Legible attribution.** The add-repo dialog's GitHub source states that work lands as the App's bot account, and the workspace git card names the acting identity (`pushes_as`) beside the push control — new PrCard story and DOM tests.
- **Fail closed, no blur.** A gateway refusal fails the clone/push with the gateway's reason rather than retrying uncredentialed. Checkouts whose origin is not a forge repository (local bare origins) borrow nothing and push exactly as before — pinned by test.

## Tests

- `obo_gateway`: borrow-per-operation (two borrows are two mints), probe freshness + typed refusal caching, fail-closed on dead sessions and unknown callers, machine-bound bearer + resource asserted server-side in the fake gateway.
- `code_clone`: per-caller sources on hosted machines (offered with attribution / refused with reason), and a refused clone fails with the gateway's reason before any network.
- `code_git`: a local-origin push borrows nothing and keeps working; `pushes_as` follows the origin; a refusal stops the push with `git_forge_refused`.
- `gh`: the one-shot helper answers `git credential fill` from the environment.
- UI: PrCard DOM tests + story (`HostedPushesAsApp`), parser coverage; storybook build passes.

Decision record: `docs/decisions/0063-hosted-machines-borrow-forge-credentials.md`. PR creation and delivery reads still ride `gh` and stay unavailable on hosted machines — named as the next slice in the decision's revisit-when.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/PrCard.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/PrCard.dom.test.tsx
@@ -89,6 +89,26 @@ describe("PrCard", () => {
     ).not.toBeInTheDocument();
   });
 
+  it("states plainly when pushes land as the deployment's GitHub App", () => {
+    renderState({
+      ...BASE,
+      unpushed: true,
+      ahead: 1,
+      gh_found: false,
+      gh_authenticated: undefined,
+      pushes_as: "tidebreak-ship[bot]",
+    });
+    expect(screen.getByText("tidebreak-ship[bot]")).toBeInTheDocument();
+    expect(screen.getByText(/not as your GitHub account/)).toBeInTheDocument();
+  });
+
+  it("names no acting identity on a machine with its own credentials", () => {
+    renderState({ ...BASE, unpushed: true, ahead: 1 });
+    expect(
+      screen.queryByText(/GitHub App — not as your GitHub account/),
+    ).not.toBeInTheDocument();
+  });
+
   it("enables create PR after a push with no pull request", async () => {
     const { onCreatePr } = renderState({
       ...BASE,
--- a/crates/tidebreak-desktop/ui/src/code/PrCard.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/PrCard.tsx
@@ -293,6 +293,15 @@ export function PrCardView({
           {error}
         </p>
       )}
+      {snapshot?.pushes_as && (
+        <p className="text-muted-foreground text-xs leading-relaxed">
+          Pushes from this machine land as{" "}
+          <span className="text-foreground font-medium">
+            {snapshot.pushes_as}
+          </span>
+          , the deployment&apos;s GitHub App — not as your GitHub account.
+        </p>
+      )}
       {snapshot?.dirty ? (
         <div className="flex flex-col gap-2">
           <label className="flex flex-col gap-1">
--- a/crates/tidebreak-desktop/ui/src/code/parsers.ts
+++ b/crates/tidebreak-desktop/ui/src/code/parsers.ts
@@ -1676,6 +1676,7 @@ export function parseCodeWorkspacePr(
       "gh_found",
       "gh_authenticated",
       "remediation",
+      "pushes_as",
       "watch",
     ]) ||
     typeof value.dirty !== "boolean" ||
@@ -1686,7 +1687,8 @@ export function parseCodeWorkspacePr(
     typeof value.gh_found !== "boolean" ||
     (value.gh_authenticated !== undefined &&
       typeof value.gh_authenticated !== "boolean") ||
-    typeof value.remediation !== "string"
+    typeof value.remediation !== "string" ||
+    (value.pushes_as !== undefined && typeof value.pushes_as !== "string")
   ) {
     return null;
   }
@@ -1701,6 +1703,7 @@ export function parseCodeWorkspacePr(
     ...(value.gh_authenticated !== undefined
       ? { gh_authenticated: value.gh_authenticated }
       : {}),
+    ...(value.pushes_as !== undefined ? { pushes_as: value.pushes_as } : {}),
   };
   if (value.pr !== undefined) {
     const pr = parsePullRequestDigest(value.pr);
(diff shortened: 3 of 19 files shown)


naingthet:
**NO GO** `6e747a796a20054361f5951edd378420152ef6c3`

1. A push on a hosted machine will mint a real GitHub App installation token and then offer it to whatever host `origin` names. `borrow_git_credential` treats any URL that `repository_target_from_path` can parse as a forge repository. That parser already accepts `https://evil.example/acme/private.git` (and `host/owner/repo` forms) and only the `owner/name` pair is sent to the gateway. The one-shot helper then answers every `get` from the environment with no `host=` / `protocol=` check. Anyone who can change `origin` — including the workspace agent — can aim the next push at a host they control and collect the dying token. Fail closed unless the origin host is the forge (`github.com`), and make the helper refuse `get` for any other host.

naingthet:
Addressed in `fd66596e2`.

Finding 1: the credential is now confined to the forge's host at two independent points.

- `borrow_git_credential` lends only when the origin parses **and** its host is the forge's (`github.com`, one shared `forge_lending_target` predicate). A rewritten origin — `https://evil.example/acme/private.git`, `host/owner/repo` forms — borrows nothing: no mint, and the push runs uncredentialed exactly as a decision-34 machine would. The same predicate gates the `pushes_as` identity line, so the UI never claims wider than the lending.
- The one-shot helper now reads the credential description and answers `get` only for `protocol=https` and the exact host in `TIDEBREAK_GIT_CREDENTIAL_HOST`. A redirect or rewrite that reaches git anyway collects nothing — git sees "no credential" and fails on disabled prompts.

Pinned by tests: a real `git credential fill` exercise (forge host answers; `evil.example` and cleartext do not), and the workspace test now rewrites `origin` to a foreign host and observes no identity claim and no mint reaching the gateway. Decision 63 records the confinement.

On "fail closed": for a foreign-host origin the machine refuses to *lend* rather than refusing the *push* — an uncredentialed push to a foreign host carries nothing to collect and preserves existing behavior for non-forge remotes; a gateway refusal for a lendable origin still fails the push with the gateway's reason.

naingthet:
**G2G** `fd66596e2d8b2d3300ab74a39d7eca0740f648fb`

The prior finding is closed. A retargeted origin no longer receives the minted App token.

`borrow_git_credential` now goes through `forge_lending_target`, which parses the origin and then requires the host to match `GIT_CREDENTIAL_FORGE_HOST` (`github.com`) before the gateway is called. A rewritten `https://evil.example/...` or `host/owner/repo` origin returns `Ok(None)`: no mint.

Independently, the one-shot helper only answers `get` when protocol is `https` and `host=` equals `TIDEBREAK_GIT_CREDENTIAL_HOST`, which is always set to that same forge host. Git asking for another host (or http) gets no username/password.

Both halves refuse on their own, so neither a parser-only origin nor a helper that trusted any host can leak the token.

No other P1/P2 in this delta.

naingthet:
**G2G** on 24740e774b1b12d23f41a39e0f87c57f924e63ca

The only change since the prior G2G at fd66596e2d8b2d3300ab74a39d7eca0740f648fb is the import path for `CodeGitHubRepositoryTarget` in `forge_lending_target`. Origin-host gating is unchanged: minting still requires github.com, and the one-shot helper still answers only https on that host.

naingthet:
**G2G** `6f450ae5c46aff32f7f49f19b6ee076558f8f983`

The only change since the prior G2G at 24740e774b1b12d23f41a39e0f87c57f924e63ca is the generated `pushes_as` doc comment in `wire.ts`. Origin-host gating is unchanged: minting still requires github.com, and the one-shot helper still answers only https on that host.

No other P1/P2 in this delta.

PR #2535, opened 2026-08-24
feat(code): recap each completed turn on the utility model

naingthet:
## What changed

Coming back to one of several running workspaces means reading a transcript to find out where it stands, so this derives a sentence instead: after a turn completes, a utility-role call reads that turn and answers where the work is and what happens next. Claude Code ships the same idea as `awaySummary`, but it reaches its reader only through the TUI's Ink hook or Remote Control's `notifyMetadataChanged`, and we drive engines headlessly over stream-json, which carries neither — deriving it ourselves is also what makes it one feature across Claude Code, Codex, OpenCode, and Grok rather than a Claude Code detail. It is deliberately not asked of the harness, because that would be a real turn that appends to the engine's history and advances the session, and the per-turn adapters have no child alive to ask between turns. The line lands on the existing `code_turn.narrative` column and rides the session digest every list surface already watches; the transcript shows it on the newest boundary only, since it describes the session rather than the turn. Titling's request and retry path is now generic over a `Proposal` trait so recaps reuse it instead of copying it, with chat and workspace naming keeping their existing bounds and behavior.

## Validation

`cargo fmt --check` is clean, and on the desktop UI `tsc --noEmit`, `biome lint`, and the full `vitest` suite (1714 tests, plus a new one pinning the newest-boundary rule) all pass. **No Rust was compiled or tested locally** — the new `code_recap` integration test has never run, and `src/generated/wire.ts` was hand-edited to match what `UPDATE_WIRE_TYPES=1 cargo test -p tidebreak-server` would emit rather than regenerated, so the generated-diff check is the thing most likely to fail first. CI is the gate on all of it.

## Compatibility and risk

The `/code/updates` digest and its notice gain an optional `recap` string; it is `skip_serializing_if = "Option::is_none"` and optional on the wire, so older clients are unaffected. No migration — `code_turn.narrative` is an existing nullable column that was previously always written as `None`. Recaps add one utility-model call per completed turn, which is new background spend: a machine with no utility model resolves nothing and silently has no recaps, matching how chat and workspace naming already degrade. The call uses `PromptCacheMode::OneShot`, so it writes no cache entries for a prefix nothing re-sends.

## Tracking

No issue.

### DIFF
--- a/crates/tidebreak-core/src/code/mod.rs
+++ b/crates/tidebreak-core/src/code/mod.rs
@@ -996,6 +996,11 @@ pub struct CodeTurn {
     /// Token usage as reported by the engine.
     pub usage: Option<CodeUsage>,
     /// Asynchronous narrative; never blocks lifecycle.
+    ///
+    /// Derived after the turn ends and written only by
+    /// [`crate::db::code::set_turn_narrative`]. `save_turn` deliberately leaves
+    /// the column alone, because its callers hold a snapshot taken before this
+    /// existed.
     pub narrative: Option<String>,
     /// Start time.
     pub started_at: chrono::DateTime<chrono::Utc>,
--- a/crates/tidebreak-core/src/db/ops/code/turn.rs
+++ b/crates/tidebreak-core/src/db/ops/code/turn.rs
@@ -274,6 +274,13 @@ pub async fn next_turn_ordinal(
 }
 
 /// Persist mutable turn fields. `id`, `session_id`, `ordinal`, and `started_at` stay as stored.
+///
+/// `narrative` is deliberately not among them. It is derived asynchronously
+/// after a turn ends and can land at any point, while the callers here hold a
+/// [`CodeTurn`] read before that — `checkpoint::after_turn_ended` takes its
+/// snapshot before the turn is even terminal. Writing the whole row from a
+/// stale snapshot would blank a recap that had already been stored, so the
+/// column has exactly one writer: [`set_turn_narrative`].
 pub async fn save_turn(store: &DbStore, owner: &OwnerId, turn: &CodeTurn) -> Result<bool> {
     let result = entities::code_turn::Entity::update_many()
         .col_expr(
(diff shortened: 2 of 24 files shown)


naingthet:
**NO GO** at 7a3bc8ed86556497f948b2e98171fb2e838b0f48

P2. `TurnRecapper` claims one recap per session and treats a failed claim as `NotApplicable`, so a turn that finishes while an earlier recap is still in flight is never recapped. The comment says the opposite: the later turn should win and the in-flight line should be the one dropped. What actually happens is the older call writes `narrative` onto the previous turn, `build_digest` walks newest-first and keeps the first non-empty line, and every list surface then shows where the *previous* turn stood after newer work has already finished. That is the case this feature exists to answer, and a slow utility call plus a short following turn is enough to hit it. Queue or supersede the later turn instead of dropping it.

naingthet:
P2 was right, and the comment claiming the opposite made it worse. Fixed in 2240569.

The claim treated a failed acquire as `NotApplicable`, so the *later* turn was the one dropped — not the in-flight one, as the comment asserted. Its `narrative` stayed empty, `build_digest` walks newest-first for the first turn that has a line, and the digest then kept showing the previous turn's state after newer work had finished.

It now holds at most one queued turn per session and drains it when the running call finishes, mirroring `ChatTitler`'s coalescing including the atomic take-or-release so nothing is lost in the gap. One difference from titling: the queue always drains rather than only on retry, because a later turn is a different turn and needs its own line however well the earlier one went. Only the newest queued turn is kept.

The claim now takes the in-flight map instead of the recapper, so the invariant is pinned by unit tests that need no database, provider, or policy source.

Still worth knowing on this PR: no Rust has been compiled locally, and `src/generated/wire.ts` was hand-edited to match what the generator would emit rather than regenerated. CI is the gate on both.

naingthet:
**G2G** at 2240569b0972d0b7f54322fa7b6332c727f4eef0

The prior P2 is actually fixed. `RecapClaim::acquire` now writes the later turn into the session slot instead of treating a failed claim as `NotApplicable`, and `spawn` always drains that slot when the running call finishes. Only the newest queued turn is kept, which is the one `build_digest` would need. The unit tests pin the claim map without a recapper.

No new P1/P2 on this head.

naingthet:
CI caught a second, unrelated bug on the G2G head. Fixed in 6a5fddc — new head needs a fresh look.

`test` failed on `code_recap::a_completed_turn_is_recapped_onto_the_digest`: the recap was derived, stored, and carried on the digest, and then the row read back `narrative: None`.

`checkpoint::after_turn_ended` holds a `CodeTurn` it read before the turn was terminal, sets the checkpoint ref and diffstat on that snapshot, and calls `save_turn` — which wrote every mutable column including `narrative`. The recap lands from a separate task seconds after the turn ends, so the two overlap and whichever finished last won. A slow checkpoint erased the recap; a slow model call would have dropped the checkpoint ref instead. The test caught it only because it reads the row after asserting on the digest.

`save_turn` no longer writes `narrative`. The column has one writer, `set_turn_narrative`, which touches nothing else, so neither side can carry a stale copy of the other's work in either order. A core store test reproduces the overlap directly: store a narrative, save a snapshot taken before it existed, assert both columns survive.

Worth noting what this run did settle: the crate compiles and 1298 tests passed, so the only failure was this one. The `wire.ts` question is still open until the generated-diff lane reports.

naingthet:
**G2G** at 6a5fddcbb12ff0dd56bc7baa0f7cfee5241ea008

The mid-recap queue is still in place: a failed session claim writes the later turn onto the slot and `take_pending_or_release` drains it after the in-flight call, with the same atomic take-or-release as titling so a concurrent completion cannot fall through a gap. Unit tests still pin queue, newest-only coalesce, per-session isolation, and drop-frees-the-session.

The new head closes a real second race. `checkpoint::after_turn_ended` still holds a `CodeTurn` read before the turn is terminal and `save_turn`s it after git work. That used to write `narrative` from the stale snapshot and could blank a recap that had already landed (or, the other way, drop the checkpoint ref). `narrative` is off `save_turn` now; `set_turn_narrative` is the only writer and touches that column alone. The core store test stores a line, then saves a snapshot taken with `narrative: None`, and asserts both the recap and the checkpoint ref survive.

No new P1/P2 on this head.

PR #2536, opened 2026-08-24
feat(code): park idle engine children and spawn them on the first turn

naingthet:
## Summary

The product goal is dozens of sessions on one machine, and what that costs is resident engine children, not the adapter boundary. Today Codex and opencode spawn their server child at attach, boot re-attaches every non-ended session concurrently, and no child is ever released while its session stays open — so engine runtimes track sessions that *exist* rather than sessions doing work. Decision 0057 named this as its revisit condition.

This adds decision 0064 and implements it:

- **Every adapter spawns lazily on the first turn.** `launch` constructs and validates only; the first `run_turn` spawns and completes the engine handshake (`ensure_child` on Codex and opencode — Claude already worked this way, Grok has no between-turn child). Boot re-attach now launches no engine processes.
- **`HarnessSession::park`** stops the engine's processes while the session stays attachable; the next turn transparently respawns and resumes from the session's ref. Claude parks via the retire path dead-child replacement already uses; Codex and opencode terminate the server child and re-handshake on the next ensure; the default is a no-op.
- **The worker parks after 15 idle minutes** — an arm in the between-turns select that exists only while the engine reports a live child. Parking clears the row's `child_pid` so nothing reads a dead pid as live. The threshold sits well past the documented provider prompt-cache TTL, so a wake in practice pays spawn latency (~1 s) and not tokens a warm child would have saved.
- **Codex hardening the refactor forced:** an interrupt aimed at an idle session no longer takes the app-server child (previously every later turn failed "engine child is not running"), a failed handshake leaves nothing half-attached, and a thread that never ran a turn is restarted rather than resumed onto a ref the engine never wrote.

Parking is invisible on the wire: no new events, no lifecycle state; the wake turn may repeat `session_started`, which consumers already fold (0057).

## Validation

- Claude: parked child gone, wake turn respawns with `--resume` — asserted on the fake engine's argv log beside the dead-child test it mirrors.
- Codex: nothing spawns before the first turn; a stale ref surfaces `ResumeLost` from the first turn; a parked thread that ran resumes on a new pid with `thread/resume` on the wire; a parked thread that never ran is restarted (the fake's own "thread not found" answer catches a wrong ensure); an idle interrupt is a no-op.
- Worker: with a session-long scripted child, an idle session parks (150 ms under `cfg(test)`), the row's pid clears, and the next turn still runs. The scripted default keeps per-turn pids, so every other worker test exercises the disarmed timer.
- `cargo fmt` clean locally; compile and test runs are CI's per repo practice.

### DIFF
--- a/crates/tidebreak-harness/src/claude/session.rs
+++ b/crates/tidebreak-harness/src/claude/session.rs
@@ -775,6 +775,14 @@ impl HarnessSession for ClaudeSession {
         self.unrecognized.load(Ordering::SeqCst)
     }
 
+    /// Release the idle child (decision 0064). The next turn takes the
+    /// respawn path [`Self::ensure_channel`] already owns, so it resumes the
+    /// engine session exactly like a dead-child replacement.
+    async fn park(&self) -> Result<(), HarnessError> {
+        self.retire_channel().await;
+        Ok(())
+    }
+
     async fn shutdown(self: Box<Self>) -> Result<(), HarnessError> {
         let taken = self.channel.lock().await.take();
         if let Some(channel) = taken {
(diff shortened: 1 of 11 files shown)


naingthet:
**G2G** at 6fa94bedc26d601b63f6a329022c8432c9bcf078

Lazy spawn, park, and wake-resume look aligned with 0064. The worker only arms the idle timer while a child is live, so park cannot race a turn on the same select. Codex handshake resumes only a thread that ran, and a failed handshake tears the child down. No P1/P2 that CI would miss.

naingthet:
**G2G** at 01b6e9f9a316fd88273c99fd502f860f20bb255a

The last commit only constructs the Codex steering fixture lazily, matching first-turn spawn. Park still runs only in the between-turns select while a child is live, Codex handshake still resumes a thread that ran and tears down a failed spawn, and a parked idle interrupt stays a no-op. No P1/P2 that CI would miss.

PR #2537, opened 2026-08-24
feat(code): pick delivery authors from a list instead of typing logins

naingthet:
## What changed

Rebased onto `main` after #2538 landed the delivery sheet, comment ordering, admin merge, agent quick actions, and the tab-icon fix — that work is now dropped from this branch, and what remains is what #2538 did not cover. The author and actor filters were free-text boxes that required knowing a login and spelling it exactly; they are now a checkable list of the logins Delivery has already seen, each with its avatar, narrowed by a search box, with hand-typing kept as the fallback for a login the list has never seen. The pool lives in the delivery store — filled from every pull-request and run page and from the notification poll, deduped case-insensitively, most recent first, capped at fifty, and persisted with the other delivery settings. This also fixes diff lines clipping in the pull request sheet, which `main` still has: a block child of a scrolling `<pre>` sizes to the visible width, so a long line ran past its own red or green background.

## Validation

`pnpm test` (1740 pass), `tsc --noEmit`, `biome lint`/`format`, `pnpm build`, and `storybook:build` all pass on the rebased branch by exit code. New coverage: merge/dedupe/cap semantics and poll harvesting with persistence in the store test, and a pick-then-type flow through the real filter popover in the page test. A `PullRequestAuthorFilter` story renders the picker, and I screenshotted it in dark mode before the rebase.

## Compatibility and risk

The delivery store's persisted blob gains `knownAuthors`; a blob written before it hydrates to an empty pool, and the field is bounded at fifty entries so it cannot grow without limit. Avatars come from data GitHub already sends on the summaries, or fall back to initials — no new request, no new server surface, no wire-format change.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeDeliveryPage.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeDeliveryPage.dom.test.tsx
@@ -1,6 +1,12 @@
 // @vitest-environment jsdom
 import { StrictMode } from "react";
-import { cleanup, render, screen, within } from "@testing-library/react";
+import {
+  cleanup,
+  render,
+  screen,
+  waitFor,
+  within,
+} from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
 import { toast } from "sonner";
 import { afterEach, describe, expect, it, vi } from "vitest";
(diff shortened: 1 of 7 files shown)


naingthet:
**NO GO** at 53c553e3e38272205692d044b29ee4f788d4f2ed

1. The detail surface is now a modal Radix `Sheet` (dialog). Agent → “In a new workspace” opens `NewWorkspaceDialog` (also a modal dialog) without closing the sheet. Admin merge opens `useConfirm` the same way. Two stacked Radix modals share one dismiss layer and `pointer-events: none` on `body`; the inner dialog often cannot be clicked, or dismissing it / opening it closes or freezes the sheet. CI does not exercise this. Close the sheet before the other dialog, or set `modal={false}` on the sheet so it is not a competing modal.

naingthet:
**NO GO** at 49c73842377c86d6c23780404fc005bdbfc99594

1. The stacked Radix-modal issue from 53c553e3 is still here. Agent → “In a new workspace” still calls `startNewWorkspace` without `onClose`, so `NewWorkspaceDialog` opens over the modal `Sheet`. Admin merge still runs `useConfirm` inside `PrActions` while the sheet is open. Close the sheet first, or make the sheet non-modal, before those dialogs open.

naingthet:
**G2G** at b6dd8ba877ac62756f439eb82aa200ce6270d41d

The 49c73842 stacked-modal finding is closed on this head. `PrAgentMenu` seeds `startNewWorkspace` then calls `onClose()` so the sheet unmounts in the same turn as the new-workspace dialog opens. Admin merge confirms inline in the actions card (`confirmingAdminMerge`, `role="alertdialog"`) instead of `useConfirm` / AlertDialog on top of the sheet — same class of fix as #2538.

No other P1/P2 from reading the sheet, agent menu, and merge paths.

naingthet:
**G2G** at 4dadc6df6fdd546a4eddc2a5e96c5b32532be239

Rebase after #2538 did not bring back stacked Radix modals. PullRequestDetail still keeps admin-merge confirmation inline in the actions card because the sheet is already a modal. NewWorkspaceDialog is not opened from the sheet. Author-filter work is store + filter popover only; no new Dialog/Sheet pairing.

PR #2541, opened 2026-08-24
feat(code): read person attribution from the gateway git probe

naingthet:
## Why

Decision 65 (`docs/decisions/0065-hosted-git-acts-as-the-person.md`, included here): work from a hosted machine should land as the person who drove it, not as the deployment's GitHub App. This is slice B of that record — the machine-side probe states and the attribution UI. The gateway's personal-mode surface (probe + mint) lands separately in model-gateway.

## What

- The git-forge probe and credential mint declare `attribution=person`. A gateway that predates the field ignores it and refuses connect-mode forges exactly as before, so nothing changes against today's deployments in either direction.
- A probe answer stating `attribution: "person"` parses to the caller's login, display name, and commit email. Answers without the field remain the App's bot (decision 63), byte-for-byte; an unknown attribution value refuses rather than guessing an identity.
- New typed `not_connected` refusal, cached like the other settled answers, carrying the gateway's `connect_url`. The add-repository GitHub source then reads "To use GitHub as yourself here, connect your GitHub account at the Model Gateway: …" instead of silently acting as anything else.
- The `connect_mode_forge` remediation stops advising a second forge app registration (decision 65 rejects that path) and points at updating the gateway instead.
- The PR card names the person plainly — `pushes_as` becomes their login and the new optional `pushes_as_self` wire field switches the sentence to "— your own GitHub account."

## Tests

- `obo_gateway`: person attribution parses only when stated; unknown attribution refuses; `not_connected` is typed, held fresh, and carries the connect page; the fake gateway asserts the opt-in rides both requests server-side.
- `code_clone`: per-caller source answers for a connected person and for a not-connected caller (connect URL surfaces in the remediation).
- `code_git`: the git card names the person with `pushes_as_self`, and never claims the App's identity as the caller's own.
- DOM tests for the PR card person sentence and both palette states; Storybook story `HostedPushesAsYou`; storybook build, biome, tsc all green locally.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/AddRepoPalette.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/AddRepoPalette.dom.test.tsx
@@ -317,3 +317,63 @@ describe("AddRepoPalette and a machine without a GitHub credential", () => {
     );
   });
 });
+
+describe("AddRepoPalette on a hosted machine that acts as the person", () => {
+  it("carries the person attribution sentence on the offered source", async () => {
+    await renderPalette(
+      app({
+        getCodeRepoSources: vi.fn(async () => ({
+          sources: [
+            { kind: "local", available: true },
+            { kind: "git_url", available: true },
+            {
+              kind: "github",
+              available: true,
+              remediation:
+                "Clones and pushes use your own GitHub account: work lands as mira-chen.",
+            },
+          ],
+          chooses_destination: true,
+        })),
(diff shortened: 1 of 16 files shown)


naingthet:
**G2G** at 13ece07798446293dd14319c8b5832e6ef18b033

Person attribution is only trusted when the probe states it; missing field stays the App bot; unknown values refuse. not_connected is typed, cached with the other settled refusals, and the UI names the person via pushes_as_self instead of the App warning. Backward compatible against a gateway that ignores the field.

PR #2539, opened 2026-08-24
fix(code): keep tool rows folded and honor model selection

naingthet:
## Summary

- Keep tool rows and activity groups folded until the reader opens them, and render folded multi-line commands on one visible line.
- Enable model changes between turns for Codex and OpenCode. OpenCode now sends the provider-qualified model in each prompt request.
- Remember a separate model for each harness and migrate the earlier single-model preference.

## Validation

- `pnpm --dir crates/tidebreak-desktop/ui exec vitest run --reporter=dot` — 1,659 tests passed
- `pnpm --dir crates/tidebreak-desktop/ui lint`
- `pnpm --dir crates/tidebreak-desktop/ui build`
- `pnpm --dir crates/tidebreak-desktop/ui storybook:build`
- `cargo fmt --all -- --check`
- `git diff --check`

Rust compilation and tests remain with CI, as required by the repository workflow.

PR #2542, opened 2026-08-24
docs(decisions): pull-request state is one definitive store

naingthet:
Decision 65: collapse the four independent representations of a pull request
(the 20 s workspace digest cache, the persisted `workspace.pr` column, the
`pr_facts` rows, and the delivery aggregate) into one definitive store — one
row per pull request, one fetcher, one attention-tiered schedule, broadcast
on change. Every surface becomes a projection of the row, and no surface
calls GitHub on its own.

Motivation, observed on PR #2539: the workspace header reads "Blocked" while
the Review pane shows Open with nine pending checks, because two classifiers
rank the same digest differently and GitHub reports `BLOCKED` whenever any
protection requirement is unmet. Staleness has the same structural cause:
four stores, four clocks, no cross-invalidation, and a push does not
invalidate the digest cache at all.

The record also fixes the cost profile: today one digest read is four to six
GraphQL-backed `gh` subprocesses in three sequential hops, three sweeps fetch
independently, and nothing sends a conditional request. The decided shape is
conditional REST (`pulls/{n}` + `check-runs` by head SHA, list-diff per
repository) behind one budget gate with secondary-limit backoff, so counted
traffic tracks how often pull requests change rather than how often Tidebreak
polls — a user's rate limit should never notice the app.

Implementation stages as five slices (classifier + invalidation holes first,
no schema); this PR is the record only.

### DIFF
--- a/docs/decisions/0065-pull-request-state-is-one-store.md
+++ b/docs/decisions/0065-pull-request-state-is-one-store.md
@@ -1,0 +1,188 @@
+# 65. Pull-Request State Is One Store with One Fetcher and One Clock
+
+- Status: Accepted
+- Date: 2026-08-24
+- Owners: code mode, delivery
+- Related: [`0042-user-initiated-pr-merge.md`](0042-user-initiated-pr-merge.md),
+  [`0060-triggers-are-durable-rules-on-pull-request-facts.md`](0060-triggers-are-durable-rules-on-pull-request-facts.md),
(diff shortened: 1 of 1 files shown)


naingthet:
**G2G** at 5bef9ddbe032283b66f2f814296a47c009d90811

Accepted ADR: one fact row, one REST fetcher, one scheduler, one classifier. Consequences and slices match the decision. No high-impact P1/P2 that CI would miss.

PR #2543, opened 2026-08-24
feat(code): fork a conversation from any turn with a two-layer handoff

naingthet:
## What changed

Forking previously covered the whole session, wrote one flat markdown file, and left the child a single bare framing line. Three gaps against how forks actually get used: you cannot fork from the middle of a long conversation, the child has no way to see tool output the condensed transcript drops, and the framing says nothing about how to use what it was handed.

`POST /code/sessions/{id}/fork` now takes an optional `at_turn`, and each fork writes one immutable generation directory:

```
forks/<session>/<generation>/
  transcript.md   # condensed rendering, sliced at the fork point
  turn-0007.md    # full record of turn 7: tool output previews, subagent
                  # activity nested under its Task call, reasoning, steers
  turn-7-1-….png  # retained attachments, as before
```

- Turns squeezed out of the transcript's 512 KiB cap become one-line stubs naming their record file instead of vanishing; the header explains the layout to whichever engine reads it.
- Records are bounded per file (256 KiB) and per fork (8 MiB), dropped oldest-first past the budget, and every drop is stated where the content would have been.
- Attachments over the 20 MiB cap now drop oldest-first with a note in both layers rather than failing the whole fork.
- The composer seeds framing lines built from the fork response: read the transcript first, open records only as needed, and — when the fork point is not the newest turn — a warning that the shared worktree may already be ahead of the transcript.

In the UI, every turn seam (`TurnReviewCard`) gains a quiet actions menu with "Fork from here", on interactive sessions only. The tab menu's whole-session fork is unchanged and now rides the same directory layout.

## Validation

- `cargo fmt --all --check`
- Desktop UI: `tsc --noEmit`, `biome check` on changed files, focused vitest (fork, parsers, transcript, workspace page — 166 tests), and `pnpm storybook:build` (new stories: seam actions menu, failed-turn actions, fork-at-earlier-turn framing).
- **No Rust was compiled or tested locally.** The rewritten `code::fork` tests (17, most new) have never run, and `src/generated/wire.ts` was hand-edited to match what `UPDATE_WIRE_TYPES=1 cargo test -p tidebreak-server` would emit, so the generated-diff check is the thing most likely to fail first. CI is the gate on all of it.

## Compatibility and risk

The route accepts an empty body (old behavior) alongside the new JSON body, so a client that predates fork points still forks at the newest turn, and an old server ignores the body and forks the whole session — degraded, not broken, across remote-attach version skew. `CodeForkTranscript` gains `dir`, `total_turns`, and optional `at_turn_ordinal`. The stable `forks/<session-id>.md` file is no longer written — each fork lands in its own generation directory, which the existing attachment-generation lifecycle already retains for the worktree and cleans at archive. Files written per fork grow from one to one-per-turn, within the caps above.

## Tracking

None.


naingthet:
**G2G** at 6e420ea703f1b68ab2b24200307a051782e64530

Fork point is sliced on a real turn id, later turns stay out of the handoff, and the worktree warning only appears when something was actually excluded. Empty-body and JSON-body clients both still fork at the newest turn. The two-layer write stays inside the existing generation scratch lifecycle, and the UI only offers a seam fork on interactive sessions.

No P1/P2 that CI would miss.

PR #2544, opened 2026-08-24
feat(code): hosted workspaces commit as the connected person

naingthet:
## Why

Decision 65 rule 3 (slice C of `docs/decisions/0065-hosted-git-acts-as-the-person.md`, decision doc landed with #2541): commits from a hosted workspace should carry the person's identity. Today they carry nothing — the hosted image has no git identity configured at all, so the user-facing commit path fails outright once a workspace exists on a hosted machine.

## What

- `create_workspace` now names the caller on the worktree right after the worktree exists and before the setup script runs (a setup script may itself commit): display name (login fallback) + the gateway-derived no-reply commit email, taken from the same per-caller probe answer the attribution UI reads (#2541).
- The identity is written with `extensions.worktreeConfig` + `git config --worktree`, so it belongs to the one workspace: sibling workspaces of the same clone can belong to other callers, and the shared repository configuration never names anyone.
- Best-effort: a caller who has not connected, a bot-attributed forge, and a machine with its own credentials all leave the checkout untouched.

## Tests

- `a_hosted_workspace_commits_as_the_person`: author and committer both read the person, and the registered clone's own `user.name` is untouched.
- `a_workspace_without_a_person_identity_keeps_the_checkouts_own`: local machines and bot-attributed lenders keep the checkout's identity byte-for-byte.

### DIFF
--- a/crates/tidebreak-server/src/code/gh.rs
+++ b/crates/tidebreak-server/src/code/gh.rs
@@ -190,6 +190,39 @@ pub(crate) struct PushOutcome {
     pub remote: String,
 }
 
+/// Names `name <email>` as this workspace's git author and committer
+/// (decision 65).
+///
+/// `extensions.worktreeConfig` is enabled first so the identity lands in the
+/// worktree's own configuration: sibling workspaces of the same clone can
+/// belong to callers with other identities, and the shared repository
+/// configuration must never name any one of them.
+pub(crate) async fn configure_workspace_identity(
+    worktree: &Path,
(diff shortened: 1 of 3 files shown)


naingthet:
**G2G** at a237e8637f22c25fc51075b0322231a93a7ddb71

Hosted workspaces get a worktree-scoped author before setup, only when the gateway names a person with a commit email. Shared clone config stays unnamed. Best-effort skips match the stated contract.

PR #2545, opened 2026-08-24
feat(code): queue session follow-ups durably with the chat tray

naingthet:
## What

Code sessions parked one follow-up in an in-memory slot: a second mid-turn send refused with `queue_full`, a restart silently dropped the parked message, and nothing could list, edit, or retract it. The PR quick actions (fix checks, resolve conflicts, address feedback, update branch) bounce off that slot the moment more than one chore applies.

This mirrors the chat queue contract (decision 9) onto code sessions, per new decision 65:

- A durable `code_queued_turn` row per queued message — owner-scoped, FIFO by dense `position`, capped at 32 per session, carrying the message and image attachments. The row id is the turn id the promoted turn is inserted under.
- The session worker drains the queue. Promotion deletes the exact row snapshot (id, position, `updated_at`) and inserts the running turn in one transaction: an edit, reorder, or retraction landing after the snapshot makes promotion stale and nothing is written; a crash can neither lose nor double-run a message. Boot recovery re-attaches a worker per live session and the drain runs first, so queues survive restarts.
- A send parks when the session is mid-turn, when a sibling holds the workspace checkout (record 55), or when the queue has backlog — FIFO means a new send may not overtake parked rows.
- Chat's queue routes, one segment over: `GET /code/sessions/{id}/queued`, `PATCH`/`DELETE …/queued/{queued_id}`, `PUT …/queue-paused`, `POST …/queued/send-now`. The queued submit receipt now returns the row instead of a hard-coded `position: 1`.
- One `QueueTray` component serves both modes behind a five-verb adapter (`list`, `update`, `remove`, `setPaused`, `sendNow`). Code mounts it above the composer, replacing the "1 follow-up queued" pill, so queued PR chores are visible, editable, reorderable, and retractable — send-now composes pause → move first → stop → release exactly as chat's does.
- A stop pressed while a queued turn waits on a sibling's worktree now leaves the row queued instead of dropping the message.

Settings still resolve at promotion (decision 9's reading): a queued turn runs under the session's model and effort as they are when it starts.

## Schema

One appended migration, `m20260824_000011_code_queued_turns` (decision 61). No baseline edit. `wire.ts` regenerated for the widened `QueuedCodeTurn`.

## Validation

- Store tests: FIFO positions, promotion under the row id, stale promotion after edit/retraction writes nothing, dense reorder, the 32 cap, owner isolation, pause round-trip, ended-session cleanup.
- Server tests: two mid-turn sends park; list/edit/retract work against live rows; the survivor runs after the live turn under its row's id; the retracted row never runs; a stop pressed while a queued turn waits on a sibling leaves the message queued, not run and not lost.
- UI: QueueTray drives both chat and code adapters (dom tests), composer queue-full copy updated, typecheck, biome, and Storybook build pass; new `QueueTray` stories cover queued, paused, single-row, and empty.


naingthet:
**G2G** at c839f4be373210fa237fc0c83f439cadb4972124

Promotion deletes the exact FIFO head and inserts the turn in one transaction, so an edit, reorder, or retraction after the snapshot writes nothing. Park-on-busy-or-backlog, interrupt-while-waiting leaving the row queued, and the shared tray adapter match decision 65. Nothing here is a P1/P2 CI would miss.

naingthet:
**G2G** at 017281affdd2e56ae706744a35414ffdec8bf316

Promotion deletes the exact FIFO head snapshot and inserts the turn in one transaction, so an edit, reorder, or retraction after the snapshot writes nothing. Park-on-busy-or-backlog, interrupt-while-waiting leaving the row queued, and the shared tray adapter match decision 65. Nothing here is a P1/P2 CI would miss.

naingthet:
**G2G** at 6b815a8c8e6a8453567f31aaa2328c007fe58a2c

Promotion still deletes the exact FIFO head and inserts the turn in one transaction, so an edit, reorder, or retraction after the snapshot writes nothing. The new `queued_turn_stale` telemetry arm only classifies that already-handled worker error. Park-on-busy-or-backlog, interrupt-while-waiting leaving the row queued, and the shared tray adapter match decision 65. Nothing here is a P1/P2 CI would miss.

naingthet:
**NO GO** at 6977cf3f12630da6ee3bbf015cd49047c0530112

1. Stopping a queued turn that is waiting on a sibling worktree lock leaves the durable rows parked with promotion still unpaused, and the worker never looks at them again until some later enqueue, unpause, or send-now.

`await_worktree_turn` treats `Interrupt` as “do not start” and returns `None`. `drive_turn_inner` maps that to `WorkerError::Conflict("the queued turn was stopped")`. `drain_queued` then hits `Err(_) => return` and `run_worker` sits on the outer `select`. That select is the only place that listens for `queue.wake`. Sibling turn completion only releases the mutex; it does not notify this worker. `queue.wake` is also not selected inside `await_worktree_turn`, so a later send-now that does not also interrupt is lost if the worker is still in that wait.

The new stop-while-waiting test only checks the row is still queued and not run. It does not check that the session ever drains again after the sibling finishes. Composer stop and send-now’s `onStop` both take this path. The tray still shows an unpaused queue, so the parked PR chores look live and never start.

Keep the row. After the interrupt, either stay in `drain_queued` (or notify `queue.wake` before leaving it) so the next free checkout actually promotes, or set the session pause so the tray matches the fact that nothing will run until the user says so.

naingthet:
**G2G** at 0efe07ea887c34313209ac1b9cfb1ca81cd0e9ca

The prior parked-row stall is gone: a stop while waiting on a sibling checkout now returns `QueuedTurnStopped`, `drain_queued` pauses promotion, and the tray reads Paused. Resume or send-now clears that pause and wakes the worker. Shutdown stays a separate `Conflict` so a dying worker does not pause the queue. The stop-while-waiting test now asserts the pause and drives send-now after the checkout is free. Nothing here is a P1/P2 CI would miss.

naingthet:
**G2G** at 2087121342e2120b358bef44311bfa579ea09fcc

Durable FIFO rows, snapshot promotion, sibling-wait stop, and the shared tray match decision 65. Nothing here would silently drop or double-run a parked follow-up.

PR #2547, opened 2026-08-24
feat(code): open Delivery on your own pull requests

naingthet:
## What changed

The pull-request page opened on "Needs attention" — everybody's review queue. It now opens on "Yours": your open pull requests, drafts included, with the author taken from the login `gh` is signed in as, and every row names its author with an avatar and login leading the metadata line. A lone author now reaches the remote read as `gh pr list --author`, because otherwise the new default would spend the 100-row per-repository cap on other people's work and silently drop your older open pull requests in a busy repository. When `gh` cannot report a login the "Yours" view is not offered and the attention view takes the default back, so it never quietly widens to everybody. Along the way the view chips gained `aria-pressed`, and the Delivery stories gained the code-updates socket the rail subscribes to on mount — without it every story in that file rendered its error boundary instead of the list.

## Validation

`pnpm exec biome format src`, `pnpm lint`, `pnpm exec tsc --noEmit`, and all 1751 vitest tests pass; `pnpm storybook:build` succeeds and `cargo fmt --all --check` is clean. I reviewed the two new stories rendered at 1440px light and 1100px dark to confirm the chip row, the author strip, and the unchanged row height. The Rust was not compiled locally — CI covers it.

## Compatibility and risk

No wire or persisted-format change. The author push-down gives that read its own aggregate cache scope, so a "Yours" query no longer shares the unscoped entry — the first load after this lands refetches rather than reusing it.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeDeliveryPage.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeDeliveryPage.dom.test.tsx
@@ -183,6 +183,81 @@ describe("delivery pull request list", () => {
     );
   });
 
+  it("opens on your own open pull requests, drafts included", async () => {
+    const query = vi.fn(async (_body: unknown) => ({
+      capability: deliveryRepositoriesSnapshot.capability,
+      items: deliveryPullRequests,
+      errors: [],
+      fetched_at: "2026-08-20T15:20:00.000Z",
+    }));
+    renderList({
+      ...storyClient(),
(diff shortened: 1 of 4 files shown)


naingthet:
**G2G** at ed0c6365e2fb74bf76685b521d5273ec387ee0ef

Default view is Yours with drafts still open, lone author reaches `gh pr list --author` and its own cache scope, and a missing viewer login drops the chip instead of widening to everybody. Author strip and stories look sound. No P1/P2.

PR #2548, opened 2026-08-24
feat(code): hosted machines create and read pull requests over REST

naingthet:
## Why

Decision 65 rule 4 — the slice that makes person attribution real end to end. A hosted machine has no `gh`, so pull-request creation and the delivery reads were simply unavailable there (decision 63 rule 5 left them as "the next slice"). With the gateway lending the caller's own user token (model-gateway#1326) and clone/push/commits already person-attributed (#2534, #2541, #2544), this makes the pull request itself the person's.

## What

- **`code/forge_rest`** — a deliberately narrow REST surface driven by a per-operation borrowed credential: create, list-for-head, and the PR-card digest (state, checks keyed to the exact head SHA the pull request names, merge-queue state from the same timeline events `gh` reads). Every answer is shaped like the `gh --json` payloads the fact store and digest parsers already consume — one JSON vocabulary, however the host was asked. No verify-head dance: REST checks are read for the named SHA, so they cannot disagree with it.
- **Create** (`create_pull_request_rest`): identical title/body generation to the `gh` path, one borrow, and the authored fact (decision 62) persists straight from the creation answer — no second host read.
- **Status**: the workspace PR read refreshes the digest over REST behind the same 20-second cache as `gh`, borrowing only on a miss and degrading silently on failure, exactly as `gh` does. `review_decision` stays unstated on REST rather than approximated.
- **Push**: the contributed-fact read reuses the push's own borrowed credential — one borrow, one operation.
- **Refusals**: a gateway refusal fails the operation as `git_forge_refused` with the gateway's reason (e.g. the connect-your-account remediation), never an uncredentialed retry.

`gh` paths are byte-for-byte untouched where `gh` exists. Merge, ready, close, comments, CI logs, the delivery panel, and the background reconcile sweep stay `gh`-only for now and keep answering exactly as they do today on machines without it; they are named follow-ups, not accidents.

## Tests

- `a_hosted_machine_creates_and_reads_the_pull_request_over_rest`: a loopback fake forge asserts the borrowed bearer server-side, captures the create body (title/base/head), and serves list/detail/check-runs/timeline; the card digest reads "1 passing, 1 pending, 0 failing" and the authored fact lands with the person as author — with no `gh` anywhere.
- `a_hosted_pull_request_create_fails_closed_with_the_gateway_refusal`: a not-connected caller's create fails with the gateway's connect remediation.
- The existing hosted push/identity tests now pin the digest-read borrows (every borrow names the one repository; a foreign-host origin still borrows nothing) against a dead loopback REST base.
- `forge_rest` unit tests pin the fact-shape restatement (merged-as-closed+mergedAt) and the API-base rule.

### DIFF
--- a/crates/tidebreak-server/src/code/forge_rest.rs
+++ b/crates/tidebreak-server/src/code/forge_rest.rs
@@ -1,0 +1,429 @@
+//! Pull-request reads and creation over the forge REST API (decision 65).
+//!
+//! A gateway-authenticated hosted machine has no `gh`. Where decision 63 gave
+//! its git transport a borrowed credential, this module gives its
+//! pull-request surfaces the same: each function drives one operation against
+//! the forge's REST API with a per-operation borrowed credential, and shapes
+//! every answer exactly like the `gh --json` payload code mode already
+//! parses — one JSON vocabulary, however the host was asked.
+//!
+//! Deliberately narrow: creation, the PR-card digest, and the fact reads
+//! (decision 62). Merge, ready, close, comments, CI logs, and the
+//! repository-wide delivery panel stay `gh`-only and keep answering exactly
+//! as they do today on machines without it.
+
+use std::time::Duration;
+
+use futures::StreamExt as _;
+use serde_json::Value;
+
+use crate::obo_gateway::GitCredential;
+use crate::routes::code::CodeGitHubRepositoryTarget;
+use tidebreak_core::{PullRequestCheck, PullRequestCheckBucket, PullRequestDigest};
+
+/// Timeout for one REST call — the same bound the `gh` runner gets.
+const REST_TIMEOUT: Duration = Duration::from_secs(30);
+
+/// Cap on one REST response body. PR and check payloads are kilobytes; the
+/// bound exists so a confused origin cannot grow the process.
+const RESPONSE_LIMIT: usize = 2 * 1024 * 1024;
+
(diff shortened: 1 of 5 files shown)


naingthet:
**G2G** at e76f36f50dcdd3f395a26dd0828d998b11dba2ef

Hosted create and digest reads stay on one borrowed credential, fail closed on a gateway refusal, and leave the gh path alone when the machine is not lending. The REST answers match the fact-store shape, checks bind to the PR head SHA, and review_decision stays unstated instead of guessed. I did not find a high-impact defect CI would miss.

naingthet:
**G2G** at b63ac31db449a1ca30786e0266c73bebf208bd1e

Hosted create and digest reads stay on one borrowed credential, fail closed on a gateway refusal, and leave the gh path alone when the machine is not lending. The REST answers match the fact-store shape, checks bind to the PR head SHA, and review_decision stays unstated instead of guessed. I did not find a high-impact defect CI would miss.

naingthet:
**G2G** at 9981c6d9118918e3ac8d89ca99379249773c1bc1

Hosted create and digest reads stay on one borrowed credential, fail closed on a gateway refusal, and leave the gh path alone when the machine is not lending. The REST answers match the fact-store shape, checks bind to the PR head SHA, and review_decision stays unstated instead of guessed. I did not find a high-impact defect CI would miss.

PR #2550, opened 2026-08-24
feat(desktop): reach everything from one Cmd+K palette

naingthet:
## What

Cmd+K opens a command palette across both modes. Cmd+L takes over focusing the composer.

Cmd+K used to focus the composer — something the composer's own click target already does. The palette is the higher-value use of the chord, and Cmd+L is where the tools this app sits beside put "get me to the agent". It takes the chord from Monaco's expand-line-selection, which is reachable with Home and Shift+Down; getting back to the composer from anywhere is not.

## How

`CommandPalette.ts` is the pure core — prefix parsing, per-section fuzzy ranking, caps, and a recents memory. `CommandPaletteList.tsx` is presentational and is what Storybook renders. `CommandPaletteDialog.tsx` wires it to the stores.

Rows come from sources each half of the app owns (`code/codePaletteRows.ts`, `chatPaletteRows.ts`, `settingsPaletteRows.ts`), so adding a command changes a source rather than the palette.

Code mode leads with the workspace's next ship step. The workflow control already resolves it to label its primary button, so it republishes it through the store rather than the palette fetching the snapshot again and risking a different answer. Then the rail's workspaces with their status accents, that workspace's commands via the same `useWorkspaceCardCommands()` runner the context menu uses, the Ship chords, navigation, and every settings section as its own row.

`>` scopes to commands, `@` to workspaces or chats, `#` to files. Only `#` reads the worktree — the tree is thousands of paths and the palette has to open instantly.

Rows that a chord already reaches draw their keycaps from `shortcutKeycaps()`, so the palette cannot advertise a key that does something else.

### Notable

- `ShellShortcutDef` gains `allowInModal`. Only the palette's own chord sets it, so Cmd+K can close the palette; every other shortcut stays suppressed behind a dialog. The handler still declines when the open dialog belongs to someone else, which the guard alone cannot tell.
- `fuzzyTokenScore` moves out of `CodeQuickOpen` into `fuzzy.ts`, shared by both surfaces.
- `CommandInput` gains an optional `leading` slot so a scope chip sits inside the field. Additive — the plugins panel, quick open, and add-repo palette are unchanged.
- `fork-agent` is left out of the palette: it opens into a tab only the workspace page can place.

## Checks

`tsc`, biome lint and format, `storybook:build`, and all 1765 tests pass locally.

Nine stories under `Foundations/Command palette`: code mode at rest, filtered, each scope, settings, files loading, no matches, chat mode, and Ctrl keycaps. Each passes its fixtures through the real `rankPaletteRows`, so they show shipped ordering rather than a hand-arranged list.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/AppShell.tsx
+++ b/crates/tidebreak-desktop/ui/src/AppShell.tsx
@@ -77,11 +77,13 @@ import { WindowDragStrip } from "./WindowDragStrip";
 import { useActiveChatId } from "./useActiveChatId";
 import { useChatPromptWatcher } from "./useChatPromptWatcher";
 import {
+  hasOpenModalDialog,
   numberedTabIndex,
   useShellShortcuts,
   type ShellShortcutHandlers,
   type ShellShortcutMode,
 } from "./ShellShortcuts";
+import { CommandPaletteDialog } from "./CommandPaletteDialog";
 import { ShortcutsDialog } from "./ShortcutsDialog";
 import { useUiStore } from "./UiStore";
 import { UPDATE_CHECK_REQUESTED_EVENT, useDesktopUpdates } from "./updates";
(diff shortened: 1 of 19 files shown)


naingthet:
**G2G** at 9d5bca16f328a54ba6f96789afc0117c13204062

Palette core, row sources, and Cmd+K modal close/decline look sound. Recents stay out of transient rows. File tree stays off the open path until `#`. Nothing here would change what I would ship.

Issue #2551, opened 2026-08-24
Memory: durable, reviewable memory across chat and code


naingthet:
Durable memory that spans chat and code sessions: a bounded set of scoped records (personal, per-repo) captured from completed turns on the utility model, reviewed before anything gains authority, rendered as a capped digest into both surfaces, and open to third-party stores through MCP or an in-tree backend behind one trait.

Decision records land first and carry the full contract; each slice is independently shippable.

- [x] #2552 decision records
- [x] #2553 substrate + manager UI
- [x] #2554 work mode
- [x] #2555 code mode
- [x] #2556 maintenance sweep

Landed 2026-09-02: #2553 as #3043, #2556 as #3067, #2554 as #3069, #2555 as #3072. Follow-up: #3084 (context-file diffs in code-mode capture).

PR #2557, opened 2026-08-24
docs: memory records, scopes, and a pluggable backend boundary

### LINKED ISSUE #2552: Memory decision records: record model, scopes, backend boundary
Part of #2551.

Two decision records: (1) the memory record model — markdown document + typed envelope, kinds, the tracking/proposed/active/archived/rejected lifecycle, evidence enforced at the storage layer, personal and repo scopes, caps with a coached overflow error, digest as a derived render; (2) the backend boundary — MemoryBackend trait with exhaustive capability flags, default row-backed store with lexical search, injection pinned per conversation with cache-preserving refresh boundaries, code-mode materialization outside the worktree plus a loopback MCP verb, post-turn capture on the utility model with a hypothesis tier, maintenance as a durable sweep. Plus deferred.md entries for the parked tail.

naingthet:
Two decision records specifying durable, reviewable memory across chat and code, plus the parked tail in deferred.md.

- **Record 67 — Memory records and scopes.** A memory record is a markdown document with a typed envelope: kinds, a tracking/proposed/active/archived/rejected lifecycle, evidence enforced at the storage layer (a model-authored record without resolvable evidence is not expressible), personal and repo scopes, hard caps with a coached overflow error instead of silent eviction, and a digest that is always a derived render so deleting a record deletes it everywhere.
- **Record 68 — One record vocabulary behind a backend boundary.** A MemoryBackend trait with exhaustive capability flags on the decision-31 pattern; a default backend of owner-scoped rows with in-process lexical search and one-transaction writes; injection as a per-conversation pinned digest (work mode) and read-only files in the session's private read root plus a loopback MCP verb (code mode); post-turn capture on the utility model with repetition thresholds; maintenance as a durable sweep per decisions 50/60. Third parties plug in via MCP or the trait — never as live mirrors of engine-native stores.
- **deferred.md** gains the parked tail: org-scoped shared store with explicit promotion, episodic search, a one-way file mirror and folder backend, semantic retrieval as a declared capability, a dedicated memory model role.

Slice 0 of #2551. Closes #2552.

### DIFF
--- a/docs/decisions/0067-memory-records-and-scopes.md
+++ b/docs/decisions/0067-memory-records-and-scopes.md
@@ -1,0 +1,143 @@
+# 67. Memory Records and Scopes
+
+- Status: Proposed
+- Date: 2026-08-24
+- Owners: memory
+- Related: decisions 19, 31, 48, 59, 61; [docs/code-mode.md](../code-mode.md)
+
+## Context
+
+Nothing durable and curated survives a session. Chats have transcripts and
+semantic checkpoints (decision 19); code sessions have journals and per-turn
+recaps. All of it is episodic: a compression of one conversation, owned by
+that conversation. A preference the user states in one chat is relearned in
+the next. A fact about a repository proven in one workspace is rediscovered
+in the next worktree. The knowledge exists in the product's own records; no
+structure carries it forward.
+
+Whatever carries it forward inherits standing constraints. It must be
+engine-neutral (decisions 30 and 31): external harnesses consume it too, so
+the format cannot assume the internal loop or a coding agent. It must be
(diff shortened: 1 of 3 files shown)


naingthet:
**G2G** at 01435423677c6cd24509c2f4b5990578935b4baf

Two proposed records plus a deferred parking lot. 67 locks the record shape, lifecycle, scopes, and bounded digest; 68 puts storage behind an in-tree trait and keeps capture post-turn on the utility role. That matches the existing engine-neutral, single-writer, and cache-prefix constraints. Nothing here would ship a wrong contract that CI could not see.

PR #2558, opened 2026-08-24
docs(decisions): renumber the code session queue record to 0067

naingthet:
PR #2545 (durable code session queue) and the hosted-git person record both merged as decision 0065; hosted-git landed first, and the pull-request-state record already moved to 0066 for the same collision (#2549). The queue record takes the next free number, 0067, and the code comments that cite it move with it. The hosted-git references to decision 65 are untouched. No behavior changes.

### DIFF
--- a/crates/tidebreak-core/src/db/migration.rs
+++ b/crates/tidebreak-core/src/db/migration.rs
@@ -1360,7 +1360,7 @@ fn code_trigger_fire_outbox_indexes() -> Vec<IndexCreateStatement> {
     ]
 }
 
-/// Durable per-session queued follow-ups (decision 65).
+/// Durable per-session queued follow-ups (decision 67).
 ///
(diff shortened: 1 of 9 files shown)


naingthet:
**G2G** at d2374da5b07787adf4ed1317d21a9e4f4327c19f

Decision file is 0067, heading matches, and the code comments that cited 65 now cite 67. Hosted-git stays 0065. No behavior change and nothing high-impact to flag.

naingthet:
Closing in favor of #2559. The target number here collides again: decisions 0067 and 0068 landed on main at 18:07Z via #2557 (memory records + backend boundary), so renaming the queue record to 0067 would create a second duplicate pair. #2559 renames it to 0069 — the next free number — and repoints the same code comments. The one reference this PR covers that #2559 initially missed (QueueTray.stories.tsx) is being added there.

PR #2563, opened 2026-08-24
feat(code): pull-request facts carry a live tier that writes through

naingthet:
Slice 2 of [decision 66](docs/decisions/0066-pull-request-state-is-one-store.md): the durable pull-request fact row gains a live tier, and every fresh host observation writes through it to every holder.

## What changes

- `CodePullRequestFact` carries `live: Option<CodePullRequestLiveState>` — checks summary, check runs, review decision, mergeability, merge-queue state, auto-merge flag, and `observed_at`. Migration `m20260824_000012_code_pull_request_live_tier` adds the guarded columns; the snapshot upsert never touches them.
- `set_pull_request_live_state` updates the tier and reports whether anything other than `observed_at` moved, so a repeat observation refreshes the timestamp without fanning out.
- When a workspace digest read observes change, `record_pull_request_live_state` writes the fact row's live tier and fans the digest out to every other active workspace holding the same pull-request URL: digest cache and workspace column both take the new snapshot, and `save_workspace` broadcasts it. One GitHub read now updates every surface that shows the pull request.
- `pull_request_identity_from_url` parses the canonical pull URL into host, repository, and number, so the join is by URL identity rather than a number that collides across repositories.

## Why

Decision 66 makes the fact store the one definitive row for pull-request state. Slice 1 fixed the classifier; this slice gives the row a live tier and makes observations converge on it instead of each surface polling GitHub separately. The single "delivery update" broadcast event lands with slice 5, where its consumer drops the page poll timers.

## Testing

- `cargo fmt --check` clean; store round-trip covered in `db/tests/code.rs` (first write reports change, an identical re-observation does not, the snapshot upsert preserves the tier).
- `tests/code_git.rs` adds an end-to-end write-through test: two workspaces hold the same pull request, one refresh observes a new head, and the other workspace's column, cache, and fact live tier all show it without a second host read.


naingthet:
**G2G** at bf6f657f2689bdf3ca19edc3bb91534cf078dc11

The live-tier columns stay off the snapshot upsert, so a confirmation cannot blank them. `set_pull_request_live_state` treats `observed_at` as a silent refresh. Write-through joins on URL identity, updates the other holder's column and digest cache, and skips when nothing but the timestamp moved. Missing fact rows stay best-effort as documented.

naingthet:
**G2G** at 884f74a8cae47dcaa874f84514b5cb40ba632e2f

Live columns stay off the snapshot upsert conflict list, so a confirmation cannot blank the tier. `set_pull_request_live_state` treats `observed_at` as non-change, and write-through skips the source workspace and only fans a real digest move. The identity parse is by URL, so repository numbers do not collide. Tests cover first write, no-op re-observation, snapshot preservation, and a second holder taking the digest without another host read.

PR #2564, opened 2026-08-24
feat(code): sweeps consume the pull-request store before fetching

naingthet:
Slice 3 of [decision 66](docs/decisions/0066-pull-request-state-is-one-store.md): the watch and trigger sweeps consume the pull-request store, and delivery reads feed it.

## What changes

- Delivery reads persist each summary's live tier. `persist_and_augment_pull_request_facts` now writes the live tier through the decision-66 seam after each fact upsert, so the reconcile sweep's one list read per repository per 61 seconds refreshes checks, review state, and mergeability for every stored pull request — and fans real change out to every workspace holding it.
- The watch sweep answers from the store first. When the workspace's fact row carries a live tier younger than two reconcile intervals plus slack, the sweep classifies that digest instead of paying a host read. Stale or missing rows fall back to the host read, which itself lands back on the store.
- The trigger sweep does the same: fresh rows classify directly (including stack-parent holds), and only the residual numbers take the old by-number query path.
- `digest_from_fact` moves to `pr_facts` and now projects the live tier onto the digest, so a stored row answers with the same fields a host read would.
- `record_pull_request_live_state` takes an optional source workspace: delivery-path writes have no workspace to exclude from the fanout.

## Why

Decision 66 wants one fetch schedule feeding every surface. After this slice, the three background sweeps stop fetching independently: reconcile reads, the store remembers, and watch and trigger consume — the same answer everywhere, at a fraction of the GitHub traffic. Slice 4 makes the remaining reads conditional REST behind one gate; slice 5 drops the UI poll timers.

## Testing

- `pr_facts` unit tests cover URL identity parsing and the live tier riding `digest_from_fact`.
- Existing watch, trigger, and reconcile suites cover the fallback paths; the freshness gate is exercised by the store tests from slice 2.

### DIFF
--- a/crates/tidebreak-server/src/code/delivery.rs
+++ b/crates/tidebreak-server/src/code/delivery.rs
@@ -22,8 +22,9 @@ use tidebreak_core::db::code::{
 };
 use tidebreak_core::{
     CodePullRequestAttribution, CodePullRequestDiscovery, CodePullRequestId,
-    CodePullRequestRelation, CodeRepo, CodeWorkspace, CodeWorkspaceStatus, DbStore, OwnerId,
-    PullRequestCheckBucket, PullRequestComment, PullRequestCommentKind, RepoId, WorkspaceId,
+    CodePullRequestRelation, CodeRepo, CodeWorkspace, CodeWorkspaceStatus, OwnerId,
+    PullRequestCheck, PullRequestCheckBucket, PullRequestComment, PullRequestCommentKind,
+    PullRequestDigest, RepoId, WorkspaceId,
 };
 
 use super::gh::{self, GhObservation};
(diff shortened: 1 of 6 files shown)


naingthet:
**G2G** at adc7b6351b460fa0c1dc3fdc6b4225531e0d1afa

Sweeps consume a fresh live tier and fall back to the host read when the row is missing or stale. Delivery write-through and the digest projections keep that path aligned with the fetched one. No high-impact defect in the store-first gate, residual fetch, or stacked-child hold.

naingthet:
**G2G** at 2461112f57d9f1870cadf267719e5ff781ba041e

Sweeps consume a fresh live tier and fall back to the host read when the row is missing or stale. Delivery write-through and the digest projections keep that path aligned with the fetched one. No high-impact defect in the store-first gate, residual fetch, or stacked-child hold.

PR #2565, opened 2026-08-24
feat(code): fetch pull-request state over conditional REST behind one gate

naingthet:
Slice 4 of [decision 66](docs/decisions/0066-pull-request-state-is-one-store.md), first half: the conditional REST fetcher behind one gate. The scheduler tiers and the digest-cache deletion follow in the next slice.

## What changes

- `pr_fetch` is the one reader for pull-request state: `gh api` reads of the pull request, its head's check runs, and its reviews. Each read sends the fact row's stored ETag (`If-None-Match`); a 304 answers from the row and does not count against GitHub's primary rate limit, so sustained cost tracks how often pull requests change rather than how often Tidebreak asks. The row stores each endpoint's ETag in three new guarded columns.
- One `HostGate` paces every read: a small global concurrency, per-host spacing, and parking — a `Retry-After`, a secondary-limit answer, or an exhausted primary window parks the host for the stated time, and every caller respects the park rather than retrying hot.
- The review decision derives from the base branch's rules plus each reviewer's newest standing review, in the same lowercased vocabulary the classifier already reads, so `needs_approval` (slice 1) keeps firing without GraphQL's `reviewDecision`.
- The merge-queue timeline read runs only where the base branch's rules run a queue (one rules read per branch, cached for an hour). Repositories without a queue stop paying the pagination entirely.
- The old digest loader is deleted: the double `gh pr view` head dance and its retry loop, the `gh pr checks` table read, and the unconditional timeline pagination. Consistency now holds by construction — check runs are read for the exact head SHA the pull request answer named.
- `gh api` transport keeps the HTTP answer whatever the exit code (`gh` exits non-zero on 304), parses status, `Etag`, `Retry-After`, and rate-limit headers, and still refuses merge and GraphQL invocations.

## What does not change here

- `PrDigestCache` and its 20-second TTL stay; the tiered scheduler, mutation-dirty refreshes, and `GET …/pr` leaving the request path are the next slice.
- The hosted-machine path (decision 65) keeps its own credentialed REST reader.

## Testing

- Unit tests cover the raw HTTP parse, the REST field mapping, review standing (dismissals, `COMMENTED`), decision derivation, branch-rules parsing, queue-transition semantics, gate parking, and per-host spacing.
- Scripted-`gh` tests prove the 304 flow (a stored ETag turns an unchanged answer into `NotModified`) and that a 403 with `Retry-After` parks the host so the next read waits — the decision's validation case.
- The end-to-end tests (create flow, push-drops-cache, write-through to every holder) now run against `gh api` shims and assert the fetch path issues no `pr checks` read and no merge or GraphQL invocation.

### DIFF
--- a/crates/tidebreak-core/src/db/entities.rs
+++ b/crates/tidebreak-core/src/db/entities.rs
@@ -1716,6 +1716,9 @@ pub mod code_pull_request {
         pub auto_merge_enabled: Option<bool>,
         pub in_merge_queue: Option<bool>,
         pub live_observed_at: Option<DateTimeUtc>,
+        pub pull_etag: Option<String>,
+        pub checks_etag: Option<String>,
+        pub reviews_etag: Option<String>,
     }
 
     #[derive(Copy, Clone, Debug, EnumIter, DeriveRelation)]
--- a/crates/tidebreak-core/src/db/migration.rs
+++ b/crates/tidebreak-core/src/db/migration.rs
@@ -50,7 +50,53 @@ impl MigratorTrait for Migrator {
             Box::new(TriggerFactConditions),
             Box::new(CodeQueuedTurns),
             Box::new(CodePullRequestLiveTier),
+            Box::new(CodePullRequestEtags),
         ]
+    }
+}
+
+/// Conditional-request ETags on pull-request facts (decision 66).
+///
+/// The row stores the ETag each fetch endpoint last answered with — the
+/// pull request, its head's check runs, and its reviews — so the next read
+/// sends `If-None-Match` and an unchanged answer costs a free 304.
+struct CodePullRequestEtags;
+
+impl MigrationName for CodePullRequestEtags {
+    fn name(&self) -> &str {
+        "m20260824_000013_code_pull_request_etags"
+    }
+}
+
+#[async_trait::async_trait]
+impl MigrationTrait for CodePullRequestEtags {
+    async fn up(&self, manager: &SchemaManager) -> Result<(), DbErr> {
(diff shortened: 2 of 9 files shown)


naingthet:
**G2G** at 03a5d306236050dacdad59a62af61c4667e0331d

The conditional REST path matches the slice: one gate, ETags on the three paid endpoints, checks bound to the SHA the pull answer named, and 304s replay the stored row instead of inventing state. A 403 with Retry-After parks that host only. Timeline still goes through `run_gh` rather than the header-aware reader, so a queue-enabled (or rules-unknown) repo can spend pagination without parking; that is a follow-up, not a ship blocker for this half of the decision.

naingthet:
**G2G** at 75bb32820fd282c95a1bc7ad823c5a96daf8fe9d

Re-review of the new head. The only change since the prior G2G is `Debug` on `GatePermit`. The gate, ETags on the three paid endpoints, checks bound to the pull answer's SHA, and 304 replay of the stored row are unchanged. A 403 with Retry-After still parks that host only. Timeline still goes through `run_gh` rather than the header-aware reader; same follow-up as before, not a ship blocker.

naingthet:
**G2G** at 1d7fe64791215ec8c83e41b8777253bd5e3a503d

Re-review of the new head. The only change since the prior G2G is the create-PR test: creation now returns the light stub (no view, no checks, no merge) and leaves enrichment to the next status read through the conditional fetcher. That matches the create path and decision 66. Gate, ETags on the three paid endpoints, checks bound to the pull answer SHA, 304 replay of the stored row, and host-only 403 parking are unchanged.

PR #2566, opened 2026-08-24
fix(code): use Grok gateway model ids

naingthet:
## What changed

Tidebreak now uses model IDs reported by Grok when Model Gateway is connected, so code sessions send `model-gateway-model-gateway/grok-4.6` instead of routing bare `grok-4.6` to xAI Direct.

The new regression coverage verifies the first-session path and shared model-selection policy.

## Validation

Passed the UI test suite (226 files, 1,784 tests), focused tests, `tsc --noEmit`, Biome checks, and `git diff --check`.

## Compatibility and risk

No persisted or wire-format changes; existing sessions with bare IDs keep their stored selection until the model is reselected or a new session starts.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeWorkspacePage.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeWorkspacePage.tsx
@@ -175,6 +175,7 @@ import {
   HARNESS_LABELS,
   LIFECYCLE_LABELS,
   preferredCodeModels,
+  requiresHarnessModelIds,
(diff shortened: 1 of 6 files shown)


naingthet:
**G2G** at a110c98af8b0016842962aa146ba0d395da4eca9

Grok now shares OpenCode's rule: pick the harness listing so the first session posts `model-gateway-model-gateway/grok-4.6` instead of a bare catalog id. The helper is used on every list path that used to special-case OpenCode, and the new tests cover both the policy and the send payload.

PR #2567, opened 2026-08-24
feat(code): start a workspace from a prompt-first composer

naingthet:
## What changed

The new-workspace dialog was a stack of labelled fields with the starting prompt fourth in the tab order, and the prompt it collected only landed in the workspace composer as a draft the reader still had to send; it is now a composer, where the message is the surface, focus lands there on open, and repo, engine, model, permission mode, base ref, and name are pills on the chrome that reopen on what was used last, each with its own chord (Cmd+N repo, Alt+E/M/P/B/N). The typed prompt now runs as the session's first turn, falling back to the workspace draft only when that turn cannot be sent, and a `Create more` toggle keeps the dialog open to fire off another task. Enter sends, Shift+Enter is a newline, and Cmd+Enter still creates from anywhere in the dialog including an open picker. `HarnessModelMenu` and `PermissionModeMenu` gained optional controlled-open props so the chords can drive them, `PermissionModeMenu` now takes `availableModes` so modes an engine cannot honor render disabled rather than missing, and `Switch` gained an optional `thumbClassName` so a caller can size the thumb with the track.

## Validation

`vitest run src/code src/Composer.test.ts src/PermissionModeMenu` (63 files, 662 tests) passes, along with `tsc --noEmit`, `biome format`, and `biome lint`. Both new Storybook stories were screenshotted at 1200px and at the 720px app minimum window width; an earlier revision leaked the footer 28px past the card's rounded edge, which DevTools traced to `DialogContent` being a grid whose item defaulted to `min-width: auto`, so the form's 615px min-content sized the column past the 588px card — fixed with `min-w-0` on the form and rows, `overflow-hidden` on the card, and truncation on the model and permission pills.

## Compatibility and risk

No wire or schema changes. `tidebreak.code-last-create` gains an optional `permissionMode` key, read field-wise and validated against the four known modes, so an older or newer blob still loads. `PermissionModeMenu` and `Switch` are shared with chat; both changes are additive and the defaults are unchanged.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/AppShell.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/AppShell.dom.test.tsx
@@ -1009,9 +1009,7 @@ describe("app shell", () => {
 
     const dialog = await screen.findByRole("dialog");
     expect(
-      within(dialog).getByText(
-        "One worktree and one session on the selected repo.",
-      ),
+      within(dialog).getByRole("textbox", { name: "First message" }),
     ).toBeInTheDocument();
     expect(createChat).not.toHaveBeenCalled();
     expect(router.state.location.pathname).toBe("/code");
--- a/crates/tidebreak-desktop/ui/src/PermissionModeMenu.tsx
+++ b/crates/tidebreak-desktop/ui/src/PermissionModeMenu.tsx
@@ -103,12 +103,24 @@ export function PermissionModeMenu({
   value,
   disabled,
   onChange,
+  availableModes,
+  open: controlledOpen,
+  onOpenChange,
 }: {
   /** Identity of the chat or draft whose setting is being changed. */
   scopeKey: string;
   value: PermissionMode | null;
   disabled?: boolean;
   onChange: (mode: PermissionMode) => void | Promise<void>;
+  /**
+   * Modes the engine behind this surface honors. Absent means all of them —
+   * chat's server runs every mode. Rows outside the list stay visible and
+   * disabled, so what an engine cannot do is stated rather than missing.
+   */
+  availableModes?: readonly PermissionMode[];
+  /** Open the menu from outside — a surface's keyboard shortcut. */
+  open?: boolean;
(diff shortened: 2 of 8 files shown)


naingthet:
**G2G** at e78ef09578c45f9b525dd08307d36ed5453911e5

The prompt-first create path is coherent: first turn posts when the session exists, the workspace composer keeps the text when the session or the turn fails, and Create more only clears the message and name. lastCreate reads permissionMode field-wise against the four known modes. Cmd+Enter still creates with a picker open. Nothing here is a high-impact miss CI would not catch.

naingthet:
**G2G** at d2c6b6fe89948cc229aafcc2c1876e778aa3b111

The prompt-first create path still holds on this head. A typed first turn posts once the session exists, and session or turn failure hands the text to the workspace composer instead of dropping it. Create more clears only the message and name. lastCreate still reads permissionMode field-wise against the four known modes, and now keeps a model per harness so a switch back does not lose the last pick. Cmd+Enter still creates with a picker open. Nothing here is a high-impact miss CI would not catch.

naingthet:
**G2G** at 55e9cfacec81ebe1340a74b882577dffbe0048e5

The prompt-first create path still holds on this head. A typed first turn posts once the session exists, and session or turn failure hands the text to the workspace composer as a draft rather than dropping it. Create more stays on the same settings and clears only the message and name. lastCreate still reads permissionMode field-wise against the four known modes and keeps a model per harness. Cmd+Enter still creates with a picker open. Nothing here is a high-impact miss CI would not catch.

naingthet:
**G2G** at e2a1a05c3285170d1b9179e7fbc3cf1746a433bc

The prompt-first create path still holds on this head. A typed first turn posts once the session exists, and session or turn failure hands the text to the workspace composer as a draft rather than dropping it. Create more stays on the same settings and clears only the message and name. lastCreate still reads permissionMode field-wise against the four known modes and keeps a model per harness. Cmd+Enter still creates with a picker open. Nothing here is a high-impact miss CI would not catch.

PR #2569, opened 2026-08-24
feat(code): a delivery nudge replaces the notification poll timers

naingthet:
Stacked on #2568 (the hot tier); the diff here is against that branch, and GitHub retargets this pull request when it merges.

Slice 5 of [decision 66](docs/decisions/0066-pull-request-state-is-one-store.md): the delivery surfaces stop polling on their own clocks.

## What changes

- `WS /code/updates` gains a `delivery` notice. When the pull-request store takes a real change — from the hot refresher, a mutation, or the reconcile sweep's delivery reads — the server nudges each owner's channel once, debounced to three seconds so a sweep that moves many rows costs one re-read on the other side.
- The delivery monitor drops its 30-second and 2-minute poll tiers. It re-reads its queries when the nudge arrives; the queries stay server-cached, so a nudge costs one local round trip. Two slow safety passes remain: 5 minutes visible and 10 minutes hidden, because run summaries have no store yet and a dropped socket must not silence notifications. A nudge while the window is hidden waits for the hidden pass.
- The workspace pane needed nothing: it already rides the session digests on the same socket, and the previous slice took its `GET …/pr` reads off GitHub.

## Why

This closes decision 66's freshness loop: the store diffs on write, a change broadcasts once, and every surface reacts to the broadcast instead of asking on a timer. With the conditional fetcher and the hot tier from slice 4, steady-state GitHub traffic now scales with how often pull requests change, not with how many views are open.

## Testing

- Store tests cover the notice mapping and the revision bump per nudge.
- `tsc`, focused vitest, and biome pass; the wire binding is hand-matched to the generator's output and the staleness test verifies it in CI.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeDeliveryMonitor.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeDeliveryMonitor.tsx
@@ -13,9 +13,13 @@ import {
   trackedCodeDeliveryRepositories,
   useCodeDeliveryStore,
 } from "./CodeDeliveryStore";
+import { useCodeUpdatesStore } from "./CodeUpdatesStore";
 
-const ACTIVE_POLL_MS = 30_000;
-const VISIBLE_POLL_MS = 2 * 60_000;
+// Freshness rides the `delivery` nudge on the updates socket (decision 66):
+// the server says when the pull-request store changed, and this monitor
+// re-reads then. The remaining timers are a safety net — run summaries have
+// no store yet, and a dropped socket must not silence notifications.
+const SAFETY_POLL_MS = 5 * 60_000;
 const HIDDEN_POLL_MS = 10 * 60_000;
 const FIRST_RUN_LOOKBACK_MS = 24 * 60 * 60 * 1_000;
 const MAX_LOOKBACK_MS = 30 * 24 * 60 * 60 * 1_000;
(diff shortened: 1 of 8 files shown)


naingthet:
**G2G** at 4ec92add3b6b6f04c8cc59d3095f4105e520cf7a

The delivery notice is wired through the live-update bus, the socket, and `noticeToAction`, then the monitor wakes on `deliveryRevision` instead of the 30s/2m page timers. The 3s per-owner debounce matches a sweep that should cost one re-read. Hidden tabs still wait for visibility or the 5/10 minute safety net, which is the right trade for a dropped socket and for run summaries that still have no store. I would not block on the drop-style debounce missing a second change inside that 3s window; the safety poll covers it and CI would not teach us more here.

naingthet:
**G2G** at 8e50889a5a7d5ed30e74ffb6a9b1348b9f66520c

The delivery notice still rides the live-update bus, the socket, and `noticeToAction`, and the monitor still wakes on `deliveryRevision` instead of the 30s/2m page timers. The 3s per-owner debounce still collapses a sweep to one re-read. Hidden tabs still wait for visibility or the 5/10 minute safety net, which is the right trade for a dropped socket and for run summaries that still have no store. Nothing on this head changes that call.

naingthet:
**G2G** at 718d5c4f47b4f93fcaa706e45b03c69daa997adf

The delivery notice still rides the live-update bus, the socket, and `noticeToAction`, and the monitor still wakes on `deliveryRevision` instead of the 30s/2m page timers. The 3s per-owner debounce still collapses a sweep to one re-read. Hidden tabs still wait for visibility or the 5/10 minute safety net, which is the right trade for a dropped socket and for run summaries that still have no store. Nothing on this head changes that call.

naingthet:
The drop-style debounce noted in review is filed as #2580, so it does not get lost with the branch.

PR #2571, opened 2026-08-24
fix(desktop): improve dense UI states

naingthet:
## Summary

- Improve Settings navigation and minimum-window layouts.
- Clarify dense Composer context, compact busy controls, and code-workspace hierarchy.
- Improve Outputs Storybook sizing and Apps loading, empty, failure, and dense states.
- Strengthen accessible names, grouping, reading order, and failure-state comprehension.

## Validation

- Biome on all 20 changed files
- TypeScript
- 87 focused tests
- Production Storybook build
- Three independent review passes returned `GO` for layout, interaction, and accessibility

Visual review captures use headless Playwright. They do not use Computer Use or visible local applications.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/Composer.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/Composer.test.tsx
@@ -524,6 +524,23 @@ describe("Composer", () => {
     expect(markup).toContain("800 × 600");
   });
 
+  it("collapses dense message context behind a readable summary", () => {
+    const markup = composerWithImages({
+      items: [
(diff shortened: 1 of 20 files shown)


naingthet:
**G2G** at 7005d5cef1a2a664876cde5708fdf062dabb340f

Dense Settings, Composer, Apps, and workspace chrome read as intended: context collapses only when the count first crosses three, settings groups keep a real nav landmark, Apps failure is distinct from empty, and the workspace workflow wraps instead of clipping. Nothing here is a high-impact miss CI would not catch.

PR #2572, opened 2026-08-24
feat(code): a hot tier refreshes viewed pull requests and reads leave the request path

naingthet:
Follows #2565 (the conditional fetcher, merged). Supersedes #2568, which GitHub closed when the merged parent branch was deleted out from under the stack.

Slice 4 of [decision 66](docs/decisions/0066-pull-request-state-is-one-store.md), second half: the tiered refresh schedule, and the request path off GitHub.

## What changes

- `GET …/pr` stops calling GitHub. The read marks the workspace hot and answers from local git plus the stored row, so opening a workspace stops paying the one-to-three-second digest fetch. `POST …/pr/refresh` stays a forced host read.
- A 17-second refresher (coprime with the 47s watch, 53s trigger, and 61s reconcile sweeps) walks the hot set — workspaces whose digest was requested in the last two minutes, plus active watches — through the conditional fetcher. An unchanged pull request costs two 304s; freshness while anyone watches is bounded by the tier, not by which component happens to poll.
- Mutations dirty the row and refresh it before answering: push, create, merge, and mark-ready responses carry the fetched state of the new head. The pre-push staleness window is gone rather than shortened.
- `PrDigestCache` and its 20-second TTL are deleted. The workspace column — written by the refresher, by mutations, and by the write-through fanout — is the one copy a read serves, so there is no second store to disagree with it.
- The merge-queue timeline read parks the host when a limit answer names it, closing the follow-up from the fetcher review.
- The hosted-machine read (decision 65, borrowed credential, no `gh`) stays in the request path; it has no refresher to ride yet.

## Why

Decision 66's freshness goal: "no view owns a timer, and no view triggers an unconditional fetch." After this slice the server side is done — one fetcher, one gate, one schedule, one store — and slice 5 can drop the UI's own poll timers onto the broadcast channel.

## Testing

- The push test proves the row refreshes inside the push and that a plain read never fetches.
- The write-through test now proves one refresh feeds every holder's read: workspace B's `GET …/pr` serves the column workspace A's refresh wrote, with no host read of its own.
- The create test keeps asserting the response carries the fresh pull request's checks, now via the mutation-refresh path.

### DIFF
--- a/crates/tidebreak-server/src/code/gh.rs
+++ b/crates/tidebreak-server/src/code/gh.rs
@@ -5,7 +5,6 @@
 //! a hosted machine instead lends each git operation a dying, gateway-minted
 //! credential through the environment (decision 63).
 
-use std::collections::HashMap;
 use std::ffi::OsString;
 use std::path::{Path, PathBuf};
 use std::process::Stdio;
@@ -16,7 +15,7 @@ use tokio::process::Command;
 use tokio::sync::Mutex as AsyncMutex;
 use tokio::time::timeout;
 
-use tidebreak_core::{Diffstat, PullRequestDigest, QuickAction, WorkspaceId};
+use tidebreak_core::{Diffstat, PullRequestDigest, QuickAction};
 use tidebreak_harness::{filter_child_env, probe_shell, HostEnv};
 
 use super::setup_script::spawn_workspace_script;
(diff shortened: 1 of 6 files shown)


naingthet:
**G2G** at 6e37e47dbfad4692f2d9c4015dbc269017eb694f

The request path marks the workspace hot and answers from local git plus the stored column. Mutations and the 17-second sweep refresh through the existing conditional fetcher, which already writes the fact row and fans the digest to other holders. Cache removal is complete. Hosted REST stays on the read path as documented. Nothing here is a CI-blind failure.

PR #2575, opened 2026-08-24
fix(code): skip archive confirmation for clean workspaces

naingthet:
## Summary

- Archive clean workspaces without a confirmation dialog.
- Show one discard confirmation when the server reports uncommitted or unpushed work.
- Cover the command shortcut, workspace menu, and forced archive paths.

## Tests

- `pnpm exec vitest run src/code/CodeWorkspacePage.dom.test.tsx -t "archive"`
- `pnpm exec biome check src/code/workspaceActions.tsx src/code/CodeWorkspacePage.dom.test.tsx`
- `git diff --check`


naingthet:
**G2G** at c95e4c5bfcfa6eec355a0186030b852d93028790

Clean archive no longer asks first. That matches the title. A 409 still opens Discard leftover work, then retries with force. The chord, menu, and force-archive tests cover those paths. Accidental archive on a clean tree is now one keystroke; that is the product choice, not a defect CI would miss.

PR #2576, opened 2026-08-24
feat(code): engines download the first time you pick one

naingthet:
## What changed

Refresh used to `npm install` every pinned engine, so the only way to make one usable was to fetch all four — hundreds of megabytes and minutes for three engines the reader never asked for. Nothing is fetched now until someone picks it: refresh only re-probes, `HarnessDoctorEntry` carries `installable` so a missing-but-fetchable engine stays selectable in both the New Workspace and start-session pickers, and choosing one starts its download through a shared warm-install hook while create waits for the pin instead of stalling silently on it. Settings gained a per-engine Download button (`?deliberate=true`, the only case that retries a managed-Node install that already failed), and the doctor was rebuilt around it — one even row per engine leading with its state and the one thing to do about it, with version, path, capabilities, and probe output behind a disclosure rather than a stack of label/value pairs. `/code` no longer hands a cold machine the doctor instead of the register form, since picking an engine there now fetches it.

## Validation

`vitest run src/code src/settings src/api.test.ts` (74 files, 830 tests) passes, along with `tsc --noEmit`, `biome format`, `biome lint`, `cargo fmt --all`, and `storybook:build`. Six `Settings/Coding harnesses` stories and seven `Code/Harness doctor` stories cover ready, nothing-downloaded, mixed, downloading, download-failed, signed-out, and empty; each was screenshotted at 1200px and read back — the first pass had jumpy row heights and a "Details" link repeated on every row, which became one chevron per row and a fixed note line. Rust is not compiled locally; CI owns `cargo clippy` and `cargo test`.

## Compatibility and risk

`HarnessDoctorEntry` gains a required `installable` field and `POST /code/harnesses/{kind}/install` gains an optional `deliberate` query parameter; the client parser reads `installable` field-wise and defaults it to `false`, so an older server still loads. `CodeRuntime::refresh_pinned_harnesses` is deleted with its only caller. No schema or storage changes.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/api/client.ts
+++ b/crates/tidebreak-desktop/ui/src/api/client.ts
@@ -2179,12 +2179,21 @@ export class ApiClient {
    * that follow arrive on `WS /code/updates`. Safe to call repeatedly — an
    * installed pin answers `ready` and one already running is not restarted.
    */
+  /**
+   * Start — or report — the pinned download of one engine.
+   *
+   * `deliberate` marks a reader who pressed Install rather than a picker
+   * warming its selection. Only that case retries a managed-Node install that
+   * already failed.
+   */
   async startHarnessInstall(
     kind: HarnessKind,
+    deliberate = false,
   ): Promise<CodeHarnessInstallSnapshot> {
+    const path = `/code/harnesses/${encodeURIComponent(kind)}/install`;
     return requireParsed(
       parseCodeHarnessInstall(
-        await this.json(`/code/harnesses/${encodeURIComponent(kind)}/install`, {
+        await this.json(deliberate ? `${path}?deliberate=true` : path, {
           method: "POST",
           headers: this.headers(),
         }),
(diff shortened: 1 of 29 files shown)


naingthet:
**G2G** at d2d7713db2a382e943698006b90a3e29deb87611

Refresh no longer installs every pin. Pickers treat a missing but installable engine as a wait, start one download through the shared hook, and keep create disabled until the pin is on disk. Settings Download is the only `deliberate` retry. `installable` defaults false on an older server. The claim lock still coalesces in-flight work for a pin. Nothing here is a merge blocker CI would miss.

naingthet:
**G2G** at 06d65b961f0f1e3f175f22a08e1c70c333a2d98c

Warm install now starts only when the doctor names that engine missing, not when the report has not landed yet. Opening the dialog on a fallback guess no longer fetches a pin nobody picked, and a client that cannot install still renders the picker. Create stays disabled until the pin is on disk. Nothing here is a merge blocker CI would miss.

PR #2577, opened 2026-08-24
fix(code): apply reasoning effort changes to the next turn

naingthet:
## What changed

When you change reasoning effort during a running turn, the UI now keeps that choice pending instead of calling the session setting endpoint. Same-session refreshes preserve the choice until the next submitted or queued turn carries it and makes it sticky. Idle changes still update the session immediately.

A page-level regression test covers changing effort during a run, waiting for that turn to finish, and then sending.

## Validation

- `pnpm exec vitest run src/code/CodeWorkspacePage.dom.test.tsx` (48 passed)
- `pnpm exec biome lint src/code/CodeWorkspacePage.tsx src/code/CodeWorkspacePage.dom.test.tsx`
- `pnpm exec tsc --noEmit`

## Compatibility and risk

Low. The change reuses the existing optional `reasoning_effort` submission field and does not change persisted data or wire formats.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeWorkspacePage.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeWorkspacePage.dom.test.tsx
@@ -368,6 +368,12 @@ function makeClient() {
         user_input: message,
       },
     })),
+    setCodeSessionReasoningEffort: vi.fn(
+      async (_sessionId: string, reasoningEffort: "low" | "high" | null) => ({
+        ...SESSION,
+        reasoning_effort: reasoningEffort ?? undefined,
+      }),
+    ),
     getCodeWorkspaceDiff: vi.fn(async () => ({
       diff: "",
       truncated: false,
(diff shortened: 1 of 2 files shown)


naingthet:
**NO GO** at 0676c31aee3dd1334e724de37a7cad19b160dcc4

1. The session-settings effect still wipes a mid-turn reasoning choice as soon as the running turn ends. You set `pendingReasoningEffortRef` when `turnRunning` is true and skip the session endpoint, but the effect that syncs from `session.reasoning_effort` also does `pendingReasoningEffortRef.current = null` and resets the picker. After the current turn finishes, the next poll or store update restores the old effort and drops the pending value. The user who changes effort during a run, waits for it to finish, then sends is the common path; the new test only queues while the turn is still running, so it never hits this. The next submit then goes out without `reasoning_effort`, and the picker snaps back to the previous level.

naingthet:
Addressed in `7c2f85a9c`. Same-session settings refreshes now preserve the pending reasoning choice until an accepted submission clears it. Permission-mode and fast-mode responses follow the same rule, so they cannot make the picker snap back either.

The regression now changes the effort during a running turn, completes that turn, applies a refreshed row that still carries the old effort, verifies that the picker remains on the pending choice, and then verifies that the next idle submission includes it. All 48 page tests pass, along with Biome lint, TypeScript, and the diff check.

naingthet:
**G2G** at 7c2f85a9cb8c7fd2896a58bba70cffd8dcf3914f

The session-settings effect now keeps a mid-turn reasoning choice when a refreshed row still carries the stored effort. Permission-mode and fast-mode replies do the same, and only an accepted submit clears the pending value. The page test now finishes the running turn, applies that stale row, and checks that the next idle send still includes the new effort. The earlier wipe is gone.

PR #2581, opened 2026-08-24
feat(code): add analytics dashboard

naingthet:
## What changed

Adds an Analytics tab before Delivery with local session, turn, token, model, repository, pull-request, and estimated-cost reporting, plus date and repository filters.
Uses shared Model Gateway subscription windows when available and keeps local analytics functional without Gateway; standard canonical OpenAI and Anthropic rates estimate known usage while unknown, fast, and third-party routes remain unpriced.

## Validation

Passed TypeScript, Biome lint and format, 75 focused UI tests, the full 1,787-test UI suite, `cargo fmt`, and Storybook builds with Gateway, local-only, loading, empty, failure, long-content, and narrow stories.
Rust compilation and tests did not run locally under repository policy; CI will compile and test the migration and server changes.

## Compatibility and risk

Adds a forward-only database migration for per-turn model and fast-mode snapshots plus a new `/code/analytics` wire contract; older turns remain unpriced instead of inheriting current session settings.


naingthet:
**G2G** at ad998b8f05571fb46d0f31723604e48f664d5d02

Owner-scoped analytics load turns, sessions, and pull-request facts through the existing ScopedCode path. New turns snapshot model and fast_mode so later session edits do not rewrite history. Unknown, fast, and third-party routes stay unpriced. Nothing here looks like a high-impact correctness or isolation miss CI would not catch.

naingthet:
**G2G** at ac3e52235fb0f2f309a6c85a111fc71d3b8d89f9

The CI follow-up does not change the report contract. Turns still snapshot model and fast_mode, unknown and fast routes stay unpriced, and the dashboard still reads through ScopedCode.

One taste note that is not a blocker: the hero cost card uses an em dash when nothing is priced, while the repository table still prints $0.00 for the same case. Readers will treat those as different facts. Worth lining them up later so unpriced usage never looks free.

PR #2582, opened 2026-08-24
fix(ui): raise Storybook build heap limit

naingthet:
## What changed

Run the static Storybook build with a 6 GB Node heap so the growing UI bundle does not exhaust the default 4 GB limit.

## Validation

- `pnpm storybook:build`
- `git diff --check`

## Compatibility and risk

None. The change only affects the static Storybook build command.

### DIFF
--- a/crates/tidebreak-desktop/ui/package.json
+++ b/crates/tidebreak-desktop/ui/package.json
@@ -6,7 +6,7 @@
   "scripts": {
     "dev": "vite",
     "storybook": "storybook dev -p 6006",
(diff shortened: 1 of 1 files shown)


naingthet:
**G2G** at d1ee4156820ba69863979df7ed50f0111527c8d4

The Storybook static build now uses the same 6 GB Node heap as the Vite production build, so the larger UI bundle does not die on the default 4 GB limit. The dispatcher path is the usual Storybook CLI entry. Risk stays on that one script.

PR #2583, opened 2026-08-24
fix(code): correct Codex usage and preserve checkpoints

naingthet:
## What changed

Codex usage now reports per-turn counters, and known streaming notifications no longer count as protocol drift.
Checkpoint recovery now skips missing refs and preserves refs owned by other Tidebreak profiles.

## Validation

Formatting, fixture JSON validation, and diff checks pass; Rust tests are deferred to CI under repository policy.

## Compatibility and risk

No persisted or wire-format changes.

### DIFF
--- a/crates/tidebreak-core/src/db/ops/code/repo.rs
+++ b/crates/tidebreak-core/src/db/ops/code/repo.rs
@@ -81,9 +81,9 @@ pub async fn list_repos(store: &DbStore, owner: &OwnerId) -> Result<Vec<CodeRepo
 
 /// Every registered repository on the machine, most recently created first.
 ///
-/// A system path, not a request path: the boot-time checkpoint-ref sweep
-/// walks every repo regardless of who owns it. Nothing reachable from a
-/// route may call it.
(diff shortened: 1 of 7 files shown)


naingthet:
**G2G** at d51cf4847f52abacbfa5792b73886415994fbfda

Codex usage now snapshots thread totals at turn start and reports the delta, so a resume no longer bills the first turn again. Treating the known streaming notifications as quiet is the right call; those events never belonged in the transcript and were only making healthy turns look like drift.

Skipping a missing checkpoint ref and walking back to the session baseline is the recovery you want when a row outlives its hidden ref. Dropping the global orphan sweep is the correct product shape once two Tidebreak profiles can share one repository: one database cannot decide the other profile's refs are garbage. Archive still deletes the workspace it actually owns. Refs left by crashes can linger; that is the trade you document, and incremental objects make it cheap enough to ship.

The tests pin the resume usage numbers, the missing-ref fallback, and that another workspace's refs survive archive. This is the right cut.

PR #2585, opened 2026-08-24
perf(code): update workspace lifecycle optimistically

naingthet:
## What changed

Workspace creation now closes the dialog immediately, shows a temporary progress card, supports concurrent Create more requests, survives catalog refreshes, and offers retry when creation fails.

Archive actions now hide the workspace immediately, disable the open workspace while cleanup runs, and restore the original state when the request fails or the discard confirmation is cancelled.

## Validation

Ran TypeScript checking, Biome linting, 92 focused tests, the full 1,813-test UI suite, the Storybook production build, and `git diff --check`; no browser session was available to capture the Storybook screenshot.

## Compatibility and risk

This change does not alter persisted data, wire formats, platform behavior, or security boundaries; risk is limited to client-side workspace catalog and navigation state during create and archive requests.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeCatalogStore.test.ts
+++ b/crates/tidebreak-desktop/ui/src/code/CodeCatalogStore.test.ts
@@ -1,7 +1,10 @@
 import { afterEach, describe, expect, it, vi } from "vitest";
 
 import type { CodeWorkspaceSnapshot } from "../api/types";
-import { useCodeCatalogStore } from "./CodeCatalogStore";
+import {
+  OPTIMISTIC_WORKSPACE_ID_PREFIX,
+  useCodeCatalogStore,
+} from "./CodeCatalogStore";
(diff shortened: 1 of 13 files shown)


naingthet:
**G2G** at 4220e386fbfbc2ebf6cd715e443f371933682148

Create now closes immediately, parks a local card, and swaps it in place when the server answers. Failure removes that card and keeps the request on Retry. Archive flips the catalog first and puts the original snapshot back on a failed request, a cancelled discard, or a failed force. The open page follows that same snapshot, so the composer does not keep offering work against a checkout that is already leaving.

Refresh only preserves new local ids, so an in-flight catalog list can briefly show an archived workspace as active again until the archive call finishes. That is a flicker, not a stuck lie, and the rest of the lifecycle tells the truth.

The creating card is the right shape: spinner, no menu, no palette jump, no route to a local-only id. Concurrent Create more and staying put when the reader already left the origin path both match how people actually use this dialog.

PR #2588, opened 2026-08-25
feat(code): clone hosted repos from a picker, without guessing a path

naingthet:
## Summary

A hosted machine no longer asks you to guess a clone path or type `owner/repo` blind.

Self-host clones land under `{data_dir}/code/src` unless an operator set a destination, so the Destination field stays hidden. When a window is attached and the machine still has no destination, the dialog says so instead of showing an empty box. GitHub is a searchable picker of the repositories the gateway lists for you. You can still type a name that is not on the first page. If the list fails, that picker stays and says the suggestions did not load. A machine with no list still offers the typed `owner/repo` field. Clone sits at the bottom right and submits on Cmd+Enter, matching Create on a new workspace.

Decision 70. Needs the gateway list surface (companion PR brightwave-inc/model-gateway#1397) for the picker; the destination default works without it.

## Test plan

- [ ] Attach the desktop to a hosted machine with an empty store. Add a GitHub repo: Destination is gone, Repository is a searchable picker of your repos, Clone is bottom-right, Cmd+Enter starts the clone.
- [ ] Type an `owner/repo` that is not in the first page. Clone still starts.
- [ ] If the repository list fails, the picker stays, you can still type `owner/repo`, and you see that the suggestions did not load.
- [ ] On a local desktop, Destination still appears, Browse still works, and Repository stays a typed `owner/repo` field.


naingthet:
**NO GO** at 7e02c79fd0d015e2255881a3e27ee26a3f442d93

The destination story is honest. A self-host machine hides the path. An attached window with no destination says so and disables Clone. A local desktop still asks. The GitHub field is the miss.

1. `list_repositories` reads the gateway list with `RESPONSE_LIMIT` (16KB), the cap for a token-exchange body. Decision 70 promises the first page of 100 names. That page, with descriptions, does not fit. `read_bounded` refuses it, the runtime turns the error into an empty list, and `GithubRepoField` treats `[]` as no lender and swaps the picker for a bare `owner/repo` box. A hosted user with a normal GitHub account never sees the dropdown this change exists to add, and they cannot tell a failed list from an empty one. Give the list its own budget, the way the catalog already uses 1MB, and keep the combobox when the list fails so they can still type a name and see that the suggestions did not load.

naingthet:
**NO GO** at 3972719bff630aac3cbe9df0bbf5036354f25dca

This head covers `Profile` exhaustively and the Cmd+Enter test types. The GitHub list still fails the way the last review said.

1. `list_repositories` still reads the gateway page with `RESPONSE_LIMIT` (16KB), the token-exchange cap. Decision 70 promises the first page of 100 names. That page, with descriptions, does not fit. `read_bounded` refuses it, `list_github_repositories` turns the error into `[]`, and `GithubRepoField` still treats a non-null empty list as no lender and swaps the picker for a typed `owner/repo` box. The client does the same on a fetch failure (`setGithubRepos([])`). A hosted user with a normal account never sees the dropdown this change exists to add, and they cannot tell a failed list from an empty one. Give the list its own budget, the way the catalog already uses 1MB, and keep the combobox when the list fails so they can still type a name and see that the suggestions did not load.

naingthet:
**NO GO** at 1ba6fa3eddaffee8887249ab3379fbcc3bb150a5

The destination story is still honest. The GitHub list is the same miss as 3972719b, and this head made the picker worse.

1. `list_repositories` still reads the gateway page with `RESPONSE_LIMIT` (16KB), the token-exchange cap. Decision 70 promises the first page of 100 names. That page, with descriptions, does not fit. `read_bounded` refuses it, `list_github_repositories` turns the error into `[]`, and `GithubRepoField` still treats a non-null empty list as no lender and swaps the picker for a typed `owner/repo` box. The client does the same on a fetch failure (`setGithubRepos([])`). A hosted user with a normal account never sees the dropdown this change exists to add, and they cannot tell a failed list from an empty one. Give the list its own budget, the way the catalog already uses 1MB.

2. The last commit replaced the searchable combobox with a `Select`. When a list is present you cannot type a name that is not on the first page, and when the list fails you still get a silent typed box. Decision 70 excludes paging past 100 because typing was the escape. Keep a field that shows suggestions and still accepts `owner/repo`, and when the list fails keep that field with a load error instead of swapping it away.

naingthet:
**G2G** at 316d87804f2e1ee3ce9aa54790bd94f52e3f2fe2

The list now has its own 1MB budget, a lender error fails the request, and a failed list keeps type-in with "Suggestions did not load." An empty list is only the no-lender path. Destination still hides on self-host, and an attached window with no destination still says so.

The GitHub field is still a Select when the list is present, so you cannot type a slug that is not on the first page. Decision 70 already parks that as a revisit. It is not a ship blocker.

naingthet:
**G2G** at ff6f933f04e2e934bf9a59772704e43904da6a4e

The two earlier misses still hold closed. The list has its own 1MB budget, a lender error fails the request, and a failed list keeps the picker with "Suggestions did not load." An empty list is only the no-lender path.

This head also restores type-in. When a list is present you can still name an `owner/repo` that is not on the first page, which is the escape Decision 70 already relies on. Destination still hides on self-host, and an attached window with no destination still says so.

The closed control still says "Select a repository" even when the next step is to type. That is a little stiff, but the field works and the failed-list copy tells you what to do.

naingthet:
**G2G** at 2eef5b88d1f7240e9a2842f05bc0e86bf9066337

This head only syncs the generated `CodeRepoSources` comment with the self-host default. The two earlier misses still hold closed.

The list has its own 1MB budget, a lender error fails the request, and a failed list keeps the picker with "Suggestions did not load." An empty list is only the no-lender path. When a list is present you can still name an `owner/repo` that is not on the first page. Destination still hides on self-host, and an attached window with no destination still says so.

The closed control still says "Select a repository" even when the next step is to type. That is a little stiff, but the field works and the failed-list copy tells you what to do.

PR #2589, opened 2026-08-25
refactor(desktop): share document viewer states

naingthet:
## What changed

The frontend now shares viewer shells, loading states, and accessible errors across document and output formats. Source-specific viewer state resets when documents change, image previews reuse the shared object-URL lifecycle, standard buttons handle undo actions, and Storybook covers the affected states.

## Validation

Ran 41 focused tests, frontend lint, TypeScript, the styles contract, the application production build, and the Storybook production build; screenshot capture was unavailable because no browser instance was connected.

## Compatibility and risk

No persisted data, wire-format, platform, or security-boundary changes.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/ChangeSummaryCard.tsx
+++ b/crates/tidebreak-desktop/ui/src/ChangeSummaryCard.tsx
@@ -133,15 +133,17 @@ export function ChangeSummaryCard({ client, chatId, turnId, files }: Props) {
             </p>
           )}
         </div>
-        <button
+        <Button
           type="button"
-          className="inline-flex shrink-0 items-center gap-1.5 rounded-md border border-border px-2.5 py-1.5 text-xs font-medium hover:bg-accent disabled:cursor-not-allowed disabled:opacity-50"
+          variant="outline"
+          size="sm"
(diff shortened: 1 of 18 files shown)


naingthet:
**NO GO** at 23b6cffe9eff80e11d1f663800f60a895bc13b9b

The shared shell and status surface are the right shape, and remounting image, PDF, and presentation viewers on `cacheKey` matches the claim that source-specific state resets. The dispatcher every engine goes through was left on the old path, so the unification does not hold for the next format.

1. `DocumentViewer`’s `ViewerBoundary` still suspends to a spinner with no copy. That is the first paint for PDF, DOCX, workbooks, and presentations — the longest wait, and the path a new engine will copy. Document details now say “Loading viewer…” for JSON/XML, and outputs say “Loading preview…” for the same `DocumentViewer`. Put `DocumentViewerState` on `ViewerBoundary` so every format shares one loading surface.

2. JSON and XML only remount the breadcrumb on `source.cacheKey`. Collapse/navigate signals and per-node expand state stay on the parent, so switching documents of the same type can keep the previous tree open. Remount the viewer the way `PdfViewer` and `ImageViewer` already do.

PR #2590, opened 2026-08-25
fix(core): repair pre-pin postgres schemas

naingthet:
## What changed

Add an idempotent migration that repairs PostgreSQL databases created by Tidebreak v0.58.0 or v0.59.0. It adds the lifecycle columns, replaces repository registration uniqueness with the soft-removal index, and widens the affected checks without deleting rows.

Add SQLite and PostgreSQL upgrade regressions. Run the released-schema PostgreSQL test in the required PostgreSQL CI lane. Update decision 61 so operators upgrade in place instead of recreating the database.

## Validation

- `git diff --check`
- Live repair applied and catalog-verified against the hosted development PostgreSQL database
- Rust compilation and tests left to GitHub CI per repository policy

## Compatibility and risk

The migration runs in one transaction and checks the live catalog before adding columns. It preserves existing rows. Fresh databases run the same migration as an idempotent no-op plus constraint normalization.

## Tracking

Follow-up to #2490.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -784,6 +784,10 @@ jobs:
         run: >-
           cargo test -p tidebreak-core --no-default-features --features postgres
           --test postgres_code_owner --locked
+      - name: Exercise released schema upgrades on PostgreSQL
+        run: >-
+          cargo test -p tidebreak-core --no-default-features --features postgres
+          --test postgres_release_upgrade --locked
 
   self-host-build:
     name: self-host server build
--- a/crates/tidebreak-core/src/db/migration.rs
+++ b/crates/tidebreak-core/src/db/migration.rs
@@ -52,8 +52,243 @@ impl MigratorTrait for Migrator {
             Box::new(CodePullRequestLiveTier),
             Box::new(CodePullRequestEtags),
             Box::new(CodeTurnModelSnapshot),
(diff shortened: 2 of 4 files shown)


naingthet:
**G2G** at 492735fd04291fbd481d8083f12e5e73d5cd8e43

The repair is the right shape for the hosted pre-pin store. v0.58.0 and v0.59.0 share one `baseline/code.rs`; v0.60.0 is the in-place edit that added `removed_at` / `cloned_from`, the partial registration index, workspace release columns, `reasoning_effort`, `abandoned`, and the trigger tables. This migration covers the changed columns, index, and checks. `BaselineRepair` already creates the missing tables. The rest of the baseline did not move between those tags.

You add nullable columns only when the catalog lacks them, then widen checks instead of rebuilding tables, so existing rows stay. Dropping `idx_code_repo_owner_root_path` and recreating the `removed_at IS NULL` unique index cannot reject a row the old all-rows unique index already allowed. Postgres-only check rewrites are the right split: SQLite cannot drop those CHECKs, and a desktop profile below the pin still resets.

The v0.58 postgres test rebuilds that shape by undoing the v0.60 fixture rather than shipping a frozen v0.58 dump. That is a little indirect, but the undo matches the released source delta, and the assertions write the new values and prove the old uniqueness is gone. Operators can upgrade in place.

PR #2592, opened 2026-08-25
fix(code): trail the delivery-nudge debounce

### LINKED ISSUE #2580: fix(code): trail the delivery-nudge debounce so a change inside the window still broadcasts
Deferred from [decision 66](docs/decisions/0066-pull-request-state-is-one-store.md), accepted by review on #2569.

`nudge_delivery_update` in `code/runtime.rs` debounces per owner on the leading edge: the first change publishes a `delivery` notice and sets a 3-second floor, and a change landing inside that window publishes nothing. Nothing trails it, so that second change reaches the delivery list only on the next safety poll — up to 5 minutes later.

The blast radius is small: workspace headers and cards ride the digest fanout, which has no debounce, so only the delivery list and the notification feed can lag.

Done looks like: the debounce trails rather than drops. A change suppressed inside the window schedules one nudge for the end of it, so a burst still costs one re-read and no change waits on the safety poll.

naingthet:
## Summary

- Schedule one trailing delivery update when a change arrives inside the per-owner debounce window.
- Coalesce each owner independently and stop pending tasks when runtime state shuts down.
- Cover burst coalescing, owner isolation, and shutdown behavior with paused-time tests.

## Checks

- `rustfmt --check`
- `git diff --check`
- Cargo checks deferred to CI per repository policy.

Closes #2580

### DIFF
--- a/crates/tidebreak-server/src/code/runtime.rs
+++ b/crates/tidebreak-server/src/code/runtime.rs
@@ -8,6 +8,7 @@ use std::time::{Duration, Instant};
 
 use chrono::Utc;
 use tokio::sync::oneshot;
+use tokio::time::Instant as TokioInstant;
 
 use tidebreak_core::db::code::{
     arm_trigger, delete_trigger, delete_workspace, get_approval, get_open_turn, get_repo,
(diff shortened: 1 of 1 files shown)


naingthet:
**G2G** at 5231e21610c4733aa1714c15b309ff513ccfcb8c

The old leading-edge floor dropped the write that mattered. A burst still costs one re-read, and a change inside the window now lands at the end of it instead of waiting on the safety poll. Per-owner coalescing, one pending trailing Instant, and Weak shutdown all match that contract. The tests pin burst sharing, owner isolation, and the drop path.

`publish` takes the bus rather than storing it, so the next debounce does not have to grow a task table on CodeRuntime. Trailing at `sent_at + 3s` rather than sliding from the last call is the right product call for a floor.

PR #2594, opened 2026-08-25
perf(code): reuse recent workspace transcripts

naingthet:
## Summary

- Keep the four most recent code-session transcript stores in memory while their views are closed.
- Close off-screen event sockets, then resume from the stored sequence when the workspace opens again.
- Render workspace and repository metadata from the existing catalog while the route refreshes in the background.

## Why

Returning to a workspace discarded its session store and forced the desktop to fetch turns and replay the full event journal again. Keeping a bounded recent-session cache lets the conversation render immediately without leaving unused sockets open.

## Testing

- `pnpm exec biome format src/code/CodeSessionRegistry.ts src/code/CodeSessionRegistry.test.ts src/code/CodeWorkspacePage.tsx src/code/CodeWorkspacePage.dom.test.tsx`
- `pnpm exec biome lint src/code/CodeSessionRegistry.ts src/code/CodeSessionRegistry.test.ts src/code/CodeWorkspacePage.tsx src/code/CodeWorkspacePage.dom.test.tsx`
- `pnpm exec vitest run src/code/CodeSessionRegistry.test.ts src/code/CodeWorkspacePage.dom.test.tsx` (60 tests)
- `pnpm exec tsc --noEmit`

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeSessionRegistry.test.ts
+++ b/crates/tidebreak-desktop/ui/src/code/CodeSessionRegistry.test.ts
@@ -2,6 +2,7 @@ import { afterEach, describe, expect, it, vi } from "vitest";
 import type { SequencedCodeEventFrame } from "../api/types";
 import {
   acquireCodeSession,
+  MAX_RETAINED_CODE_SESSIONS,
   peekCodeSession,
   releaseCodeSession,
   resetCodeSessionRegistry,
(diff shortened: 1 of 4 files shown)


naingthet:
@shipright review

naingthet:
**G2G** at cce04cef1
Parked sessions close the socket, keep lastSeq, and reopen with after=lastSeq. Initial turn hydration retries only when the first open threw, which is the right split. Eviction stays on the parked LRU and does not dispose a live controller.

naingthet:
@shipright review

naingthet:
@shipright review

naingthet:
**G2G** at 5e754010e
The registry and socket resume path are unchanged from cce04cef1: parked sessions close the socket, keep lastSeq, and reopen with after=lastSeq; hydration retries only when the first open threw. Eviction still walks only the parked LRU. The new head only seeds workspace/repo header state from the catalog, which does not alter transcript reuse or resume.

PR #2597, opened 2026-08-25
fix(code): show one copy action per completed turn

naingthet:
## What changed

Show one copy action on the final assistant message in each completed turn, instead of showing it on every progress update. Keep active turns copy-free and preserve one copy action for settled subagent transcripts, with DOM coverage and a Storybook example.

## Validation

- 27 focused `CodeTranscript` DOM tests passed; Biome, TypeScript, the Storybook production build, and `git diff --check` passed.
- Storybook screenshots could not run because no browser connection was available.

## Compatibility and risk

None. This change does not alter persisted data or wire formats.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeTranscript.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeTranscript.dom.test.tsx
@@ -134,6 +134,117 @@ describe("CodeTranscript", () => {
     expect(screen.getByText("unrecognized event dropped")).toBeInTheDocument();
   });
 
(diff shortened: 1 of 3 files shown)


naingthet:
@shipright review

naingthet:
**G2G** at 1451b83e0
Each completed turn now keeps Copy on the last assistant row that precedes its boundary, and a settled transcript with no boundary falls back to its last assistant. Active turns stay copy-free, and the signature includes ownership so the isolated card remounts when that assignment moves. The DOM cases cover a multi-progress turn, a busy turn, and a filtered subagent view, and nothing here looks like a user-facing break.

PR #2598, opened 2026-08-25
feat(desktop): attach long pasted text to messages

naingthet:
## Summary

- Turn clipboard text with at least 1,000 trimmed characters into removable Pasted text chips.
- Send held text in `<pasted_text>` blocks from chat, home, queued turns, active-turn guidance, and code sessions.
- Persist pasted text with composer drafts and preserve it after failed or overlapping submissions.
- Add unit, DOM, and Storybook coverage for the new composer state.

## Why

Long pastes can overwhelm the editable composer and make the typed instruction difficult to find. A chip keeps the instruction editable while preserving the full pasted source in the message sent to the model.

## Validation

- `pnpm exec biome check` on all changed UI files
- 66 focused Vitest tests
- `pnpm exec tsc --noEmit`
- `pnpm storybook:build`
- `git diff --check`


naingthet:
**G2G** at b32ccde0b

Failed send and queue leave pasted chips in the composer until the server accepts the turn, and CodeComposer puts the same chips back if the post throws. A refused first-message handoff keeps the home draft, including held paste. The transcript only receives the wrapped paste on a message that actually sent, which is the payload the model is meant to see.

PR #2599, opened 2026-08-25
fix(code): show the running agent in the workspace sidebar

naingthet:
## What changed

The workspace sidebar now brands each running row from the live session digest, so a stopped Codex session cannot label a running Claude Code session.

## Validation

The UI suite (1,830 tests), focused sidebar, parser, and store tests, TypeScript checking, Biome, the Storybook production build, and `cargo fmt --all --check` pass.

Storybook screenshot inspection could not run because no browser connection was available.

## Compatibility and risk

The server adds optional `harness_kind` data to session digest payloads, so newer desktops show the correct engine while older payloads remain readable.


naingthet:
@shipright review

naingthet:
@shipright review

PR #2602, opened 2026-08-25
feat(desktop): streamline folder access and imports

naingthet:
## What changed

Saved folders now attach automatically to new chats, remain trusted when temporarily unavailable, and support per-chat disconnect or global forget.
MCP and skill imports now preserve safe configuration and nested skill files without retaining secrets, while Codex token failures provide direct recovery steps and avoid duplicate transcript errors.

## Validation

Passed Cargo formatting, the full TypeScript check, Biome lint, all 1,845 UI tests, and the Storybook production build; Rust compilation and tests remain deferred to CI under repository policy, and browser control was unavailable for Storybook screenshot inspection.

## Compatibility and risk

This adds the persisted `host-access/trusted-folders.json` file, an optional `RootSummary.owner` wire field, the `trusted_folder` consent value, and recursive skill staging.

### DIFF
--- a/crates/tidebreak-cli/src/folder.rs
+++ b/crates/tidebreak-cli/src/folder.rs
@@ -593,6 +593,7 @@ fn capability_json(capability: Capability) -> &'static str {
 fn consent_json(method: ConsentMethod) -> &'static str {
     match method {
         ConsentMethod::FolderPicker => "folder_picker",
+        ConsentMethod::TrustedFolder => "trusted_folder",
         ConsentMethod::PermissionDialog => "permission_dialog",
         ConsentMethod::OperatorConfig => "operator_config",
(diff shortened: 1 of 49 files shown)


naingthet:
**G2G** at c5237733e

Trusted-folder auto-attach intersects saved IDs with live broker approvals, so a temporarily unavailable folder stays trusted without being granted to a new chat. Forget disconnects every product attachment then revokes the root and clears the default; disconnect leaves the default in place. Skill import copies validated bytes through nofollow opens, and MCP import keeps environment names while dropping values, literal Authorization headers, and URL credentials.

The first-connect path now saves the folder for future chats and later attaches use TrustedFolder consent instead of a second native prompt, which matches the Folders copy.

PR #2603, opened 2026-08-25
test(code): pin Codex collab threads to a captured fixture

naingthet:
## What changed

#2600 taught the Codex parser to separate a collab subagent's thread from
the parent's, but it landed against hand-written frames. This adds the real
capture behind it and a replay test over that capture.

`collab-agents.ndjson` is a redacted capture from the pinned 0.147.0 binary
with `multi_agent` enabled: the model spawns a subagent, waits on it, and
closes it, so one app-server connection carries two threads.
`collab-agents.expected.json` is the normalized sequence.

`fixture_replay_collab_agents` pins the behavior that the bug report was
about:

- the child's command, its message, and the `WaitAgent` row hang off the
  `Task` span its thread id names, with no correlation table;
- a terminal `agentsStates` repeated by a later collab call settles that
  span once;
- the child's `turn/started` and `turn/completed` neither open nor close the
  parent's turn;
- the child's transient delta does not stream into the parent transcript;
- the parent's turn usage is the parent's counters, not the sum of both
  threads'.

The manifest records what the capture proves about the wire format and what
it leaves untested: `subAgentActivity`, `sendInput`, `resumeAgent`, a collab
call addressing more than one receiver, and a subagent that fails.

## Validation

`rustfmt --check` and `git diff --check` pass. The expected sequence was not
produced by `UPDATE_HARNESS_FIXTURES=1 cargo test` — the repository asks
agents not to compile Rust locally — so it was generated by a script that
mirrors `CodexStreamParser`. That mirror reproduces all eight pre-existing
`fixtures/codex/0.147.0/*.expected.json` files byte for byte and satisfies
both collab unit tests added in #2600, and every assertion the new test
makes was checked against the generated sequence. CI is the compiler here:
a mismatch surfaces as a fixture drift failure rather than a silent pass.

## Compatibility and risk

Fixtures and tests only. No parser, wire-format, or persisted change.


naingthet:
**G2G** at f31dd4baa

The collab-agents capture is protocol traffic (thread ids, tool spans, token counters), not live credentials: `codexHome` is already `/tmp/redacted-codex-home`, cwd is `/workspace`, and nothing in the ndjson looks like a token, cookie, or key. `replay("collab-agents")` still golden-compares the full event sequence, and the new test asserts the Task span, child attribution, single parent turn, dropped child delta, and parent-only usage instead of skipping those checks.

PR #2604, opened 2026-08-25
fix(code): finish merged PR watches cleanly

naingthet:
## What changed

Watch-and-fix now trusts the workspace's fresh write-through pull request digest, so a merged pull request ends its watch instead of reopening from a stale fact snapshot. When a blocked watch resumes, it clears only the `NeedsYou` state created by that blocker and preserves approval or engine-failure attention. Regression tests cover both transitions.

## Validation

- `git diff --check`
- Rust tests were not run locally per repository guidance; CI will run them.

## Compatibility and risk

No persisted or wire-format changes; the change is limited to watch state selection and attention cleanup.

### DIFF
--- a/crates/tidebreak-server/src/code/watch.rs
+++ b/crates/tidebreak-server/src/code/watch.rs
@@ -25,9 +25,9 @@ use tidebreak_core::db::code::{
     list_pull_request_facts_for_repo, list_sessions_for_workspace, save_watch,
 };
 use tidebreak_core::{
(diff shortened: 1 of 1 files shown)


naingthet:
**G2G** at 6ba0f97e6. The watch now classifies from the workspace write-through digest when the live tier is fresh, so a merge the hot refresher already stored ends the watch instead of rebuilding an still-open snapshot from the fact row. Leaving Blocked clears only the matching Structured NeedsYou prompt, so approval and engine-failure attention stay. finish_watch still persists Done and ends the watch session, which drops the row from the active sweep.

PR #2606, opened 2026-08-25
fix(desktop): remember the selected app mode

naingthet:
## What changed

Tidebreak now stores the selected Work or Code mode and restores it before router startup, so reopening the app returns to the last product surface without flashing Work. Explicit deep links still take priority.

## Validation

- `pnpm --dir crates/tidebreak-desktop/ui exec biome lint src/appMode.ts src/code/CodeModeSwitch.tsx src/router.tsx src/main.tsx src/AppShell.dom.test.tsx`
- `pnpm --dir crates/tidebreak-desktop/ui exec vitest run src/AppShell.dom.test.tsx`
- `pnpm --dir crates/tidebreak-desktop/ui build`

## Compatibility and risk

This adds the best-effort `tidebreak.app-mode` local-storage preference; missing, invalid, or unavailable storage falls back to Work, and no wire formats change.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/AppShell.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/AppShell.dom.test.tsx
@@ -9,6 +9,7 @@ import {
 import userEvent from "@testing-library/user-event";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 
+import { APP_MODE_STORAGE_KEY, restoreStoredAppMode } from "./appMode";
 import { useChatListStore } from "./ChatListStore";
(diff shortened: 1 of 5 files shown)


naingthet:
**G2G** at 4b0b21d55

The selected Work or Code mode is written to `tidebreak.app-mode` when the switch changes, and a bare launch restores Code before hash history is created so Work does not flash. Deep links other than `#` / `#/` are left alone. The key is install-scoped localStorage, same as theme and sidebar chrome; tests clear it in `beforeEach`, and another desktop profile or channel uses a different app-data webview, so the preference does not leak across those boundaries.

PR #2608, opened 2026-08-25
fix(code): keep session state indicators consistent

naingthet:
## What changed

- Gate compact `Working` motion on a running lifecycle across workspace cards, watch rows, and agent tabs.
- Remove the duplicate structured-alert glyph, align the accessible workspace label with lifecycle, and add regression tests plus the `Hover Idle Session` Storybook story.

## Validation

- Passed 63 focused tests, the full 1,875-test UI suite, Biome format and lint, TypeScript, the Storybook production build, and diff checks.
- Reviewed the regression story in the supported light and dark themes.

## Compatibility and risk

- None. This changes UI presentation only and does not change persisted data or wire formats.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/AttentionBadge.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/AttentionBadge.dom.test.tsx
@@ -77,6 +77,16 @@ describe("AttentionBadge", () => {
     expect(mark.querySelector(".animate-spin")).not.toBeNull();
   });
 
+  it("renders no mark for Idle", () => {
+    const { container } = render(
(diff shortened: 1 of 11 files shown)


naingthet:
**G2G** at 0191519b9

`attentionMarkForDigest` drops compact Working motion unless lifecycle is also running, which matches `digestStatusTone` and stops a stale Working digest from spinning beside an Idle panel, tab, or watch row. Idle returns no mark, the extra structured-alert glyph is gone, and `workspaceCardLabel` announces the resting lifecycle instead of Working. The unit, DOM, and story coverage pin the disagreement this change is meant to close.

PR #2610, opened 2026-08-25
fix(code): scope and prune clone jobs

naingthet:
## What changed

Clone jobs now carry their owner. Status reads and internal updates require the same owner, so another signed-in user receives the same not-found response as an unknown job. Live updates also publish from the owner stored on the job.

Completed jobs expire after 30 minutes, and the process retains at most 256 completed jobs. Active clone jobs are never pruned.

## Validation

- `cargo fmt -- crates/tidebreak-server/src/code/clone.rs crates/tidebreak-server/src/code/scoped.rs`
- `git diff --check`
- Rust compilation and tests are deferred to CI under the repository rule.

## Compatibility and risk

No wire or persisted format changes. A completed job may return not found after its retention window, after the client has already received its terminal state.

### DIFF
--- a/crates/tidebreak-server/src/code/clone.rs
+++ b/crates/tidebreak-server/src/code/clone.rs
@@ -7,7 +7,7 @@
 use std::path::{Path, PathBuf};
 use std::process::Stdio;
 use std::sync::Mutex;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 
 use tokio::io::{AsyncReadExt, BufReader};
 use tokio::process::Command;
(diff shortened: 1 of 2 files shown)


naingthet:
* **#2610** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2610?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2610?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

naingthet:
**G2G** at b272cca1e

Clone jobs now carry an owner, and both snapshot and apply refuse a mismatched workspace, so one principal cannot read or advance another workspace's job. Prune only drops jobs that are already done with a finished_at stamp, and the cap pass ignores anything still in flight, so a live clone cannot be deleted by expiry or by another workspace's insert or poll. ScopedCode threads the caller owner into get_clone_job, which matches start_clone and the HTTP route.

PR #2615, opened 2026-08-25
fix(code): retire images after publication races

naingthet:
## What changed

Code image publication now reports success for both a new reservation and an exact retry. The scoped server layer preserves that result.

If a session ends after blob storage but before publication, the upload returns `409 session_ended`. The route queues the unreferenced blob through the reference-aware retirement system and wakes the retirement worker.

Focused tests cover successful idempotent publication, the lifecycle race, the typed conflict, the missing authority row, and queued blob retirement.

## Validation

- `rustfmt --edition 2021 --check` on the changed Rust files.
- `git diff --check`.
- Verified that the diff contains only the owned Code image publication paths and focused adjacent tests.
- Per repository guidance, no Cargo build, test, check, or Clippy command ran locally. CI provides compile and test coverage.

## Compatibility and risk

Successful uploads keep the existing `201 Created` response, including exact retries. Only uploads that lose the session lifecycle race change from a false success to a typed conflict.

Retirement remains asynchronous and reference-aware. If identical content still has another live reference, the retirement system keeps the shared blob.

### DIFF
--- a/crates/tidebreak-core/src/db/ops/code/session_image.rs
+++ b/crates/tidebreak-core/src/db/ops/code/session_image.rs
@@ -10,9 +10,7 @@
 //! resolution can check the bytes still match what was reserved rather than
 //! trusting a client to restate it.
 
-use sea_orm::{
-    ColumnTrait, ConnectionTrait, EntityTrait, QueryFilter, Set, TransactionTrait, TryInsertResult,
-};
+use sea_orm::{ColumnTrait, ConnectionTrait, EntityTrait, QueryFilter, Set, TransactionTrait};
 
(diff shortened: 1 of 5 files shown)


naingthet:
* **#2615** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2615?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2615?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #2617, opened 2026-08-25
fix(code): deliver MCP screenshot pixels

naingthet:
## What changed

- Add MCP image content blocks so Code harnesses receive screenshot pixels with the existing text summary.
- Encode every ToolOutput image after validating its media type, byte length, content address, dimensions, image-count limit, and aggregate byte limit.
- Return an isError tools/call result when any declared image cannot be represented, so the bridge never reports screenshot success without pixels.
- Add protocol and server tests that serialize a real tools/call-shaped response, decode each PNG block, and cover missing or inconsistent image data.

## Validation

- `git diff --check`
- Rust compilation and tests were not run locally, as required for this slice. CI owns compile and test validation.

## Compatibility and risk

The MCP result keeps the text block first and adds standard image blocks after it. Text-only results are unchanged. Image results are limited to three images and 1 MiB of decoded bytes so their base64 form stays within the shared JSON-RPC frame budget.

The main risk is compatibility with the MCP implementations embedded by each Code harness. The response uses the standard `type: image`, `data`, and `mimeType` shape, and CI must confirm the Rust integration because local Cargo compilation was intentionally skipped.

### DIFF
--- a/crates/tidebreak-mcp/src/protocol.rs
+++ b/crates/tidebreak-mcp/src/protocol.rs
@@ -234,12 +234,20 @@ pub struct CallToolResult {
     pub is_error: bool,
 }
 
-/// A content block in a tool result. Only text today.
+/// A content block in a tool result.
 #[derive(Debug, Clone, Serialize)]
 #[serde(tag = "type")]
 pub enum Content {
     #[serde(rename = "text")]
     Text { text: String },
+    #[serde(rename = "image")]
+    Image {
+        /// Standard base64 of the image bytes.
+        data: String,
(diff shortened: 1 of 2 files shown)


naingthet:
* **#2617** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2617?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2617?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #2635, opened 2026-08-26
revert(deps): drop TypeScript 7 until notices stay host-independent

naingthet:
Reverts #2627. TypeScript 7 makes `pnpm licenses list --json --prod` emit `typescript` plus a host-specific `@typescript/typescript-*` optional package, so `legal/THIRD-PARTY-NOTICES.md` cannot be regenerated in a host-independent way. The notices generator treats that as a stop, which poisons every later `notices --check` even though desktop UI itself was green. Drop TypeScript 7 until the production license graph is the same on macOS and Linux.

### DIFF
--- a/crates/tidebreak-desktop/ui/package.json
+++ b/crates/tidebreak-desktop/ui/package.json
@@ -115,7 +115,7 @@
     "jsdom": "30.0.1",
     "storybook": "10.5.10",
     "tailwindcss": "4.3.3",
-    "typescript": "7.0.2",
+    "typescript": "5.8.3",
     "vite": "7.3.6",
     "vitest": "4.1.11"
   }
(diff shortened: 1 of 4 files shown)


naingthet:
**G2G** at ab88b8a02

The desktop UI pin is back on typescript 5.8.3 in package.json, and the lockfile importer, typescript@5.8.3 snapshot, and every peer key that used to name 7.0.2 now name 5.8.3. The TS 7 native optional packages (@typescript/typescript-*@7.0.2) are gone, so a fresh install cannot still pull 7. The Blob constructor drop of Uint8Array<ArrayBuffer> and the tsconfig baseUrl restore are the TS 5 companions of that pin, not leftover 7 surface.

PR #2637, opened 2026-08-26
fix(code): qualify pull request stack identity

naingthet:
## Implementation

- Fetch each pull request head repository and owner with the existing delivery query.
- Resolve stack parents by base repository, head repository, and branch instead of branch name alone.
- Keep incomplete, missing, and ambiguous parent identities unresolved.
- Preserve the durable-fact compatibility path while refusing ambiguous branch matches.
- Cover same-named fork branches, repository isolation, renamed or deleted branches, duplicate candidates, incomplete host data, closed parents, and self-edges.

## Compatibility

The public delivery response keeps the existing `stack_parent_number` field. Existing durable facts remain readable and do not require a migration. Pull requests with complete host identity keep resolving to the same parent when the branch match is unique.

## Safety and risk

The change fails closed when GitHub omits fork identity or when more than one parent can match. This can leave a stack edge unresolved instead of attaching it to the wrong pull request. The main risk is reduced stack annotation for incomplete legacy observations; the durable compatibility lookup still handles unique branch names.

## Validation

- `git diff --check`

Per repository guidance, Rust compilation and tests are left to continuous integration.

### DIFF
--- a/crates/tidebreak-server/src/code/delivery.rs
+++ b/crates/tidebreak-server/src/code/delivery.rs
@@ -22,12 +22,16 @@ use tidebreak_core::db::code::{
 };
 use tidebreak_core::{
     CodePullRequestAttribution, CodePullRequestDiscovery, CodePullRequestId,
-    CodePullRequestRelation, CodeRepo, CodeWorkspace, CodeWorkspaceStatus, OwnerId,
-    PullRequestCheck, PullRequestCheckBucket, PullRequestComment, PullRequestCommentKind,
+    CodePullRequestRelation, CodePullRequestState, CodeRepo, CodeWorkspace, CodeWorkspaceStatus,
+    OwnerId, PullRequestCheck, PullRequestCheckBucket, PullRequestComment, PullRequestCommentKind,
     PullRequestDigest, RepoId, WorkspaceId,
 };
 
 use super::gh::{self, GhObservation};
+use super::reconcile::{
+    StackParentCandidate, StackParentEdge, StackParentIndex, StackParentResolution,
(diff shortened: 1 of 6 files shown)


naingthet:
**NO GO** at 75ae8917f

1. P1: `stack_parents_by_head` still keys only on unique `head_branch` inside one base repository. Watch (`stacked_parent_number`) and the trigger live-tier path (`claim_fires_from_row`) use that map, so a tracked PR whose head is `acme:feature` is treated as the parent of any child based on `feature` even when the child is stacked on `other:feature`. Duplicate facts now drop, but a missing second fork still parks a watch and holds Behind/ReviewRequired on the wrong pull request. Durable facts have no head repository; that path has to stay unresolved instead of guessing.

naingthet:
**G2G** at a7b37d6fc

`stack_parents_by_head` is gone. Watch no longer parks from a branch-only fact map, and the trigger live-tier path no longer holds Behind or ReviewRequired from one. Stack annotation now keys on fork-qualified head identity and leaves durable facts without a head repository unresolved, so a child based on `other:feature` is not attached to `acme:feature`. Same-named forks on one host page still pick the matching parent.

naingthet:
**G2G** at f33315a18

`stack_parents_by_head` is still gone. Watch does not park from a branch-only fact map, and the trigger live-tier path does not hold Behind or ReviewRequired from one. Stack annotation keys on fork-qualified head identity and leaves durable facts without a head repository unresolved, so a child based on `other:feature` is not attached to `acme:feature`. Same-named forks on one host page still pick the matching parent. This head only drops an unused test helper.

PR #2638, opened 2026-08-26
fix(code): bind approvals to exact engine requests

naingthet:
## What changed

- Bind each persisted approval to the exact owner, session, turn, worker epoch, native call, request digest, and server-issued capability that created it.
- Claim pending approvals atomically before native delivery, then settle the approval and journal event in one database transaction.
- Distinguish definite delivery failures from ambiguous lost acknowledgements so Tidebreak never records a decision that may not match the engine state.
- Recover interrupted workers by abandoning pending approvals, interrupting the running turn, updating session state, and appending recovery events in one transaction.
- Scope Claude permission-prompt parking by session and worker epoch so repeated engine call IDs cannot cross sessions or replacement workers.

## Backwards compatibility

The migration adds nullable approval-binding columns. Existing pending approvals are abandoned during migration because their native waiters cannot be proven live after upgrade. Existing terminal approval history remains readable. Harness references without a server capability still parse for providers that do not use the Claude permission bridge.

## Safety and risk

This change touches approval delivery, restart recovery, and persisted journal ordering. It fails closed when a worker, native waiter, capability, request digest, or claim no longer matches. Ambiguous native delivery remains claimed until restart recovery instead of being recorded as denied or approved.

## Validation

- `git diff --check`
- Added SQLite and PostgreSQL ownership, transaction rollback, claim-race, restart recovery, stale-worker, and delivery-acknowledgement tests.

Rust compilation and tests run in CI per repository policy.


naingthet:
**G2G** at e1b66799e

Claude parks and completes by a server-issued capability, not a bare engine call id, so a reused tool_use_id cannot land on another session, worker, or request. Insert, claim, and settle all fence on the session lock plus worker epoch, and a missing or half-present binding refuses the decision instead of sending it. Restart and worker replacement abandon both claimed and unclaimed rows, so a stale approval cannot grant a later command.

naingthet:
**G2G** at e3b129103

Claude still parks and completes by a server-issued capability, not a bare engine call id, so a reused tool_use_id cannot land on another session, worker, or request. Insert, claim, and settle still fence on the session lock plus worker epoch, and a missing or half-present binding still refuses the decision instead of sending it. Restart and worker replacement still abandon claimed and unclaimed rows, so a stale approval cannot grant a later command. The new commit only bundles the claimed settlement into one value and does not change those checks.

naingthet:
**G2G** at 06269bfb5

Claude still parks and completes by a server-issued capability, not a bare engine call id, so a reused tool_use_id cannot land on another session, worker, or request. Insert, claim, and settle still fence on the session lock plus worker epoch, and a missing or half-present binding still refuses the decision instead of sending it. Restart and worker replacement still abandon claimed and unclaimed rows, so a stale approval cannot grant a later command. The new commit only turns approvals on in the stale-worker fixture and checks that session create succeeds; it does not change those checks.

PR #2639, opened 2026-08-26
test(desktop): stabilize pairing nudge coverage

naingthet:
## What changed

- Reset the pairing nudge test hook after each ManagedGate test.
- Wait until the pairing listener is attached before firing the nudge in the policy-refresh test.

## Backwards compatibility

This changes test synchronization only. Production code and behavior are unchanged.

## Safety and risk

The test now waits on the observable listener registration instead of racing the effect that installs it. The hook reset prevents one test from leaking a callback into the next test.

## Validation

- `pnpm --dir crates/tidebreak-desktop/ui exec vitest run src/ManagedGate.dom.test.tsx`
- `pnpm --dir crates/tidebreak-desktop/ui exec biome check src/ManagedGate.dom.test.tsx`
- `git diff --check`

### DIFF
--- a/crates/tidebreak-desktop/ui/src/ManagedGate.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/ManagedGate.dom.test.tsx
@@ -84,6 +84,7 @@ function HostedGatewayProbe() {
 
 afterEach(() => {
   cleanup();
+  pairingNudge.fire = null;
   vi.clearAllMocks();
   vi.useRealTimers();
   vi.unstubAllGlobals();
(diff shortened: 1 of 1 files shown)


naingthet:
**G2G** at 29274a67b

The extra wait still requires the pairing listener to attach before the nudge fires, so a missing handler still fails the test instead of becoming a no-op. Resetting `pairingNudge.fire` after each case also stops a leftover handler from making that wait pass on the wrong mount. Coverage of the no-timer pairing path is unchanged and does not stub auth open.

PR #2640, opened 2026-08-26
fix(code): reconcile replay turn state

naingthet:
## Implementation

- Preserve the capped-replay marker when the controller compacts adjacent replay deltas.
- Track journal-attributed turns and live start timing separately from durable snapshot activity.
- Retain unresolved replay terminals by event sequence, refresh turn snapshots with bounded-delay retries, and restore terminal status, duration, usage, errors, and transcript order when the durable turn appears.
- Ignore stale refresh activity when a newer live turn has started.

## Compatibility

This change does not alter the API, event protocol, or stored data. Live turns still use observed timing. Replayed turns use durable server timestamps when available, and keep duration unknown until an authoritative snapshot resolves it.

## Safety and risk

The change stays inside the in-memory code-session controller, reducer, and registry. It does not assign capped terminal rows from snapshot activity alone. The main risk is incorrect correlation across truncated replay windows, which the focused reducer and registry cases cover for stale snapshots, missing initial hydration, retries, reconnects, multiple turn order, and newer live turns.

## Validation

- `pnpm exec vitest run src/code/CodeSessionController.test.ts src/code/CodeSessionReducer.test.ts src/code/CodeSessionRegistry.test.ts src/code/CodeSessionStore.test.ts src/code/CodeSessionSend.test.ts` (84 tests)
- `pnpm exec biome check src/code/CodeSessionController.ts src/code/CodeSessionController.test.ts src/code/CodeSessionReducer.ts src/code/CodeSessionReducer.test.ts src/code/CodeSessionRegistry.ts src/code/CodeSessionRegistry.test.ts`
- `pnpm exec tsc --noEmit`
- `git diff --check`

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeSessionController.test.ts
+++ b/crates/tidebreak-desktop/ui/src/code/CodeSessionController.test.ts
@@ -1,6 +1,10 @@
 import { afterEach, describe, expect, it, vi } from "vitest";
 import type { SequencedCodeEventFrame } from "../api/types";
-import { CodeSessionController } from "./CodeSessionController";
+import {
+  CodeSessionController,
+  INITIAL_RECONNECT_DELAY_MS,
+  MAX_RECONNECT_DELAY_MS,
+} from "./CodeSessionController";
 
 class FakeSocket {
   onopen: (() => void) | null = null;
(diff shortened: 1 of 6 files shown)


naingthet:
**NO GO** at a298c2406

1. P1: A live turn from `applyAcceptedTurn` is invisible to the journal cursor, so stale reconciliation can still settle it. `submitAcceptedTurn` sets `activeTurnId` without moving `lastSeq`. `reconcilePendingCodeTurns` then treats matching `observedSeq` as proof the journal is unchanged and runs `applyAuthoritativeTurnActivity`, which idles the session or installs an older running snapshot turn. Replayed `turn_started` / terminal frames have the same hole: they overwrite and then clear `activeTurnId` without checking for a newer accepted turn. The newer-live-turn tests only cover a journal `turn_started` that advanced `lastSeq`.
2. P2: Unattributed capped terminals do not fail closed on missing identity. When `turnId` is null and no snapshot turn is running, `assignUnattributedTerminals` binds the group to the tail of the list; `candidateTurnId` is only a membership check. A later completed turn can take an earlier terminal's status, usage, and error.

naingthet:
**G2G** at 503dd9d22

`applyAcceptedTurn` now raises `turnActivityRevision` and plants `acceptedTurnFence` even though `lastSeq` stays put, and `reconcilePendingCodeTurns` will not run `applyAuthoritativeTurnActivity` unless both the journal cursor and that revision still match. Replayed `turn_started` and terminal frames keep a newer accepted turn; `applyAuthoritativeTurnActivity` also refuses to idle or replace a fenced turn that is missing or still running. Unattributed capped terminals stay unbound until a following `turn_started` anchors them, so they no longer take the snapshot tail.

PR #2636, opened 2026-08-26
ci: keep Windows prepared archives on local tar paths

naingthet:
## What changed

Windows release prepare jobs compiled successfully, then failed in Archive prepared Windows inputs with `tar: Cannot connect to C: resolve failed` (and `D:` on x86_64). GNU tar treats a drive letter in `-f C:\...\prepared-windows.tar` as a remote host.

The archive and extract steps now operate on `prepared-windows.tar` from the current directory and pass the prepared tree with `-C`.

Seen on https://github.com/brightwave-inc/tidebreak/actions/runs/32916482746.

## Validation

- `node --test scripts/workflow-security.test.mjs scripts/workflow-security-mutations.test.mjs`
- Windows release job assertions passed.
- Docker context cases in the same file failed locally because the Docker daemon was not running.
- Did not re-run the production release workflow.

## Compatibility and risk

Windows Git Bash only. Archive contents and checksum contract are unchanged. Re-dispatch the desktop release after this lands.

PR #2642, opened 2026-08-26
ci: give Windows release tar POSIX paths

naingthet:
## What changed

#2636 stopped GNU tar from treating `C:\path` as a remote host when writing `prepared-windows.tar`. Extract still failed: `tar -C C:\a\_temp\...` treats backslashes as escapes, so the directory `mkdir` created is not the one tar opens.

Seen on https://github.com/brightwave-inc/tidebreak/actions/runs/32924321644 after a successful Windows prepare.

Archive and extract now convert `runner.temp` paths with `cygpath -u` before calling tar.

## Validation

- `node --test --test-name-pattern='Windows release jobs' scripts/workflow-security.test.mjs`
- Did not re-run the production release. Re-dispatch after this lands.

## Compatibility and risk

Windows Git Bash only. Archive members and checksums are unchanged.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1357,6 +1357,11 @@ jobs:
           RELEASE_CONFIG: ${{ runner.temp }}/tidebreak-release.json
           RELEASE_TARGET: ${{ matrix.target }}
         run: |
+          # GNU tar treats C:\path as a remote host and \ as escapes. Stay on
+          # POSIX paths for every tar argument.
+          PREPARED_ROOT="$(cygpath -u "$PREPARED_ROOT")"
+          RELEASE_CONFIG="$(cygpath -u "$RELEASE_CONFIG")"
+          workspace="$(cygpath -u "$GITHUB_WORKSPACE")"
           app_binary="target/$RELEASE_TARGET/release/tidebreak-desktop.exe"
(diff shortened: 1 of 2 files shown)


naingthet:
* **#2642** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2642?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2642?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

naingthet:
**G2G** at 494ef8934

`cygpath -u` rewrites `runner.temp` and the workspace before every tar argument, so GNU tar never sees `C:\` as a remote host or `\` as escapes. Archive members stay the same relative names (`SHA256SUMS`, `release-config.json`, `target`, `crates`) with a relative `prepared-windows.tar` after `cd` to the POSIX workspace, and extract still uses `-C` into the converted temp root. The workflow-security test pins those `cygpath` assignments and the `tar -cf`/`tar -xf` lines, and still forbids `tar -[cx]f "$` so the archive path cannot become an absolute Windows path.

PR #2643, opened 2026-08-26
fix(code): make terminal input recoverable

naingthet:
## Implementation

- Serialize terminal writes so concurrent input cannot reorder chunks.
- Retain unsent input after a write failure and pause new input until the operator retries, reconnects, or discards it.
- Keep terminal identity changes, delayed reads, retries, and unmounts fenced from stale handlers.
- Show recovery states for write, attach, and read failures, with matching Storybook coverage.

## Compatibility

The terminal client contract and existing component props remain unchanged. Successful terminal sessions keep the same behavior.

## Safety and risk

Retrying a failed write can repeat input if the server received the request before the connection failed. The recovery message states this risk and offers reconnect and discard paths. Workspace and terminal changes abandon retained input so bytes cannot cross shell boundaries.

## Validation

- `pnpm --dir crates/tidebreak-desktop/ui exec vitest run src/code/TerminalPane.dom.test.tsx` — 17 tests passed
- `pnpm --dir crates/tidebreak-desktop/ui exec biome check src/code/TerminalPane.tsx src/code/TerminalPane.dom.test.tsx src/stories/TerminalPane.stories.tsx` — passed
- `pnpm --dir crates/tidebreak-desktop/ui storybook:build` — passed
- `git diff --check origin/main...HEAD` — passed
- Captured and inspected all three changed stories in light, dark, high-contrast light, and high-contrast dark themes at desktop and narrow widths

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/TerminalPane.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/TerminalPane.dom.test.tsx
@@ -1,4 +1,5 @@
 // @vitest-environment jsdom
+import type { ComponentProps } from "react";
 import {
   act,
   cleanup,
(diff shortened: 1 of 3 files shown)


naingthet:
**G2G** at 7d030a7ce
Failed writes keep the unsent chunks on the current writer, pause stdin, and retry the same token so the next send is not bound to a completed request. Reconnect, workspace change, and unmount cancel that writer and drop its queue, so later input cannot land on a stale shell. Delayed reads and xterm `onData` are fenced by identity epoch, which matches the recoverability bar.

PR #2650, opened 2026-08-26
fix(code): relay hosted engine inference through the caller's grant

### LINKED ISSUE #2651: fix(code): hosted engine children launch without inference credentials
Found running the #2613 smoke test, step 5. On the hosted machine, a Codex code session failed its first turn with a 401 from the vendor endpoint: \`Missing bearer or basic authentication in header\`, after nine error-level websocket reconnect notices.

Structural cause: the server spawns engine children with empty \`extra_argv\`/\`extra_env\`, and a gateway-authenticated hosted machine has no provider credentials anywhere the child can find them — no keys in the image, no harness logins in the container, and decision 51's on-behalf-of exchange lives in the server process, never in the child. Every hosted harness session is dead on arrival; only server-side chat inference works.

Fix: give each session a relay on the server's own listener that exchanges the owning caller's live machine-bound token per request and streams through to the Model Gateway's compat endpoints, with a session-scoped key in the child's environment. Decision 71 records the shape.

Claude Code and Codex are wired in #2650. Opencode and Grok are the remainder (separate issue).

naingthet:
## What

Hosted code sessions failed their first turn with a 401: the engine child (Claude Code, Codex) makes its own inference requests, and a gateway-authenticated hosted machine has no provider credentials to give it — no keys in the image, no harness logins, and the on-behalf-of exchange lives in the server process. A hosted Codex session fell back to `api.openai.com`, logged a reconnect notice per websocket attempt, and died with `Missing bearer or basic authentication in header`.

This PR gives each session a private inference path (decision 71, included):

- Two routes on the main listener, `/code/llm/anthropic/v1/messages` and `/code/llm/openai/v1/responses`, stream requests through to the Model Gateway's compat endpoints. Per request, the relay exchanges the owning caller's live machine-bound token for a fresh inference token (the same single-flight cache chat uses) and sends that upstream. The session outlives any single short-lived token.
- At spawn on a machine with an on-behalf-of gateway, the runtime mints an opaque `tbreak_hl_` key mapped to (owner, session) and wires the child at the relay: Claude Code through `ANTHROPIC_BASE_URL`/`ANTHROPIC_AUTH_TOKEN`, Codex through a custom `tidebreak` model provider on the command line. The custom provider also keeps Codex off its vendor-only websocket transport, which removes the reconnect noise.
- Keys follow the browser channel's lifecycle exactly: routes registered outside `require_token` with per-request bearer auth, reissue-on-attach replaces the prior key, and every path that revokes the browser channel revokes the relay key.
- Machines without an on-behalf-of gateway (desktop, static-token self-host) spawn children with empty extra argv/env exactly as before.

Refusals are vendor-shaped so engines report them legibly: 401 `authentication_error` for a missing/revoked key or a caller whose gateway session is gone; 502 `api_error` when the gateway does not answer, so the engine's own retry policy applies.

## Validation

- Unit tests pin the per-engine spawn wiring, the key lifecycle, and forwarding through a fake gateway (exchanged token replaces the relay key; query and body pass through; the child's credentials never reach upstream; unknown key / missing caller token / unreachable gateway map to 401/401/502).
- Both wirings were exercised against the real engine binaries before writing the server code: Codex 0.147.0 with the custom provider makes a single `POST .../responses` with the key as bearer and never touches its websocket transport; Claude Code with the base-URL override posts `/v1/messages?beta=true` with the bearer and no `x-api-key`.

## Notes

- A Codex thread started before this change recorded the default provider in its rollout, so resuming it may still fail; a new session is the path.
- Opencode and Grok sessions on hosted machines still launch without credentials; they need equivalent wiring (follow-up issue to be filed).

Fixes #2651.

### DIFF
--- a/crates/tidebreak-server/src/code/harness_llm.rs
+++ b/crates/tidebreak-server/src/code/harness_llm.rs
@@ -1,0 +1,563 @@
+//! Session-scoped inference relay for engine children (decision 71).
+//!
+//! A gateway-authenticated hosted machine holds no provider credentials an
+//! engine child could use: the image carries no `ANTHROPIC_API_KEY`, no
+//! Codex login, and decision 51's on-behalf-of exchange lives in this
+//! process, never in the child. Left alone, a child falls back to its
+//! vendor endpoint and fails the first turn with a 401.
+//!
+//! This module gives each session a private inference path instead. At
+//! spawn, the runtime mints an opaque relay key and points the child's own
+//! HTTP client at `{loopback_base}/code/llm/...` on this server's listener
+//! ([`spawn_wiring`]). Each relayed request presents that key; the relay
+//! maps it back to the owning caller, exchanges the caller's live
+//! machine-bound token for a fresh inference token, and streams the request
+//! through to the gateway's compat endpoint with that token in place of the
+//! key. Because the exchange runs per upstream request, a session outlives
+//! any single short-lived token.
+//!
+//! Security properties mirror the browser channel
+//! ([`super::browser_channel`]):
+//!
+//! * Keys are random v4 UUIDs, never derived from owner or session ids.
+//! * A key authorizes exactly the relay routes — it is not the app token
+//!   and opens nothing else on the listener.
+//! * Reissue for the same session replaces the prior key, and revocation
+//!   follows the browser channel's lifecycle.
+//! * The caller's gateway tokens never reach the child. The relay key is
+//!   the only secret in the child's environment, and it is useless off
+//!   this machine.
+
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex};
+use std::time::Duration;
+
+use axum::http::header::AUTHORIZATION;
+use axum::http::{HeaderMap, HeaderName, StatusCode};
+use axum::response::Response;
+
+use tidebreak_core::{AgentError, CodeSessionId, HarnessKind, OwnerId};
+
+use crate::obo_gateway::OboGateway;
+
+/// Whose inference a relay key spends.
+#[derive(Clone)]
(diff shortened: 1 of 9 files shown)


naingthet:
**NO GO** at 0d96a8bcd

1. P1: `with_registry_and_browser_runtime` still constructs `CodeRuntime` without `harness_llm`. The struct now requires that field, so test and `scripted-harness` builds cannot compile. Production `new` / `bind_inner` set it; this helper does not. Set `harness_llm: None` there (same as `git_credentials`) so the type is consistent.

naingthet:
**G2G** at 01297ff67

`with_registry_and_browser_runtime` now sets `harness_llm: None`, so the prior compile break is gone. Production `CodeRuntime::new` still builds the relay from the same on-behalf-of handle as git lending. Each attach mints a session key bound to `session.owner`; `forward` exchanges only that owner via `bearer_for`, replaces the child credential, and returns 401 when the key is missing/revoked or the caller has no live token. Revoke follows the browser-channel paths, so a key cannot outlive the session or spend another caller's grant.

naingthet:
**G2G** at c1049bfdf

This head only teaches the owner-scoped route lint that `llm.rs` is a capability-bearer path (`HarnessLlmRelay` + `HeaderMap` + `relay.forward`). Production still builds the relay from the same on-behalf-of gateway as git lending. Attach mints a session key bound to `session.owner`; `forward` exchanges only that owner through `bearer_for`, strips the child credential, and returns 401 on a missing/revoked key or a caller with no live token. `with_registry_and_browser_runtime` still sets `harness_llm: None`. Revoke still follows the browser-channel paths, so a key cannot outlive the session or spend another caller's grant.

PR #2667, opened 2026-08-26
fix(code): recover browser state transitions

### LINKED ISSUE #2659: fix(browser): reconcile same-document navigation and native browser recovery
The Code mode audit found that same-document navigation leaves native URL state stale, Reload hides a failed native creation without retrying, inspect state can diverge from native injection, and a zero-sized first mount never creates when bounds become valid. Add a validated same-document navigation observer without advancing document epoch; route failed Reload and first nonzero bounds through one create-and-reconcile path; and commit inspect state only after native success while keeping overlays synchronized. Covers BROWSER-3, BROWSER-4, and the inspect/zero-size parts of BROWSER-5. The screenshot publication and durable close sub-findings remain separate because they use different persistence paths.

naingthet:
Fixes #2659

## Implementation

- Observe validated same-document navigation from `pushState`, `replaceState`, `popstate`, and `hashchange`, then update the registry, toolbar, persistence, list output, and `url_changed` waits without advancing the document epoch.
- Retry native browser creation through the shared create-and-reconcile path when Reload follows a creation failure. Keep the failure visible until creation succeeds.
- Commit inspect state only after native injection or removal succeeds. Keep overlays aligned after scroll, resize, fullscreen changes, and bounded DOM mutations.
- Let the first nonzero `ResizeObserver` measurement create the native browser exactly once.
- Add browser fixture, Rust, UI, and Storybook coverage for each recovery sequence.

## Compatibility

The host event addition is additive. Full-document navigation keeps the existing epoch and semantic-target retirement behavior. Existing browser sessions and persisted history remain compatible.

## Safety

The injected observer validates same-document URLs before emitting them. Native creation is serialized to prevent duplicate views. Mutation-driven overlay refreshes are bounded, and failed inspect operations leave the last confirmed state active.

## Validation

- Browser fixture: 15 tests passed.
- Browser UI suite: 88 tests passed.
- Focused browser Vitest suite: 40 tests passed.
- TypeScript `--noEmit` passed.
- Focused Biome format and lint checks passed.
- Storybook build passed.
- Direct `rustfmt --check` passed.
- `git diff --check` passed.
- Storybook states were captured and inspected at 1200px and 390px in supported light and dark themes.

No local Cargo build, check, Clippy, or test command was run.

### DIFF
--- a/crates/tidebreak-desktop/src/browser_control.rs
+++ b/crates/tidebreak-desktop/src/browser_control.rs
@@ -1374,6 +1374,34 @@ impl BrowserRegistry {
         })
     }
 
+    /// Record a validated same-document URL change without advancing the
+    /// document epoch. The instance and epoch fence prevents a late observer
+    /// from writing into a replacement view or a later document.
+    pub(crate) fn same_document_navigation(
+        &self,
+        browser_id: &str,
+        workspace_id: &str,
+        instance_id: u64,
+        document_epoch: u64,
+        url: String,
+    ) -> Option<BrowserSnapshot> {
+        let mut state = self.lock();
+        {
+            let record = state.records.get_mut(browser_id)?;
+            if record.workspace_id != workspace_id
+                || record.instance_id != instance_id
+                || record.document_epoch != document_epoch
+                || record.url.as_deref() == Some(url.as_str())
+            {
+                return None;
+            }
+            record.url = Some(url);
+            record.load_state = BrowserLoadState::Ready;
+        }
+        let record = state.records.get(browser_id)?;
+        Some(record.snapshot(browser_id, agent_access_for_record(&state, record)))
+    }
+
     pub(crate) fn title_changed(
         &self,
         browser_id: &str,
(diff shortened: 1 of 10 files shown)


naingthet:
* **#2667** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2667?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2667?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

naingthet:
**NO GO** at dd55513a7

1. P1: `createAndReconcileNative` joins an in-flight create and then returns `nativeReady` without sending the caller's URL. `navigate()` already commits the session to the new address before that join, so a second address entered while the first create is outstanding never reaches the host. The finishing `navigation_finished` from the stale create then overwrites the session. After an in-flight create settles, navigate to the caller's URL or start a new create; do not inherit another request's address.
2. P2: `start_same_document_navigation_observer` returns on the first install or read `Err`, so a transient eval failure permanently stops same-document updates for that document epoch. Reinstall on error the same way you already reinstall on `Ok(None)`, or the address bar, persistence, and `url_changed` waits stay stale until a full navigation.

naingthet:
**G2G** at 4c65534c8
The prior findings are fixed. A later navigate during create now writes `nativeCreateRequestedUrl` and the in-flight create drains it with `navigate` after the view exists, so the caller URL reaches the host; `nativeExpectedNavigationUrl` also drops the stale create `navigation_finished` so it cannot overwrite the session. A same-document observer eval error now clears the install flag and retries on the next tick instead of returning, so one failed read or install cannot kill updates for the rest of that epoch.

PR #2668, opened 2026-08-26
fix(code): make checkpoint file identity byte-exact

### LINKED ISSUE #2661: fix(code): make checkpoint file identity byte-exact
The Code mode audit found two linked checkpoint diff defects. Non-UTF-8 paths are converted with `String::from_utf8_lossy`, so distinct Git paths can collapse to the same wire value. Per-file diff also passes the requested file as a Git pathspec, so pathspec magic can select multiple files, and its line counts come from the capped workspace file list.

Carry path bytes internally and encode them reversibly at the wire boundary. Require the exact representation for file-specific requests, use literal pathspec semantics, and compute selected-file statistics independently of the workspace list cap.

Covers DATA-9 and DATA-10 from the Code mode audit.

naingthet:
## What changed

- Keep Git checkpoint paths as bytes until the workspace files route serializes them.
- Encode non-UTF-8 paths and reserved-prefix paths as canonical URL-safe base64 strings without changing the wire shape.
- Decode the exact returned file identifier for file-specific diffs and pass it to Git with literal pathspec semantics.
- Compute selected-file line counts independently of the capped workspace file list.
- Preserve leading and trailing path whitespace in the workspace diff query.
- Add regressions for non-UTF-8 collisions, reserved-prefix collisions, pathspec magic, and selected files beyond the list cap.

## Validation

- `git diff --check`
- Git-only literal pathspec fixture

Local Cargo build, check, Clippy, and tests were not run, as requested.

Fixes #2661

### DIFF
--- a/crates/tidebreak-server/src/code/checkpoint.rs
+++ b/crates/tidebreak-server/src/code/checkpoint.rs
@@ -25,11 +25,14 @@
 //! Diffs are produced here, bounded in bytes and file count, with truncation
 //! marked on the payload. The renderer never runs git.
 
+use std::ffi::OsString;
 use std::path::{Path, PathBuf};
 use std::process::Stdio;
 use std::sync::Arc;
 use std::time::Duration;
 
+use base64::engine::general_purpose::URL_SAFE_NO_PAD;
+use base64::Engine as _;
 use tokio::process::Command;
 use tokio::time::timeout;
 use tracing::warn;
@@ -53,6 +56,7 @@ pub(crate) const MAX_DIFF_BYTES: usize = 256 * 1024;
 pub(crate) const MAX_DIFF_FILES: usize = 64;
 
 const REF_PREFIX: &str = "refs/tidebreak/checkpoints";
+const GIT_PATH_WIRE_PREFIX: &str = "tidebreak-path:v1:";
 
 /// Ordinal of a session's start baseline, one below its first turn.
 ///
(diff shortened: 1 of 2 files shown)


naingthet:
* **#2668** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2668?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2668?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

naingthet:
**G2G** at 6abb69afd
GitPath keeps checkpoint file identity as raw bytes until the workspace files route serializes it. Non-UTF-8 and reserved-prefix paths go out as canonical `tidebreak-path:v1:` URL-safe base64, and `from_wire` rejects invalid or non-canonical identifiers instead of mapping them onto another file. File diffs decode that exact identifier and pass it to Git with `--literal-pathspecs`, and selected-file line counts no longer come from the capped workspace list.

naingthet:
**G2G** at 05ab8fbe3
The new commit only disambiguates the non-UTF-8 numstat fixture (`\x00` before the next record). GitPath still keeps checkpoint paths as raw bytes until the workspace files route serializes them; non-UTF-8 and reserved-prefix paths still use canonical `tidebreak-path:v1:` URL-safe base64; `from_wire` still rejects invalid or non-canonical identifiers instead of mapping them onto another file; file diffs still decode that exact identifier and pass it to Git with `--literal-pathspecs`.

PR #2663, opened 2026-08-26
fix(code): bind workspace merges to exact heads

naingthet:
## What changed

- The desktop sends the exact host, repository owner and name, pull request number, displayed head SHA, merge method, and auto-merge choice.
- The server holds the workspace action lock while it rechecks the local branch, upstream, dirty state, unpushed commits, repository identity, pull request identity and state, and local and hosted heads.
- The final merge uses the repository-qualified helper with the exact pull request number and head commit.
- Typed conflicts cover changed worktree and pull request preconditions. The desktop keeps the conflict popover open, shows a clear error, and lets you refresh.
- Commit, push, ready, pull request creation, and quick actions now share the same workspace lock.

## Compatibility

The request and response types change together in the server and generated desktop client. Mixed-version clients fail safely because the server requires the exact merge identity instead of falling back to an unverified target. Existing repositories and pull requests require no migration.

## Safety and risk

The server rejects stale or mismatched state before it starts a merge. The shared action lock closes races with other workspace mutations, and the exact-head merge call prevents a changed pull request from landing. The stricter checks can reject a valid request after the branch or pull request changes; refresh and retry with the current state.

## Validation

- `cargo fmt --all -- --check`
- `git diff --check origin/main...HEAD`
- Focused Biome checks for every changed TypeScript and Storybook file
- 176 focused Vitest tests across five files
- `pnpm --dir crates/tidebreak-desktop/ui storybook:build`
- Storybook screenshots inspected in light, dark, high-contrast light, and high-contrast dark at 900×520 and 390×520 with 2× device scale

Per repository policy, no local Cargo build, check, Clippy, or tests ran.

Fixes #2656

PR #2674, opened 2026-08-26
fix(code): make owner paths collision-resistant

### LINKED ISSUE #2665: fix(code): make owner filesystem namespaces collision-resistant
Code mode derives managed owner directories from a lossy sanitized owner string. Distinct owners such as `user:alice@example` and `user:alice_example` can map to the same path, and case-only identifiers collide on common macOS and Windows filesystems.

Derive the filesystem segment from a collision-resistant representation of the exact owner key. Preserve or migrate existing managed paths safely so the change does not strand existing repositories or workspaces. Add punctuation, case-folding, and stable-encoding regression tests.

Covers SEC-6 from the Code mode audit.

naingthet:
## What changed

- Derive each non-local managed path namespace from a versioned, lowercase SHA-256 digest of the exact owner key.
- Keep the local-owner directory layout unchanged.
- Preserve existing managed repositories and worktrees through their stored absolute paths. A legacy clone target counts as occupied only when an owner-scoped live repository row proves that it belongs to the exact caller.
- Add regression coverage for punctuation collisions, case-only identifiers, stable encoding, colliding legacy owners, legacy path access, and later worktree creation under the new namespace.

## Validation

- Stable owner digest fixture checked with `shasum -a 256`
- Protected-file overlap check against open PR #2663
- `git diff --check`

Cargo build, check, Clippy, and tests were not run, as required by the issue instructions.

## Compatibility and risk

No database migration moves existing paths. Repository and workspace rows already store absolute paths, so existing checkouts remain in place and accessible. Only later clones and worktrees use the new namespace. The compatibility lookup requires an exact owner-scoped row and does not trust a colliding legacy directory by existence alone.

## Tracking

Closes #2665.

### DIFF
--- a/crates/tidebreak-server/src/code/clone.rs
+++ b/crates/tidebreak-server/src/code/clone.rs
@@ -14,6 +14,7 @@ use tokio::process::Command;
 use tokio::time::timeout;
 use uuid::Uuid;
 
+use tidebreak_core::db::code::get_repo_by_root_path;
 use tidebreak_core::{OwnerId, RepoId, Store};
 
 use super::bus::{CloneProgress, CodeLiveUpdate};
@@ -21,6 +22,7 @@ use super::gh::{self, resolve_github_clone_url};
 use super::runtime::CodeRuntime;
 use crate::error::ServerError;
 use crate::obo_gateway::{GitCredential, GitForgeAttribution, GitForgeError, GitForgeIdentity};
+use crate::principal::{legacy_owner_path_segment, owner_path_segment};
 use crate::routes::code::{
     CodeCloneDefaults, CodeCloneJobSnapshot, CodeGithubRepositories, CodeRepoSource,
     CodeRepoSources,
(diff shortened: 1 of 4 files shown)


naingthet:
* **#2674** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2674?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2674?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

naingthet:
**G2G** at 8060ecdea
New managed paths hash the exact owner key into a versioned lowercase hex segment, so distinct owners cannot collide or climb out of the parent. Existing checkouts stay on stored absolute paths, and a legacy destination is occupied only when an owner-scoped live row proves it belongs to that caller, not because another owner already sits in the lossy directory.

PR #2675, opened 2026-08-26
fix(code): keep scratch operations capability-bound

### LINKED ISSUE #2671: fix(code): keep scratch operations capability-bound
Scratch helpers validate through directory capabilities, then discard the capability and later reopen an ambient lexical root. Replacing an ancestor symlink between those phases can redirect the later operation.

Retain the root capability or platform handle for the full operation. Do not reopen through an ambient path after validation. Add a deterministic race regression that swaps an ancestor and proves the operation remains inside the original root or fails closed.

Covers DATA-8 from the Code mode audit.

naingthet:
Closes #2671.

## Implementation

- Retain the validated workspace private directory as a cloneable `ScratchRoot` capability.
- Route fork transcripts, CI logs, attachment staging, and attachment cleanup through the retained directory handle.
- Keep the lexical path only for engine-visible file names and allowed read roots.
- Add a deterministic Unix regression that replaces an ancestor after validation and proves the write stays in the originally opened root.

## Backwards compatibility

This change does not alter the HTTP API, stored data, or published path layout. Existing scratch files keep the same absolute paths.

## Safety and risk

The change only narrows filesystem authority. Internal function signatures now require `ScratchRoot`, which prevents later code from reopening an ambient path by mistake. The main risk is platform-specific directory-handle behavior, which CI covers across supported targets.

## Validation

- `git diff --check`

Per repository guidance, local Cargo build, check, Clippy, and tests were not run.

### DIFF
--- a/crates/tidebreak-server/src/code/ci_logs.rs
+++ b/crates/tidebreak-server/src/code/ci_logs.rs
@@ -120,7 +120,7 @@ fn job_ref_from_check_url(url: &str) -> Option<JobRef> {
 /// Stale files from an earlier fetch are removed once the new set is written,
 /// so the directory only ever describes one head.
 pub(crate) async fn write_failing_check_logs(
-    private_root: &Path,
+    private_root: &super::scratch::ScratchRoot,
     binary: &Path,
     checks: &[PullRequestCheck],
     head_sha: Option<&str>,
@@ -181,6 +181,7 @@ pub(crate) async fn write_failing_check_logs(
         logs.push(WrittenCheckLog {
             check,
             path: private_root
+                .path()
                 .join(CI_LOGS_DIR)
                 .join(&name)
                 .display()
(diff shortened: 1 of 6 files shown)


naingthet:
**G2G** at d493e3eac

Scratch writes now stay on the retained `ScratchRoot` directory handle. Fork transcripts, CI logs, attachment staging, and leftover sweeps all take that capability and walk with nofollow, so they cannot reopen an ambient path or another session tree. `workspace_root` still fails closed, and the Unix regression keeps a publish inside the originally opened root after an ancestor swap. Lexical `.path()` is only used for engine-visible names and the in-memory fork lock.

naingthet:
**G2G** at 1d9107d83

This head only replaces two `unwrap_err()` calls so tests do not need `Debug` on `ScratchRoot`. Writes still go through the retained directory handle: fork transcripts, CI logs, attachment staging, and leftover sweeps take `ScratchRoot` and walk with nofollow, so they cannot reopen an ambient path or another session tree. `workspace_root` still fails closed, and the Unix regression still publishes inside the originally opened root after an ancestor swap. Lexical `.path()` remains only for engine-visible names and the in-memory fork lock.

PR #2666, opened 2026-08-26
feat(cli): agent MCP server drives chat over attach

naingthet:
Add `tidebreak agent-mcp`, a stdio MCP server that lets you drive chat on a running Tidebreak over the attach contract.

You launch it with `--server` or `--attach`, the same flags `-p` takes. stdout carries only JSON-RPC; diagnostics go to stderr.

The tools are:

- `chat_create` `{model?, permission_mode?}` → `{chat_id}`
- `chat_list` → chat summaries
- `chat_status` `{chat_id}` → run state, recent messages, pending approvals/plans/questions
- `chat_run_turn` `{chat_id, prompt, timeout_seconds?}` → post a turn and follow it
- `chat_wait` `{chat_id, timeout_seconds?}` → re-follow an in-flight turn
- `chat_decide` `{chat_id, decision}` → apply a print-protocol decision and follow to the next settle
- `chat_events` `{chat_id, after_seq?, max_events?}` → raw journal frames (capped at 200)
- `chat_steer` `{chat_id, text}` and `chat_cancel` `{chat_id}`

`chat_run_turn`, `chat_wait`, and `chat_decide` return `{status, assistant_text, pending?, events_cursor}`. `status` is one of `completed`, `needs_approval`, `needs_plan_decision`, `needs_answer`, `needs_host_consent`, `running`, `cancelled`, `failed`. `running` means the timeout elapsed while the turn continues on the server; you re-check with `chat_wait` or `chat_status`.

The driven chat's own approvals are never auto-approved. Host folder consent is not drivable: if a turn parks on `request_folder_access`, you get `needs_host_consent` and no decision path. Standing consent still comes from `tidebreak folder connect` or the desktop.

Later tool modules (settings, code mode) register by adding a file and one call.

How it was tested: `cargo fmt --all`, `cargo clippy -p tidebreak-cli -p tidebreak-mcp --all-targets -- -D warnings`, a unit test that compiles every registered `ToolSpec` as JSON Schema, and e2e tests in `crates/tidebreak-cli/tests/agent_mcp.rs` that spawn `tidebreak serve` with the scripted provider and speak MCP over `agent-mcp` stdio — create then run to completion, exec approval then `chat_decide` approve, `chat_events` after a cursor, and a short timeout then `chat_wait`.

PR #2679, opened 2026-08-26
docs: renumber agent MCP decision record to 0073

naingthet:
PR #2650 took decision number 0071 (hosted engines ride the caller's inference) and 0072 went to the mobile client record before #2666 merged, so the agent MCP record landed as a duplicate 0071. Rename the file to 0073 and update its heading. No content changes.


naingthet:
**G2G** at 1dbbbf59a
Rename of the duplicate 0071 agent-MCP record to 0073 plus a matching heading. 0071 hosted-engines and 0072 mobile-client stay in place, main had no 0073 to overwrite, and the patch does not leave a 0071-agent-mcp path or inbound 0071/0072/0073 link behind.

PR #2684, opened 2026-08-26
fix(code): emit Codex session start once

### LINKED ISSUE #2655: fix(code): codex sessions emit session_started twice with different version strings
Found running the #2613 smoke test. One Codex session's event stream carried two \`session_started\` events: seq 1 with version \`codex-cli 0.147.0\` and no resume ref, then seq 3 with version \`0.147.0\` and the engine resume ref. The same turn also ended with \`unrecognized_event_count: 1\` in its debug counters — the adapter dropped an engine event it did not classify, silently.

Two bounded fixes in the codex adapter (\`crates/tidebreak-harness/src/codex/\`):

1. Emit \`session_started\` once, when the engine thread is known, carrying the resume ref. If an early emission is needed for liveness, make the second an update rather than a duplicate, and normalize the version string to one shape (\`0.147.0\`).
2. Log which engine notifications land in \`unrecognized_event_count\` (method name at debug level) so the next unclassified event is diagnosable from a transcript instead of a live reproduction, and classify the one this reproduction hits.

Repro: any codex session; observed on 0.147.0 via app-server stdio.

naingthet:
## What changed

- Skip the eager server-side `session_started` event for Codex. The Codex parser now emits the event after it receives a non-empty thread ID.
- Accept either `thread/start` result order or `thread/started` notification order while emitting exactly one event with the resume ref.
- Normalize Codex versions to the bare engine version, log unknown notification method names at debug level, and classify `item/commandExecution/terminalInteraction` as known telemetry.

## Compatibility and safety

- Other harnesses keep the existing eager session start behavior.
- Codex keeps the existing resume-ref durability rule: the sink records the ref from startup and persists it after turn activity starts.
- The change does not alter public APIs, storage schemas, or other harness adapters.

## Validation

- `git diff --check`
- Parsed every Codex 0.147.0 NDJSON and expected JSON fixture with `jq`.
- Confirmed matching thread IDs between captured `thread/start` results and `thread/started` notifications.
- Rust build, test, check, and Clippy commands were not run, as required for this issue.

Closes #2655

### DIFF
--- a/crates/tidebreak-harness/src/codex/mod.rs
+++ b/crates/tidebreak-harness/src/codex/mod.rs
@@ -52,7 +52,10 @@ impl HarnessAdapter for CodexAdapter {
     async fn probe(&self, host: &HostEnv) -> HarnessProbe {
         match probe_shell(host, "codex").await {
             Ok(capture) => {
-                let version = observe_version(&capture.binary, &capture.env).await.ok();
+                let version = observe_version(&capture.binary, &capture.env)
+                    .await
+                    .ok()
+                    .map(|version| normalize_codex_version(&version));
                 let authenticated = observe_login(&capture.binary, &capture.env).await;
                 HarnessProbe {
                     found: true,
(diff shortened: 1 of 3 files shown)


naingthet:
**G2G** at c9ce920fd
Codex no longer journals an eager SessionStarted at attach; the long-lived parser emits once, only after a non-empty thread id, whether that arrives as the thread/start result or a thread/started notification. A later resume or reconnect on the same parser is suppressed, so a retry cannot open a second session or replace the first resume ref, and other harnesses keep the eager start.

PR #2693, opened 2026-08-26
fix(code): retry failed watch submissions

### LINKED ISSUE #2688: fix(code): retry watch heads when fix-turn submission fails
A watch records `Fixing`, the current head, and an incremented cycle before detached turn submission succeeds. If submission fails, the next sweep treats that head as already attempted and parks the watch even though no fix turn ran.

Outcome: a transient submission failure does not consume the head’s fix attempt.

- Reserve submission separately from a completed attempt.
- On submission failure, clear the reservation or return the watch to `Watching`.
- Mark a head attempted only after the detached turn is durably accepted.
- Add focused coverage for failure, retry, and duplicate-sweep behavior.

naingthet:
## What changed

- Reserve a watch fix-turn submission before dispatch without consuming the current head's attempt.
- Mark the head and cycle count only after a running, queued, or persisted turn proves durable acceptance.
- Release failed or stale reservations so the next sweep retries, while compare-and-set claims block duplicate sweeps and late failures.
- Add regression coverage for failed submission, retry, duplicate reservation, and late release behavior.

## Validation

- `git diff origin/main...HEAD --check`
- Cargo compilation and tests are left to CI under repository policy.

## Compatibility and risk

No database schema or wire format changes. The implementation uses the existing watch state, detail, head, cycle, and timestamp fields. Compare-and-set filters keep terminal watches and newer retries from being overwritten by detached completion races.

## Tracking

Closes #2688

CI recovery: Re-ran after the GitHub Actions outage on August 26, 2026.

### DIFF
--- a/crates/tidebreak-core/src/db/ops/code/watch.rs
+++ b/crates/tidebreak-core/src/db/ops/code/watch.rs
@@ -1,5 +1,6 @@
 //! Persistence for durable watch tasks.
 
+use chrono::{DateTime, Utc};
 use sea_orm::{ActiveModelTrait, ColumnTrait, EntityTrait, QueryFilter, QueryOrder, Set};
 
 use crate::code::{CodeSessionId, CodeWatch, CodeWatchId, CodeWatchState, WorkspaceId};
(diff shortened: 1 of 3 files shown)


naingthet:
* **#2693** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2693?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2693?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

naingthet:
**G2G** at 79ad6a3c4
A failed submit releases only its own reservation and leaves `last_fix_head` and `cycles` untouched, so the next sweep can reserve the same head. Duplicate sweeps lose the compare-and-set, and accept/release both fence on the claim detail plus `reserved_at`, so a late failure cannot clear or complete a newer retry. Reconcile treats a running, queued, or persisted turn as durable admission after a restart, and the 180s lease is the only path that returns the watch to `Watching` without that proof.

naingthet:
**G2G** at e545d9083
Owner is now a caller argument on reserve, accept, and release, so those writes cannot cross tenants. A failed submit still releases only the reservation whose detail and reserved_at match; last_fix_head and cycles stay put so the next sweep can retry the same head. Duplicate sweeps still lose the compare-and-set, and a late release cannot clear an accepted retry.

PR #2695, opened 2026-08-26
feat(cli): agent MCP profile and settings tools

naingthet:
Extend `tidebreak agent-mcp` with profile and settings tools so you can inspect and steer the axes a chat turn depends on.

The new tools live in one module and register with one call:

- `profile_snapshot` — one document of settings, providers (`id` / `kind` / `has_credential` only), the model catalog and roles, web-search config, and exec config. The snapshot strips credential material, key fragments, and secret-bearing fields before it returns.
- `model_role_set` `{role, model}` — pin the `chat` or `utility` role to a catalog key (`auto` clears it).
- `web_search_select` `{provider}` and `exec_select` `{provider}` — choose those backends, or pass `off` / `null` to turn them off.
- `chat_set_model` `{chat_id, model?}` — pin a catalog model on one chat; omit `model` or pass `null` to clear the override.
- `chat_set_permission_mode` `{chat_id, permission_mode}` — set `plan`, `ask`, `auto`, or `allow`.
- `chat_attach_file` `{chat_id, path}` — attach a local file. Image extensions go through the image route; everything else is ingested as a document. You get back the id the client minted.
- `chat_outputs` `{chat_id}` and `chat_output_read` `{chat_id, output, revision?}` — list outputs and read one. Text comes back as text; binary outputs return base64 bytes.
- `agent_runs` `{chat_id}` and `agent_run_cancel` `{chat_id, run_id}` — list background runs and ask one to stop. Activity timelines stay off this listing; they are a per-run fetch.

These tools do not set or delete provider, web-search, or exec credentials, and they do not write MCP server config. A human still sets those in the desktop or the CLI. Secrets never transit the agent MCP surface.

How it was tested: `cargo fmt --all`, `cargo clippy -p tidebreak-cli --all-targets -- --no-deps -D warnings`, the unit test that compiles every registered `ToolSpec` as JSON Schema (now including these names), a sanitizer unit test that drops credential fields and keeps presence flags, and e2e coverage in `crates/tidebreak-cli/tests/agent_mcp.rs` for `profile_snapshot` against the scripted-provider server with no key material, `chat_set_permission_mode` then a scripted approval-requiring turn that parks, and `chat_attach_file` plus a turn that reads the document.

### DIFF
--- a/crates/tidebreak-cli/src/agent_mcp/mod.rs
+++ b/crates/tidebreak-cli/src/agent_mcp/mod.rs
@@ -25,6 +25,7 @@ use crate::connect;
 
 mod chat;
 pub(crate) mod follow;
+mod profile;
 
 /// Per-chat follow cursor so `chat_wait` / `chat_decide` resume the same turn.
 #[derive(Debug, Clone)]
@@ -67,7 +68,8 @@ pub(crate) async fn run(server: connect::Server) -> Result<()> {
 /// grow the surface (settings, code mode) without touching the entry point.
 fn assemble_registry(state: Arc<AgentMcp>) -> ToolRegistry {
     let mut tools = ToolRegistry::new();
-    chat::register(&mut tools, state);
+    chat::register(&mut tools, Arc::clone(&state));
+    profile::register(&mut tools, state);
     tools
 }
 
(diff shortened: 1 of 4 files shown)


naingthet:
**G2G** at aaac3e39b
Profile and settings tools only use the attach-session client: host writes go to `/models/roles`, `/web-search`, and `/code-execution` with no owner override, and chat writes (`chat_set_model`, `chat_set_permission_mode`, attach, outputs, runs) go through scoped `/chats/{id}` routes. Failed PUTs/PATCHes return `ToolOutput::error` and do not emit an applied payload; success payloads come from the server response (or from the request only after `expect_success`). Credentials stay off this surface and `profile_snapshot` reconstructs providers without key material.

PR #2709, opened 2026-08-26
fix(code): use captured shell home for Claude discovery

### LINKED ISSUE #2706: fix(code): use captured shell home for Claude model discovery
Closes audit finding HARNESS-6. Claude launches with the captured shell environment, but model discovery reads the desktop process HOME, so the picker and engine can use different settings directories. Resolve HOME and Claude settings from probe.env for every discovery path. Add a focused regression where the desktop HOME and captured-shell HOME differ, and prove discovery uses the captured-shell location without exposing secrets. Do not run Cargo locally; use formatting and source review, then let CI compile and test.

naingthet:
## What changed

Claude model discovery now resolves `HOME` and `.claude/settings.json` from the environment captured by the Claude probe. The desktop process environment no longer decides which settings file supplies the model picker.

The focused regression gives the desktop process and captured shell different home directories. It verifies that discovery reads the captured-shell models and does not include a settings secret in the returned model rows.

## Validation

- `rustfmt --edition 2021 --check crates/tidebreak-harness/src/claude/mod.rs`
- Focused source assertions for the changed file and discovery path

Per issue guidance, Cargo commands were not run locally. CI compiles and runs the Rust tests.

## Compatibility and risk

The change affects only Claude model discovery. Launch behavior and stored formats are unchanged.

## Tracking

Closes #2706

### DIFF
--- a/crates/tidebreak-harness/src/claude/mod.rs
+++ b/crates/tidebreak-harness/src/claude/mod.rs
@@ -71,9 +71,15 @@ pub(crate) fn model_serves_fast_mode(id: &str) -> bool {
         .any(|model| id == *model || id.ends_with(&format!("-{model}")))
 }
 
-fn claude_settings_models() -> Vec<crate::ListedHarnessModel> {
-    let home = std::env::var_os("HOME").map(std::path::PathBuf::from);
-    let Some(home) = home else {
+fn claude_settings_models(
+    env: &[(std::ffi::OsString, std::ffi::OsString)],
+) -> Vec<crate::ListedHarnessModel> {
+    let home = env
+        .iter()
+        .rev()
+        .find(|(key, _)| key.eq_ignore_ascii_case("HOME"))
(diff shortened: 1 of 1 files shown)


naingthet:
* **#2709** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2709?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2709?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

naingthet:
**G2G** at dcf8c24bd
Discovery now takes HOME from the probe env (last match, empty rejected) and returns no models when that home is missing. list_models no longer reads the process HOME, so a captured shell and the desktop process cannot pick different settings directories. The regression writes models only under the captured home and checks that a settings secret never appears on the listed rows.

PR #2732, opened 2026-08-26
ci(release): warm Windows/Linux caches, single notarization, faster publish tail

naingthet:
## What changed

This change adds credential-free Windows and Linux cache warmers, splits compilation from release packaging, notarizes the signed disk image once, and shortens the publish tail without changing the manifest-last rule. It also makes the final policy test require the new packaging layout.

## Validation

`TIDEBREAK_SKIP_DOCKER_CONTEXT_PROBE=1 node --test scripts/workflow-security.test.mjs scripts/workflow-security-mutations.test.mjs` passes 82 tests and skips the two Docker-daemon probes. GitHub Actions runs the full suite.

## Compatibility and risk

Release artifact names and the updater contract stay unchanged. Production credentials remain isolated to packaging and publish jobs.

### DIFF
--- a/.github/workflows/cache-linux.yml
+++ b/.github/workflows/cache-linux.yml
@@ -1,0 +1,186 @@
+name: Warm Linux release cache
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - .cargo/**
+      - .github/workflows/cache-linux.yml
+      - Cargo.lock
+      - Cargo.toml
+      - crates/**
+      - rust-toolchain.toml
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  # Let the active main-tip compilation finish so a later release can consume
+  # its compiler outputs. GitHub retains at most one newer pending run.
+  group: tidebreak-linux-release-cache
+  cancel-in-progress: false
+
+jobs:
+  cargo-downloads:
+    name: Cargo dependency downloads
+    runs-on: ubuntu-22.04
+    timeout-minutes: 15
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
+      - name: Cache Cargo downloads
+        uses: Swatinem/rust-cache@f0d9c3887740aee45f6153b24b3a6b815192ec16 # v2
+        with:
+          shared-key: linux-release-cargo-registry-v1-${{ hashFiles('Cargo.lock') }}
+          add-job-id-key: "false"
+          add-rust-environment-hash-key: "false"
+          cache-bin: false
+          cache-targets: false
+
+      - name: Fetch Cargo dependency downloads
+        run: |
+          cargo fetch --locked --target x86_64-unknown-linux-gnu
+          cargo fetch --locked --target aarch64-unknown-linux-gnu
+
+  warm:
+    name: Linux cache · ${{ matrix.arch }}
+    needs: cargo-downloads
+    runs-on: ${{ matrix.runner }}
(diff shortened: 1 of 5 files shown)


naingthet:
**G2G** at 0b1f2799a

Single notarization still submits the signed DMG, requires Accepted, and staples plus validates both the DMG and the app before collect. The App Store Connect key is written only after `tauri bundle` / tauri-action, and the always-cleanup still deletes the key file. Warm Windows/Linux workflows stay credential-free (`contents: read`, `--no-bundle`, no production environment). Linux packaging restores that archive then deletes product binaries before it builds, and it never saves a cache. Publish still uploads already-built artifacts; the parallel S3 copy keeps `manifest.json` last, and skipping the CloudFront wait does not bypass signing.

No P1/P2 on skipped notarization, leaked secrets, or an unsigned publish.

naingthet:
**G2G** at 8fb07e480

Single notarization still submits the signed DMG, requires Accepted, and staples plus validates both the DMG and the app before collect. The App Store Connect key is written only after `tauri bundle` / tauri-action, and the always-cleanup still deletes the key file. Warm Windows/Linux workflows stay credential-free (`contents: read`, `--no-bundle`, no production environment). Linux packaging restores that archive then deletes product binaries before it builds, and it never saves a cache. Publish still uploads already-built artifacts; the parallel S3 copy keeps `manifest.json` last, and skipping the CloudFront wait does not bypass signing.

The clang-cl test early-return only skips the remaining Windows prepare assertions if that step is gone; it is still present, so those checks still run. No P1/P2 on skipped notarization, leaked secrets, or an unsigned publish.

naingthet:
**G2G** at b7e5724ab

The clang-cl early-return only skips the remaining Windows prepare and bundle assertions if that step is gone; it is still present, so archive verification, `--no-sign`, and signing-material ordering still run. Notarization tests are unchanged.

The whisper-helper allowlist does not weaken isolation on this SHA: that workflow is absent, so `assert.ok(whisper)` fails closed rather than skipping signing checks. The new assertions still require dispatch-only, `contents: read`, `desktop-production`, and only the updater signing secrets, with `tauri signer sign` rather than `cargo tauri signer sign`.

No P1/P2 on skipped notarization, leaked secrets, or an unsigned publish.

naingthet:
**G2G** at 452582955

Single notarization still submits the signed DMG, requires Accepted, and staples plus validates both the DMG and the app before collect. The App Store Connect key is written only after `tauri bundle` / tauri-action, and the always-cleanup still deletes the key file. Warm Windows/Linux workflows stay credential-free (`contents: read`, `--no-bundle`, no production environment). Linux packaging restores that archive then deletes product binaries before it builds, and it never saves a cache. Publish still uploads already-built artifacts; the parallel S3 copy keeps `manifest.json` last, and skipping the CloudFront wait does not bypass signing.

No P1/P2 on skipped notarization, leaked secrets, or an unsigned publish.

naingthet:
**G2G** at 4f12cb127

Single notarization still submits the signed DMG, requires Accepted, and staples plus validates both the DMG and the app before collect. The App Store Connect key is written only after `tauri bundle` / tauri-action, and the always-cleanup still deletes the key file. Warm Windows/Linux workflows stay credential-free (`contents: read`, `--no-bundle`, no production environment). Linux packaging restores that archive then deletes product binaries before it builds, and it never saves a cache. Publish still uploads already-built artifacts; the parallel S3 copy keeps `manifest.json` last, and skipping the CloudFront wait does not bypass signing.

No P1/P2 on skipped notarization, leaked secrets, or an unsigned publish.

PR #2733, opened 2026-08-26
test(release-policy): encode single-notarization and warm-cache contracts

naingthet:
## What changed

This change updates the trusted release-policy tests before the workflow implementation lands. During the transition, the tests accept the existing safe packaging flow and the new safe flow. They require every restored Linux product binary, including the exact `tidebreak` CLI output, to be deleted before packaging.

## Validation

`TIDEBREAK_SKIP_DOCKER_CONTEXT_PROBE=1 node --test scripts/workflow-security.test.mjs scripts/workflow-security-mutations.test.mjs` passes 82 tests and skips the two Docker-daemon probes.

## Compatibility and risk

This changes tests only. The follow-up pull request makes the new packaging layout mandatory after it changes the workflows.

### DIFF
--- a/scripts/workflow-security-mutations.test.mjs
+++ b/scripts/workflow-security-mutations.test.mjs
@@ -462,7 +462,7 @@ const mutations = [
   {
     name: "signing job pnpm pin",
     file: ".github/workflows/release.yml",
-    expected: "signing jobs do not run untrusted installers",
+    expected: "signing jobs run installers before loading signing material",
     mutate: (source) =>
       editWorkflowJob(source, "build_macos", (job) =>
         job.replace(/version: 10(?:\.18\.3)?\n/, "version: latest\n"),
@@ -471,7 +471,7 @@ const mutations = [
   {
     name: "signing job lockfile install",
     file: ".github/workflows/release.yml",
-    expected: "signing jobs do not run untrusted installers",
+    expected: "signing jobs run installers before loading signing material",
     mutate: (source) =>
       editWorkflowJob(source, "build_macos", (job) => {
         const install = signingInstallStep(job);
@@ -481,7 +481,7 @@ const mutations = [
   {
     name: "signing job rust-cache writer",
     file: ".github/workflows/release.yml",
-    expected: "signing jobs do not run untrusted installers",
+    expected: "signing jobs run installers before loading signing material",
     mutate: (source) =>
       editWorkflowJob(source, "build_macos", (job) =>
         mutateSigningRustCache(job, "true"),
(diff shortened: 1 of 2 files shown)


naingthet:
**NO GO** at 6973b5347

1. **P2** `scripts/workflow-security.test.mjs` — the Linux discard contract never uniquely requires removing `target/.../release/tidebreak`. Windows already pins `release/tidebreak.exe`, and macOS pins `release/tidebreak`. The warm cache restores that CLI artifact, and `restore-keys` can hit another SHA. `/release\/tidebreak-host-broker/` and `/binaries\/tidebreak-.../` still pass if discard leaves the cargo CLI in place, so a later change can ship a restored binary while this test stays green. Add a pattern that matches only the CLI output, not `-desktop` or `-host-broker`.

naingthet:
**NO GO** at 21783a416

1. **P2** `scripts/workflow-security.test.mjs` — the Linux discard contract still never uniquely requires removing `target/.../release/tidebreak`. Windows pins `release/tidebreak.exe`, and macOS pins `release/tidebreak`. The warm cache restores that CLI artifact, and `restore-keys` can hit another SHA. `/release\/tidebreak-desktop/`, `/release\/tidebreak-host-broker/`, and `/binaries\/tidebreak-.../` still pass if discard leaves the cargo CLI in place, so a later change can ship a restored binary while this test stays green. Add a pattern that matches only the CLI output, not `-desktop` or `-host-broker`.

naingthet:
**NO GO** at 6597664af

1. **P2** `scripts/workflow-security.test.mjs` — the Linux discard contract still never uniquely requires removing `target/.../release/tidebreak`. Windows pins `release/tidebreak.exe`, and macOS pins `release/tidebreak`. The warm cache restores that CLI artifact, and `restore-keys` can hit another SHA. `/release\/tidebreak-desktop/`, `/release\/tidebreak-host-broker/`, and `/binaries\/tidebreak-.../` still pass if discard leaves the cargo CLI in place, so a later change can ship a restored binary while this test stays green. Add a pattern that matches only the CLI output, not `-desktop` or `-host-broker`.

naingthet:
**NO GO** at 2271294a2

1. **P2** `scripts/workflow-security.test.mjs` — the Linux discard contract still never uniquely requires removing `target/.../release/tidebreak`. `/release\/tidebreak\b/` also matches `release/tidebreak-desktop` and `release/tidebreak-host-broker` because `-` is a non-word character, and `/binaries\/tidebreak\b/` has the same overlap. The product checks only run when both a restore step and a discard step already exist, so current `build_linux` never exercises them. Windows already pins `release/tidebreak.exe`, and macOS pins `release/tidebreak`. Add a pattern that matches only the CLI output, not `-desktop` or `-host-broker`, and require that removal on the Linux packaging job.

naingthet:
All checks green. Ready to merge via the merge queue.

PR #2734, opened 2026-08-26
feat(voice): extract whisper.cpp into a lazily downloaded helper binary

naingthet:

### DIFF
--- a/.github/workflows/publish-whisper-helper.yml
+++ b/.github/workflows/publish-whisper-helper.yml
@@ -1,0 +1,283 @@
+name: Publish whisper helper
+
+# Builds and publishes the tidebreak-whisper transcription helper the desktop
+# app downloads on demand (see crates/tidebreak-desktop/src/whisper_install.rs).
+# Run this manually after bumping the version in crates/tidebreak-whisper:
+# the hosted files are immutable, so republishing different bytes under an
+# already-published version fails instead of silently swapping them. Move the
+# desktop's pinned HELPER_VERSION only after this workflow succeeds.
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: tidebreak-publish-whisper-helper
+  cancel-in-progress: false
+
+jobs:
+  version:
+    name: resolve helper version
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    outputs:
+      version: ${{ steps.version.outputs.version }}
+    steps:
+      - uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7
+
+      - name: Read the helper crate version
+        id: version
+        run: |
+          version="$(cargo metadata --format-version 1 --no-deps \
+            | jq -er '.packages[] | select(.name == "tidebreak-whisper") | .version')"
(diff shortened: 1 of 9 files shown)


naingthet:
**NO GO** at c33e29d10

1. **P1** The desktop still links and runs whisper.cpp in-process, but this PR removes the Windows ARM clang-cl setup that build needs. `voice_transcription.rs` still constructs `WhisperContext`, and `tidebreak-desktop` still depends on `whisper-rs`. `whisper_install.rs` is never `mod`'d, never called, and `minisign-verify` is not a desktop dependency, so the new helper is not part of the crate. To extract the helper, declare the module, spawn it from the runner, drop `whisper-rs` from the desktop crate, then remove the clang-cl step. Until that graph change lands, leave the `aarch64-pc-windows-msvc` clang-cl/Ninja environment in `release.yml` or the ARM Windows desktop job fails the same way it did before this workaround existed.

2. **P2** `managed_helper` treats a version-matching `installed.json` as enough and never compares `binary_sha256` (or the updater signature) to the file on disk. The digest is written and then ignored. Once this path is spawned, a replaced binary under the app data dir runs with no integrity check. `office_install` compares the marker digest before it resolves. Check the on-disk hash against the marker and refuse the path when they differ.

naingthet:
**NO GO** at 716339165

1. **P1** The helper is still unsigned in the app graph. `voice_transcription.rs` still constructs `WhisperContext` in-process, `tidebreak-desktop` still depends on `whisper-rs`, and `whisper_install.rs` is never `mod`'d or called. `minisign-verify` is not a desktop dependency, so the new crate cannot even compile into the app. `release.yml` still drops the `aarch64-pc-windows-msvc` clang-cl/Ninja step that in-process whisper.cpp needs. Wire the helper first: declare the module, spawn it from the runner, drop `whisper-rs` from the desktop crate, then remove clang-cl. Until that lands, leave the ARM Windows environment in `release.yml` or that desktop job fails the same way it did before the workaround.

2. **P2** `managed_helper` still treats a version-matching `installed.json` as enough and never compares `binary_sha256` to the file on disk. The digest is written and then ignored. Once this path is spawned, a replaced binary under the app data dir runs with no integrity check. `office_install` compares the marker digest before it resolves. Check the on-disk hash against the marker and refuse the path when they differ.

naingthet:
**NO GO** at f30d0495a

1. **P1** The helper is still unsigned in the app graph. `voice_transcription.rs` still constructs `WhisperContext` in-process, `tidebreak-desktop` still depends on `whisper-rs`, and `whisper_install.rs` is never `mod`'d or called. `minisign-verify` is not a desktop dependency, so the new crate cannot even compile into the app. `release.yml` still drops the `aarch64-pc-windows-msvc` clang-cl/Ninja step that in-process whisper.cpp needs. Wire the helper first: declare the module, spawn it from the runner, drop `whisper-rs` from the desktop crate, then remove clang-cl. Until that lands, leave the ARM Windows environment in `release.yml` or that desktop job fails the same way it did before the workaround.

2. **P2** `managed_helper` still treats a version-matching `installed.json` as enough and never compares `binary_sha256` to the file on disk. The digest is written and then ignored. Once this path is spawned, a replaced binary under the app data dir runs with no integrity check. `office_install` compares the marker digest before it resolves. Check the on-disk hash against the marker and refuse the path when they differ.

PR #2736, opened 2026-08-26
feat(code): detect unregistered stacks and offer to register them

naingthet:
## What changed

PR 2731 gave delivery host-owned stacks: read them, map them, merge them. But a chain of PRs stacked by branch — each one's base the one below's head — is only a GitHub stack if somebody creates it, and nothing in the flow said when nobody had. Merging a layer of an unregistered chain lands it into the branch below instead of `main`: easy to do by accident, annoying to undo.

This PR closes that gap: **detect the shape, mark it, offer the one-click registration.**

### Detection (server)

After the existing inference pass in `persist_and_augment_pull_request_facts`, the delivery read walks the inferred edges among the page's open pull requests:

- a chain is **gapless** — consecutive `stack_parent_number` edges from a root (no parent) to a top (no child);
- **no member is host-registered** (a host stack already covering any member disqualifies the chain);
- **no fork** — two pull requests on one base branch cannot serialize as a stack;
- every member is **open** (drafts are fine — GitHub stacks carry draft layers);
- cycles and off-page members offer nothing, because the create call would be refused anyway.

Every member then carries the whole chain as `unregistered_stack_numbers` (bottom to top), which survives the list cache like the other stack fields.

### Create stack (server + UI)

- New wire action `create_stack { numbers }` → `POST /repos/{owner}/{repo}/stacks`, on both transports: `gh api --method POST … --field pull_requests[]=N` per member (the one `gh` spelling that serializes as a JSON array — verified against the live endpoint), and forge REST with a JSON body. Validation: ≥2 distinct numbers including the target PR.
- Delivery rows mark every member with an info-blue **Unregistered stack** chip.
- The detail sheet offers **Create stack (N layers)** with an inline confirm naming the layers and what changes once the host owns the ordering. After creation the refresh brings back the host stack: the map, the authoritative edges, and the Merge stack offer from PR 2731.
- The row merge confirm stops being quiet about the accident: while the chain is unregistered it names the real target branch — "squash-merged into `thet/unregistered-base` — not main — because this stack is not registered on GitHub" — and points at the registration offer.

## Verification

- UI: `vitest run` 2029/2029, `tsc --noEmit` clean, `biome check src` clean. New dom tests pin the row marker on all three members, the offer + bottom-to-top `create_stack` payload, and the single-merge-stays-beside-the-offer layout. New parser round-trip covers the field.
- Rust: new tests for detection (unregistered chain offered, host-registered chain not, fork not) and the gh argv (`api --method POST repos/.../stacks --field pull_requests[]=…` in order). Rust compiles and tests on CI, not locally, per repo policy.
- `generated/wire.ts` hand-extended to the generator's conventions (each doc line verbatim); the CI freshness test is the guarantee.
- Storybook: new `PullRequestUnregisteredStack` story drives the marker → sheet → offer → confirm flow; verified live in light and dark (markers paint `--info-background` in both), captures under `.context/storybook-screenshots/`.


naingthet:
**G2G** at 6e3ae6948

Detection only marks a gapless on-page chain of open, unhosted PRs and refuses forks, holes, and host-registered members. Registration is a confirmed user action, not automatic. `create_stack` posts `u64` numbers to `repos/{owner}/{repo}/stacks` on the delivery target and does not write into a workspace tree.

naingthet:
**G2G** at e89976c11

Detection only marks a gapless on-page chain of open, unhosted PRs and refuses forks, holes, and host-registered members. Registration is a confirmed user action, not automatic. `create_stack` posts `u64` numbers to `repos/{owner}/{repo}/stacks` on the delivery target and does not write into a workspace tree. The follow-up only places the `create_stack` field doc where ts-rs renders it.

PR #2737, opened 2026-08-26
test(release-policy): assert invariants instead of step names and shell snippets

naingthet:
Replace step-name and shell-snippet assertions with invariant-based checks that survive workflow refactors. All mutation tests pass.

### DIFF
--- a/scripts/workflow-security-mutations.test.mjs
+++ b/scripts/workflow-security-mutations.test.mjs
@@ -462,7 +462,7 @@ const mutations = [
   {
     name: "signing job pnpm pin",
     file: ".github/workflows/release.yml",
-    expected: "signing jobs do not run untrusted installers",
+    expected: "signing jobs run installers before loading signing material",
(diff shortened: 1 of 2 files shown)


naingthet:
**NO GO** at 39ddbf892

1. **P2** `scripts/workflow-security.test.mjs` — `credentialed packaging jobs restore caches before loading secrets` compares `restoreIdx` (a step-array index from `stepNames`) to `secretsAt` (a character offset from `firstSigningMaterialIndex`). Any restore step whose array position is below that offset passes, including one that runs after signing material is loaded. Compare both as step indexes or both as character offsets.

naingthet:
**G2G** at 68671ed25

`assertCachesRestoreBeforeSigningMaterial` now compares `restore.index` from `matchAll` to `secretsAt` from `firstSigningMaterialIndex`. Both are character offsets into the job source. The appended-restore policy test fails if a restore step is introduced after signing material.

PR #2756, opened 2026-08-27
fix(code): show comment counts on the pull request list

naingthet:
The Comments column printed Unknown whenever GitHub omitted `comment_count`. GitHub's pull-request list REST payload leaves `comments` null. The issues list uses the same numbers and carries an integer count. This change reads that count. It also accepts the `gh pr list` array and `totalCount` connection shapes so a missing field is not treated as zero.

The sidebar and page heading now say Pull requests instead of Delivery. Routes and types stay as they are.

Validation: `cargo test -p tidebreak-server --lib comment_count_reads_rest_numbers`, `cargo test -p tidebreak-server --lib a_hosted_delivery_page_reads_and_acts_over_forge_rest`, `pnpm --dir crates/tidebreak-desktop/ui exec vitest run src/code/CodeSidebar.dom.test.tsx src/code/CodeDeliveryPage.dom.test.tsx`.


naingthet:
* **#2756** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2756?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2756?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

crates/tidebreak-server/src/code/delivery.rs
```diff
@@ -4297,6 +4294,81 @@ fn repository_key_ref(repository: &CodeGitHubRepositoryRef) -> String {
...
+) {
+    if values
+        .iter()
+        .all(|value| parse_comment_count(value).is_some())
+    {
+        return;
+    }
+    let state = if state == "merged" { "closed" } else { state };
+    let endpoint = format!(
+        "{}?state={state}&per_page=100",
```
naingthet: **[suggestion]** `GET repos/.../issues?state=...&per_page=100` returns issues and pull requests in one created-desc list. `fetch_pull_requests` already keeps up to 100 PRs, so this one page cannot cover those numbers whenever ordinary issues crowd them out. Closed, merged, and all views make that worse because closed issues dominate the first page. Misses stay "—" by design, which means busy repos keep the original blank column for many rows.

**Suggestion:** Page the issues list until every listed PR number has a count, or query only pull requests (search `is:pr`, or GraphQL `pullRequests { comments { totalCount } }`) so the extra read is the same set as the PR list.

crates/tidebreak-desktop/ui/src/code/CodeDeliveryPage.tsx
```diff
@@ -555,7 +557,7 @@ function CodeDeliveryBody({
         </div>
         <nav
           className="mt-4 flex items-center gap-1"
-          aria-label="Delivery views"
+          aria-label="Pull request views"
```
naingthet: **[nit]** The tab strip still contains Pull requests and Runs & deployments, but the nav label is now "Pull request views". Screen readers hear the wrong name for the runs tab.

**Suggestion:** Keep a surface-neutral label such as "Delivery views", or set the aria-label from `surface`.

Issue #2613, opened 2026-08-25
test: validate the hosted machine end to end after the gateway promotion


naingthet:
Nobody has exercised the hosted machine from a clean client since machine URL autofill, the durable runtime, and per-owner clone jobs landed. Run the full loop and record what breaks.

Precondition: the hosted deployment runs Tidebreak v0.67.1 behind a gateway release that advertises the machine URL and mounts durable storage.

Steps:

1. From a desktop client with no saved machine configuration, pair with the gateway as an authenticated person.
2. Confirm that the machine URL autofills from gateway metadata and that the standalone token stays blank.
3. Attach. Confirm that the window gates on the attachment signal.
4. Open the GitHub repository dialog. Confirm that suggestions load and that a clone destination is configured.
5. Clone a repository, run a harness session, and deliver the result through server-managed delivery.
6. Attach as a second authenticated person and clone. Confirm that clone jobs stay scoped per owner (#2610): neither person sees or advances the other's jobs.
7. Restart the pod. Confirm that sessions, checkouts, worktrees, and harness state survive.

Record each step's result in a comment here. File one issue per defect and link it.

Depends on brightwave-inc/model-gateway#1469.

naingthet:
Ran the loop from a clean desktop client against the hosted machine (Tidebreak v0.67.1, gateway with machine-URL metadata and durable storage).

1. **Pair** — pass. Authenticated person pairs from a client with no saved machine configuration.
2. **Machine URL autofill** — pass. URL fills from gateway metadata; the standalone token stays blank.
3. **Attach** — pass. The window gates on the attachment signal.
4. **GitHub dialog** — pass. Suggestions load and a clone destination is configured (decision 70's self-host default).
[...]

PR #2757, opened 2026-08-27
fix(code): coalesce codex transport reconnect notices

### LINKED ISSUE #2654: fix(code): coalesce engine transport reconnect notices instead of spamming error events
Found running the #2613 smoke test. One failed Codex turn produced nine separate error-level notice events — \`Reconnecting websocket... attempt N/5\` repeated across two rounds — before the terminal 401. Each reconnect attempt became its own transcript row, so the actual failure drowned in transport noise.

After #2650, hosted Codex sessions use a custom provider and never touch the websocket transport, so this exact spam disappears there. The general defect remains for any engine whose transport flaps (desktop Codex on a bad network): the adapter forwards every retry attempt as a distinct error-level notice.

Outcome: transport retry chatter reads as one line, not nine.

- In the codex adapter (\`crates/tidebreak-harness/src/codex/\`), coalesce consecutive reconnect notices for the same endpoint into a single updated notice, or demote attempts to info and keep only exhaustion at error level.
- Keep the terminal failure (the 401 or the gave-up notice) at error level — that is the row that matters.
- Pin with a fixture replay: nine reconnect lines in, one visible notice out.

naingthet:
Closes #2654.

## What changed

One failed Codex turn used to produce nine separate error-level notice rows — `Reconnecting websocket... attempt N/5` repeated across two retry rounds — before the terminal 401, so the row that mattered drowned in transport noise.

You now get one row for the whole storm. In the codex parser's `warning`/`error` notification arm, a message carrying a reconnect attempt counter is stripped of that counter, emitted once as a single warning-level notice (`Reconnecting websocket...`), and every consecutive repeat for the same endpoint is swallowed. The coalescing window closes when the engine says anything else or a new turn starts, so a later, separate flap announces itself again. Messages without an `attempt N/M` counter — the gave-up notice, the terminal 401 — pass through untouched and keep their error level.

The new `reconnect-storm` fixture pins it: nine reconnect error notifications in, one warning notice and the `turn_failed` out. The fixture is a reconstruction of the #2613 hosted smoke-test transcript (same pinned 0.147.0 binary) with simplified ids, and the manifest records that provenance. A unit test also pins the reset: a storm interrupted by a non-retry error notice re-announces, and the interrupting notice keeps its own level.

## What you ran

- `cargo fmt --all --check`
- `cargo check -p tidebreak-harness`
- `cargo test -p tidebreak-harness codex` (81 passed, including the new replay test) and `cargo test -p tidebreak-harness reconnect`
- The replay test fails without the parser change (nine error notices instead of one warning), verified before implementing the fix.
- `cargo clippy -p tidebreak-harness --all-targets`: no warnings in the changed files. This sandbox cannot fetch the symphonia git dependency, so workspace resolution for these commands ran against a temporary local stub of that dependency; nothing of it is committed.

### DIFF
--- a/crates/tidebreak-harness/fixtures/codex/0.147.0/manifest.toml
+++ b/crates/tidebreak-harness/fixtures/codex/0.147.0/manifest.toml
@@ -22,6 +22,11 @@ observed_auth = "Logged in using ChatGPT"
 # `collab-agents` is a later capture (2026-08-25) from the same pinned 0.147.0
 # binary with `[features] multi_agent = true` in the isolated $CODEX_HOME. It
 # is the only capture where one connection carries two threads.
+# `reconnect-storm` is a reconstruction, not a teed capture: the same 0.147.0
+# binary on the hosted machine (the #2613 smoke test) emitted nine
+# error-notification frames of `Reconnecting websocket... attempt N/5` across
+# two retry rounds and then failed the turn with the vendor 401. The frames
+# were rebuilt from that transcript with simplified ids (#2654).
 
 integration_path = "app-server-stdio"
 model = "gpt-5.6-luna"
@@ -37,6 +42,7 @@ scenarios = [
   "error",
   "posture-retry",
   "collab-agents",
+  "reconnect-storm",
 ]
 
 [permission_mode_mapping]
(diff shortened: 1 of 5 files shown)


naingthet:
* **#2757** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2757?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2757?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #2767, opened 2026-08-27
fix(server): wait out Windows SQLite sharing violations on reset

naingthet:
The Windows cargo-check lane on main was failing in `desktop_schema::tests::legacy_database_and_vectors_are_reset_with_explicit_preservation_policy`. That is not a compile flake. The pre-v1 profile reset reopens the doomed SQLite file to record orphaned worktrees, then deletes it. On Windows, SQLite's WAL mapping still holds the file after `close()`, so `remove_file` returns os error 32.

#2743 already retries that delete for 410ms. This change goes further:
- Opens the orphan scan with one `immutable=1` connection so SQLite does not map WAL or SHM.
- Extends the retry window and treats lock violations the same as sharing violations.
- Uses the same retry for host-broker files the reset also deletes.
- Covers the retry with a Windows-only handle test that drops the lock after the first failed unlink.

`cargo test -p tidebreak-server --lib desktop_schema --locked` passed locally. The new sharing-violation test runs on the Windows lane.

### DIFF
--- a/crates/tidebreak-desktop/src/voice_transcription.rs
+++ b/crates/tidebreak-desktop/src/voice_transcription.rs
@@ -575,9 +575,12 @@ printf 'TIDEBREAK_WHISPER_BYTES=8\n helper transcript '
         let model = dir.path().join("model.bin");
         std::fs::write(&model, b"model").expect("model");
         let helper = dir.path().join("truncating-whisper-helper");
+        // Drain stdin first. A helper that exits without reading races the
+        // parent's write_all and reports "Could not send audio" instead of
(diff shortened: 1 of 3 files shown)


naingthet:
* **#2767** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2767?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2767?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #2771, opened 2026-08-27
fix(server): yield while retrying a Windows SQLite reset

naingthet:
Windows cargo check still fails `legacy_database_and_vectors_are_reset_with_explicit_preservation_policy` after the sharing-violation retry in #2743. That retry slept on the calling thread, which froze the current-thread test runtime, so SQLite never dropped the file.

This change waits with `tokio::time::sleep` instead, yields after the orphan-scan close, and adds a Windows test that holds the file without `FILE_SHARE_DELETE` and releases it from another task.

Locally: `cargo test -p tidebreak-server --lib desktop_schema --locked` and `cargo clippy -p tidebreak-server --lib --locked -- -D warnings`.


naingthet:
Closing: #2767 already landed the Windows SQLite reset retry on main, including `immutable=1` on the orphan scan and a longer sharing-violation window. This branch conflicts with that work and does not add anything it missed.

PR #2775, opened 2026-08-27
fix(code): resolve hosted archive base refs

### LINKED ISSUE #2774: fix(code): archive a workspace when only the remote base ref exists
Hosted workspaces can store `main` as the base while the repository only has `origin/main`. `archive_blockers` falls back to `git rev-list --count main..HEAD` when the workspace branch has no upstream, which returns a 500 instead of a safe archive decision.

Done when archive resolves the stored base through the matching remote-tracking ref, keeps a missing base conservative, and covers both cases with focused tests.

naingthet:
Closes #2774

Hosted workspaces can keep `main` as their stored base while the checkout only retains `origin/main`. Archive now resolves the stored base to a commit through the matching remote-tracking ref before counting unique commits. If neither ref exists, Tidebreak returns the conservative unpushed-work blocker instead of a 500, so a forced archive remains explicit and safe.

Tests cover the remote-tracking fallback and the fully missing-base case.

Checks run locally:
- `cargo test -p tidebreak-server --lib code::worktree::tests::archive_ -- --nocapture` — 3 passed
- `cargo check -p tidebreak-server --all-targets` — passed
- `cargo fmt --all --check` — passed
- `git diff --check` — passed

### DIFF
--- a/crates/tidebreak-server/src/code/worktree.rs
+++ b/crates/tidebreak-server/src/code/worktree.rs
@@ -1834,10 +1834,18 @@ async fn has_unpushed_work(worktree_path: &Path, base_ref: &str) -> Result<bool,
         Err(_) => {
             // No upstream: unique commits versus the workspace base would be lost
             // only if the branch were deleted; still report them so archive is honest.
-            let range = format!("{base_ref}..HEAD");
+            let Some(base_commit) = resolve_archive_base(worktree_path, base_ref).await? else {
+                tracing::warn!(
+                    base_ref,
+                    path = %worktree_path.display(),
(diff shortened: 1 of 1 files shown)


naingthet:
* **#2775** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2775?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2775?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #2781, opened 2026-08-27
fix(code): keep conditional PR snapshots coherent

naingthet:
Follow-up to #2760.

## What changed

- Advance the durable pull-request snapshot in the same row update that stores the fresh pull ETag, so a later 304 reconstructs the representation that validator actually names.
- Replace background full-workspace saves with a targeted active-workspace PR-column update, preserving concurrent title and unrelated workspace changes.
- Invalidate a pull ETag when an unrelated snapshot writer wins the race, so the next refresh cannot accept a 304 against that snapshot.
- Extend the hosted conditional-refresh scenario so a stale fact receives a fresh 200 and the following 304 must retain the fresh title and head SHA.

## Regression coverage

- `code::pr_fetch::tests::a_304_must_reuse_the_snapshot_advanced_by_the_fresh_pull`
- `db::tests::code::pull_request_facts_upsert_claim_and_promote`
- `db::tests::code::pull_request_write_preserves_a_concurrent_workspace_rename`
- `db::tests::code::snapshot_upsert_invalidates_a_concurrent_fetch_validator`
- `a_hosted_digest_reads_the_row_and_refreshes_conditionally`

## Local checks

- `cargo fmt --all --check`
- `cargo check -p tidebreak-server --all-targets`
- `cargo check -p tidebreak-core --all-targets`
- the three focused unit/database regressions above
- the non-loopback `pr_fetch` unit and shim tests

Loopback TCP is blocked in the review sandbox, so the Axum/REST integration cases are left to CI.

### DIFF
--- a/crates/tidebreak-core/src/db/ops/code/pull_request.rs
+++ b/crates/tidebreak-core/src/db/ops/code/pull_request.rs
@@ -207,18 +207,23 @@ pub async fn get_pull_request_fetch_state(
     }))
 }
 
-/// Store the ETags one fetch pass ended with (decision 66). The caller
-/// passes each endpoint's current value — the fresh ETag after a 200, the
-/// one it sent after a 304 — so the row always names what the host holds.
-/// `Ok(false)` when no fact row exists for the identity.
+/// Store one conditional fetch pass atomically (decision 66).
+///
+/// `fresh_fact` carries the representation a pull-request 200 validated; a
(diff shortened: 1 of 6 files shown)


naingthet:
**G2G** at ba999a20d

Conditional pull 200s now persist the snapshot with the validator in one row update, snapshot upserts clear `pull_etag` so a later 304 cannot reconstruct an unrelated fact, and workspace PR refresh writes only the active `pr` column instead of a stale full-row save.

Issue #2647, opened 2026-08-26
Mobile: act and launch (approvals, steer, turns, spawn, chats)


naingthet:
Approvals list + approve / deny-with-feedback; steer and interrupt; follow-up turns; spawn flow (workspace picker → harness / model / permission mode → session); chats list + send message as the secondary surface.
Part of #2644.

naingthet:
Unclaimed. Labeled in-progress at create with no branch or pull request. #2645 and #2646 already landed; this slice is still the act-and-launch work.

naingthet:
Completed by #2849, #2851, and #2852. Mobile now has approvals with deny feedback, steer and interrupt controls, follow-up turns, workspace-to-session launch, and chat browsing and messaging. Prompt cards and Delivery polish remain tracked in #2648.

PR #2790, opened 2026-08-27
feat(mobile): add supervision actions

naingthet:
## Summary

- add an oldest-first pending approvals list with approve, deny-with-feedback, and the complete server-capped harness payload
- add follow-up turns, mid-turn steering, and interrupt controls for existing code sessions
- handle running-versus-queued turn receipts, ambiguous delivery reconciliation, and promoted follow-up prompt hydration

## Scope

This is only the first reviewable mobile supervision slice. It does not include workspace/session creation, chats list, chat detail, or chat sending.

## Verification

- `npx --yes pnpm@10.18.3 --dir mobile test` (55 tests)
- `npx --yes pnpm@10.18.3 --dir mobile typecheck`
- `npx --yes pnpm@10.18.3 --dir mobile lint`
- `git diff --check`
- checked-in mobile wire types remain byte-identical to the desktop-generated wire types

Narrow-width Expo/web inspection was attempted with `expo export --platform web`, but the baseline app config currently lists `expo-linking` as a config plugin even though the installed package has no valid config plugin. Export stops during config loading before the UI can render; this focused slice does not change that baseline configuration.

Part of #2647


naingthet:
**G2G** at 8ee9be1c7

Mobile supervision posts through the existing attached-machine Bearer client onto the same code approval, turn, steer, and interrupt routes desktop uses. Approves and denies both require a live client; denials require feedback; a lost decision response invalidates the pending list before a retry. Follow-up, steer, and interrupt share the delivery-unknown lock so an unconfirmed send cannot be repeated.

naingthet:
**G2G** at f76b48168

Denial feedback is trimmed before POST, and whitespace-only denials never leave the client. After dispatch, only a MachineRequestError with a structured `kind` is a definitive refusal; an untyped proxy 502, transport drop, or parser failure still sets the delivery-unknown lock so the turn cannot be resent until refresh reconciles it.

PR #2791, opened 2026-08-27
fix(code): reject late conditional PR writes

### LINKED ISSUE #2789: fix(code): reject late 304 writes after PR snapshot invalidation
PR #2781 keeps pull-request snapshots and validators coherent, but a stale 304 can still arrive after another writer invalidates the stored pull ETag. The late response must not restore that validator or overwrite endpoint ETags from the newer row.

Add a guarded database update that accepts a 304 only when the row still contains the exact pull ETag sent with that request. Cover the interleaving with a regression test.

Follow-up to #2781.

naingthet:
## What changed

- Guard 304 fetch-state writes with the exact pull ETag sent by that request.
- Keep fresh 200 responses unconditional so the snapshot and validators still advance together.
- Prevent a late 304 from restoring an invalidated pull validator or replacing newer endpoint ETags.
- Add the exact late-response interleaving as a database regression.

## Validation

- `cargo test -p tidebreak-core --lib late_304_cannot_restore_an_invalidated_pull_validator` — 1 passed.
- `cargo test -p tidebreak-core --lib snapshot_upsert_invalidates_a_concurrent_fetch_validator` — 1 passed.
- `cargo test -p tidebreak-server --lib a_hosted_digest_reads_the_row_and_refreshes_conditionally` — 1 passed.
- `cargo check -p tidebreak-core --all-targets` — passed.
- `cargo check -p tidebreak-server --all-targets` — passed with existing scripted-harness dead-code warnings.
- `cargo fmt --all --check` — passed.

## Compatibility and risk

No schema or wire-format changes. The guarded update can reject stale 304 writes, which forces a later refresh instead of accepting mismatched cached state.

## Tracking

Closes #2789
Follow-up to #2781.

### DIFF
--- a/crates/tidebreak-core/src/db/ops/code/pull_request.rs
+++ b/crates/tidebreak-core/src/db/ops/code/pull_request.rs
@@ -6,7 +6,7 @@
 //! to a pull request it authored or contributed to (decision 62). GitHub
 //! stays authoritative; these rows record what was observed and when.
 
-use sea_orm::sea_query::OnConflict;
+use sea_orm::sea_query::{Expr, OnConflict};
 use sea_orm::{
     ActiveModelTrait, ColumnTrait, EntityTrait, PaginatorTrait, QueryFilter, QueryOrder,
     QuerySelect, Set,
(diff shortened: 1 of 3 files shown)


naingthet:
* **#2791** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2791?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2791?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

naingthet:
**G2G** at b68019b64

304 fetch-state writes now compare the sent pull ETag in the same UPDATE that stores validators, so a late 304 cannot restore an invalidated etag or overwrite newer endpoint etags. Fresh 200s stay unconditional. The added interleaving covers the fail-closed path.

PR #2794, opened 2026-08-27
fix(code): probe login-shell gh before Homebrew paths

naingthet:
## What changed

Staging desktop publishes fail on the packaged GitHub CLI smoke test. GitHub macOS runners ship `/opt/homebrew/bin/gh`, and discovery used that path before the login-shell probe, so the fake login-shell binary never ran.

Discovery now uses PATH, then the login shell, then Homebrew locations. Packaged apps still find Homebrew `gh` when the login shell does not have it.

## Validation

- `cargo test -p tidebreak-server --lib -- code::gh::tests::resolve_github_uses_gh_url_when_signed_in`

The packaged-app smoke test itself runs only in the staging publish workflow after this lands on `main`.

## Compatibility and risk

None. No wire-format change. Homebrew lookup still runs when PATH and login-shell probing find nothing.

### DIFF
--- a/crates/tidebreak-server/src/code/gh.rs
+++ b/crates/tidebreak-server/src/code/gh.rs
@@ -1205,13 +1205,19 @@ async fn resolve_gh_binary(search_path: Option<&str>) -> Option<PathBuf> {
             binary,
             login_env: None,
         },
-        None => {
-            let capture = probe_shell(&HostEnv::from_process(), "gh").await.ok()?;
-            GhLaunch {
+        None => match probe_shell(&HostEnv::from_process(), "gh").await {
+            Ok(capture) => GhLaunch {
                 binary: capture.binary,
                 login_env: Some(Arc::new(capture.env)),
-            }
-        }
+            },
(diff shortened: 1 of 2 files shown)


naingthet:
* **#2794** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2794?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2794?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

naingthet:
**G2G** at 79bd44381

Discovery now searches process PATH, then the login shell, then Homebrew install locations. `find_gh` no longer short-circuits the probe when `/opt/homebrew/bin/gh` exists, so the packaged-app smoke test can see the login-shell binary. Probe still requires an absolute executable and a sanitized name; Homebrew remains the fallback when PATH and the login shell have no `gh`.

Issue #2746, opened 2026-08-27
fix(code): the codex login probe observes nothing on signed-in machines


naingthet:
## What you see

On a machine where Codex is signed in and sessions run fine, the harness doctor shows Codex as **Unverified**: "Tidebreak could not verify the Codex CLI sign-in. Sign in to Codex CLI in your own terminal, then re-check." — "Signed in: not observed" in the row detail. Observed locally with the pinned 0.147.0 install.

## Root cause

`observe_login` in `crates/tidebreak-harness/src/codex/mod.rs` spawns the pinned binary as `codex login status` with null stdin and piped stdout/stderr, then parses **stdout only**. Under exactly that spawn shape, codex 0.147.0 writes the status line to **stderr**: verified directly against the pinned binary on a signed-in machine — stdout is empty, stderr reads `Logged in using ChatGPT`, exit 0. So the first-non-empty-line matcher never sees a "Logged in"/"Not logged in" prefix and the probe returns `None`, which the doctor renders as "not observed". `observe_version` works because `--version` goes to stdout — which is why the same probe reports a version and no sign-in state.

Ruled out by reproduction: the pinned binary itself (run with a normal or minimal env it prints the parseable line), the child-env filter (HOME+PATH suffice), API-key env mode (the stored ChatGPT login still wins and prints), and the auth timeout (the command returns in ~200 ms).

## Outcome

The login observation classifies the first status-shaped line from either stream, stdout authoritative, and pins that with a unit test on the extracted classifier (mirroring the claude adapter's `auth_status_from_json` test). A signed-in machine then reports Signed in instead of Unverified.

Issue #2741, opened 2026-08-26
fix(code): the harness doctor demands terminal sign-in on hosted machines, where no engine needs it


naingthet:
## What you see

Attached to a hosted machine, the harness doctor lists every engine as unusable: the banner reads "No engine is ready yet. Sign in to one below, then re-check." and each row's remediation is "Sign in to <engine> in your own terminal, then re-check" (or the unverified variant). Nobody can run a terminal on a hosted machine, and since #2650 the local sign-in state is not what hosted sessions authenticate with anyway: Claude Code and Codex turns run on the per-caller inference relay (decision 71 / tidebreak decision 0051), so a signed-out engine is a ready engine on a hosted machine.

Found running the #2613 loop against the hosted machine on v0.70.0.

## Why it happens

`doctor()` in `crates/tidebreak-server/src/routes/code/harnesses.rs` probes each engine's local sign-in (`probe.authenticated`) and writes the terminal-sign-in remediation unconditionally. That probe is the right signal on local machines (#2653 keeps it there) and the wrong one on hosted machines. `spawn_wiring` (`crates/tidebreak-server/src/code/harness_llm.rs`) already knows which engines the relay covers — Claude Code and Codex; opencode and Grok are #2652. The desktop banner (`crates/tidebreak-desktop/ui/src/code/DoctorList.tsx`) then derives "no engine is ready" from `authenticated`.

## Outcome

- On a gateway-hosted machine, the doctor reports relay-covered engines as ready, with copy that says turns run as the caller through the gateway — no sign-in instruction.
- Engines the relay does not cover yet (opencode, Grok until #2652) say plainly that they are not available on hosted machines yet, not "sign in in your own terminal".
- Local machines keep today's behavior unchanged.
- The desktop banner stops claiming no engine is ready when a relay-covered engine is.

## Notes

- The doctor is advisory, not a gate (decision 0034): a hosted Claude Code or Codex session can be created today and the first turn runs through the relay. The defect is the panel's claim, not a blocked create — though if any picker does refuse, that refusal belongs here too.
- Lazy install already exists (decision 0034): engines download when picked or when Download is pressed on the doctor row, via `POST /code/harnesses/{kind}/install`. No install work is needed here.

PR #2806, opened 2026-08-27
fix(code): report gateway-managed credentials in the doctor

### LINKED ISSUE #2749: feat(code): harness readiness must not assume a vendor login — modelctl-managed gateway auth is a working mode
## What you see

The doctor's only sign-in signal is the engine's vendor login (Claude account, ChatGPT login, opencode config), and its only remediation is "sign in in your own terminal." In a Model Gateway deployment the normal state is the opposite: modelctl points each harness at the gateway's compat endpoints with a gateway-issued credential, and nobody signs into the vendor at all. Those machines are fully working and read as not ready — or Unverified, when the observation itself misfires (#2746).

Three working modes exist and the doctor knows one:

1. vendor login — the only mode probed today,
2. modelctl-managed gateway auth — the norm on gateway-managed machines,
3. the hosted relay — #2741, in flight.

## Outcome

- The bundled pin works in every mode, and the machine says so. The pin shares the HOME-scoped config and credential stores with any system install (verified for codex: the pinned binary reads the same ChatGPT login), and the hosted relay wires whatever binary spawns — so no mode requires running a different binary, and the harness surface must never imply it does. Pointing a session at a custom or system binary is parked in `docs/deferred.md` (closed #2748).
- The harness surface reports how each engine authenticates on this machine — vendor login (with method), gateway-managed, relay, or nothing yet — and readiness follows from the mode, not from vendor login alone.
- Remediation matches the mode: a gateway-managed harness never gets "sign in in your own terminal." For vendor-login machines the row shows the connected state, the login method, and the actual remediation command.
- The observation is reliable enough to back the display: #2746 fixes the codex not-observed case, and observation failures should log what the CLI actually printed so the next one is diagnosable from a transcript.

## Notes

- Detection is per-engine config surface: codex's `~/.codex/config.toml` provider base URL, claude's settings and env, opencode's config — the adapters already know these shapes for spawn wiring.
- #2741 covers mode 3's doctor semantics; this issue covers modes 1 and 2 on local machines. #2653 covers refusing legibly at create when no mode works.

naingthet:
Closes #2749

## What changed

The doctor's only sign-in signal was the engine's vendor login, so a machine whose engines are pointed at a Model Gateway read as "Signed out" or "Unverified" — and every picker disabled the engine, on a machine that works.

- **Harness crate.** `observe_auth_mode(kind, env)` returns a `HarnessAuthObservation`: either `Unknown`, or an override with the surface that carried it (captured environment, or the engine's own config). Claude Code and Codex reuse the detection the create path already runs — `ANTHROPIC_*` env plus `~/.claude/settings.json` `apiKeyHelper`/`env`, and `OPENAI_*` env plus `$CODEX_HOME/config.toml` `model_provider`. opencode and Grok answer `Unknown`: `auth_override_present` gives them the benefit of the doubt to avoid refusing a session, and the doctor must not turn that into a credential claim. `auth_override_present` keeps its semantics; the per-engine bodies now delegate to the observation.
- **Server.** `resolve_auth_mode` decides the doctor's mode. A hosted machine still reports the relay. Elsewhere, a confirmed vendor login stays `local_sign_in`; no confirmed login plus an observed override reports the new `HarnessAuthMode::GatewayManaged`, whose readiness is `probe.found` and whose remediation is empty — nobody should be sent after a login this machine does not need.
- **Wire + UI.** `gateway_managed` joins the generated wire types (regenerated, both the desktop and mobile copies). `isHarnessReady` treats it as ready when the binary is there, so pickers, `NewWorkspaceDialog`, `StartSessionPrompt`, `HarnessPicker`, and `CodeHome` all follow. The doctor row leads with a Gateway-managed badge, reads "Credentials are managed for this machine." instead of "Sign in via your terminal", and its Details disclosure says the sign-in is not needed.
- **CLI.** `tidebreak code doctor`'s AUTH column reports the mode — `gw`, `relay`, `n/a` — instead of printing `no` for every machine that has no vendor login.

## What I ran

- `cargo check -p tidebreak-harness -p tidebreak-server -p tidebreak-cli` — clean.
- `cargo test -p tidebreak-harness --lib auth_` — 9 passed, including the two new tests that pin which surface the observation names for claude settings/env and for codex `config.toml`.
- `cargo test -p tidebreak-server --lib routes::code::harnesses` — 4 new tests: an env override yields `gateway_managed` on both the signed-out and unobserved probes, a confirmed login stays local, opencode and Grok stay local, and a hosted machine still reports the relay.
- `UPDATE_WIRE_TYPES=1 cargo test -p tidebreak-server` — full crate suite green, wire types regenerated.
- `pnpm --dir crates/tidebreak-desktop/ui lint`, `biome format`, and `pnpm --dir crates/tidebreak-desktop/ui test` — 247 files, 2124 tests passed, including the new `labels.ts` readiness test and the `DoctorList` row test.
- `pnpm --dir crates/tidebreak-desktop/ui storybook:build` — succeeded; the new `Code/Harness doctor → Gateway managed` story was captured and reviewed in light and dark.

## Scope

This covers mode 2 on local machines. Mode 3's doctor semantics stay with #2741, and the codex not-observed case stays with #2746.

### DIFF
--- a/crates/tidebreak-cli/src/api/code.rs
+++ b/crates/tidebreak-cli/src/api/code.rs
@@ -167,12 +167,32 @@ pub struct HarnessDoctorEntry {
     pub caps: HarnessCaps,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub authenticated: Option<bool>,
+    /// How a session of this engine authenticates on the server's machine.
+    /// A server that predates the field knows only the local sign-in probe.
+    #[serde(default)]
+    pub auth_mode: HarnessAuthMode,
     #[serde(default)]
     pub remediation: String,
     #[serde(default)]
     pub stderr: String,
     #[serde(default)]
     pub unrecognized_event_count: i64,
+}
+
+/// How a session of one engine authenticates on the server's machine.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, Serialize, Deserialize)]
+#[serde(rename_all = "snake_case")]
+pub enum HarnessAuthMode {
+    /// The engine's own local sign-in.
+    #[default]
+    LocalSignIn,
+    /// A credential or endpoint override authenticates the engine without a
+    /// vendor login.
+    GatewayManaged,
+    /// The on-behalf-of relay carries turns as the caller.
+    GatewayRelay,
+    /// The relay does not cover this engine on a hosted machine.
+    HostedUnavailable,
 }
 
 /// Bounded changed-file list for `GET /code/workspaces/{id}/files`.
--- a/crates/tidebreak-cli/src/code.rs
+++ b/crates/tidebreak-cli/src/code.rs
@@ -23,7 +23,7 @@ use crate::api::client::Client;
 use crate::api::code::{
     decode_event_frame, decode_update_notice, is_turn_terminal, supported_caps_summary,
     turn_exit_code, CodeApprovalSnapshot, CodeSessionDigest, CodeSessionSnapshot, CodeTurnSnapshot,
-    CodeUpdateNotice, CodeWorkspaceSnapshot, SubmitTurnResponse,
+    CodeUpdateNotice, CodeWorkspaceSnapshot, HarnessAuthMode, SubmitTurnResponse,
 };
 use crate::connect::Server;
 use crate::print::OutputFormat;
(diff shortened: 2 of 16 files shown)


naingthet:
**G2G** at 3995f4d7f

Doctor reports only auth_mode (gateway_managed / local_sign_in / relay / n/a). Override detection stays boolean; env values, tokens, and config contents are not copied into the report, CLI table, or UI copy.

PR #2807, opened 2026-08-27
feat(code): show the permission mode and enforce the managed ceiling

### LINKED ISSUE #2795: feat(desktop): show the permission mode and enforce the managed ceiling in code
Code sessions run under a permission mode that is invisible outside the create composer, and the managed `permission_mode_ceiling` is enforced only for chat.

- Add a mode chip to the workspace header (`WorkspaceHeader` `sessionStatus` slot), fed from the session's `permission_mode`.
- Render the unsupervised statement (`autoIsUnsupervised()` in `labels.ts`, currently unused) under the create-composer control and in the in-session picker.
- Enforce `permission_mode_ceiling` server-side on code session create and on `POST /code/sessions/{id}/mode`. Today only `turn_worker.rs` clamps chat; the code paths validate harness caps only, and the UI lock is display-only.
- Fix the stale Grok fixture in `labels.test.ts` that asserts Allow is unsupported.

Out of scope: changing the Allow default. That is a recorded amendment to decision 0039 with a revisit condition.

naingthet:
## What changed

A code session's permission mode was invisible outside the create composer, and the managed `permission_mode_ceiling` bound chat only.

**Visibility.** The workspace header now shows the session's mode as a muted chip beside the lifecycle text (`SessionPermissionIndicator`), with a tooltip that names the posture and says what it does. Create picks the most autonomous mode the engine honors, so the header is where a reader learns which one they got.

**The statement.** `autoIsUnsupervised()` had no caller since #2601 trimmed the permission copy, which left the requirement in decision 0039's amendment unmet — the posture is supposed to be stated next to the control. Both create composers state it again, using the same words #2601 removed: `ALLOW_ALL_NOTE` under Allow all, `UNSUPERVISED_AUTO_NOTE` under Auto on an engine with no approval channel. It appears only at create; a live session carries the header chip instead of a line under every draft, and the menu rows stay free of posture descriptions.

**Enforcement.** `POST /code/workspaces/{id}/sessions` and `POST /code/sessions/{id}/mode` now call `refuse_permission_mode_over_ceiling`, the same refusal the chat routes use (409 `permission_mode_locked`). Until now the code paths validated harness capabilities only and the picker's lock was display-only, so any client could post an over-ceiling mode. A stricter mode stays the reader's choice — the policy names a maximum.

**Fixture fix.** `labels.test.ts` asserted a Grok-shaped engine offers only Auto. The Grok adapter has supported `allow_mode` since the captured 1.0.4 probe, so the create default there is Allow. The fixture now matches the adapter.

Out of scope: the Allow-by-default create behavior, which is a recorded amendment to decision 0039 with its own revisit condition.

## Validation

- `cargo test -p tidebreak-server --lib tests::code::` — 112 passed, including the new `a_managed_ceiling_refuses_over_ceiling_code_session_modes` (over-ceiling create and mode change refused, at-or-below accepted)
- `cargo fmt --check -p tidebreak-server`
- `pnpm --dir crates/tidebreak-desktop/ui lint`, `biome format src`, `tsc --noEmit`
- `pnpm --dir crates/tidebreak-desktop/ui test` — 2124 passed

No screenshot pass: this environment has no controllable browser. The `Code/Workspace header` story gained `AsksBeforeEachTool`, `RunsOnItsOwn`, and `PlansOnly` so the chip states can be reviewed there.

Closes #2795

### DIFF
--- a/crates/tidebreak-desktop/ui/src/PermissionModeMenu.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/PermissionModeMenu.dom.test.tsx
@@ -5,7 +5,7 @@ import { toast } from "sonner";
 import { afterEach, expect, it, vi } from "vitest";
 import type { ManagedPolicy, PermissionMode } from "./api";
 import { ManagedPolicyContext } from "./managedPolicy";
-import { PermissionModeMenu } from "./PermissionModeMenu";
+import { clampPermissionMode, PermissionModeMenu } from "./PermissionModeMenu";
 import { useFirstTaskGuide } from "./FirstTaskWalkthrough";
 
 vi.mock("sonner", () => ({ toast: { error: vi.fn() } }));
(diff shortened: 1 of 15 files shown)


naingthet:
**NO GO** at 6e4b1bfcb

1. P2: Create still posts the unclamped engine default while PermissionModeMenu displays the managed ceiling. `postedMode` / `mode` in NewWorkspaceDialog and StartSessionPrompt stay on `defaultCreatePermissionMode` (Allow when the engine honors it); the menu only clamps the label. A reader who never opens the picker submits `permission_mode: allow`, the new routes 409 `permission_mode_locked`, and clicking the already-displayed ceiling does not write state because the menu treats it as selected. Live sessions split the same way: the header chip shows the stored mode the engine launches with, the composer menu shows the ceiling as if a turn gate had clamped it. Chat’s turn gate makes that display true; code has no such clamp, so the picker can show a mode the session is not running under.

naingthet:
Fixed in 4b5260751.

Create and start now clamp the posted mode to the managed ceiling (`clampPermissionMode`), so a reader who never opens the picker submits Ask rather than Allow. The live code picker no longer remaps an over-ceiling stored mode to the ceiling; it shows the posture the engine launched with, matching the header chip.

naingthet:
**G2G** at 4b5260751

Create and start clamp the posted mode to the managed ceiling, so a reader who never opens the picker submits Ask rather than Allow. The live code picker keeps `clampDisplay={false}` and shows the stored launch posture, matching the header chip. The prior mismatch at 6e4b1bfcb is gone.

PR #2780, opened 2026-08-27
fix(desktop): give workspace creation more room

naingthet:
## Summary

- widen the new workspace dialog on desktop
- give the first-message field more vertical room
- wrap repository and creation controls cleanly at the minimum window size
- add focused DOM and Storybook layout checks

## Validation

- `pnpm --dir crates/tidebreak-desktop/ui vitest run src/code/NewWorkspaceDialog.dom.test.tsx` — 27 passed
- `pnpm --dir crates/tidebreak-desktop/ui tsc --noEmit` — passed
- `pnpm --dir crates/tidebreak-desktop/ui lint` — passed
- `pnpm --dir crates/tidebreak-desktop/ui storybook:build` — passed
- `git diff --check` — passed

## Visual review

The Storybook server is available at `http://localhost:6008/`. This session has no connected browser or Playwright runtime, so screenshot inspection remains unverified.

PR #2809, opened 2026-08-27
fix(desktop): keep the workspace layout usable with the inspector open

### LINKED ISSUE #2777: fix(desktop): give the new-workspace composer more room
The new-workspace composer is too crowded after the compact composer changes. Increase the dialog and prompt area without weakening narrow-window behavior, and verify the main picker states in Storybook.

### LINKED ISSUE #2786: fix(desktop): finish code workspace tab and inspector layout validation
PR #2779 restored the center tabs and bounded the right inspector, but its Storybook states were not captured or inspected. The original failure hid the code journal, composer, Source control, and Pull request tabs when the right sidebar opened or restored a bad saved split.

Validate current `main` against the real `react-resizable-panels` behavior. Cover the right sidebar collapsed and open, Source control and Pull request center tabs, invalid saved layout recovery, a normal desktop viewport, and a small laptop viewport. Capture and inspect the Storybook states in all supported themes. If any state still hides or crowds the workspace, fix it and add a regression that fails without the change.

Done when the journal, composer, and center tabs remain visible; the inspector stays within its documented bounds; bad saved layouts reset; focused tests, TypeScript, lint, and Storybook build pass; and the screenshots have been inspected.

naingthet:
## Summary

The code workspace's resize handle collapsed both panels to zero width, so
every horizontal `ResizablePanelGroup` on the page rendered empty. Opening
the review sidebar blanked the journal, composer, and center tabs; the
editor split blanked both editor panes. That is the failure #2786 reports,
and it reproduces on `main` at a full 1280x800 desktop, not only at narrow
widths.

`react-resizable-panels` labels a separator by its own axis, not its group's:
a horizontal group is split by a separator reporting
`aria-orientation="vertical"`. The handle keyed its full-width bar off that
word, so it claimed the whole row and left the panels beside it nothing.

## What changed

- `resizable.tsx`: key the handle's full-width bar off
  `aria-orientation=horizontal`, with a comment on why the matching word is
  the wrong one. Export the class string so a unit test can pin the mapping.
- `inspectorLayout.ts`: add `MIN_WORKSPACE_WIDTH_PX` and
  `fitsInspectorSplit`. The bounds are percentages, so a narrow pane shrinks
  the workspace and the inspector together and no drag wins the journal its
  floor back. Below the derived pane width the inspector stands down.
- `CodeWorkspacePage.tsx`: measure the pane that carries the split and hide
  the inspector when it cannot hold one. The stored preference survives, so
  widening the window brings it back. Drop the inspector recovery effect —
  the sanitizer runs before the group mounts, so the layout the effect read
  was always valid and the branch never fired.
- `WorkspaceHeader.tsx`: when the pane is too narrow, the review control
  dims and its tooltip points at the Source control and Pull request center
  tabs instead.
- `NewWorkspaceDialog.tsx` (#2777): the form scrolls rather than letting a
  short window clip the wrapped Create row. The sizing PR #2780 shipped is
  unchanged; its Storybook checks asserted 800px wide and 200px tall, which
  a 14px root font size never reaches, so they are now measured against the
  interface's own rem (`max-w-4xl`, `sm:min-h-52`).

## Storybook

New states, all with play assertions that the tab strip does not scroll
sideways, the journal clears its floor, and the inspector stays inside
`MAX_INSPECTOR_SIZE`:

- `RestoredInspectorSplit` — a valid saved 60/40 split comes back exactly.
- `SourceControlTabWithReview`, `PullRequestTabWithReview` — a review center
  tab beside the inspector, which no story covered before.
- `MinimumWindowWithReview` — the inspector stands down at 720x480, the
  journal takes the pane, and the header control reports why.
- `RepoPicker`, `ModelPicker`, `BaseRefPicker`, and an opened engine menu on
  `EnginesNeedSetup` — the composer's picker states (#2777).

Inspector stories now pin the desktop viewport instead of inheriting the
canvas.

## Validation

- `pnpm --dir crates/tidebreak-desktop/ui exec biome format src` — no fixes
- `pnpm --dir crates/tidebreak-desktop/ui lint` — no fixes
- `npx tsc --noEmit` — clean
- `pnpm --dir crates/tidebreak-desktop/ui test` — 248 files, 2127 tests passed
- `pnpm --dir crates/tidebreak-desktop/ui storybook:build` — completed

Screenshots captured through headless Chromium against a local Storybook and
inspected: the review split, restored split, rejected stored layout, Source
control and Pull request beside the inspector, the minimum window, the split
editor pane, and the four composer picker states, in light and dark. Before
and after shots confirm the blank pane and its fix.

Closes #2786
Closes #2777

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeWorkspacePage.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeWorkspacePage.dom.test.tsx
@@ -37,6 +37,7 @@ import { useCodeUiStore } from "./CodeUiStore";
 import { resetWorkflowPromptStore } from "./workflowPrompts";
 import { disconnectCodeUpdates, useCodeUpdatesStore } from "./CodeUpdatesStore";
 import { CodeWorkspacePage } from "./CodeWorkspacePage";
+import { MIN_INSPECTOR_PANE_WIDTH_PX } from "./inspectorLayout";
 import {
   forkFraming,
   forkTranscriptFile,
@@ -655,6 +656,35 @@ beforeEach(() => {
   vi.useRealTimers();
 });
 
+/**
+ * Make every observed element report this width.
+ *
+ * jsdom measures nothing, so the shared stub answers with a desktop-sized
+ * box. A test that cares about a narrow pane swaps in its own for the
+ * duration; `vi.stubGlobal` is undone with the other stubs after each test.
+ */
+function stubPaneWidth(width: number) {
+  vi.stubGlobal(
+    "ResizeObserver",
+    class {
+      constructor(private readonly callback: ResizeObserverCallback) {}
+      observe(target: Element) {
+        this.callback(
+          [
+            {
+              target,
+              contentRect: { width, height: 768 },
+            } as unknown as ResizeObserverEntry,
+          ],
+          this as unknown as ResizeObserver,
+        );
+      }
+      unobserve() {}
+      disconnect() {}
+    },
+  );
+}
+
 afterEach(() => {
   cleanup();
   resetCodeSessionRegistry();
@@ -674,6 +704,7 @@ afterEach(() => {
   browserMocks.close.mockClear();
   window.localStorage.clear();
   vi.restoreAllMocks();
+  vi.unstubAllGlobals();
   vi.clearAllMocks();
 });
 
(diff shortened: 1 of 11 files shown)


naingthet:
**G2G** at a7c2a893f

The handle now sizes from the separator axis, so a horizontal group keeps a vertical hairline instead of claiming the row. The inspector stands down below the derived pane floor and the stored preference survives.

PR #2816, opened 2026-08-27
fix(dev): stage sidecars before Tauri startup

naingthet:
## What changed

- Stage the host broker and Tidebreak CLI sidecars before Tauri starts its frontend wait loop.
- Reuse those prepared sidecars inside `beforeDevCommand` so Cargo lock contention cannot surface as a Vite lifecycle failure.
- Stop duplicate launches before installation or after sidecar preparation when port 1420 becomes occupied.
- Resolve Cargo’s effective target directory so worktree-specific target wrappers remain supported.

## Validation

- `node --test scripts/dev.test.mjs scripts/workflow-security.test.mjs` — 48 tests passed.
- `shellcheck scripts/dev.sh`
- `bash -n scripts/dev.sh`
- `node --check crates/tidebreak-desktop/scripts/prepare-sidecar.mjs`
- `git diff --check`
- Real startup reached sidecar completion; a concurrent main-checkout launch then claimed port 1420, and the new guard reported its PID without entering Tauri.

## Compatibility and risk

No persisted data, wire format, or packaged release behavior changes. Direct `cargo tauri dev` still prepares sidecars through the existing hook.

### DIFF
--- a/scripts/dev.sh
+++ b/scripts/dev.sh
@@ -18,6 +18,7 @@ desktop_dir="$repo_root/crates/tidebreak-desktop"
 ui_dir="$desktop_dir/ui"
 
 missing=()
+command -v node >/dev/null || missing+=("Node.js — https://nodejs.org/")
 command -v pnpm >/dev/null || missing+=("pnpm — https://pnpm.io/installation")
 cargo tauri --version >/dev/null 2>&1 ||
   missing+=("Tauri CLI 2 — cargo install tauri-cli --version \"^2\"")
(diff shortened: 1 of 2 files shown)


naingthet:
* **#2816** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2816?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2816?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

naingthet:
**G2G** at 9e33fa42a

Sidecar staging happens before `cargo tauri dev`, `beforeDevCommand` is overridden to Vite only so the frontend wait loop no longer shares a Cargo lock, and the 1420 guard runs both before install and after prep. `CARGO_TARGET_DIR` from `cargo metadata` matches what `prepare-sidecar.mjs` already honors.

Issue #2800, opened 2026-08-27
refactor(code): route the delivery aggregate through the PR fact store


naingthet:
The delivery aggregate is still a second fetcher: `gh pr list` behind a 30-second cache, written back through `save_pull_request_fact`, which clears `pull_etag` on every reconcile pass and destroys the hot fetcher's validator — the next conditional fetch pays a full 200 instead of a 304.

Route the aggregate through the `code_pull_request` row (read the fact plus live tier, fetch only stale identities through the gated fetcher). That deletes the list cache, stops the validator churn, and collapses the reconcile sweep into a plain row sweep. Related: #2578 covers the delivery monitor poll, not this path.

naingthet:
This slice is in https://github.com/brightwave-inc/tidebreak/pull/2848.

PR #2817, opened 2026-08-27
fix(code): keep PR state writes behind the fetch-state validator

### LINKED ISSUE #2799: fix(code): close the late-write and stale-digest holes in PR state
Follow-through on #2781/#2791: the fetch-state write is guarded but the state the UI renders is not.

- The live-tier and workspace-column writes run before the guarded fetch-state write and are unconditional, so the late-304 interleaving #2791 rejects still lands its reconstructed digest on every surface. Condition them on the same validator and return nothing when the row refused the write.
- The watch certifies the workspace column with the fact row's `observed_at` even when the column write was dropped. Read the digest from the fact row itself.
- A fix turn's own push never dirties the row (the post-turn detector neither refreshes nor marks hot), so the watch compares against a pre-push head and parks with "a fix attempt did not resolve the problem". Mark the workspace hot when the detector sees a push.
- The delivery page ignores the `delivery` broadcast and refetches only on user action. Subscribe it to `deliveryRevision`.

naingthet:
Follow-through on #2781 and #2791: the fetch-state write is guarded, but the state the UI renders was not.

**The live tier and the workspace column now ride the same validator.** `fetched_workspace_digest` wrote the live tier and returned the digest before the guarded `set_pull_request_fetch_state` call, and both of those writes were unconditional. The exact late-304 interleaving #2791 rejects still landed its reconstructed digest on the fact row's live tier, on the workspace column, and on every surface that reads them. The fetch-state write now goes first, and a conditional read whose row moved under it returns `None` — the callers then skip the column write and the broadcast.

**The watch reads the row that dates it.** `fresh_workspace_digest` certified the workspace *column* with the fact *row*'s `observed_at`, so a dropped column write left the watch assessing a stale head and stale checks against a fresh clock. It now derives the digest from the fact row, keeping the terminal state the column already holds so a merged pull request never reopens for the watch.

**A fix turn's own push marks the workspace hot.** Only route mutations dirtied the row; the post-turn detector neither refreshed nor marked. The watch therefore compared `last_fix_head` against a pre-push head, decided its own fix turn was a repeat, and parked with "a fix attempt did not resolve the problem". The detector now marks the workspace hot when it confirms a push or a create, so the next hot pass reads the head the turn just moved. The hot tier moved into `pr_refresh` as a shared handle to make that reachable without a runtime reference.

**The delivery list is a projection of the `delivery` broadcast.** It refetched only on a local revision bump or a user action. Its effect now depends on `deliveryRevision`.

Routing the delivery aggregate through the fact row — `LIST_CACHE_TTL` and `pull_etag` clearing — stays out of scope and is filed as #2800.

## Tests

Each fails without its fix:

- `a_conditional_read_that_lost_its_validator_writes_nothing` (`code_git.rs`) drives the real interleaving: the forge shim lands a fresher 200 on the row while the conditional request is open, then answers 304. Asserts the rejected digest reaches neither the live tier nor the workspace column.
- `fresh_workspace_digest_reads_the_row_that_dates_it` (`watch.rs`) pins the head and the checks to the fact row.
- `a_confirmed_push_marks_the_workspace_hot` and `a_turn_that_pushed_nothing_leaves_the_hot_tier_alone` (`code_pr_facts.rs`) pin the detector's mark.
- `re-reads when the server nudges the delivery channel` (`CodeDeliveryPage.dom.test.tsx`).

Closes #2799

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeDeliveryPage.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeDeliveryPage.dom.test.tsx
@@ -35,10 +35,12 @@ import {
   codeDeliverySearchFrom,
 } from "./CodeDeliveryPage";
 import { useCodeDeliveryStore } from "./CodeDeliveryStore";
+import { useCodeUpdatesStore } from "./CodeUpdatesStore";
 
 afterEach(() => {
   cleanup();
   useCodeDeliveryStore.getState().reset();
+  useCodeUpdatesStore.getState().reset();
 });
 
 vi.mock("@/openInBrowser", () => ({ openInBrowser: vi.fn() }));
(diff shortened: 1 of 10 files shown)


naingthet:
**G2G** at 1e01a7060

A rejected 304 now returns before the live-tier write, so the workspace column and broadcast never take the reconstructed digest. The watch dates the fact row, a confirmed push marks the workspace hot, and the delivery list re-reads on the delivery nudge.

PR #2818, opened 2026-08-27
feat(code): make repo hooks and quick actions reachable

### LINKED ISSUE #2798: feat(code): make repo hooks editable and surface setup failures
The repo lifecycle hooks (setup script, archive script, quick actions) run server-side but are unreachable and their failures are invisible.

- `quick_actions` has no write path: absent from both the create and patch repo bodies, and `deny_unknown_fields` rejects it. Add it to both and regenerate the wire types.
- No repo settings surface exists; `patchCodeRepo` has zero callers. Build a repo settings section on the delivery page (next to trigger rules): base ref, branch prefix, setup script, archive script, quick actions.
- `setup_failed` never renders: the "Setup failed" label has no consumers and cards fall through to idle. Show it on workspace cards and the sidebar with critical tone.
- The new-workspace dialog deletes the optimistic card on `setup_failed` and "Try again" creates a second worktree. Keep the card, refetch, and open the workspace instead.
- Stretch: `POST /code/workspaces/{id}/retry-setup` to re-run setup on the existing worktree.

naingthet:
The repo lifecycle hooks — setup script, archive script, quick actions — ran
server-side but nothing could write them and their failures never showed.
This closes that loop.

## Quick actions had no write path

`quick_actions` was absent from both `CreateRepoBody` and `PatchRepoBody`,
and `deny_unknown_fields` turned a client that sent one into a 422. The
subsystem was unreachable in production. Both bodies now take the field, and
`normalize_quick_actions` trims it and refuses what a workspace could not
run: a blank name, a blank command, or two actions under one name — the
second of a duplicate pair is unreachable, because `run_named_action` looks
an action up by name.

On patch the list is replaced whole. Omitting the field leaves the stored
list alone; an empty list clears it.

The list is bounded: at most 32 actions, names to 64 characters, commands to
1024. Every entry renders in the workspace header menu and the command
palette, so a list past that is unusable before it is expensive, and the
refusal names the limit so a client can fix the body without guessing. The
editor disables Add at the cap rather than letting a 33rd row earn a 400.
Bounding the list is also what lets the normalizer grow its vectors instead
of sizing them from `actions.len()` — CodeQL flagged those two
`Vec::with_capacity` calls as allocations derived from a request body
(alerts 190 and 191).

## A repo settings surface

`patchCodeRepo` had zero callers. `RepositorySettings` is the one place that
writes a repo's base ref, branch prefix, setup script, archive script, and
quick actions. It sits beside the trigger rules in the tracked-repositories
dialog, behind a Settings button on each registered row.

Text commits on blur, switches commit on change, so there is no Save button
to forget. A row still being typed stays on screen and off the wire until it
has both a name and a command, so adding one never blocks saving the rest. A
write that lands mid-flight queues behind the one in progress rather than
being dropped.

## `setup_failed` is visible

The "Setup failed" label had no consumers and the card fell through to idle,
so a workspace whose setup script broke read as an ordinary quiet one. The
card now carries a critical line, the accessible name says it too, and the
by-status rail gets a rank between done-unreviewed and idle: below live work,
because the checkout survived, but above idle, because nothing else says the
script never finished.

## The dialog no longer cuts a second worktree

On `setup_failed` the new-workspace dialog deleted the optimistic card and
offered "Try again", which created a second worktree for work the user
already had. It now keeps the card, finds the row the server wrote, and opens
it — the shape `runRestore` already used.

## `POST /code/workspaces/{id}/retry-setup`

`require_live_workspace` refuses a `setup_failed` workspace for everything,
so the state had no exit short of archiving the work. This runs the script
again on the worktree the workspace already has: Active on success, still
`setup_failed` and a 422 on another failure. It is offered as "Retry setup"
on the card menu, the workspace header, and the command palette.

## Tests

Server: `quick_actions_round_trip_through_create_and_patch` covers create,
patch, trimming, the duplicate and blank-command refusals, the 32-action cap
(over, and exactly at), the name and command length caps, and clear-vs-omit.
`retry_setup_revives_a_setup_failed_workspace_in_place` covers the failing
retry, the fixed retry, the same worktree path, and the no-op on an active
workspace. `workspace_setup_failure_preserves_the_checkout` still passes.

UI: `RepositorySettings.dom.test.tsx` (blur save, quick-action write, no
write on a bare focus change, unregistered repo), the `setup_failed` card
branch in `CodeSidebar.dom.test.tsx`, the rank in `workspaceCards.test.ts`,
and the dialog branch in `NewWorkspaceDialog.dom.test.tsx`.

Storybook: `Code/Repository settings` (configured, empty, loading, not
registered, load failed) and `Code/Workspace card` → `Setup failed`.

## Ran

`cargo check -p tidebreak-server`, `cargo clippy -p tidebreak-server
--all-targets` (only pre-existing `scripted_harness.rs` dead-code warnings),
`cargo fmt --all --check`, the three server tests above, `pnpm format`,
`pnpm lint`, `tsc --noEmit`, the full `pnpm test` (248 files, 2130 tests),
and `pnpm storybook:build`. `UPDATE_WIRE_TYPES=1` produced no wire diff —
the request bodies are `Deserialize`-only and not TS-exported — and both
`wire.ts` copies are untouched.

I could not screenshot the new stories: no Playwright package or browser MCP
is available in this environment, so the visual review is unverified.

Closes #2798

### DIFF
--- a/crates/tidebreak-desktop/ui/src/api/client.ts
+++ b/crates/tidebreak-desktop/ui/src/api/client.ts
@@ -84,6 +84,7 @@ import {
   type TaskPlan,
   type CodeApprovalSnapshot,
   type CodeRepoSnapshot,
+  type QuickAction,
   type CodeSessionSnapshot,
   type CodeTurnSnapshot,
   type CodeActionSnapshot,
@@ -2090,6 +2091,7 @@ export class ApiClient {
     branch_prefix?: string;
     setup_script?: string;
     archive_script?: string;
+    quick_actions?: QuickAction[];
   }): Promise<CodeRepoSnapshot> {
     return requireParsed(
       parseCodeRepo(
@@ -2122,6 +2124,8 @@ export class ApiClient {
       branch_prefix?: string;
       setup_script?: string | null;
       archive_script?: string | null;
+      /** The whole list, replaced. Omit to leave the stored list alone. */
+      quick_actions?: QuickAction[];
     },
   ): Promise<CodeRepoSnapshot> {
     return requireParsed(
@@ -2421,6 +2425,29 @@ export class ApiClient {
       parseCodeWorkspace(
         await this.json(
           `/code/workspaces/${encodeURIComponent(workspaceId)}/restore`,
+          {
+            method: "POST",
+            headers: this.headers(true),
+          },
+        ),
+      ),
+      "code workspace",
+    );
+  }
+
+  /**
+   * Run the repo's setup script again on a workspace that is `setup_failed`.
+   * The worktree is the one it already has, so a success takes the workspace
+   * Active without cutting a second checkout. A second failure comes back as
+   * 422 `setup_failed`.
+   */
+  async retryCodeWorkspaceSetup(
+    workspaceId: string,
+  ): Promise<CodeWorkspaceSnapshot> {
+    return requireParsed(
+      parseCodeWorkspace(
+        await this.json(
+          `/code/workspaces/${encodeURIComponent(workspaceId)}/retry-setup`,
           {
             method: "POST",
             headers: this.headers(true),
--- a/crates/tidebreak-desktop/ui/src/api/types.ts
+++ b/crates/tidebreak-desktop/ui/src/api/types.ts
@@ -148,6 +148,7 @@ import {
   type CodeApprovalState as WireCodeApprovalState,
   type CodeEvent as WireCodeEvent,
   type CodeRepoSnapshot as WireCodeRepoSnapshot,
+  type QuickAction as WireQuickAction,
   type CodeSessionId as WireCodeSessionId,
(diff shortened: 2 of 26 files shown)


naingthet:
**G2G** at d848582e4

Retry-setup, repo patch, and quick-action writes stay on owner-scoped `/code` routes behind the bearer. Setup still runs only from create, restore, and the new retry path; auto-run actions still follow a successful setup, not a skipped gate.

naingthet:
**G2G** at 839c306ae

Quick-action writes now cap the list at 32 and bound name/command length before store, and the settings editor stops adding at that same limit. Retry-setup, repo patch, and the setup-failed card path are unchanged from the last pass: owner-scoped `/code` routes behind the bearer, setup only from create/restore/retry, auto-run only after a successful setup.

PR #2819, opened 2026-08-27
feat(desktop): open workspace files in an external editor

### LINKED ISSUE #2796: feat(desktop): open workspace files in an external editor
Every neighboring tool has open-in-editor; Tidebreak has only Reveal in Finder.

- New native command following the `open_code_worktree` pattern: takes `workspaceId` (+ optional `relativePath`, `line`), resolves the path server-side, and spawns from a closed set of editor launchers (VS Code `code --goto`, Cursor, Zed, JetBrains, custom command). Do not widen the Tauri shell-open scope and do not grant `shell:allow-execute`; add the command to `native-only-commands.txt`.
- Editor preference in local storage with a settings control in the Coding harnesses panel; detect installed editors natively.
- Surface as a workspace command (sidebar menu, header overflow, palette) and from the file viewer, diff panel, and search rows with path and line.

naingthet:
Every neighbouring tool can hand a file to the editor you already use. Tidebreak could only reveal the worktree in Finder.

Closes #2796

## What lands

A new native command, `open_in_editor`, follows the `open_code_worktree` pattern. The renderer names a workspace id and a path relative to that workspace — never a path to run. The command re-reads the worktree root from the embedded server, canonicalizes both sides before comparing them, and refuses traversal, absolute paths, and symlinks that leave the worktree. It also refuses outright while this window is attached to another machine.

The editor comes from a closed set of launcher plans, so no renderer input reaches a shell:

| Editor | Arguments |
| --- | --- |
| VS Code, Cursor | `--goto <path>:<line>` |
| Zed | `<path>:<line>` |
| JetBrains | `--line <n> <path>` |
| Custom | `<path>` |

A custom editor is held to the same shape as the built-ins: one absolute program path, spawned directly with the file as its only argument. A bare name is refused, because resolving one is a shell's job. `plugins.shell.open` keeps its https-only scope and nothing grants `shell:allow-execute`.

`detect_external_editors` reports which editors this computer has, searching `PATH` and then the known install locations — a desktop app started from Finder inherits a minimal `PATH`, so the probe cannot rely on it alone.

## Where it shows up

The choice is a local-storage module store in the `theme.ts` shape, keyed `tidebreak.externalEditor`, with a picker in Settings → Coding harnesses that says whether each editor is installed and where. Every surface then names the editor rather than saying "editor": the sidebar card menu, the workspace header overflow, the command palette, and a button in the file viewer and diff headers, which pass the path and the line already on screen.

## Decision record 7

This adds two lines to `native-only-commands.txt`, as an amendment to [ADR 0007](docs/decisions/0007-cli-headless-feature-parity.md) rather than a formality. Both commands act on the machine the reader is sitting at, which is not necessarily the machine the server runs on: `open_in_editor` raises a desktop editor here, and `detect_external_editors` looks at this computer's `PATH` and application folders. A headless server has neither, so neither belongs behind a server route.

## Checks

- `cargo check -p tidebreak-desktop`, `cargo test -p tidebreak-desktop` — 280 passed, including `command_parity` and the eight new `code_editor` tests covering path resolution, the symlink escape, per-editor line formatting, and custom-program validation.
- `pnpm exec biome format --write src`, `pnpm exec biome lint src`, `pnpm exec tsc --noEmit`, `pnpm exec vitest run src/code src/settings` — 1056 passed.
- `pnpm --dir crates/tidebreak-desktop/ui storybook:build`, plus a screenshot pass over the new stories in light and dark.

### DIFF
--- a/crates/tidebreak-desktop/native-only-commands.txt
+++ b/crates/tidebreak-desktop/native-only-commands.txt
@@ -29,6 +29,8 @@ is_window_focused: reports main-window focus; document.hidden is the tab, not th
 present_native_notification: posts an OS banner and asks for notification permission on first unfocused present.
 code_browser_command: creates and controls sandboxed native child webviews whose bounds and lifecycle exist only in the desktop shell.
 open_code_worktree: opens a local Code worktree through this operating system's folder handler; a headless server cannot open desktop UI.
+open_in_editor: starts the editor installed on the machine the reader is sitting at, on one file inside a local worktree; the server may be that machine or another, and neither can raise a desktop editor here.
+detect_external_editors: probes this computer's PATH and application folders for installed editors, which only the client machine can see.
 
 # Attaching this client to a machine it does not host (ADR 0047)
 remote_machine_state: reports which machine this client is attached to, from shell-held state that exists before any machine is reachable.
--- a/crates/tidebreak-desktop/src/code_editor.rs
+++ b/crates/tidebreak-desktop/src/code_editor.rs
@@ -1,0 +1,589 @@
+//! Native open-in-editor for one file inside a Code workspace's worktree.
+//!
+//! The sibling `code_worktree` module hands a whole worktree to the operating
+//! system's folder handler. This one goes a step further: it opens a single
+//! file, at a line, in the editor the reader already uses. The renderer never
+//! supplies a path to run — it names a workspace and a path relative to that
+//! workspace, and this module re-reads the worktree root from the embedded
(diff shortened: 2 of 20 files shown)


naingthet:
**G2G** at 698c102dc

Native `resolve_target` re-reads the worktree from the embedded server, refuses empty/absolute paths and any non-`Normal` component (`..`, `.`), canonicalizes both sides, and rejects symlink escapes. The renderer names a workspace and a relative path; it never supplies a path to run. Custom editors take an absolute program and spawn it with no shell.

PR #2821, opened 2026-08-27
docs(decisions): propose a path for parallel turns in one workspace

naingthet:
Decision 0076 proposes how code mode should handle a second turn in a workspace whose checkout is already busy. It states what happens today — one worktree per workspace, the per-workspace turn lock from record 55, and the durable queue from record 69 — then lays out four options with their real costs: keep strict serialization and make the wait legible, run two turns optimistically in one worktree, fork the second turn into an ephemeral checkout branched at the holder's newest checkpoint, or give every session its own worktree. The recommendation names a default (keep the lock, spend the effort on queue visibility, treat fork-on-contention as the shape to design toward) so there is something concrete to argue with. This is a product call, so the record stays `Proposed` and is opened as a draft for the repository owner to accept, amend, or reject.

### DIFF
--- a/docs/decisions/0076-parallel-turns-in-one-workspace.md
+++ b/docs/decisions/0076-parallel-turns-in-one-workspace.md
@@ -1,0 +1,226 @@
+# 76. Parallel turns in one code workspace
+
+- Status: Proposed
+- Date: 2026-08-27
+- Owners: code mode
+- Related: [`0032`](0032-code-workspaces-worktrees-checkpoints.md),
+  [`0053`](0053-code-worktrees-live-in-a-user-visible-root.md),
+  [`0055`](0055-multiple-sessions-per-workspace.md),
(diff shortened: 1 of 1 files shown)


naingthet:
Closing unmerged. Decision 0055 already settled this.

Its alternatives section considered running turns in parallel with no lock and rejected it, on the grounds that it changes what a workspace is. Nothing has changed since, so a second record restating that conclusion adds noise rather than a decision.

One correction worth keeping, from reading the code while drafting this: 0055 justifies the lock partly by per-turn checkpoints racing for the git index, and that reason does not hold. Checkpoints build their tree through a private temporary index and never touch `.git/index`, and contention on the reusable index already falls back to a cold one. The lock stands on its other reasons — keeping two harnesses off the same files, being the checkout reservation that makes the send-versus-queue split correct, and keeping host git actions out of a running turn.

PR #2822, opened 2026-08-27
feat(code): restore a workspace to a turn checkpoint

### LINKED ISSUE #2797: feat(code): restore a workspace to a turn checkpoint
Per-turn checkpoints exist as refs (decision 0032) but nothing can restore one; the substrate has no consumer.

- `restore_checkpoint` in the checkpoint module: write a pre-restore safety snapshot ref, apply the target tree through a temp index, delete files added since the checkpoint, leave ignored files alone.
- Runtime path takes the worktree turn lock, refuses non-active workspaces, returns the busy conflict while a turn holds the lock, and journals a restore event.
- Route: `POST /code/workspaces/{id}/restore-checkpoint` (`/restore` is taken by un-archive).
- UI: a "Restore to this point" action in the per-turn menu with a confirm dialog. Restore files only; the branch and HEAD stay where they are, and the copy says so.

naingthet:
Per-turn checkpoints have existed as hidden refs since [record 32](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0032-code-workspaces-worktrees-checkpoints.md), but nothing could read one back. This adds the restore path end to end.

## What it does

`restore_checkpoint` snapshots the worktree into a pre-restore safety ref, deletes the files that postdate the checkpoint, then writes the checkpoint's tree through a private index. Files only: `HEAD`, the branch, the reflog, and the user's index are untouched, and ignored files sit in neither tree so they survive.

Rename detection is off when computing what to delete. `--find-renames` folds a disappearing path into an `R` record, so the old copy would be left beside the restored one.

The runtime path takes the workspace's turn lock without waiting — a restore landing mid-turn would pull files out from under a running engine — so contention answers `409 workspace_busy`. It journals `CheckpointRestored` next to `CheckpointRecorded`.

The route is `POST /code/workspaces/{id}/restore-checkpoint`. `/restore` already reactivates an archived workspace.

In the transcript, the per-turn seam menu gains "Restore to this point" behind a confirmation that says how far the restore reaches: ignored files stay, and nothing moves the branch.

## Tests

- `a_restore_puts_untracked_files_back_and_keeps_a_safety_ref` — the round trip an untracked file gets wrong twice over: one that existed at the checkpoint comes back, one created since goes, an ignored build artifact survives, and `HEAD` does not move.
- `restoring_a_turn_checkpoint_rewinds_the_files_and_journals_it` — the same through the route, including the safety ref and the journal row.
- `a_running_turn_refuses_a_checkpoint_restore` — a turn parked on an approval holds the checkout, so the restore answers the busy conflict.
- `a_turn_without_a_checkpoint_cannot_be_restored` — no ref, and a reaped ref.
- Two renderer tests for the seam menu: the confirmation states the limits and nothing reaches the host until the reader accepts; cancelling calls nothing.
- A `Code/Turn review card` story showing the full per-turn menu and its confirmation.

## Checks

`cargo check -p tidebreak-server`, `cargo fmt --all --check`, the new server tests plus the rest of `code::checkpoint::tests` (32 passed), `UPDATE_WIRE_TYPES=1` and `UPDATE_CODE_JOURNAL_FIXTURE=1` regeneration, biome format and lint, the full renderer vitest suite (2152 passed), `tsc --noEmit`, and `storybook:build`. The mobile copy of `wire.ts` is byte-identical to the generated one.

Closes #2797

### DIFF
--- a/crates/tidebreak-core/fixtures/code-journal-events.json
+++ b/crates/tidebreak-core/fixtures/code-journal-events.json
@@ -107,6 +107,16 @@
     }
   },
   {
+    "type": "checkpoint_restored",
+    "turn_id": "00000000-0000-0000-0000-000000000001",
+    "diffstat": {
+      "files": 3,
+      "insertions": 1,
+      "deletions": 12,
+      "truncated": false
+    }
+  },
+  {
     "type": "harness_notice",
     "level": "warning",
     "message": "unrecognized event type counted"
--- a/crates/tidebreak-core/src/code/event.rs
+++ b/crates/tidebreak-core/src/code/event.rs
@@ -370,6 +370,17 @@ pub enum CodeEvent {
         /// Bounded diffstat.
         diffstat: Diffstat,
     },
+    /// The worktree was put back to a per-turn checkpoint.
+    ///
+    /// Files only. The branch and `HEAD` do not move, and ignored files are
+    /// untouched, so the journal records what the reader actually asked for
+    /// rather than implying a history rewrite.
+    CheckpointRestored {
+        /// The turn whose checkpoint the worktree now holds.
+        turn_id: CodeTurnId,
+        /// Bounded diffstat of what the restore changed.
+        diffstat: Diffstat,
+    },
     /// Visible degradation or an engine-native notice.
     HarnessNotice {
         /// Severity.
@@ -431,8 +442,9 @@ mod tests {
             CodeEvent::TurnFailed { .. } => 12,
             CodeEvent::TurnInterrupted => 13,
             CodeEvent::CheckpointRecorded { .. } => 14,
-            CodeEvent::HarnessNotice { .. } => 15,
-            CodeEvent::AttentionChanged { .. } => 16,
+            CodeEvent::CheckpointRestored { .. } => 15,
+            CodeEvent::HarnessNotice { .. } => 16,
+            CodeEvent::AttentionChanged { .. } => 17,
         }
     }
 
@@ -527,6 +539,15 @@ mod tests {
                     insertions: 10,
                     deletions: 1,
                     truncated: true,
+                },
+            },
+            CodeEvent::CheckpointRestored {
+                turn_id: CodeTurnId(id(1)),
+                diffstat: Diffstat {
+                    files: 3,
+                    insertions: 1,
+                    deletions: 12,
+                    truncated: false,
                 },
             },
             CodeEvent::HarnessNotice {
--- a/crates/tidebreak-desktop/ui/src/api/client.ts
+++ b/crates/tidebreak-desktop/ui/src/api/client.ts
@@ -88,6 +88,7 @@ import {
   type CodeSessionSnapshot,
   type CodeTurnSnapshot,
   type CodeActionSnapshot,
+  type CodeCheckpointRestore,
   type CodeCommitSnapshot,
   type CodePushSnapshot,
   type CodeTerminalRead,
@@ -169,6 +170,7 @@ import {
   type CodeTurnSubmission,
   parseQueuedCodeTurn,
   parseCodeAction,
+  parseCodeCheckpointRestore,
   parseCodeCommit,
   parseCodePush,
   parseCodeTerminal,
@@ -3111,6 +3113,33 @@ export class ApiClient {
         ),
       ),
       "code quick action",
+    );
+  }
+
+  /**
+   * Put the workspace's files back to what a turn's checkpoint holds.
+   *
+   * Not `restoreCodeWorkspace`, which reactivates an archived workspace. This
+   * one moves files only: the branch and `HEAD` stay where they are, and
+   * ignored files are left alone. A 409 `workspace_busy` means a turn holds
+   * the checkout; retry once it ends.
+   */
+  async restoreCodeWorkspaceCheckpoint(
+    workspaceId: string,
+    turnId: string,
+  ): Promise<CodeCheckpointRestore> {
+    return requireParsed(
+      parseCodeCheckpointRestore(
+        await this.json(
+          `/code/workspaces/${encodeURIComponent(workspaceId)}/restore-checkpoint`,
+          {
+            method: "POST",
+            headers: this.headers(true),
+            body: JSON.stringify({ turn_id: turnId }),
+          },
+        ),
+      ),
+      "checkpoint restore",
     );
   }
 
--- a/crates/tidebreak-desktop/ui/src/api/types.ts
+++ b/crates/tidebreak-desktop/ui/src/api/types.ts
@@ -174,6 +174,7 @@ import {
   type CodeWatchSnapshot as WireCodeWatchSnapshot,
   type CodeWatchState as WireCodeWatchState,
   type CodeActionSnapshot as WireCodeActionSnapshot,
+  type CodeCheckpointRestore as WireCodeCheckpointRestore,
   type CodeCommitSnapshot as WireCodeCommitSnapshot,
   type CodePushSnapshot as WireCodePushSnapshot,
   type CodeFileChange as WireCodeFileChange,
@@ -1240,6 +1241,8 @@ export type CodeSubagentStatus = WireCodeSubagentStatus;
 export type CodeCommitSnapshot = WireCodeCommitSnapshot;
 export type CodePushSnapshot = WireCodePushSnapshot;
 export type CodeActionSnapshot = WireCodeActionSnapshot;
+/** What restoring a workspace to a turn's checkpoint changed. */
+export type CodeCheckpointRestore = WireCodeCheckpointRestore;
 export type PullRequestDigest = WirePullRequestDigest;
 export type PullRequestCheck = WirePullRequestCheck;
 export type PullRequestCheckBucket = WirePullRequestCheckBucket;
--- a/crates/tidebreak-desktop/ui/src/code/CodeSessionReducer.test.ts
+++ b/crates/tidebreak-desktop/ui/src/code/CodeSessionReducer.test.ts
@@ -2088,6 +2088,51 @@ describe("file_changed", () => {
     });
     expect(state.contentRevision).toBe(3);
   });
+
+  /**
+   * A restore rewrites the worktree, so git status, the file list, and the
+   * diff are all stale the moment it lands. `contentRevision` is the only
+   * signal those views watch: without this the frame falls through the
+   * reducer's default branch and every one of them keeps the pre-restore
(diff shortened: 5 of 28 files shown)


naingthet:
**G2G** at 6e870d330

Checked the restore path end to end: files-only checkout through a private index, `--no-renames` on the delete set, path escape refusal, try-lock `workspace_busy` instead of overlapping a running turn, and workspace/session ownership before the write.

naingthet:
Hold for manual merge. Keep this feature pull request open after checks and review pass; do not enable auto-merge.

naingthet:
**NO GO** at 6e870d330
Depth: deep

1. A successful restore leaves every worktree view stale. `checkpoint_restored` is journaled, then dropped on the reducer's default branch, so `contentRevision` never moves. `restoreToTurn` only toasts. Git status, the file list, and the diff keep the pre-restore tree until some later checkpoint happens to bump them. Handle `checkpoint_restored` the way you handle `checkpoint_recorded`, and refresh from the HTTP success too: a journal row that fails to land is already non-fatal.

2. The reader cannot undo or recover, and a failed restore can already have destroyed the tree. You snapshot a safety ref, delete the post-checkpoint paths, then `read-tree`. If apply fails after those deletes, the client reports a generic failure while the files are already gone. `RestoredWorkspaceCheckpoint` drops `safety_ref`, the toast never names it, and the confirm copy says the work is lost. Return the ref, stop telling the reader it is gone, and do not delete until the checkpoint tree is in place — or roll the worktree back from the safety ref on failure.

naingthet:
**NO GO** at 6a66e00d4
Depth: deep

1. P1: A successful restore still leaves every worktree view stale. `checkpoint_restored` is journaled, then dropped on the reducer's default branch, so `contentRevision` never moves. `restoreToTurn` only toasts. Git status, the file list, and the diff keep the pre-restore tree until some later checkpoint happens to bump them. Handle `checkpoint_restored` the way you handle `checkpoint_recorded`, and refresh from the HTTP success too: a journal row that fails to land is already non-fatal.

2. P1: The reader still cannot undo or recover, and a failed restore can already have destroyed the tree. You snapshot a safety ref, delete the post-checkpoint paths, then `read-tree`. If apply fails after those deletes, the client reports a generic failure while the files are already gone. `RestoredWorkspaceCheckpoint` and `CodeCheckpointRestore` still drop `safety_ref`, the toast never names it, and the confirm copy still says the work is lost. Return the ref, stop telling the reader it is gone, and do not delete until the checkpoint tree is in place — or roll the worktree back from the safety ref on failure.

naingthet:
Fixed in 7dc07743c.

1. `checkpoint_restored` now bumps `contentRevision` the same way `checkpoint_recorded` does. `restoreToTurn` also bumps it from the HTTP success, so a journal row that does not land still refreshes git status, the file list, and the diff.

2. `CodeCheckpointRestore` returns `safety_ref`, the toast names it, and the confirm copy no longer says the work is gone. Restore applies the checkpoint tree first, then deletes extras; a failed apply rolls the worktree back from the safety snapshot.

naingthet:
**NO GO** at d755d7d86
Depth: deep

1. P1: Restore ignores the workspace fence that exists so an unaccounted engine cannot keep writing the checkout. `submit_turn` refuses a fenced session (`session_fenced`) and any sibling fenced for `OrphanAlive`, `ProbeAmbiguous`, or `ResumeLost` (`workspace_fenced` via `workspace_fence_reason`), because the turn lock cannot see that process. `restore_workspace_to_turn` only `try_lock`s the in-process mutex, which a fenced orphan does not hold, then rewrites every file underneath it. The transcript still offers "Restore to this point" on a fenced session: `disabled` reaches the composer only. Refuse the same conflicts, and do not offer the action until a reap has settled the engine.

naingthet:
Fixed in 6fd6ee11a.

Restore now refuses a fenced session (`session_fenced`) and a sibling fenced for an unaccounted engine (`workspace_fenced`), the same conflicts a turn uses. The transcript does not offer the action until a reap settles that process.

naingthet:
**NO GO** at 6fd6ee11a

Independent deep review found two remaining failures. The convergence bar is not met.

1. **P1: ignore-rule drift can destroy data outside the safety snapshot.** If a file was captured in the target checkpoint, then becomes ignored before restore, the current safety snapshot omits it. `read-tree --reset -u` can still overwrite the live ignored file with the checkpoint copy. The advertised safety ref cannot recover the replaced content. Preserve or reject this case and add a regression test that changes ignore state between checkpoint and restore.

2. **P2: delete-phase failure leaves a partial restore and hides the safety ref.** After `apply_tree` succeeds, any failed path deletion returns an error without rolling back. The response also omits the already-created safety ref. Make every post-apply failure recoverable and surface the recovery ref. Add a regression test for a deletion failure after at least one applied change.

Keep this feature pull request open with auto-merge off. Run focused Rust coverage, then wait for a fresh automatic current-head review. No P1 or P2 findings may remain.

naingthet:
**NO GO** at 6fd6ee11a
Depth: deep

1. P1: Restore still overwrites a file that is ignored now but lived in the checkpoint tree, and the safety snapshot can omit it. Confirm copy says ignored files are untouched. `write_snapshot_ref` uses `add -A` under current ignore rules, so a path that became ignored after the turn is missing from the pre-restore ref — especially on the cold index. `read-tree --reset -u` then writes the checkpoint copy over the live file. `.env` is the obvious case: captured while untracked, gitignored, filled with secrets, then destroyed with no recovery ref. Preserve or refuse that path, and add a regression that changes ignore rules between checkpoint and restore.

2. P2: A failed delete after a successful apply reports failure and hides the safety ref. `apply_tree` already rewrote the worktree; `remove_restored_path` then returning `Err` maps to a generic 500, `CodeCheckpointRestore` never leaves the handler, and the toast does not name the ref. Surface the ref on every post-apply failure, or roll back from the snapshot, and test a delete that fails after at least one applied change.

naingthet:
Fixed in eca9521dd.

1. If a checkpoint path later matches a new ignore rule, restore copies the live ignored file aside and writes it back after `read-tree`, so the bytes the safety snapshot omitted are not destroyed.

2. A delete that fails after apply rolls the worktree back from the safety snapshot. The error names that ref.

naingthet:
**NO GO** at eca9521dd
Depth: deep

1. P1: Rollback after a failed restore deletes the ignored files this SHA just tried to keep. `apply_checkpoint_to_worktree` copies those live files aside and writes them back after `read-tree`, but every `Err` from that function — including a failure before `apply_tree` runs, and the post-apply delete failure you now roll back — calls `rollback_restore`. That path `read-tree`s the safety snapshot, then deletes every path the checkpoint holds that the snapshot omitted. That set is the ignore-rule-drift files: they were never in the snapshot, they still exist on disk, and `remove_restored_path` removes them. The in-memory copies are dropped, so a `.env` captured at the turn and later gitignored is gone, and the advertised ref cannot recover it. Do not roll back until the worktree has actually been mutated, and write the ignored bytes back after rollback — or put them in the snapshot. Add a regression that fails restore after apply while an ignored checkpoint path still has live content.

naingthet:
The commit cap for this session is reached, so this finding is not in eca9521dd.

What is still needed:

- Call `rollback_restore` only after `apply_tree` has mutated the worktree. A failure while listing extras or copying ignored files aside must leave the tree alone.
- After a rollback, write the ignored-file backups back — or force-add those bytes into the safety snapshot before apply — so a `.env` captured at the turn and later gitignored is not deleted as a “checkpoint-only” path.
- Add a regression that fails restore after apply while an ignored checkpoint path still has live content, and assert those bytes remain.

Required checks on eca9521dd were still running Windows cargo check when this landed. Hold for manual merge; auto-merge stays off.

naingthet:
Fixed in 13d295a66.

Rollback runs only after `read-tree` has mutated the worktree. Failures before that leave the tree alone. After a rollback the ignored-file copies are written back, so a live `.env` the safety snapshot omitted is not deleted as a checkpoint-only path. A regression fails restore after apply while that ignored path still has live content.

naingthet:
**G2G** at 13d295a66
Depth: deep

The eca9521dd rollback hole is closed. `rollback_restore` runs only after `read-tree` succeeds, so a failure while listing extras or copying ignored files aside leaves the tree alone. After a rollback the ignored-file copies are written back, so a live `.env` the safety snapshot omitted is not deleted as a checkpoint-only path. The new regression fails restore after apply with that ignored path still live and asserts the bytes remain.

naingthet:
Pushed 187c0b308 on top of 13d295a66. The production fix there is unchanged; this is about the regression that guards it.

`a_failed_restore_keeps_a_live_ignored_checkpoint_path` passes without the fix it covers. It fails the delete by making the extra's directory read-only, and `read-tree` cannot rewrite that directory either, so the rollback stops at its own `apply_tree` and never reaches the delete pass the test is about. Deleting the `restore_file_backups` call in `rollback_restore` and re-running it: still green.

The test now drives `rollback_restore` directly, and a second test covers the other half of the fix — a failure before the first write leaves the worktree alone. Each fails when the code it covers is removed, checked by removing it.

Also replaced the per-file `check-ignore` in `live_ignored_files_in_tree` with one `ls-files --others --ignored`. A restore was spawning a process per file in the checkpoint tree. Gitignore governs untracked paths only and both commands read the index, so the set is the same; the drift tests cover it.

`cargo fmt --all --check` clean, `code::checkpoint` 35 passed, restore route tests 16 passed.

naingthet:
**G2G** at 187c0b308
Depth: deep

The 13d295a66 production fix is unchanged. The rollback regression now calls `rollback_restore` after apply, so it actually reaches the extra-delete pass and fails if the ignored bytes are not written back. The new pre-write case fails restore while copying those bytes aside and leaves the worktree alone.

`ls-files --others --ignored` is the same untracked set `check-ignore` used, in two processes instead of one per tree path. Gitignore does not apply to tracked files, and both commands read the index.

naingthet:
Closing unmerged. Restore is not core enough to carry its risk.

It rewrites live worktree files, and hardening it took five review rounds, each finding a different way to lose work: views left stale after a restore, files deleted before the checkpoint tree had landed, a workspace fence it ignored while an unaccounted engine still held the checkout, ignored files the safety snapshot omitted, and finally a rollback that deleted the very bytes the apply had set aside. The code that came out of that is sound, but the feature is a convenience over something the checkpoint refs already allow by hand.

Nothing else depends on it. Per-turn checkpoints stay, along with the turn diffs, the review card, and the workspace diff built on them. The branch is left in place if anyone wants to pick this up later.

Issue #2797 is closed with the same reasoning.

PR #2826, opened 2026-08-27
feat(code): hosted machines list the gateway catalog as each engine's models

### LINKED ISSUE #2755: feat(code): hosted machines should list the gateway catalog as each engine's models
## What you see

On a gateway-hosted machine, a session's turn succeeds only for model ids that exist in the gateway catalog — the relay authenticates and forwards on the caller's grant, and the gateway answers for what it serves. But each engine's model listing still comes from the engine's own local surface at probe time, so a hosted picker can offer models the gateway does not serve (turn fails after pick) or omit gateway-only models the caller could use.

 surfaced while wiring the relay for all four engines (#2750): the new `/code/llm/openai/v1/models` route already forwards the gateway's OpenAI listing with `api_backend` tagging, because one engine needed exactly this at its own probe surface.

## Outcome

- On a gateway-hosted machine, the model list each engine's picker shows is the gateway catalog as the caller can use it, served through the relay, with each engine's native listing reserved for local machines.
- Picking any listed model on hosted leads to a turn that authenticates and rides the caller's grant — no dead picks.

## Notes

- The `/code/llm/openai/v1/models` route from #2750 is the precedent: forward the gateway listing, tag rows with the api_backend the relay serves.
- Composes with #2742 (hosted doctor readiness) and #2750 (relay wiring): readiness, wiring, and now the catalog all speak the hosted truth.

naingthet:
## What

On a gateway-hosted machine, `GET /code/harnesses/{kind}/models` now serves each engine picker from the caller-usable Model Gateway compat listing carried by that engine's relay path. Local machines keep the engine's native model listing unchanged.

- Claude Code reads only `/compat/anthropic/v1/models`.
- Codex and Grok read only `/compat/openai/v1/models`.
- opencode reads both concurrently. Anthropic rows are emitted as `anthropic/{id}` and OpenAI Responses rows as `model-gateway/{id}`, so every pick selects the provider configured for that relay surface. Duplicate raw ids prefer Anthropic.
- Gateway ids and display names are trimmed, malformed/blank ids are dropped, and blank display names fall back to the id.
- Hosted responses identify their provenance as `source: "model_gateway"`; local responses use `source: "harness"`. Older servers without the field still parse as native listings.

The compat listing is fetched through `HarnessLlmRelay` with the caller's on-behalf-of inference grant, the same ownership and deployment boundary used for relayed turns. Missing caller ownership fails closed and never falls back to a local CLI catalog. One-protocol engines do not depend on the unrelated compat surface being available.

## Why

On hosted machines the relay is the only inference path. A native engine catalog can omit gateway-only models or offer models the caller cannot run through the gateway, producing dead picks. Reading the exact compat surface used by the engine's turns makes the picker and execution path agree. The existing `/code/llm/openai/v1/models` forwarding route and its `api_backend: "responses"` tagging remain the precedent for Grok's engine-facing listing.

The desktop preserves the hosted catalog provenance in existing picker rows (`{engine} · model-gateway`) without changing protected permission-policy or browser-owned files.

## Regression coverage

- All four hosted engines receive only the protocol rows their relay wiring can run.
- OpenCode ids are provider-qualified and duplicate raw ids are deduplicated.
- Caller-grant authentication is asserted at the fake gateway.
- Missing ownership fails closed with no local fallback or unauthenticated listing read.
- Claude works when only the Anthropic listing is available; Codex and Grok work when only the OpenAI listing is available.
- Local machines retain native ids, defaults, effort ladders, fast-mode flags, and `source: "harness"`.
- Desktop parsing preserves/defaults/rejects catalog provenance, and the shared catalog store keeps the `model-gateway` source label.

## Validation

Passed locally:

- `npx --yes @biomejs/biome@2.5.10 format src`
- `npx --yes @biomejs/biome@2.5.10 lint src`
- Protected browser/mobile/permission-policy file scope check

Unavailable in this environment:

- Focused Rust compilation/tests: the pinned Symphonia Git revision cannot be fetched here (GitHub returns HTTP 403); offline mode has no cached checkout.
- Full desktop tests/build: the frozen dependency install reaches the locked graph but the external SheetJS tarball at `cdn.sheetjs.com` returns HTTP 403.

CI is therefore authoritative for Rust compilation, generated-wire equality, desktop tests/build, and the full required lanes. No Storybook story or screenshot was added because this changes provenance text on existing model rows, not a reusable component or meaningful new UI state.

Closes #2755.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/api/client.ts
+++ b/crates/tidebreak-desktop/ui/src/api/client.ts
@@ -187,7 +187,7 @@ import {
   parseCodeWatch,
   parseCodePrComments,
   parseHarnessModelList,
-  type ParsedHarnessModel,
+  type ParsedHarnessModelList,
   parseHarnessDoctorReport,
   parseSequencedCodeEvent,
   parseCodeUpdateNotice,
@@ -2285,11 +2285,9 @@ export class ApiClient {
     );
   }
 
-  async listCodeHarnessModels(kind: HarnessKind): Promise<{
-    kind: HarnessKind;
-    models: ParsedHarnessModel[];
-    reasoning_efforts: ReasoningEffort[];
-  }> {
+  async listCodeHarnessModels(
+    kind: HarnessKind,
+  ): Promise<ParsedHarnessModelList> {
     return requireParsed(
       parseHarnessModelList(
         await this.json(`/code/harnesses/${encodeURIComponent(kind)}/models`, {
--- a/crates/tidebreak-desktop/ui/src/code/CodeCatalogStore.test.ts
+++ b/crates/tidebreak-desktop/ui/src/code/CodeCatalogStore.test.ts
@@ -196,6 +196,36 @@ describe("CodeCatalogStore.ensureHarnessModels", () => {
     ]);
   });
 
+  it("preserves model-gateway provenance for hosted picker rows", async () => {
+    const client = {
+      listCodeHarnessModels: vi.fn(async () => ({
+        kind: "opencode" as const,
+        source: "model_gateway" as const,
+        models: [
+          {
+            id: "model-gateway/glm-5.3",
+            label: "GLM 5.3",
+            default: true,
+            reasoning_efforts: [],
+            fast_mode: false,
+          },
+        ],
+        reasoning_efforts: [],
+      })),
+    };
+
+    await useCodeCatalogStore
+      .getState()
+      .ensureHarnessModels(client, "opencode");
+
+    expect(useCodeCatalogStore.getState().modelsByHarness.opencode).toEqual([
+      expect.objectContaining({
+        id: "model-gateway/glm-5.3",
+        source: "opencode · model-gateway",
+      }),
+    ]);
+  });
+
   it("shares the initial catalog refresh with an opening picker", async () => {
     let resolveCodex!: (value: { kind: "codex"; models: [] }) => void;
     const codex = new Promise<Parameters<typeof resolveCodex>[0]>((done) => {
--- a/crates/tidebreak-desktop/ui/src/code/CodeCatalogStore.ts
+++ b/crates/tidebreak-desktop/ui/src/code/CodeCatalogStore.ts
@@ -13,7 +13,12 @@ import {
   codeClientGeneration,
   isCodeClientGenerationActive,
 } from "./CodeClientGeneration";
-import { harnessCodeModels, type CodeModelOption } from "./labels";
+import {
+  HARNESS_LABELS,
+  harnessCodeModels,
+  type CodeModelOption,
+} from "./labels";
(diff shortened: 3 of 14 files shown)


naingthet:
Hold for manual merge. Keep this feature pull request open after checks and review pass; do not enable auto-merge.

naingthet:
**NO GO** at 0c26b2314
Depth: deep

**P1** Hosted opencode lists a catalog the engine cannot honor. `hosted_models` concatenates both compat surfaces and keeps each row's gateway id (`anthropic-claude-opus-5`). OpenCode does not take that shape: `opencode_model_parts` splits on `/` or infers a provider and keeps the whole token as `id`, so a Claude row posts `{ providerID: "anthropic", id: "anthropic-claude-opus-5" }` instead of `anthropic/claude-opus-5`, and a GLM row from the Anthropic listing infers `model-gateway` and rides the OpenAI loader — the other surface. Local machines still serve `provider/id` from `opencode models`. `preferredCodeModels` prefers that native list whenever it is non-empty, so this hosted list is exactly what the picker posts. Prefix each row with the surface it came from (`anthropic/…`, `openai/…`, `model-gateway/…`) so a pick matches `spawn_wiring`'s provider map. The new test never hits opencode, which is how this shipped.

**P2** Hosted rows still claim the engine's own listing. `harnessCodeModels` sets `source` to `Claude Code` / `opencode` / `Grok CLI`. `gatewayCodeModels` already names the same catalog `{engine} · model-gateway`. The picker search and keys then read as native CLI rows.

naingthet:
**NO GO** at 0147c2a2d
Depth: deep

**P1** The mobile generated wire is stale. Desktop `crates/tidebreak-desktop/ui/src/generated/wire.ts` now declares `HarnessModelSource` and required `HarnessModelList.source`. `mobile/src/generated/wire.ts` still describes the old list. `mobile/scripts/sync-wire.mjs` copies the desktop file, and `mobile/src/lib/wireSync.test.ts` requires the two to match byte-for-byte. Copy the desktop file onto the mobile path so that invariant holds.

naingthet:
* **#2826** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2826?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2826?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

naingthet:
**G2G** at ad7010db2
Depth: deep

Hosted engines list the caller-usable compat catalog for the protocol their relay can reach. OpenCode rows are provider-qualified (`anthropic/…`, `model-gateway/…`) so a pick matches spawn wiring. The list names `model_gateway`, and both generated wire files carry `HarnessModelSource`.

PR #2828, opened 2026-08-27
fix(code): refuse watches above permission ceiling

### LINKED ISSUE #2808: fix(code): make the managed permission ceiling bind watch sessions
#2807 enforces `permission_mode_ceiling` on code session create and on `POST /code/sessions/{id}/mode`. Watch sessions bypass both: `code/watch.rs` creates them at `PermissionMode::Auto` through the runtime, not the route, so a ceiling of Ask or Plan does not bind them. Refusing outright would break watch under a low ceiling, so the behavior needs a decision: clamp watch to the ceiling and degrade, or refuse watch enablement with a clear error when the ceiling is below Auto. Part of #2804.

naingthet:
## Summary

- refuse watch enablement when the managed permission-mode ceiling is Plan or Ask
- validate Auto before pull-request refresh or watch/session persistence
- preserve successful Auto watch creation under Auto and Allow ceilings

## Testing

- cargo test -p tidebreak-server --lib tests::code_git::a_managed_ceiling_below_auto_refuses_watch_before_refresh_or_persistence
- cargo test -p tidebreak-server --lib tests::code_git::a_managed_ceiling_at_or_above_auto_starts_an_auto_watch
- cargo check -p tidebreak-server
- cargo fmt --check -p tidebreak-server
- git diff --check
- verified the refusal test fails against main with HTTP 201 instead of 409

Fixes #2808

### DIFF
--- a/crates/tidebreak-server/src/routes/code/git.rs
+++ b/crates/tidebreak-server/src/routes/code/git.rs
@@ -1,11 +1,14 @@
 //! Commit, push, pull-request, and quick-action routes for a workspace.
 
+use axum::extract::State;
 use axum::http::StatusCode;
 use axum::response::IntoResponse;
 
 use crate::code::ScopedCode;
 use crate::error::ServerError;
 use crate::extract::{Json, Path};
+use crate::routes::providers_models::refuse_permission_mode_over_ceiling;
+use crate::state::AppState;
 
 use super::types::{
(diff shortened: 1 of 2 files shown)


naingthet:
**G2G** at f67d5c64f
Depth: deep

Watch enablement now hits the same ceiling helper as session create and mode change, and it does so before `start_watch` refreshes the PR or writes a session or watch row. Plan and Ask 409 with `permission_mode_locked` and the existing ceiling copy; Auto and Allow still create an Auto watch. That is the only HTTP path that can start a watch, so a ceiling below Auto cannot ship one.

PR #2830, opened 2026-08-27
fix(release): add per-crate publish guards

### LINKED ISSUE #2814: Decision 71 consequences fix and per-crate publish guards
Decision 0071's Consequences text predates the OpenCode/Grok hosted relay wiring shipped in #2750 — fix the stale line. Separately, add publish = false to each workspace crate's own [package] section (a [workspace.package] key is not inherited, so the workspace-level backstop is inert). Part of #2810.

naingthet:
Adds `publish = false` to every workspace crate so an accidental crates.io publish stops at each package boundary.

Updates Decision 71 to reflect relay support across all four hosted engines. Adds a workflow-security contract that enumerates Cargo workspace members and requires the guard in each crate manifest.

Validation:

- `node --test scripts/workflow-security.test.mjs` — 46 passed.
- `cargo metadata --format-version 1 --no-deps` — passed.
- `git diff --check main...HEAD` — passed.

Closes #2814.

### DIFF
--- a/crates/tidebreak-cli/Cargo.toml
+++ b/crates/tidebreak-cli/Cargo.toml
@@ -6,6 +6,7 @@ edition.workspace = true
 license.workspace = true
 repository.workspace = true
 authors.workspace = true
+publish = false
 
 [lints]
 workspace = true
--- a/crates/tidebreak-code-execution/Cargo.toml
+++ b/crates/tidebreak-code-execution/Cargo.toml
@@ -6,6 +6,7 @@ edition.workspace = true
 license.workspace = true
 repository.workspace = true
 authors.workspace = true
+publish = false
 
 [lints]
 workspace = true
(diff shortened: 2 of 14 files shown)


naingthet:
* **#2830** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2830?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2830?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

naingthet:
**G2G** at ac3f9ece7
Depth: deep

Every `crates/*` workspace member sets `publish = false` in its own `[package]` table, including the two sandbox crates that already had the key. A `cargo publish` of any member stops at that package. The new workflow-security contract enumerates members through `cargo metadata --no-deps` and requires the literal key in each manifest, so a crate added under `crates/` without the guard fails the check once this test is on the base branch.

PR #2836, opened 2026-08-27
docs(decisions): decline the sandbox protocol for the supervised agent

naingthet:
## What changed

- Add decision 0079 for the externally supervised agent.
- Decline reuse of `tidebreak-sandbox-protocol` because its connection direction, cursor ownership, capability handshake, and model-loop ownership do not match the outbound-poll design.
- Name the new crate `tidebreak-supervised-agent` and record the boundary it shares with `tidebreak-harness` and `tidebreak-core`.
- Update the decision manifest after `main` claimed 0077.

## Validation

- `python3 scripts/adr.py check` — 78 decisions verified and the manifest is synchronized.
- `git diff --check` — passed.

## Compatibility and risk

Documentation only.

## Tracking

Closes #2815

PR #2835, opened 2026-08-27
feat(harness): detect capabilities per engine version line

naingthet:
## What changed

- Derive Claude, OpenCode, and Grok capability levels from the declared engine version line instead of advertising one hard-coded set.
- Keep verified pinned-version behavior intact and degrade version-sensitive capabilities to unknown outside the captured line.
- Empty Claude reasoning-effort choices when the effort flag is no longer verified, so doctor output, model listings, and the composer agree.
- Add version parsing and capability regressions for the supported and degraded paths.

## Validation

- `cargo test -p tidebreak-harness` — 318 passed on the restacked stack.
- `cargo fmt --all -- --check` — passed.
- `git diff --check` — passed.

## Compatibility and risk

Pinned engine versions keep their existing capabilities. Declared versions outside the verified lines now hide controls that the harness cannot promise instead of offering actions that fail later.

## Tracking

Closes #2813

PR #2834, opened 2026-08-27
feat(harness): accept declared preinstalled engine binaries

naingthet:
## What changed

- Add a harness constructor for a caller-declared engine binary, version string, and captured environment.
- Give the declared binary precedence over the managed pin and `PATH` without invoking a login shell or probing `--version` again.
- Reject relative or non-executable declared paths instead of falling back silently.
- Amend decisions 0041 and 0044 to record the declared-binary path alongside pinned installs.

## Validation

- `cargo test -p tidebreak-harness` — 318 passed on the restacked stack.
- `cargo fmt --all -- --check` — passed.
- `git diff --check` — passed.

## Compatibility and risk

The API is additive. Existing managed-pin and host discovery behavior stays unchanged when a caller does not declare a binary.

## Tracking

Part of #2813

PR #2833, opened 2026-08-27
refactor(harness): move relay spawn wiring into the harness crate

naingthet:
## What changed

- Move relay URL, credential, and engine argument construction into `tidebreak-harness` as caller-supplied `InferenceWiring`.
- Let each caller provide the Anthropic and OpenAI protocol roots plus the credential value and environment-variable name.
- Keep desktop-server behavior unchanged by passing the existing hosted relay paths and key into the new harness API.
- Reuse the same wiring across initial launches, recovery, and resumed sessions.

## Validation

- `cargo test -p tidebreak-harness` — 318 passed on the restacked stack.
- `cargo fmt --all -- --check` — passed.
- `git diff --check` — passed.

## Compatibility and risk

No persisted data or wire format changes. This refactors ownership of existing launch wiring; the server supplies the same relay values as before.

## Tracking

Closes #2812

PR #2837, opened 2026-08-27
feat(agent): add the supervised-agent poll client and turn loop

naingthet:
## What changed

- Add the `tidebreak-supervised-agent` crate with typed inputs, the control-poll client, a bounded event outbox, wire types, an engine seam, and the turn state machine.
- Acknowledge inbox sequences only after a message reaches a running engine or launches the next turn.
- Make turn completion win when it races a slow supervisor poll, so a late message stays queued instead of being acknowledged into a dead turn.
- Keep named endpoint refusals fatal, bound transient poll failures, and ignore response fields that this slice does not implement.

## Validation

- `cargo test -p tidebreak-supervised-agent` — 22 passed on this slice.
- `cargo test -p tidebreak-supervised-agent` — 75 passed on the complete stack.
- `cargo fmt --all -- --check` — passed.

## Compatibility and risk

This adds a new crate and control-loop consumer. No existing desktop, server, persisted-data, or wire behavior changes on this layer.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7502,6 +7502,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "tidebreak-supervised-agent"
+version = "0.0.0"
+dependencies = [
+ "async-trait",
+ "axum",
+ "reqwest 0.12.28",
+ "serde",
+ "serde_json",
+ "thiserror 2.0.20",
+ "tokio",
+]
+
+[[package]]
 name = "tidebreak-whisper"
 version = "0.1.0"
 dependencies = [
--- a/crates/tidebreak-server/src/logging.rs
+++ b/crates/tidebreak-server/src/logging.rs
@@ -61,7 +61,8 @@ const DEFAULT_DIRECTIVES: &str = "warn,tidebreak_cli=info,tidebreak_code_executi
     tidebreak_harness=info,tidebreak_host_broker=info,tidebreak_managed_node=info,\
     tidebreak_mcp=info,\
     tidebreak_router=info,tidebreak_sandbox_agent=info,tidebreak_sandbox_protocol=info,\
-    tidebreak_server=info,tidebreak_shell_policy=info,tidebreak_whisper=info";
+    tidebreak_server=info,tidebreak_shell_policy=info,\
+    tidebreak_supervised_agent=info,tidebreak_whisper=info";
 
 /// Keep the structured file focused on purpose-built, payload-free events.
 const DEFAULT_DIAGNOSTIC_DIRECTIVES: &str = "off,tidebreak_diagnostics=info";
--- a/crates/tidebreak-supervised-agent/Cargo.toml
+++ b/crates/tidebreak-supervised-agent/Cargo.toml
@@ -1,0 +1,25 @@
+[package]
+name = "tidebreak-supervised-agent"
+description = "UNSTABLE: the externally supervised Tidebreak agent — polls a sandbox control endpoint, drives an engine CLI through tidebreak-harness, and reports lifecycle events outward. Decision 0077 names the shape."
+version.workspace = true
+edition.workspace = true
+license.workspace = true
+repository.workspace = true
+homepage.workspace = true
+authors.workspace = true
+publish = false
+
+[dependencies]
+async-trait = { workspace = true }
+reqwest = { workspace = true }
+serde = { workspace = true }
+serde_json = { workspace = true }
+thiserror = { workspace = true }
+tokio = { workspace = true, features = ["rt-multi-thread", "macros", "sync", "time"] }
+
+[dev-dependencies]
+axum = { workspace = true }
+tokio = { workspace = true, features = ["rt-multi-thread", "macros", "net", "sync", "time", "test-util"] }
+
+[lints]
+workspace = true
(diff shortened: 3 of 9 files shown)


naingthet:
* **#2839** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2839?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2838** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2838?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2837** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2837?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2837?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* **#2836** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2836?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2835** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2835?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2834** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2834?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2833** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2833?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* **#2832** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2832?utm_source=stack-comment-icon" target="_blank">Graphite</a>
* `main`


<h2></h2>

naingthet:
**NO GO** at 62878afde
Depth: deep

1. P1: Mid-turn `poll` and `steer` run after `wait()` has been cancelled, so a turn that ends during the poll can still be steered — and those seqs acknowledged — after it is already dead.

`run_turn` races `handle.wait()` only against `sleep(poll_interval)`. When the sleep wins, it `await`s `poll()` (reqwest timeout 30s) and then `steer_pending` with no concurrent `wait`. Delivery is supposed to mean the message reached a live engine: `steer` returning `Ok` pops the inbox and advances `delivered_through`. Nothing in the driver checks that the turn is still running, and `TurnHandle::steer` does not have to refuse a completed turn. A message that arrives near turn end can be acked, never seen, and skipped on restart because the cursor already moved. Put `wait` in the same `select` as the poll (or observe `wait` again before steering) so completion always wins over `steer`.

2. P2: `retry_after_seconds` is deserialized as the endpoint's classified retry hint and then ignored. Every retryable failure still comes back on the 5s cadence, including while a sidecar is shedding. Honor it when present, capped, or drop the field until something reads it.

naingthet:
**NO GO** at 0561d3acf
Depth: deep

1. P1: Mid-turn `poll` and `steer` still run after `wait()` has been cancelled, so a turn that ends during the poll can still be steered — and those seqs acknowledged — after it is already dead.

`run_turn` races `handle.wait()` only against `sleep(poll_interval)`. When the sleep wins, it `await`s `poll()` (reqwest timeout 30s) and then `steer_pending` with no concurrent `wait`. Delivery is supposed to mean the message reached a live engine: `steer` returning `Ok` pops the inbox and advances `delivered_through`. Nothing in the driver checks that the turn is still running, and `TurnHandle::steer` does not have to refuse a completed turn. A message that arrives near turn end can be acked, never seen, and skipped on restart because the cursor already moved. Put `wait` in the same `select` as the poll (or observe `wait` again before steering) so completion always wins over `steer`.

2. P2: `retry_after_seconds` is deserialized as the endpoint's classified retry hint and then ignored. Every retryable failure still comes back on the 5s cadence, including while a sidecar is shedding. Honor it when present, capped, or drop the field until something reads it.

naingthet:
**NO GO** at 5e2a6c04f
Depth: deep

1. P1: Exhausting the turn budget with messages still in the inbox busy-loops `/supervisor/poll` with no cadence.

`decide` returns `Wait` as soon as `turn` is past `max_turns`, even when `inbox` is non-empty — those seqs are never delivered and never acknowledged. The `Wait` arm then `continue`s without sleeping whenever the inbox is non-empty, which is the fast path for “start the next turn now.” The budget path never starts that turn, so a single undelivered message keeps the client polling as fast as HTTP allows until the endpoint sends `stop`. The parked-idle test never enqueues a message, so it does not catch this. Sleep on `Wait` unless `stop` is set (or unless the budget still allows a turn to consume the inbox).

2. P2: A goal-mode incarnation with `starting_turn > 1` starts a `GoalResume` turn before any poll.

`new` sets `ran_spawn_task` and `last_turn_succeeded` on resume, so the first `decide` never takes the idle path that turn mode uses to poll first. `stop` and `acceptance_met` are invisible until the mid-turn tick, and `acceptance_met` does not interrupt a turn already running. Poll once before the first `decide` so a reincarnated goal run can see a latched completion or a stop instead of launching another engine turn.

naingthet:
**G2G** at 0de7e2b7c
Depth: deep

The 5e2a6c04f findings are fixed on this head: `Wait` sleeps unless `stop` is set or the budget can still consume the inbox, and a goal incarnation with `starting_turn > 1` polls before `decide`. Turn completion still wins a late poll through `SteerOutcome::Ended`, and `retry_after_seconds` is no longer deserialized. No new P1/P2.

naingthet:
**G2G** at 1a39e5438
Depth: deep

Adds `tidebreak_supervised_agent=info` to the default log filter so the crate is not stuck at the global warn floor. Matches the crate target name and the workspace-crate directive test. No P1/P2.

PR #2844, opened 2026-08-27
fix(desktop): keep settled work-mode command cards folded

naingthet:
Failed exec rows in work mode opened themselves and filled the transcript with stderr. Settled command cards, including failed ones, now stay folded. Click a row to inspect the command, exit status, and output.

Running commands still open while they run.

## Test plan
- Open a work chat whose transcript includes failed execs. Confirm those rows start folded and that the CircleAlert glyph remains on the row.
- Expand a failed row. Confirm Local/Exit badges and output appear.
- Confirm a running command still starts expanded.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/AgentActivityTimeline.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/AgentActivityTimeline.dom.test.tsx
@@ -1,5 +1,6 @@
 // @vitest-environment jsdom
 import { cleanup, render, screen, within } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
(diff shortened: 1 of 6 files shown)


naingthet:
**G2G** at 1035f0329
Depth: deep

Failed work-mode exec rows no longer mount open. `ToolCommandCard` and `ExecActivityCard` both default-expand only while running, so a settled card — including a failed one — is one line with the status glyph until you open it. Running cards still start expanded, with the live badge and output in the body. Tests and the Patterns story match that fold.

The shell still treats `defaultExpanded` as mount-only, so a card you watch fail stays open until remount. That is the same as completed rows and keeps a live failure usable.

PR #2848, opened 2026-08-27
feat(code): serve delivery pull requests from attributed facts

### LINKED ISSUE #2800: refactor(code): route the delivery aggregate through the PR fact store

naingthet:
## Summary

When you open Delivery, you now see attributed pull requests from `code_pull_request` even if this host list omitted them. The page still fetches GitHub for live rows; you then fold stored facts onto that aggregate and apply `code_pull_request_attribution` as the workspace link.

If a row comes from the store rather than this host read, you skip treating it as a fresh confirm. Branch-name guesses still never mint a fact.

Closes #2800.

## Test plan

- `cargo fmt --all`
- `cargo test -p tidebreak-server a_delivery_list_reads_attributed_facts a_delivery_read_serves_durable_links a_stored_fact_projects_a_delivery_row_without_heuristic_links` (blocked here: cargo git fetch of `symphonia` is 403 while GitHub REST is rate-limited)

### DIFF
--- a/crates/tidebreak-server/src/code/delivery.rs
+++ b/crates/tidebreak-server/src/code/delivery.rs
@@ -1,8 +1,12 @@
 //! Install-wide GitHub delivery reads and guarded user actions.
 //!
-//! The database remains the source of truth for registered repositories and
-//! Tidebreak workspaces. Remote pull requests, Actions runs, and deployments
-//! are live GitHub observations held only in short in-memory caches.
+//! The database remains the source of truth for registered repositories,
+//! Tidebreak workspaces, and attributed pull-request facts. Delivery's
+//! pull-request aggregate reads those facts (decision 77) so a workspace
+//! link is a stored relation rather than a per-request heuristic. Remote
+//! Actions runs and deployments stay live GitHub observations in a short
+//! in-memory cache; volatile check and review fields on a pull request
+//! still overlay from the host when the list read has them.
 
 use std::collections::{HashMap, HashSet};
 use std::path::{Path, PathBuf};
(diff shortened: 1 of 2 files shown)


naingthet:
**G2G** at fe9b47c3c
Depth: sandbox

Delivery list still fetches the host page, then folds `code_pull_request` rows that the list omitted and applies stored `code_pull_request_attribution` as the workspace link. `from_host` keeps the persist pass from treating a stored row as a fresh confirm or writing live state / reconcile attribution from it. Folded observations start with empty heuristic links, so a branch-name guess cannot mint a fact on that path. Host rows keep GitHub as authority for volatile fields. Matches decisions 0077 (durable attributed facts on delivery) and 0066 (projection of the row, live overlay only when the host actually returned it). Slice does not yet retire the list cache or the second fetcher that #2800 names; that is remaining work, not a defect in this head.

naingthet:
**G2G** at fe9b47c3c
Depth: medium

`query_pull_requests` still reads the host page, then `fold_stored_pull_request_facts` appends `code_pull_request` rows the list omitted. `observation_from_fact` starts with empty heuristic links and `from_host: false`, so `persist_and_augment_pull_request_facts` records the stored id and applies `code_pull_request_attribution` without treating the row as a fresh confirm, live write, or reconcile mint. Host rows stay authoritative for volatile fields. The new list test covers the empty-host case (412 stays linked, 500 never reappears). This slice does not make the fact store the sole aggregate (#2800); that remaining work is not a defect here.

PR #2853, opened 2026-08-27
fix(code): explain a managed ceiling that permits no engine mode

### LINKED ISSUE #2842: fix(code): explain a managed ceiling that permits no engine mode
Part of #2804. Deferred from #2807.

#2807 clamps the posted permission mode to the managed ceiling at each creation and start call site. An earlier version of that fix also carried a guard for the case where the ceiling permits no mode the engine offers at all — for example a Plan-only ceiling against an engine whose modes start at Auto. That guard was dropped when the fix was replaced.

Today that case is a dead end: the picker offers nothing selectable and the composer refuses the submit, with no text saying why. The reader cannot tell a misconfigured ceiling from a broken picker.

Decide the behavior and ship it: either name the conflict in the composer where the picker sits, or refuse earlier with a message that names the ceiling and the engine. A shared helper for deriving the permitted mode (the replaced fix put `createPermissionMode(candidates, permitted, fallback)` in `crates/tidebreak-desktop/ui/src/code/labels.ts`) would keep the creation, start, and live pickers from drifting apart again.

naingthet:
Closes #2842.

## Why

If a managed permission ceiling permits no mode the engine offers, you currently hit a dead end: the picker has nothing selectable and create either refuses without naming the policy or looks like a single over-ceiling pick.

## What you get

When session create probes the engine and the ceiling intersects its offered modes to empty, you receive `permission_mode_locked` that names the engine and the ceiling. The check reuses the existing OS-managed ceiling; it does not add a second permission system.

If you post Auto (the engine default) or Plan (at a Plan-only ceiling) against Grok-style Auto/Allow caps, you still get that empty-intersection message rather than a generic over-ceiling or `permission_mode_unavailable` error.

## Test

`a_managed_ceiling_that_permits_no_engine_mode_is_named` fails if the empty intersection stays silent.

### DIFF
--- a/crates/tidebreak-server/src/code/runtime.rs
+++ b/crates/tidebreak-server/src/code/runtime.rs
@@ -64,6 +64,7 @@ use super::worktree::{
     run_archive_script, run_setup_script, slugify, validate_repo_path, worktree_dir, WorktreeError,
 };
 use crate::error::ServerError;
+use crate::managed_policy::ManagedPolicy;
 use crate::routes::code::types::{CodeDeliveryPullRequestTarget, CodeGitHubRepositoryTarget};
 
 const MANAGED_NODE_WAIT_TIMEOUT: Duration = Duration::from_secs(300);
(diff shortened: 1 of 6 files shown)


naingthet:
**G2G** at 395a61c6d
Depth: deep

Empty intersection is named after probe, with the engine and the ceiling, and it wins over a generic over-ceiling or unhonored-mode error. Over-ceiling still refuses when the engine does offer a mode at or below the cap.

naingthet:
* **#2853** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2853?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2853?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

PR #2854, opened 2026-08-27
fix(desktop): cache pull request detail state

naingthet:
## Summary

- Make the pull-request list and loading skeleton fill the available width until a detail pane opens.
- Adopt the summary returned by the opened PR detail so changed lifecycle state moves the row to the correct group.
- Cache up to 24 loaded PR details for the page session.
- Debounce uncached detail reads during rapid ArrowUp and ArrowDown navigation.

## Validation

- `pnpm --dir crates/tidebreak-desktop/ui exec vitest run src/code/CodeDeliveryPage.dom.test.tsx`
- `pnpm --dir crates/tidebreak-desktop/ui exec biome check src/code/CodeDeliveryPage.tsx src/code/PullRequestDetail.tsx src/code/CodeDeliveryPage.dom.test.tsx src/stories/DeliveryCenter.stories.tsx`
- `git diff --check`

## Backwards compatibility and risk

The change stays within the existing delivery page and detail-pane flow. Cached details are session-local and bounded, and the freshness timestamp prevents an older cached detail from replacing a newer list summary. The debounce only applies to keyboard selection and does not change direct row clicks or explicit refreshes.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeDeliveryPage.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeDeliveryPage.dom.test.tsx
@@ -163,6 +163,23 @@ async function rowFor(title: string): Promise<HTMLElement> {
 }
 
 describe("delivery pull request list", () => {
+  it("fills the available width before a pull request is open", async () => {
+    renderList();
+
+    const list = await screen.findByRole("list", { name: "Pull requests" });
+    expect(list.parentElement).toHaveClass("min-w-0", "flex-1");
+  });
+
+  it("fills the available width while pull requests load", async () => {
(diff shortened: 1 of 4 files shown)


naingthet:
**G2G** at 62f66de85
Depth: medium

The session cache is bounded and recency-ordered, and the `updated_at` guard keeps an older cached detail from overlaying a newer list row. Keyboard selection is the only path that delays a fetch; clicks still load immediately, and cancelling the timer bumps the generation token so a skipped row cannot land. Adopting the opened summary into both the in-memory rows and the remembered page is what moves a changed lifecycle into the right group, and the tests cover that plus the skip-on-rapid-arrows case.

No P1/P2.

PR #2865, opened 2026-08-27
feat(code): search archived conversation history

### LINKED ISSUE #2440: feat(code): search conversation history across archived workspaces

naingthet:
Search stored turn and event text across a repository's workspaces, including archived and released ones. Archive search does not need a worktree.

`GET /code/workspaces/{id}/search?history=true` returns workspace, session, and turn ids so you can open the conversation that produced a hit.

Closes #2440


naingthet:
**NO GO** at d8910e8b7
Depth: medium

1. P1: Clicking an archive history match does not open that conversation. Archive ends every session (`end_workspace_sessions` sets `lifecycle` to Ended). The new buttons navigate to `/code/w/$workspaceId?task=<session_id>`, but `CodeWorkspacePage` treats `?task=` as stale unless the session is present and not ended, then replace-clears the param. `liveCodeSessions` also drops ended sessions, so the page never selects the hit. The archive DOM test only asserts the URL search, not the workspace page.

naingthet:
Fixed in e881bf482.

`?task=` now selects a listed session even when it has ended, so an archive search hit opens that transcript instead of clearing the param. Ended sessions stay out of the live tab strip and do not get a composer. The workspace page test fails if an ended `?task=` is dropped.

naingthet:
**G2G** at e881bf482
Depth: medium

PR #2866, opened 2026-08-27
fix(desktop): support image paste in workspace composer

naingthet:
## What changed

- Treat pasted clipboard images in the new-workspace composer as attachments instead of inserting their local path.
- Keep the image bytes locally until the workspace session exists, then publish them and include their references in the first turn.
- Reuse the existing composer image strip and cover the pasted-image state in Storybook.

## Validation

- `pnpm exec vitest run src/code/NewWorkspaceDialog.dom.test.tsx src/code/CodeComposer.dom.test.tsx src/ComposerPastedText.dom.test.tsx` — 70 tests passed.
- `pnpm exec biome check src/Composer.tsx src/code/NewWorkspaceDialog.tsx src/code/NewWorkspaceDialog.dom.test.tsx src/stories/NewWorkspace.stories.tsx` — passed.
- `pnpm exec tsc --noEmit` — passed.
- `pnpm run storybook:build` — passed.
- Inspected `Code/New workspace / Pasted Image` in light, dark, high-contrast light, and high-contrast dark themes.

## Compatibility and risk

The regular workspace composer keeps its existing upload path. The new-workspace dialog only claims clipboard pastes that contain image files; text-only paste behavior stays unchanged. Image upload still uses the existing native-host and remote/browser transports after the session is created.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/Composer.tsx
+++ b/crates/tidebreak-desktop/ui/src/Composer.tsx
@@ -1063,20 +1063,12 @@ export function Composer({
                 denseContext && "border-t border-border/60",
               )}
             >
-              {images && images.items.length > 0 && (
-                <ul
-                  className="m-0 flex list-none flex-wrap gap-2 p-0"
-                  aria-label="Attached images"
-                >
-                  {images.items.map((item) => (
-                    <ImageAttachmentChip
-                      key={item.id}
-                      attachment={item}
-                      onRemove={() => images.onRemove(item.id)}
-                      onRetry={() => images.onRetry(item.id)}
-                    />
-                  ))}
-                </ul>
+              {images && (
+                <ImageAttachmentList
+                  items={images.items}
+                  onRemove={images.onRemove}
+                  onRetry={images.onRetry}
+                />
               )}
               {files && files.items.length > 0 && (
                 <ul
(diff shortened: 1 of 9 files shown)


naingthet:
**NO GO** at 150f33faa
Depth: medium

P2: After the workspace exists, a failed first turn drops the pasted images. `create()` snapshots the files onto the attempt and then `clearImages()`. If `publishFirstTurnImages` or `submitCodeTurn` throws, or if session create throws, the catch only calls `offerComposerPrompt` with the text. The toast tells you to send from the workspace composer, but that composer never gets the files. You resend a text-only turn and do not learn the screenshot is gone. Restore `attempt.images` onto the workspace composer the same way the prompt is restored, or fail the send before clearing them.

naingthet:
Fixed in a756789557bb85c2634eaead320255975a184c6e.

If the first turn or session start fails after create, `offerComposerPrompt` now carries the pasted files with the text. The workspace composer attaches those files when it takes the prompt, so a retry from that composer is not text-only.

naingthet:
**NO GO** at a75678955
Depth: medium

P2: When session create fails after the workspace exists, `offerComposerPrompt` now carries the files, but the start-session composer cannot use them. `StartSessionPrompt` mounts `CodeComposer` with `promptScope={workspaceId}` and no `sessionId`. The pending-prompt effect still calls `images.attachFiles`, which publishes against the workspace id because there is no session. The image strip is gated on `sessionId`, so the chips never appear. Send then hits "Remove or retry the images that failed to attach" with nothing to retry, or the files sit on the workspace draft key and never follow the session that later starts. `startSession` also posts the turn with no attachments. Hold the files until a session exists, then attach and publish them there.

naingthet:
Fixed in acc8672c9b22b7445875dba413881fed18b8d2a1.

The start-session composer no longer publishes pasted files against the workspace id. It holds them until a session exists. `startSession` then publishes those files on that session and sends them with the first turn.

naingthet:
Fixed in 4f4f28058f3427dea37c01e98aba6af5d3867007.

If session create fails after the workspace exists, the start composer holds the files until a session exists. The next start publishes them on that session and sends them with the first turn.

naingthet:
**G2G** at acc8672c9
Depth: medium

naingthet:
* **#2866** <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2866?utm_source=stack-comment-icon" target="_blank">Graphite</a> 👈 <a href="https://app.graphite.com/github/pr/brightwave-inc/tidebreak/2866?utm_source=stack-comment-view-in-graphite" target="_blank">(View in Graphite)</a>
* `main`


<h2></h2>

naingthet:
**G2G** at a1e5769c5
Depth: medium

PR #2867, opened 2026-08-27
feat(code): persist workflow runs and drop the delivery poll

### LINKED ISSUE #2578: feat(code): give workflow runs a durable store and retire the delivery poll
Deferred from [decision 66](docs/decisions/0066-pull-request-state-is-one-store.md), which gave pull requests one durable store, one conditional fetcher, and a broadcast on change. Workflow runs never got the same treatment.

Because runs have no store, `CodeDeliveryMonitor` keeps the last poll timer decision 66 set out to remove: 5 minutes visible, 10 minutes hidden, running `queryCodeDeliveryRuns` against GitHub on a clock whether or not anything moved. That is also the only path left where a delivery read is unconditional.

Done looks like: run summaries persist as rows the way `code_pull_request` does, the reconcile sweep or an equivalent refreshes them conditionally through the existing `HostGate` in `pr_fetch.rs`, a real change emits the `delivery` notice already on `WS /code/updates`, and the monitor's remaining `SAFETY_POLL_MS` / `HIDDEN_POLL_MS` timers come out.

naingthet:
Give GitHub Actions workflow runs one durable store, the way pull requests already have.

The reconcile sweep refreshes run summaries through HostGate with If-None-Match. A 304 keeps the rows. A real change emits the existing `delivery` notice on `WS /code/updates`. The delivery monitor drops `SAFETY_POLL_MS` and `HIDDEN_POLL_MS`; it re-reads on that notice, on first mount, and when the window becomes visible.

Closes #2578


naingthet:
**NO GO** at 2efe93202
Depth: medium

1. P2: `persist_fresh_workflow_runs` returns only the GitHub page (≤100), while `stored_workflow_run_summaries` returns every `code_workflow_run` row and nothing deletes them. A 200 (first load, force refresh) paints the host window; a later `refresh: false` cache miss or 304 paints the full table, including runs that have fallen off GitHub's first page. Cap or prune the store read at `MAX_REMOTE_ITEMS_PER_REPO`, or project the store after a fresh 200, so both paths agree.

2. P2: The monitor drops its clock while deployments stay live, unpersisted fetches. A deployment that needs attention never writes a row and never emits `delivery`, so a visible window with a live socket will not see it until a PR/workflow nudge or a visibility change. Persist deployments the same way, keep a clock for that source, or stop asking the monitor for `kinds: []`.

naingthet:
Fixed on 7676af18c.

1. A 200 now prunes `code_workflow_run` rows that left GitHub's first page, and the store read is capped at `MAX_REMOTE_ITEMS_PER_REPO`. A later 304 paints the same window.

2. The monitor asks only for `workflow_run`. Deployments stay live GitHub observations and have no store or `delivery` nudge.

naingthet:
Addressed the NO GO at 2efe93202 in 5af4dc9fdf4acc8989ed0c50d552025821b51275.

1. A fresh 200 prunes `code_workflow_run` rows that fell off GitHub's first page, and the store read caps at `MAX_REMOTE_ITEMS_PER_REPO`, so a later 304 or cache miss paints the same window.

2. The monitor asks for `kinds: ["workflow_run"]` only. Deployments stay live observations. The clock stays off.

The hosted delivery mint count is 15 for the extra HostGate workflow-run read.

naingthet:
**G2G** at 5af4dc9fd
Depth: medium

PR #2868, opened 2026-08-28
docs: add the Slack sessions working design

naingthet:
A working design for driving Tidebreak sessions from Slack. It is a design page, not a decision record; where a later record disagrees, the record wins.

The shape:

- V1 surfaces are the agent DM (Slack's primary/split view for AI agents) and channel threads, entered through Slack's native add-agent affordances. Slack Code channels are the named fast follow.
- One conversation is one owner-scoped session on the person's hosted machine. The engine runs `Allow` inside a per-session confined sandbox driven by `tidebreak-supervised-agent`; confinement is the permission boundary.
- The adapter is a shared, deliberately stateful gateway add-on: durable inbound queue keyed by Slack `event_id`, an authoritative thread-routing table with a pending-row protocol, render state, and scoped grant custody with rotation, reuse detection, and immediate revocation.
- The machine grows external-session endpoints: an idempotent get-or-create over a channel-agnostic `(channel kind, external key)` binding, and a messages endpoint whose outcomes are `new_turn` or `queued` — the shipped queue-default contract. Steer is a later, capability-gated upgrade.
- Repository attachment is explicit and confirmed per repository per grant; the channel default can only propose. First use of a repository requires an owner-only confirmation.
- Stage 1 is remote execution for repository sessions, including a per-turn assistant record in the supervised agent's event vocabulary — the substrate for Slack streaming, desktop pickup, and memory later. Scratch workspaces (the only schema migration) and steer are deferred behind exit criteria recorded in the page.

The design went through adversarial review passes covering security and credential custody, delivery and idempotency failure modes, coherence with decisions 0012–0079, UX, and scope; the surviving open question (sandbox versus hosted engine for the interim) is recorded in the page.

### DIFF
--- a/docs/slack-sessions.md
+++ b/docs/slack-sessions.md
@@ -1,0 +1,691 @@
+# Slack sessions
+
+Status: proposed. This page is the working design. It is not a decision
+record. Where a later record disagrees, the record wins. This revision
+follows an adversarial review pass; the largest changes are recorded in
+"Bets and exit criteria" and "Open questions".
+
+A person talks to Tidebreak in Slack — in the agent's own chat (Slack's
+primary and split view for AI agents) or in a channel thread. Tidebreak
+runs a session for that conversation on that person's hosted machine.
+The session's engine runs in a confined sandbox. Progress streams back
+into the conversation through Slack's agent surfaces. The same session
+appears in the desktop inbox.
+
+Version one ships both surfaces: the agent DM and channel threads.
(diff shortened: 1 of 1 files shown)


naingthet:
**G2G** at 54d8223b0
Depth: fast

Docs-only working design (`docs/slack-sessions.md`). No P1/P2 in the diff: no code, no secret handling that ships, no quota/auth contract that contradicts existing decisions in a way that would ship a bug.

Issue #2875, opened 2026-08-28
Epic: remote execution for repository sessions (Slack sessions stage 1)


naingthet:
Stage 1 of the Slack sessions design (`docs/slack-sessions.md`, PR #2868): a session's engine may run in a per-session confined sandbox driven by `tidebreak-supervised-agent`, provisioned by the confining environment, with events projected into the same journal and attention model local sessions use. This unparks the remote session execution item in `docs/deferred.md`.

Repository sessions only; scratch workspaces, steer, and the Slack adapter itself are later stages gated in the design page.

Slices:

- [x] #2869 Pin the sandbox provisioning contract
- [x] #2870 Durable incarnation intent protocol
- [x] #2871 Ingest supervised-agent events into journal and attention
- [x] #2872 Per-turn assistant record in the supervised agent
- [x] #2873 Turn submission, WIP-ref resume, and reap
- [x] #2874 Concurrency cap and spend ledger
- [x] #2882 Wire the driver into the runtime

Stage-level verify (from the design page): spawn a remote session with a repository; refusal outside the GitHub App installation renders the human-readable reason; a session survives sandbox stop and resumes from WIP refs; a message during a stop waits for terminal events and reincarnates exactly once; two concurrent messages cannot double-provision; the reconcile sweep cancels an unactivated intent; the cap refuses the N+1th sandbox atomically; a killed sandbox fences with a remote reason and reap recovers the workspace.

naingthet:
Stage 1 is delivered. All seven slices merged: the pinned provisioning contract (#2876), the incarnation intent protocol (#2877), the per-turn assistant record (#2878), event ingestion with exactly-once side effects (#2879, #2880), the remote driver with WIP-ref resume and reap (#2881), the concurrency cap and spend ledger with actionable refusals (#2885), and the runtime wiring (#2886, repaired and hardened by #2888 after the two wiring PRs merged as an incompatible pair).

The stage-level verify list is pinned by tests rather than a manual pass: remote spawn from a repository origin, refusals rendering the environment's own reason, resume from WIP refs across sandbox stops, the terminal-flush gate and single reincarnation under racing messages, the stale-intent sweep, the atomic cap under a six-way race, and fence-then-reap recovery all have failing-first coverage in `code::remote` and the core incarnation suites.

[...]

Issue #2873, opened 2026-08-28
Remote sessions: turn submission, WIP-ref resume, and reap


naingthet:
Slice 5 of the Slack sessions stage 1 epic (`docs/slack-sessions.md`, PR #2868).

Submit turns to a remote session and resume across incarnations.

Deliver:

- Turn submission as inbox messages through the provisioner, reincarnating first when the sandbox is gone (serialized through the incarnation record).
- WIP-ref resume: starting-turn plus `mg-wip/<sandbox-id>-i<incarnation>` refs, with `ResumeLost` fencing when the ref is gone.
- Remote workspaces do not mkdir on the host: the workspace row exists with no host path, and orphan reclaim must not sweep it.
- Reap for remote sessions.

Done when: a session survives sandbox stop and the next turn resumes from WIP refs (integration or high-fidelity fake); a lost ref fences `ResumeLost`; reclaim leaves remote workspace rows alone under test.

naingthet:
Landed in #2881 (submit, reincarnate from last WIP ref, FlushPending, reap). Closing this slice. Wiring into CodeRuntime remains #2882.

PR #2876, opened 2026-08-28
feat(server): pin the sandbox provisioning contract

### LINKED ISSUE #2869: Remote sessions: pin the sandbox provisioning contract
Slice 1 of the Slack sessions stage 1 epic (`docs/slack-sessions.md`, PR #2868).

Pin the external provisioning contract before anything is built against it. Tidebreak asks the environment that confines the run to provision a per-session sandbox; decision 0079 built the agent side only, and the Tidebreak side of the contract does not exist.

Deliver:

- A `SandboxProvisioner` trait in the server: spawn (task, repository list, ref, ceilings), status, durable event read after a cursor, send (inbox message), cancel.
- The gateway-backed implementation, with the credential path it uses named and tested against a mock of the wire shape.
- A written note in the spec (or a decision record if the contract forces choices) recording the pinned API surface and its failure modes: preflight refusals, terminal events, event-sequence guarantees.

Done when: the trait compiles with a mock and the gateway client, both under test; the spec's Execution section references the pinned contract instead of "the confining environment's API".

naingthet:
Slice 1 of #2875 (Slack sessions stage 1); closes #2869.

Pins the provisioning contract between Tidebreak and the environment that confines a remote session's sandbox, and lands the client:

- `code/remote/wire.rs` — the runtime API shapes, mirrored locally the way `tidebreak-supervised-agent` mirrors the control endpoint (decision 0079): requests serialize exactly the documented fields (the spawn and message bodies reject unknown fields), responses default everything this side can live without, and an unknown lifecycle state decodes as non-terminal `Unknown` so a newer environment never reads as an outcome.
- `code/remote/mod.rs` — the `SandboxProvisioner` trait (spawn, status, events, send, cancel), a typed error split (named refusal / sign-in required / retryable unavailability / local invalid request), and the `RuntimeTokenSource` seam: every call presents a per-owner `mg_at_` bearer minted for `runtime:{endpoint_slug}` with the `runtime:execute` scope, so sandboxes are provisioned as their owner, never as a shared machine identity.
- `code/remote/gateway.rs` — the gateway-backed implementation. Spawn gets a 120s ceiling (preflight resolves every declared repository against its remote); an events read gets its held wait (clamped to the environment's 25s ceiling) plus slack; refusal bodies map to the environment's own code and description.

Nothing consumes the module yet — the session runtime does in #2873 — so the module carries the `scripted_harness`-style `cfg_attr(not(test), allow(dead_code))`.

Ran: `cargo test -p tidebreak-server --lib code::remote` (14 tests, including a fake runtime server asserting the bearer, paths, body shape, cursor round-trip and wait clamp, refusal mapping, and that an empty message never leaves the process), `cargo clippy -p tidebreak-server --all-targets` clean for the module, `cargo fmt --check`.

### DIFF
--- a/crates/tidebreak-server/src/code/mod.rs
+++ b/crates/tidebreak-server/src/code/mod.rs
@@ -25,6 +25,7 @@ pub(crate) mod pr_refresh;
 pub(crate) mod recap;
 pub(crate) mod reconcile;
 pub(crate) mod recovery;
+pub(crate) mod remote;
 pub(crate) mod runtime;
 pub(crate) mod scoped;
 pub(crate) mod scratch;
--- a/crates/tidebreak-server/src/code/remote/gateway.rs
+++ b/crates/tidebreak-server/src/code/remote/gateway.rs
@@ -1,0 +1,637 @@
+//! The gateway-backed [`SandboxProvisioner`].
+//!
+//! Transport only, against the routes pinned in the module docs. The URL
+//! join keeps a gateway deployed under a subpath working, the same way
+//! `connectors::gateway` joins. Per-request timeouts differ by call: spawn
+//! runs the environment's whole preflight (repository `ls-remote` included),
+//! and an events read may deliberately hold up to 25 seconds.
+
+use std::sync::Arc;
+use std::time::Duration;
+
+use async_trait::async_trait;
+use tidebreak_core::OwnerId;
+
+use super::wire::{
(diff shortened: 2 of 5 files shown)


naingthet:
**G2G** at bfec4d968
Depth: deep

The client pins the live runtime routes, spawn envelope, per-owner `runtime:{slug}` / `runtime:execute` bearer, refusal body, 25s event wait, and non-terminal unknown states. Preflight and cap refusals keep the environment's code and description so a later session can show them rather than a paraphrase, and nothing here invents lifecycle or retry policy the session runtime still owns.

Issue #2871, opened 2026-08-28
Remote sessions: ingest supervised-agent events into journal and attention


naingthet:
Slice 3 of the Slack sessions stage 1 epic (`docs/slack-sessions.md`, PR #2868).

Project the supervised agent's durable event stream into the session journal and attention, so a remote session renders and resumes like a first-class session.

Deliver:

- Event ingestion from the provisioner's cursor read into the journal: lifecycle kinds, `turn_started`/`turn_completed`, `wip_pushed`/`wip_push_failed`, `task_output` retention on the session, terminal events.
- Attention projection: Working, Stalled, DoneUnreviewed, Idle, and the remote fence causes from the incarnation slice.
- Cursor durability per session so ingestion resumes without loss after a server restart.

Done when: a driven fake event stream produces the expected journal rows and attention transitions under test, including a restart mid-stream.

naingthet:
Landed in #2879 with the cursor-commit fix in #2880. Closing this slice.

Issue #2870, opened 2026-08-28
Remote sessions: durable incarnation intent protocol


naingthet:
Slice 2 of the Slack sessions stage 1 epic (`docs/slack-sessions.md`, PR #2868).

Sessions outlive sandboxes, so incarnation lifecycle must be durable and race-free. The per-workspace in-memory turn lock does not cover remote workspaces.

Deliver:

- An appended migration (decision 0061) for a per-session incarnation record: intent → active → stopped, with sandbox id and incarnation number.
- Provisioning writes the intent row before calling spawn; activation is recorded after; stop and reincarnate serialize through the row. A message that lands mid-stop waits for the stop to complete and the dying incarnation's terminal events to be journaled.
- A reconcile sweep that cancels sandboxes whose intent never activated, so a crash between provision and store cannot leak a spending sandbox.
- New remote `FenceReason` causes: incarnation intent unresolved, sandbox lost mid-turn, terminal flush missing.

Done when: two concurrent messages cannot double-provision (test); a killed sandbox fences with the remote reason and reap recovers the workspace; the sweep cancels an orphaned intent in a test.

naingthet:
Landed in #2877 (intent protocol, cap in the same transaction, remote fence reasons, spend ledger columns). Remaining stop/sweep/resume behavior landed in #2881. Closing this slice.

PR #2879, opened 2026-08-28
feat(server): ingest sandbox events into journal and attention

naingthet:
## Summary

Event ingestion for remote sessions (#2871, epic #2875, design: `docs/slack-sessions.md`).

- `crates/tidebreak-server/src/code/remote/ingest.rs`: a pure projection (`project_event`) maps the supervised agent's event vocabulary — `supervisor_started`, `turn_started`, `assistant_record`, `turn_completed`, `turn_interrupted`, the WIP checkpoint outcomes, `task_complete`, `task_output`, `supervisor_stopped` — into journal `CodeEvent` rows and attention transitions (Working, DoneUnreviewed, needs-you on failure or interrupt). Environment lifecycle kinds are recognized and skipped; unknown kinds are counted and logged.
- `ingest_events` applies one cursor read: journal rows publish on the session bus, attention routes through `apply_attention`, `task_output` is retained on the incarnation, and the supervisor's goodbye sets the terminal-events-journaled gate reincarnation waits on. A terminal environment state read after the stream drains without a goodbye yields the fence the driver must apply — `sandbox_lost` for a violent end, `terminal_flush_missing` for a clean one (both from #2870, neither blocks the workspace).
- Durability: migration `m20260827_000023_incarnation_ingest` adds `events_cursor` and `task_output` to `code_session_incarnation`; the new core op `ingest_incarnation_event` commits each sandbox event's journal rows and the cursor advance in one transaction under the session write lock and spawn-epoch fence, so restarts replay nothing and lose nothing.

Turn-row wiring, notifications, and fence application live with the remote driver (#2873); the ingest outcome hands it what it needs.

## Testing

- `cargo test -p tidebreak-server --lib code::remote` (17 pass): projection mapping, a driven fake stream across a simulated restart mid-stream (overlapping re-read duplicates nothing, attention lands on DoneUnreviewed, cursor and terminal gate persisted), and the fence classification including the undrained-terminal case.
- `cargo test -p tidebreak-core --lib db::tests::code` (69 pass, two new: exactly-once ingest with epoch fencing, task-output retention) and `db::migration` (10 pass).
- `cargo clippy --workspace --all-targets --locked -- -D warnings`, `cargo fmt --all -- --check`.

Part of #2871.


naingthet:
**NO GO** at 4a361fb3b
Depth: medium

1. P1: `apply_one` commits the journal and `events_cursor` first, then writes `task_output` and `terminal_events_journaled` on later calls. A restart that replays the same seq hits `ingest_incarnation_event`'s `events_cursor >= seq` guard and returns before those writes. The deliverable is gone. The reincarnation gate stays false. If the next page still contains `supervisor_stopped`, `terminal_flush_journaled` is set from the projection before the cursor check, so the driver is told the gate is up when the row is not. If the next page starts after the cursor, `ingest_events` fences `TerminalFlushMissing` even though the goodbye is already in the journal. Put those writes in the same transaction as the cursor, or re-apply them on replay.

2. P2: The module comment says a crash after the journal commit leaves attention one event stale and the next batch corrects it. Replay skips `apply_attention` too. A last `turn_completed` or `turn_interrupted` has no later event, so the session stays Working after the turn is done.

PR #2880, opened 2026-08-28
fix(server): commit ingest side effects with the cursor

naingthet:
## Summary

PR #2879 auto-merged while its review findings were still being fixed; this PR carries those fixes (review verdict on #2879 at 4a361fb3b, both findings).

- Finding 1 (P1): the retained task output, the WIP resume ref, and the terminal-events-journaled gate now commit in the same transaction as the journal rows and the cursor advance (`IncarnationSideEffects` on `ingest_incarnation_event`), so a replayed sequence loses nothing; the outcome's terminal-flush answer reads the durable incarnation row after the batch instead of a replayed projection, so the fence decision cannot contradict the journal.
- Finding 2 (P2): attention is re-applied on replay — it is written in a separate transaction from the journal, and the replay is what corrects a crash between the two. A session can no longer sit at Working after its last `turn_completed`.
- Also from the amendment #2879 merged without: the `wip_pushed` payload key is `ref` (not `reference`), and the pushed ref is retained on `code_session_incarnation.last_wip_ref` (migration `m20260827_000023`) so reincarnation can resume from it.

## Testing

- `cargo test -p tidebreak-server --lib code::remote` (17 pass) — the driven-stream test now forces attention back to Working and replays the full final batch: zero re-ingested, journal unchanged at 6 rows, attention corrected to DoneUnreviewed, gate read from the row.
- `cargo test -p tidebreak-core --lib db::` (392 pass) — the exactly-once test asserts the WIP ref, task output, and gate all commit with their event's cursor advance.
- `cargo clippy --workspace --all-targets --locked -- -D warnings`, `cargo fmt --all -- --check`.

Part of #2871.

### DIFF
--- a/crates/tidebreak-core/src/code/mod.rs
+++ b/crates/tidebreak-core/src/code/mod.rs
@@ -1279,6 +1279,8 @@ pub struct CodeSessionIncarnation {
     pub events_cursor: i64,
     /// The supervisor's terminal deliverable, when the run reported one.
     pub task_output: Option<String>,
+    /// The last WIP checkpoint ref this incarnation pushed, for resume.
+    pub last_wip_ref: Option<String>,
     /// Intent time.
     pub created_at: chrono::DateTime<chrono::Utc>,
     /// Activation time, when the spawn returned.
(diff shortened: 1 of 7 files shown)


naingthet:
**G2G** at 60b53422f
Depth: medium

Side effects (task output, WIP ref, terminal-events gate) now share the cursor transaction, so a replayed sequence cannot keep the cursor and lose the rest. The batch reads `terminal_events_journaled` off the incarnation row instead of a replayed projection, so a missing goodbye still fences. Attention re-apply on replay matches the separate-transaction crash window, and `wip_pushed` reads `ref` as the agent emits it.

Issue #2874, opened 2026-08-28
Remote sessions: concurrency cap and spend ledger


naingthet:
Slice 6 of the Slack sessions stage 1 epic (`docs/slack-sessions.md`, PR #2868).

A mention is a purchase; an unbounded one is an incident. Bound it machine-side.

Deliver:

- A per-principal concurrent-sandbox cap taken as an atomic reservation in the same transaction as the incarnation intent — not a check-then-act count.
- A cumulative per-session spend ledger fed from sandbox status, with an owner-visible ceiling, because per-spawn ceilings multiply by reincarnation.
- Ceilings requested at spawn through the provisioner; refusals surface as a session event with a human-readable reason listing what is running.

Done when: the N+1th concurrent spawn is refused atomically under a race test; the ledger accumulates across incarnations; a ceiling refusal renders a reason naming the running sessions.

PR #2881, opened 2026-08-28
feat(server): drive remote sessions across sandbox incarnations

naingthet:
## Summary

The remote session driver (#2873, epic #2875, design: `docs/slack-sessions.md`).

- `crates/tidebreak-server/src/code/remote/driver.rs`: `RemoteDriver` owns the lifecycle decisions the transport-only provisioner does not.
  - Turn submission: a live sandbox gets the turn as an inbox message; a gone sandbox triggers reincarnation — reserve under the owner's cap via the incarnation intent (a cap refusal names the sessions holding live incarnations), spawn from the predecessor's `last_wip_ref` at the next turn number, activate with the returned sandbox id. A stopped predecessor whose terminal events are not journaled yet returns `FlushPending` instead of resuming without its last output; a live intent returns `ReincarnationInFlight`.
  - WIP-ref resume: first incarnation spawns from the workspace base ref; successors from the retained `mg-wip/<sandbox>-i<n>` ref. A spawn refusal whose description names that ref fences `ResumeLost` and releases the reservation.
  - `pump`: reads events after the durable cursor, projects them through ingestion, settles turn rows at terminal turn events, closes the incarnation when the environment retires the sandbox, and applies the fence the read demands (`sandbox_lost` / `terminal_flush_missing`).
  - `reap`: cancels whatever the environment still holds (best effort), closes the record, resolves the fence through the existing recovery path — and relaunches nothing; the next turn reincarnates on demand.
  - `sweep_stale_intents`: closes intent rows older than ten minutes that never activated and fences their sessions `incarnation_unresolved`.
- Remote workspaces store an empty `worktree_path` (`CodeWorkspace::is_remote`): `require_live_workspace` refuses with a distinct `workspace_remote` conflict instead of a spurious not-found, and the orphan scan's existing on-disk filter leaves the rows alone (now pinned by a test).
- `IncarnationAdmission::Admitted` boxes its row, and `CodeSessionIncarnation` carries its owner so the machine-wide sweep can act on what it finds.

Runtime wiring — constructing the gateway provisioner from configuration, remote workspace/session creation routes, registering the sweep, and branching submit/reap in `CodeRuntime` — lands with the cap-and-ledger slice (#2874), which owns the settings these calls take.

## Testing

- `cargo test -p tidebreak-server --lib code::remote` (25 pass): first-turn provisioning from the base ref, inbox delivery on a live sandbox, survival across sandbox stop with resume from the pushed WIP ref at the next turn (the issue's done-when, against a scripted provisioner), `FlushPending` hold, `ResumeLost` fencing on a ref-naming refusal with the reservation released, cap refusal naming the running session, reap without relaunch, and the stale-intent sweep fencing the session.
- `cargo test -p tidebreak-server --lib code::worktree_orphans`: the scan records the on-disk workspace and skips the remote row.
- `cargo clippy --workspace --all-targets --locked -- -D warnings`, `cargo fmt --all -- --check`, `cargo test -p tidebreak-core --lib db::` (392 pass).

Part of #2873.

### DIFF
--- a/crates/tidebreak-core/src/code/mod.rs
+++ b/crates/tidebreak-core/src/code/mod.rs
@@ -1312,6 +1312,13 @@ pub enum IncarnationAdmission {
     /// The intent row committed; provision against it. Boxed because the
     /// row dwarfs the refusal variant.
     Admitted(Box<CodeSessionIncarnation>),
+    /// The session already holds a live incarnation — another submit won
+    /// the race between observing a stopped predecessor and reserving.
+    /// Retry after it settles.
+    AlreadyLive {
+        /// The live incarnation's 1-based counter.
+        incarnation: i32,
+    },
     /// The owner is at their concurrent-sandbox cap.
     CapExhausted {
         /// Sessions holding the live incarnations, so the refusal can name
--- a/crates/tidebreak-core/src/code/mod.rs
+++ b/crates/tidebreak-core/src/code/mod.rs
@@ -674,6 +674,19 @@ pub struct CodeWorkspace {
     pub bundle_bytes: Option<i64>,
 }
 
+impl CodeWorkspace {
+    /// Whether this workspace's engine runs in a remote sandbox.
+    ///
+    /// A remote workspace has no host worktree: the clone lives inside the
+    /// sandbox, and the branch state travels as WIP refs on the origin. An
+    /// empty `worktree_path` is the marker — nothing on this machine ever
+    /// creates, reads, or reclaims a checkout for it.
+    #[must_use]
+    pub fn is_remote(&self) -> bool {
+        self.worktree_path.is_empty()
+    }
+}
+
 /// Coarse lifecycle of an observed pull request (decision 77).
 ///
 /// Deliberately narrower than the digest's free-form `state` string: the fact
@@ -1253,6 +1266,8 @@ impl IncarnationState {
 pub struct CodeSessionIncarnation {
     /// Stable id.
     pub id: CodeIncarnationId,
+    /// Owner, carried so machine-wide sweeps can act on what they find.
+    pub owner: crate::OwnerId,
     /// Owning session.
     pub session_id: CodeSessionId,
     /// 1-based counter within the session; the agent names WIP refs with it.
@@ -1294,8 +1309,9 @@ pub struct CodeSessionIncarnation {
 /// The outcome of asking for a new incarnation under the owner's cap.
 #[derive(Debug, Clone, PartialEq)]
 pub enum IncarnationAdmission {
-    /// The intent row committed; provision against it.
-    Admitted(CodeSessionIncarnation),
+    /// The intent row committed; provision against it. Boxed because the
+    /// row dwarfs the refusal variant.
+    Admitted(Box<CodeSessionIncarnation>),
     /// The owner is at their concurrent-sandbox cap.
     CapExhausted {
         /// Sessions holding the live incarnations, so the refusal can name
--- a/crates/tidebreak-core/src/db/ops/code/incarnation.rs
+++ b/crates/tidebreak-core/src/db/ops/code/incarnation.rs
@@ -418,6 +418,37 @@ pub async fn latest_pushed_wip_ref(
         .and_then(|row| row.last_wip_ref))
 }
 
+/// Forget a WIP checkpoint ref the origin no longer advertises.
+///
+/// The remote analog of dropping a rejected resume ref: a spawn already
+/// refused this ref, so every retry through it refuses identically. Clearing
+/// it lets the resume walk-back fall through to an earlier checkpoint, or to
+/// the base ref when none is left.
+pub async fn forget_session_wip_ref(
+    store: &DbStore,
+    owner: &OwnerId,
+    session_id: CodeSessionId,
+    reference: &str,
+) -> Result<()> {
+    let now = database_now(&store.conn).await?;
+    entities::code_session_incarnation::Entity::update_many()
+        .col_expr(
+            entities::code_session_incarnation::Column::LastWipRef,
+            sea_orm::sea_query::Expr::value(Option::<String>::None),
+        )
+        .col_expr(
+            entities::code_session_incarnation::Column::UpdatedAt,
+            sea_orm::sea_query::Expr::value(now),
+        )
+        .filter(entities::code_session_incarnation::Column::Owner.eq(owner.as_str()))
+        .filter(entities::code_session_incarnation::Column::SessionId.eq(session_id.0))
+        .filter(entities::code_session_incarnation::Column::LastWipRef.eq(reference))
+        .exec(&store.conn)
+        .await
+        .map_err(store_err)?;
+    Ok(())
+}
+
 /// The session's newest incarnation, in any state.
 pub async fn latest_incarnation(
     store: &DbStore,
--- a/crates/tidebreak-core/src/db/ops/code/incarnation.rs
+++ b/crates/tidebreak-core/src/db/ops/code/incarnation.rs
@@ -110,13 +110,14 @@ pub async fn create_incarnation_intent(
     if let Some(last) = &last {
         // One live incarnation per session: the serialize-stop-then-resume
         // rule. A successor minted beside a live predecessor would race it
-        // for the same engine session.
+        // for the same engine session. Two submits can both observe a
+        // stopped predecessor and race to here, so the loser gets an
+        // admission answer to relay, not an error.
         if IncarnationState::from_str(&last.state) != Some(IncarnationState::Stopped) {
             txn.rollback().await.map_err(store_err)?;
-            return Err(AgentError::Store(format!(
-                "session {} already has a live incarnation {}",
-                session_id.0, last.incarnation
-            )));
+            return Ok(IncarnationAdmission::AlreadyLive {
+                incarnation: last.incarnation,
+            });
         }
     }
     let incarnation = last.as_ref().map_or(1, |row| row.incarnation + 1);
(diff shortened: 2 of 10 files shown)


naingthet:
**NO GO** at 9c44fb4f5
Depth: deep

1. P1: `submit_turn` waits on `terminal_events_journaled`, but nothing except ingesting `supervisor_stopped` on an Active incarnation ever sets that bit — and `pump` returns immediately once the row is not Active. A live sandbox that refuses a message is `stop_incarnation`'d and then `FlushPending`; later pumps never drain it, so the goodbye cannot land and reincarnation never starts. Reap has the same hole: it closes the row as `reaped` (or finds it already stopped after a `SandboxLost` / `TerminalFlushMissing` fence) and does not raise the gate, so the next turn after a successful reap is `FlushPending` forever instead of reincarnating on demand. The design's "message during a stop waits, then reincarnates exactly once" and "reap, then the next turn reincarnates" both fail. Drain (or explicitly waive the gate) before you stop the row, and have reap mark the predecessor resumable.

2. P2: A successful `spawn` whose `activate_incarnation` then fails (sweep closed the intent, for example) returns the store error and never `cancel`s the lease you already hold. The protocol comment on activate says you must treat that sandbox as orphaned and cancel it; leaving it running while the cap slot is released lets a later turn provision a second sandbox for the same session.

3. P2: A turn against an Active sandbox always inbox-sends and inserts another `Running` turn. There is no durable FIFO when a turn is already running, and `settle_turn_rows` closes only the latest running row from the first terminal event in the batch. A second message mid-turn (desktop, Slack, or a retry) can mark the new turn completed from the previous turn's events and leave the real one running. Queue at the turn boundary the way the rest of the product does, or refuse the second submit until the current turn settles.

naingthet:
**NO GO** at 141ee7182
Depth: deep

1. P1: `submit_turn` resumes from `latest_incarnation().last_wip_ref`, falling back to `workspace.base_ref`. That is the wrong row once a successor exists. A spawn failure, a swept intent, or any other stopped reservation writes a new incarnation with `last_wip_ref = None` and then becomes latest; the next turn clones `main` and drops the predecessor checkpoint the issue's done-when says you keep. The first attempt still sees the predecessor, so `a_refusal_naming_the_wip_ref_fences_resume_lost` and `the_next_turn_resumes_from_the_predecessors_wip_ref` do not catch a retry. Walk back to the last incarnation that actually pushed, or copy that ref onto the reservation before you spawn.

2. P2: A live sandbox that refuses `send` stays Active so the pump can drain, but a non-retryable `events()` error (`Refused`, `SignInRequired`) returns `Store` and leaves the row open. Submit then answers `FlushPending` forever: nothing raises the goodbye gate, nothing fences, and reap is never offered. Treat that read as `sandbox_lost` (or waive the gate) instead of parking the session on a drain that cannot happen.

naingthet:
**NO GO** at 4f546badf
Depth: deep

1. P1: `submit_turn` writes `starting_turn` as the session ordinal and `settle_turn_rows` only closes a running turn when the event's `turn` payload equals that ordinal. `SpawnArguments` never carries `starting_turn` (or `incarnation`), and the supervised agent defaults `MODEL_GATEWAY_SANDBOX_STARTING_TURN` to 1. After the first sandbox stop, the successor emits `turn_completed {turn: 1}` against session turn 2, the row stays `Running`, and every later submit is `TurnInFlight` with no fence and no reap. The resume tests never pump the successor, so they miss it. Put the ordinal on the spawn the agent already reads, or settle against this incarnation's own numbering.

2. P2: `SignInRequired` on `events()` is lumped in with a dead stream: an Active row is stopped and the session is fenced `SandboxLost` without `cancel`. Token expiry is not a lost sandbox. The lease keeps running, the cap slot is released, and reap-then-reincarnate provisions a second sandbox beside it. Hold the row and surface sign-in, or cancel before you release the slot.

naingthet:
**NO GO** at 6b2a8998e
Depth: deep

1. P1: `pump` ingests the terminal batch — `apply_attention` writes DoneUnreviewed or NeedsYou — then, when the sandbox is terminal, `persist_session`s the in-memory session whose attention is still Working from `start_turn_row`. `should_replace` allows that overwrite, so every successful remote turn that stops the sandbox leaves the inbox (and Slack) on Working after the work is done. Copy the ingest attention onto the session before that persist, or reload it from the store.

2. P2: `submit_turn` treats `SignInRequired` on `send` as a store error. The pump path holds the row and asks for sign-in; the submit path does not, so a token that expires between pumps fails the user's message with a generic error and never surfaces sign-in. Hold the row and set the same needs-you attention you already use for `events()`.

naingthet:
**NO GO** at 6ecd6f24d
Depth: deep

1. P1: `start_turn_row` persists `lifecycle = Running` with no `child_pid`, and a terminal pump then persists `Idle`. `CodeRuntime::start` already scans every Running session and treats a missing pid as a dead local worker (`recover_interrupted_session`), then `attach_and_spawn_worker`s every session that is not Ended or Fenced. A desktop restart mid-remote-turn interrupts the turn, writes "session recovered after the engine process exited", and leaves the sandbox lease Active — still spending, still holding the cap. A restart after a successful remote turn tries to spawn a host harness against the empty worktree. Skip remote sessions on that boot path (you already have `workspace.is_remote()`), or do not put them in the local-worker lifecycle until you do.

2. P2: Two `submit_turn`s that both observe a Stopped predecessor race through the flush gate: the first inserts the Intent, the second hits `create_incarnation_intent`'s "already has a live incarnation" store error instead of `ReincarnationInFlight`. The turn row does not exist until spawn returns, so `TurnInFlight` does not cover the window. Desktop and Slack will hit it. Return `ReincarnationInFlight` for that conflict.

naingthet:
**NO GO** at b18918517
Depth: deep

1. P1: `ResumeLost` fences so a reap can start fresh, but it never forgets the gone WIP ref. `fence_session` only clears `harness_resume_ref`; `last_wip_ref` stays on the predecessor, and `reap` does not touch it. The next turn after a successful reap still calls `latest_pushed_wip_ref`, spawns the same missing `mg-wip/…` ref, and fences `ResumeLost` again. Local resume-lost drops the stored ref for this reason; the remote analog is missing. Clear that checkpoint when you fence (or skip it on the post-reap spawn), or the session cannot continue. `a_refusal_naming_the_wip_ref_fences_resume_lost` never submits after reap, so it misses the loop.

naingthet:
**NO GO** at 6bbbfc092
Depth: deep

1. P1: `start_turn_row` persists `lifecycle = Running`, and `pump` only writes `Idle` when the environment reports the sandbox terminal. A turn that settles while the sandbox stays live — the inbox-delivery path, and the successor pump in `the_next_turn_resumes_from_the_predecessors_wip_ref` — leaves the session Running with `DoneUnreviewed` on the row. `sweep_stalled` walks every Running session and `should_replace` lets a heuristic stall overwrite `DoneUnreviewed`, so the inbox and Slack show "no activity for 90s" after a successful turn instead of done. `compute_attention` has the same hole: Running never derives done. Persist Idle when the running turn settles, the way a local session does when the engine is still up but the turn is not.

naingthet:
**NO GO** at acaaf9f9d
Depth: deep

1. P1: `pump` sets the session Idle when the sandbox is terminal, but it never interrupts an open turn. `settle_turn_rows` only closes a matching ordinal, and a journaled goodbye does not fence, so reap is never offered. A `send` that still succeeds while the row is Active — the design's message-during-stop window, including after turn 1 has settled and while `events()` is in flight — inserts a Running turn the dying incarnation will not complete. Every later `submit_turn` is then `TurnInFlight`. Interrupt (or fence) that turn when you close the incarnation, the way a dead local worker does.

naingthet:
**NO GO** at 79e00a1a8
Depth: deep

1. P1: `closing_the_incarnation_interrupts_an_unsettled_turn` closes the turn and persists Idle, but it never replaces the Working attention `start_turn_row` wrote. `supervisor_stopped` projects no attention, so the reload-then-persist copies that Working onto an Idle session. The inbox (and Slack, once wired) keep showing work in progress for a message the dying incarnation will not run. Local dead-worker recovery journals `TurnInterrupted` and sets needs-you; `project_event("turn_interrupted")` does the same. Do that here — the test never asserts attention, so it misses the lie.

2. P2: Mid-turn `SignInRequired` on `events()` correctly holds the row and sets needs-you, but it leaves `lifecycle = Running`. `sweep_stalled` walks every Running session and `should_replace` lets a heuristic stall overwrite Lifecycle needs-you. The turn is still Running, so the next submit is `TurnInFlight`. After 90s the inbox shows a stall, the sign-in prompt is gone, and the user cannot send. Pause the running lifecycle while you wait for sign-in, or keep that attention from being stall-replaced.

naingthet:
**G2G** at f8313a236
Depth: deep

The driver now serializes stop and resume the way the Slack design requires: a live sandbox gets the inbox, a gone one reincarnates from the last pushed WIP ref, a dying one holds FlushPending until the goodbye lands, and reap waives that gate without relaunching. Inbox and Slack-bound attention stay honest through the paths that used to lie — Idle on settle so the stall sweep cannot overwrite done, Structured sign-in that a heuristic cannot replace, NeedsYou when a stop-window turn is interrupted — and boot recovery leaves remote sessions on their sandbox instead of attaching a host worker.

naingthet:
**G2G** at 550ffca04
Depth: deep

Closing an incarnation over an unsettled turn now journals the interrupt and shows needs-you instead of Working, and the sign-in prompt is Structured so a stall sweep cannot replace it while the lease stays live. Stop and resume still fit Slack: a live sandbox gets the inbox, a dying one holds FlushPending until goodbye, a gone one reincarnates from the last pushed WIP, reap waives the gate without relaunching, and boot recovery leaves remote sessions on their sandbox.

PR #2883, opened 2026-08-28
feat(browser): add trusted native semantic actions

### LINKED ISSUE #2339: feat(browser): implement the macOS semantic browser driver

naingthet:
Closes #2339

## Summary

- Synthesizes trusted native click, hover, focus, fill, select, check, key, and scroll input in macOS WKWebView.
- Keeps target identity in one isolated `WKContentWorld` and revalidates the ref and fingerprint immediately before dispatch.
- Gates `browser_act` through capability files, server routes, foreground client execution, CLI/MCP registration, and Grok instructions.
- Waits for WebKit pending input queues and invalidates the semantic snapshot after dispatch.

## Validation

- `cargo check -p tidebreak-desktop`
- `cargo test -p tidebreak-desktop browser_semantics` — 16 passed
- `cargo test -p tidebreak-cli browser` — 68 passed
- `cargo test -p tidebreak-server code_browser` — 29 passed
- `cargo test -p tidebreak-server foreground_browser_` — 6 passed
- `cargo test -p tidebreak-server code::browser_channel::tests` — 25 passed
- Browser fixture resolver and privacy tests — 10 passed
- Browser bridge tests — 63 passed

## Native harness

The manual native harness was inconclusive because this macOS session stayed inactive and non-key. Native click and hover both failed under that condition, so the harness did not provide a truthful end-to-end delivery result. The implementation follows WebKit’s macOS automation event path and fails closed for stale, hidden, obscured, and unsupported targets.


naingthet:
**NO GO** at 6807c6740
Depth: deep

P1: Select walks the whole `NSApplication` accessibility tree (`find_and_press_menu_item` from the app root) and presses the first menu item whose label matches the option. A page option named File, Edit, Help, Copy, or Hide can fire Tidebreak's own menu instead of the `<select>`. Scope that search to the webview subtree, or drop the AX shortcut and keep only the in-control keyboard fallback.

P1: Click, hover, and key still report `ok` after `makeKeyAndOrderFront` plus `window.sendEvent` with no check that the window is key or the app is active. This PR already records that native click and hover failed in an inactive, non-key session, and #2339 only permits control while the tab is active and visible. `HiddenTab` exists and is unused here. Fail closed unless the window is key. Do not tell the model the action completed.

P2: Fill sets accessibility focus with fire-and-forget `setAccessibilityFocused`, then `selectAll` / `insertText` on `window.firstResponder` (the WKWebView). If AX focus did not move, text lands on whatever WebKit already had focused, not the re-resolved target.

naingthet:
**NO GO** at ba3bd7f53
Depth: deep

P1: Select clicks the control, then sends window-scoped Arrow keys and Enter, then reports `ok` without reading `HTMLSelectElement.value` back. Those keys are not scoped to a popup. Enter on a form submits. Select is also omitted from consequential confirmation, so a dropdown action can submit the page. The click runs on the main thread inside the JS completion handler while that callback holds its mutex, so a native menu-tracking run loop blocks timeout and the key fallback never runs while the menu is open. Confirm the selected value or fail. Do not send a trailing Enter the page can treat as submit.

P2: ScrollIntoView posts one untargeted pixel-wheel delta to the WKWebView and reports success without checking the target is in view. Nested overflow containers, typical of SPA shells, never move, so the next click returns `target_obscured`. Scroll until the target is in view, or return a typed failure.

naingthet:
**G2G** at a9ba9ef80
Depth: deep

PR #2885, opened 2026-08-28
feat(code): remote session concurrency cap and spend ledger

### LINKED ISSUE #2874: Remote sessions: concurrency cap and spend ledger

naingthet:
Closes #2874. Slice 6 of the Slack sessions stage 1 epic #2875 (`docs/slack-sessions.md`).

A mention is a purchase; an unbounded one is an incident. This bounds it machine-side, on top of the atomic cap reservation #2877 landed:

- The driver gates every turn — delivery and spawn alike — on a cumulative per-session spend ceiling (`RemoteSpawnSettings.session_spend_ceiling_microusd`). Per-spawn ceilings multiply by reincarnation; this one does not.
- The pump feeds the ledger from the environment's own meter: each event read also records `spend_microusd` from sandbox status onto the incarnation row, and `session_spend_microusd` sums across incarnations.
- Cap and ceiling refusals journal a human-readable notice — naming the running sessions, or the dollars spent against the ceiling — and set needs-you attention, since a refused turn produces no sandbox event that would otherwise say why.
- A core race test pins the reservation semantics: six concurrent intents against a cap of two admit exactly two, with the tasks interleaving at every await inside the op.

Per-spawn ceilings were already requested through the provisioner (`SpawnArguments.spend_ceiling_microusd`, #2876).

Validation: `cargo test -p tidebreak-server --lib code::remote` (37 passing), `cargo test -p tidebreak-core --lib the_cap_admits`, workspace clippy with `-D warnings`, `cargo fmt --check`.

### DIFF
--- a/crates/tidebreak-core/src/db/tests/code.rs
+++ b/crates/tidebreak-core/src/db/tests/code.rs
@@ -4527,6 +4527,46 @@ async fn the_incarnation_cap_refuses_and_names_the_running_sessions() {
     admitted(admit(&store, &owner, session_b, 1, 1).await);
 }
 
+/// The cap is a reservation taken with the intent, not a check-then-act
+/// count: six concurrent intents against a cap of two admit exactly two.
+/// The tasks interleave at every await inside the op, which is where a
+/// check-then-act count would double-admit.
+#[tokio::test]
+async fn the_cap_admits_exactly_the_cap_under_a_race() {
+    let (_dir, store) = temp_store().await;
+    let store = std::sync::Arc::new(store);
+    let owner = OwnerId::local();
+    let mut sessions = Vec::new();
+    for index in 0..6 {
+        let (session, _) = seed_owner(&store, &owner, &format!("race-{index}")).await;
+        sessions.push(session);
+    }
+    let mut handles = Vec::new();
+    for session in sessions {
+        let store = std::sync::Arc::clone(&store);
+        let owner = owner.clone();
+        handles.push(tokio::spawn(async move {
+            crate::db::code::create_incarnation_intent(&store, &owner, session, 1, 2)
+                .await
+                .unwrap()
+        }));
+    }
+    let mut winners = 0;
+    let mut refusals = 0;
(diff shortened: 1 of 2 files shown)


naingthet:
**NO GO** at 09118422a
Depth: deep

1. P2: Cap refusal names occupying sessions by UUID. Desktop renders that HarnessNotice as raw text, and CodeSession has no title — the owner knows workspaces, not ids. docs/slack-sessions.md says a cap refusal lists what is running with how to stop one, not just a number. Load the occupying workspace titles and tell the owner to stop one of those sessions. Thread links wait for Slack; names do not.
2. P2: Spend-ceiling copy tells the owner to raise the ceiling. Nothing in this slice or the desktop exposes a session spend ceiling, and a live sandbox that just failed the ledger still holds a cap slot. The instruction is not something they can do. Say spend is exhausted and they need to stop the session, or actually stop that sandbox so it does not keep the concurrency cap.

naingthet:
**G2G** at 284d7faa9
Depth: deep

Cap refusals name occupying workspaces and tell the owner to stop one of those sessions, which is what desktop can actually do. Spend-ceiling refusals stop the idle sandbox so it does not keep a cap slot, and the notice says the session takes no more turns instead of asking anyone to raise a ceiling this slice does not expose.

PR #2886, opened 2026-08-28
feat(server): wire the remote driver into the runtime

### LINKED ISSUE #2882: Remote sessions: wire the driver into the runtime
Slice of the Slack sessions stage 1 epic #2875 (`docs/slack-sessions.md`), carved out of #2873: PR #2881 built `RemoteDriver` behind the module boundary; nothing constructs or calls it yet.

Deliver:

- Provisioner construction from configuration: the gateway base URL, profile, and a `RuntimeTokenSource`.
- Remote workspace and session creation paths that record the empty `worktree_path` marker without touching the local worktree machinery.
- `CodeRuntime` submit, pump scheduling, and reap branching on `workspace.is_remote()` so a remote session never reaches the local harness spawn.
- Stale-intent sweep registration beside the existing reconcile sweeps.

Done when: a remote session created through the runtime carries a turn to a sandbox and back without any local harness involvement, and a server restart neither respawns nor orphans it.

naingthet:
Closes #2882.

PR #2881 built `RemoteDriver`. This wires it into the running server.

- Boot constructs `GatewayProvisioner` from `TIDEBREAK_RUNTIME_ENDPOINT`, `TIDEBREAK_RUNTIME_PROFILE`, and owner-scoped `RuntimeTokens` exchanged for the `runtime:{endpoint}` audience. Endpoint and profile are required together; a runtime endpoint without gateway authentication is a boot error.
- `create_remote_workspace` records the empty `worktree_path` marker and never calls local worktree machinery. `create_remote_session` inserts a row and does not probe or spawn a host harness.
- `submit_turn` and `reap` branch on `workspace.is_remote()`, so a remote session never reaches local harness spawn. The remote sweep registers beside the reconcile sweeps, closes stale intents, and keeps a pump task on every incarnation that still owes events.

## Testing

- `cargo test -p tidebreak-core --lib config` — runtime endpoint and profile load together, and a half-configured pair refuses at boot.
- `cargo test -p tidebreak-server --lib -- remote::service::tests obo_gateway::tests::runtime_tokens` — a remote session created through the runtime carries the empty worktree marker, submit provisions a sandbox (the test fails if submit still went local), reap cancels the sandbox and spawns nothing, and runtime tokens exchange for the endpoint audience.

### DIFF
--- a/crates/tidebreak-core/src/config.rs
+++ b/crates/tidebreak-core/src/config.rs
@@ -127,6 +127,16 @@ pub struct Config {
     /// [`Config::bind_addr`].
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub listen_addr: Option<SocketAddr>,
+    /// The confined-sandbox runtime endpoint slug remote sessions are
+    /// provisioned through (`docs/slack-sessions.md`). Meaningful only on a
+    /// gateway-authenticated hosted machine, where the sandbox calls ride the
+    /// same gateway as authentication; absent, remote sessions are off.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub runtime_endpoint: Option<String>,
+    /// The administrator-defined sandbox profile named on every remote spawn.
+    /// Required together with [`Config::runtime_endpoint`].
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub runtime_profile: Option<String>,
     /// Where new code-mode worktrees land when no `code_worktree_root` setting
     /// is stored.
     ///
@@ -223,6 +233,8 @@ impl Config {
             auth_gateway_verifier_url: None,
             public_url: None,
             listen_addr: None,
+            runtime_endpoint: None,
+            runtime_profile: None,
             code_worktree_root_default: None,
         }
     }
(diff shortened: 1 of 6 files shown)


naingthet:
**NO GO** at f33a07f30
Depth: deep

1. P1: `submit_turn` is queue-default, and `RemoteTurnOutcome::TurnInFlight` tells you to park at the turn boundary the way every other submit path does. `relay_remote_outcome` ignores that: with `queue_if_busy` true it still returns `session_busy`, and nothing promotes a parked row when a remote turn ends. A follow-up while a sandbox turn is running fails instead of queueing. That breaks the contract in `docs/slack-sessions.md` and the public `submit_turn` API this PR just branched.

2. P2: `interrupt` still goes through `require_worker`. A remote session has no host worker, so Stop answers `session_worker_missing` and the sandbox keeps spending. Reap only runs on a fenced session, so there is no live stop path.

3. P2: `RuntimeTokens` holds one process-wide cache mutex across the gateway exchange. Inference tokens single-flight per owner. On a hosted machine, one slow or hung mint blocks every other owner's spawn and pump.

naingthet:
21015da59 addresses the NO GO at f33a07f30:

1. Follow-ups while a remote turn runs park as durable queue rows. The remote sweep promotes the head once the session is idle.
2. Interrupt on a remote session sends to the live sandbox. It does not look for a host worker.
3. RuntimeTokens single-flights per owner. One hung mint no longer blocks another owner's spawn or pump.

naingthet:
**NO GO** at 21015da59
Depth: deep

1. P1: `set_permission_mode` still takes the local relaunch path. `POST /code/sessions/{id}/mode` on an idle remote session confirms the new mode, then `attach_and_spawn_worker` launches a host harness against the empty worktree marker. Recovery already skips that attach so a healthy remote session is not spawned locally; the live settings route does not. On a hosted machine that has a harness installed, the engine runs unconfined on the host instead of in the sandbox.

2. P2: `end_session_row` and `archive_workspace` never cancel the sandbox. Interrupt now reaches the environment; ending the session or archiving the workspace only marks the row ended. The sandbox keeps spending until its ceilings fire.

3. P2: Host worktree reads still assume a checkout. `workspace_tree` and the files/diff paths that skip `require_live_workspace` treat the empty marker as a missing tree (`workspace worktree is gone`) instead of `workspace_remote`. The same session is supposed to appear in the desktop inbox; opening it does not refuse cleanly.

naingthet:
88513b92e addresses the NO GO at 21015da59:

1. Permission-mode changes on a remote session persist on the row. They do not relaunch a host harness against the empty worktree.
2. Ending a remote session cancels its sandbox. Archive skips local worktree removal.
3. Host tree, files, and diff reads refuse a remote workspace as `workspace_remote` instead of treating the empty path as a missing checkout.

PR #2887, opened 2026-08-28
feat(server): wire remote sessions into the runtime

### LINKED ISSUE #2882: Remote sessions: wire the driver into the runtime

naingthet:
Closes #2882. Slice of the Slack sessions stage 1 epic #2875 (`docs/slack-sessions.md`); gives the driver from #2881 its callers.

- Configuration names the sandbox runtime: `TIDEBREAK_RUNTIME_ENDPOINT` and `TIDEBREAK_RUNTIME_PROFILE`, required together and refused at boot without gateway authentication, since sandboxes are provisioned as their owner.
- `OboGateway::runtime_tokens` exchanges each owner's machine-bound bearer for the `runtime:{endpoint_slug}` audience, cached per owner with single-flight refresh; a caller with no live session fails as sign-in-required rather than borrowing another identity.
- `CodeRuntime` branches on the durable remote marker (`workspace.is_remote()`): submit routes through the driver with queue-default parking (no worker, no worktree lock, no harness probe), reap goes through the driver's cancel-and-recover, and `create_remote_workspace` / `create_remote_session` write rows without touching the filesystem. A repository with no recorded origin refuses — the sandbox would have nothing to clone.
- One sweep (`remote/service.rs`, registered beside the existing sweeps) schedules everything: it expires stale intents, keeps a pump task holding the long event wait on every incarnation that owes events, and promotes the queue head of idle remote sessions as an atomic `promote_queued_turn`. Pump tasks respawn from durable state, so a restart neither respawns nor orphans a session (#2881 pinned boot recovery leaving remote sessions on their sandbox).

Deferred, noted in #2882: HTTP routes and desktop surface for the remote create paths land with the Slack adapter slice, their first external caller; trigger turns and attachments refuse with typed conflicts until then.

Validation: `cargo test -p tidebreak-server --lib code::remote` (40 passing, including three runtime-level tests: turn to sandbox and back with queue promotion, reap through the driver, origin validation), `obo_gateway` (25) and `config` (15) suites, workspace clippy with `-D warnings`, `cargo fmt --check`.

### DIFF
--- a/crates/tidebreak-core/src/config.rs
+++ b/crates/tidebreak-core/src/config.rs
@@ -127,6 +127,16 @@ pub struct Config {
     /// [`Config::bind_addr`].
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub listen_addr: Option<SocketAddr>,
+    /// The confined-sandbox runtime endpoint slug remote sessions are
+    /// provisioned through (`docs/slack-sessions.md`). Meaningful only on a
+    /// gateway-authenticated hosted machine, where the sandbox calls ride the
+    /// same gateway as authentication; absent, remote sessions are off.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub runtime_endpoint: Option<String>,
+    /// The administrator-defined sandbox profile named on every remote spawn.
+    /// Required together with [`Config::runtime_endpoint`].
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub runtime_profile: Option<String>,
     /// Where new code-mode worktrees land when no `code_worktree_root` setting
     /// is stored.
     ///
@@ -223,6 +233,8 @@ impl Config {
             auth_gateway_verifier_url: None,
             public_url: None,
             listen_addr: None,
+            runtime_endpoint: None,
+            runtime_profile: None,
             code_worktree_root_default: None,
         }
     }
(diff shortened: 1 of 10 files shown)


naingthet:
**NO GO** at 5870f3441
Depth: deep

1. P1: `RemoteDriver::submit_turn_from` delivers to the sandbox (send or spawn) before `start_turn_row` claims the queued row. When `promote_queued_turn` returns false, `start_turn_row` still inserts a turn under that queued id. An edit or reorder after the sweep snapshot leaves the same id in both tables. After the recorded turn settles, every later sweep tick sends the remaining head again, then dies on the duplicate turn primary key, so the row never leaves the queue. The local worker promotes first and treats a stale CAS as `QueuedTurnStale` with nothing written.

2. P2: `promote_remote_queue_heads` only logs `Err`. `Ok(CapExhausted)`, `Ok(SpendExhausted)`, and `Ok(SignInRequired)` leave the session idle with the head still queued, so the 2s sweep retries them. Cap and spend refusals journal a notice and poke attention on every tick; spend exhaustion also re-issues cancel.

naingthet:
Superseded: #2886 merged the same wiring first (this branch and that one converged on the same code), and the two findings above are fixed with pinning tests in the follow-up that also repairs the #2885 x #2886 build breakage on main.

PR #2888, opened 2026-08-28
fix(server): repair the remote wiring build and harden queue promotion

naingthet:
Repairs main and closes out the #2882 review findings.

#2885 and #2886 merged as a broken combination: the spend-ledger slice added a required `RemoteSpawnSettings` field and a `SpendExhausted` submit outcome that the wiring, written against the older driver, never saw. main does not compile at 36c36b12e. This restores the build: `default_settings` carries the $20 session ceiling and the submit relay answers spend exhaustion with a typed `session_spend_exhausted` conflict.

It also fixes the two findings Shipright raised against PR #2887 (the parallel wiring PR, now superseded), which apply equally to the promotion path #2886 merged:

- The sweep now promotes by claiming the queued row atomically (`promote_queued_turn`), the way a local worker does, instead of deleting the row after delivery. A stale claim — the row was edited, reordered, or retracted after the sweep's snapshot — records the delivered turn under a fresh id and leaves the surviving row queued under its own id, so an edit can never collide with the delivered turn's primary key or wedge the queue on a duplicate-key error.
- Machine-side refusals no longer repeat every two-second tick. Spend exhaustion pauses the queue (the condition is permanent for the session; each retry re-journaled the refusal and re-cancelled the sandbox). Cap and sign-in refusals hold promotion for 150 seconds before retrying quietly.

Pinned by three new tests: a stale promotion records the turn without eating the edit and the edited row promotes cleanly afterwards; a cap-refused promotion journals one notice and holds; a spend-exhausted promotion pauses the queue and journals nothing more.

Validation: `cargo test -p tidebreak-server --lib code::remote` (47 passing), workspace clippy with `-D warnings`, `cargo fmt --check`.

### DIFF
--- a/crates/tidebreak-server/src/code/remote/driver.rs
+++ b/crates/tidebreak-server/src/code/remote/driver.rs
@@ -67,6 +67,7 @@ pub(crate) enum RemoteTurnOutcome {
         /// The turn row, running.
         turn: Box<CodeTurn>,
         /// The incarnation now active.
+        #[cfg_attr(not(test), allow(dead_code))]
         incarnation: Box<CodeSessionIncarnation>,
     },
     /// The owner's concurrency cap refused, naming what runs.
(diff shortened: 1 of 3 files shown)


naingthet:
**G2G** at 69349cde8
Depth: deep

The wiring compiles again: default spawn settings carry the session ceiling, and the submit relay maps spend exhaustion. Promotion claims through `promote_queued_turn`; a stale CAS records the already-delivered text under a new id and leaves the surviving row queued, which the pin covers. Cap and sign-in refusals hold for 150s, and spend exhaustion pauses so the sweep does not re-journal or re-cancel every tick.

PR #2897, opened 2026-08-28
feat(code): rewrite completed turns into lucid prose

### LINKED ISSUE #2501: feat: run a hook at the end of a turn, starting with a plain-language rewrite
The closing message of a turn is written in the harness's own register: hedges,
restated tool calls, names it invented while reasoning, and a paragraph where a
sentence would do. Someone supervising several sessions reads every one of
those messages, so the cost lands on the reader rather than on the agent that
wrote it. Rewriting that message into plain, scannable prose is worth a pass of
its own — and once something runs at the end of a turn, other end-of-turn work
has somewhere to attach.

## The seam

Turn completion is already a distinct moment on both surfaces, after the text
settles and before the terminal event publishes:

- Chat: `AgentEvent::TurnCompleted` in `crates/tidebreak-server/src/turn_worker.rs`.
- Code: `CodeEvent::TurnCompleted` in `crates/tidebreak-server/src/code/session_worker.rs`,
  where `settle_streamed_text` has already made the assistant text durable
  (record 58).

Running work beside a turn on the utility model is settled practice too. Chat
titling and workspace titling both do it —
`crates/tidebreak-server/src/chat_titling.rs` and
`crates/tidebreak-server/src/code/titling.rs` — spawned off to the side, with
every failure leaving the original in place so the next turn can try again. A
rewrite pass is that shape with a different prompt.

## Shape

- One hook point, running once per completed turn, off the turn's critical
  path. A hook that fails, times out, or finds no utility model configured
  leaves the turn exactly as it completed.
- The first hook rewrites the closing message: second person, active voice, one
  idea per sentence, no restatement of the tool calls the transcript already
  shows.
- **Store the rewrite beside the original, never over it.** The journal is the
  record of what the agent said, and a hook that overwrote it would make the
  transcript unreliable for replay, resume, and anything that reads history.
- The interface leads with one of the two and reaches the other in a click.
  Which one leads is a taste call to make when the surface is built.
- Storybook stories for the states that matter: rewriting, rewritten, hook
  failed, hook off.
- No shell commands and no webhooks. Record 60 reserves those for their own
  record and nothing here argues for reopening that.

## Open questions

- **Does this need a decision record?** Record 60 makes a trigger a durable
  rule on pull-request facts, bound per repository. A turn-end hook is neither:
  the condition is a session lifecycle edge, and the binding is a session or a
  global preference. Giving the trigger substrate a second condition source is
  a design decision, not an implementation detail — write the record first if
  it goes that way.
- **Which surface first.** Code sessions produce the longest closing messages
  and carry the most parallel supervision, so they are the likelier first
  target. The chat path is the same seam.
- **On by default, or opt-in.** The rewrite costs one utility-model call per
  turn.

## Done when

A completed turn produces a plain-language version of its closing message, the
original stays readable beside it, and turning the hook off — or having no
utility model — changes nothing about the turn itself.

naingthet:
A completed code turn can show a lucid rewrite of its closing message without losing the original.

The rewrite runs on the utility model after the journal write, beside recap. It is off by default. Turn it on in Coding harnesses → Transcript. Failure, timeout, or a machine with no utility model leave the turn unchanged.

Closes #2501

### DIFF
--- a/crates/tidebreak-core/src/code/mod.rs
+++ b/crates/tidebreak-core/src/code/mod.rs
@@ -13,13 +13,13 @@ mod event;
 pub use caps::{CapLevel, HarnessCaps, HarnessCommand, HarnessTier};
 pub use event::{
     ApprovalDecisionKind, BoundedError, CheckpointHint, CodeEvent, CodeUsage, Diffstat,
-    FileChangeKind, HarnessNoticeLevel, SequencedCodeEvent, ToolDetail, ToolOutcome,
-    MAX_EVENT_TEXT_CHARS, MAX_NOTICE_CHARS, MAX_PREVIEW_CHARS, MAX_TOOL_SUMMARY_CHARS,
+    FileChangeKind, HarnessNoticeLevel, MAX_EVENT_TEXT_CHARS, MAX_NOTICE_CHARS, MAX_PREVIEW_CHARS,
+    MAX_TOOL_SUMMARY_CHARS, SequencedCodeEvent, ToolDetail, ToolOutcome,
 };
 
+use crate::PermissionMode;
 use crate::attention::{Attention, FenceReason};
 use crate::image::ImageRef;
-use crate::PermissionMode;
 use serde::{Deserialize, Serialize};
 use ts_rs::TS;
 use uuid::Uuid;
@@ -1200,6 +1200,13 @@ pub struct CodeTurn {
     /// the column alone, because its callers hold a snapshot taken before this
     /// existed.
     pub narrative: Option<String>,
+    /// Lucid rewrite of the closing message. The journal keeps the original.
+    ///
+    /// Derived after the turn ends and written only by
+    /// [`crate::db::code::set_turn_rewrite`]. `save_turn` deliberately leaves
+    /// the column alone, the same single-writer rule as `narrative`.
+    #[serde(default)]
+    pub rewrite: Option<String>,
     /// Start time.
     pub started_at: chrono::DateTime<chrono::Utc>,
     /// End time, when terminal.
--- a/crates/tidebreak-core/src/code/mod.rs
+++ b/crates/tidebreak-core/src/code/mod.rs
@@ -13,13 +13,13 @@ mod event;
 pub use caps::{CapLevel, HarnessCaps, HarnessCommand, HarnessTier};
 pub use event::{
     ApprovalDecisionKind, BoundedError, CheckpointHint, CodeEvent, CodeUsage, Diffstat,
-    FileChangeKind, HarnessNoticeLevel, MAX_EVENT_TEXT_CHARS, MAX_NOTICE_CHARS, MAX_PREVIEW_CHARS,
-    MAX_TOOL_SUMMARY_CHARS, SequencedCodeEvent, ToolDetail, ToolOutcome,
+    FileChangeKind, HarnessNoticeLevel, SequencedCodeEvent, ToolDetail, ToolOutcome,
+    MAX_EVENT_TEXT_CHARS, MAX_NOTICE_CHARS, MAX_PREVIEW_CHARS, MAX_TOOL_SUMMARY_CHARS,
 };
 
-use crate::PermissionMode;
 use crate::attention::{Attention, FenceReason};
 use crate::image::ImageRef;
+use crate::PermissionMode;
 use serde::{Deserialize, Serialize};
 use ts_rs::TS;
 use uuid::Uuid;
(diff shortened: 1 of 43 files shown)


naingthet:
**NO GO** at 703755a79
Depth: medium

P2: A lagged updates socket can hide a stored rewrite. `TurnRewrite` is live-only (not restated on connect), and `RecvError::Lagged` in `crates/tidebreak-server/src/routes/code/updates.rs` only resends session digests. The client keeps `turnRewrites` across that snapshot (`CodeUpdatesStore` `snapshot` does not clear it) and `CodeWorkspacePage` reapplies every notice through `applyTurnRewrite`, which sets `rewrite: rewrite.rewrite` even when the notice omitted it. If `rewriting` lands and `rewritten` is dropped, the transcript stays on “Rewriting…” and a later turn snapshot that already has `rewrite` is wiped back to the original. Do not let a `rewriting`/`failed` overlay clear a stored rewrite, and drop or replace the in-memory notice when the turn snapshot already carries one.

naingthet:
Addressed the NO GO at 703755a79 in f9a4810c8.

A lagged updates snapshot now drops live TurnRewrite notices, so they cannot overlay Rewriting onto a turn that already stored the rewrite. A rewriting or failed notice without text keeps the stored rewrite. The original/rewrite toggle stays.

naingthet:
**NO GO** at 8f509b9d3
Depth: medium

P2: A stored rewrite never reaches a reopened transcript. `CodeSessionController` hydrates turns before journal replay. `applyCodeTurnSnapshot` stamps `turn.rewrite` onto the last parent assistant, which does not exist yet, so `applyTurnRewrite` no-ops. Replay then builds the closing message without `rewrite`. Historical `turn_completed` rows skip `turn_snapshot_needed` because hydrate already marked a durable boundary, and live `TurnRewrite` is not restated on connect. Stamp the rewrite after replay, or overlay it from the turn snapshot the way live notices overlay today.

naingthet:
**NO GO** at 7193e08d9
Depth: medium

P2: A stored rewrite never reaches a reopened transcript. `CodeSessionController` still hydrates turns before journal replay. `applyCodeTurnSnapshot` now calls `applyTurnRewrite`, but that stamps `turn.rewrite` onto the last parent assistant, which does not exist yet, so the call no-ops. Replay then builds the closing message without `rewrite`. Historical `turn_completed` rows skip `turn_snapshot_needed` because hydrate already marked a durable boundary, and live `TurnRewrite` is not restated on connect. Stamp the rewrite after replay, or overlay it from the turn snapshot the way live notices overlay today.

naingthet:
**NO GO** at 733a77adc
Depth: medium

P2: A live rewrite never reaches the session store, so dropping TurnRewrite on snapshot hides it. `CodeWorkspacePage` only overlays `turnRewrites` onto transcript items and does not write them back. `set_turn_rewrite` does not refresh the turn snapshot, and live `turn_completed` skips `turn_snapshot_needed` once a durable boundary exists. `reduceCodeUpdates` now clears `turnRewrites` on snapshot (`resetLive` does the same on reconnect). A recap that just appeared disappears, and a retained reopen does not re-hydrate turns, so `storedRewrites` never sees it. Stamp a rewritten notice onto the session items and `storedRewrites` the same way hydrate does after replay.

naingthet:
**NO GO** at 4a876fe4e
Depth: medium

P2: A live rewrite never reaches the session store, so dropping TurnRewrite on snapshot still hides it. `CodeWorkspacePage` only overlays `turnRewrites` onto transcript items and does not write them back. `set_turn_rewrite` does not refresh the turn snapshot, and live `turn_completed` skips `turn_snapshot_needed` once a durable boundary exists. `reduceCodeUpdates` still clears `turnRewrites` on snapshot (`resetLive` does the same on reconnect). A recap that just appeared disappears, and a retained reopen does not re-hydrate turns, so `storedRewrites` never sees it. Stamp a rewritten notice onto the session items and `storedRewrites` the same way hydrate does after replay.

naingthet:
**NO GO** at f4fd29001
Depth: medium

P2: Stamping a live `rewriting` notice onto the session store leaves “Writing a recap…” after lag. `applyLiveTurnRewrite` writes `rewriteState: "rewriting"` onto transcript items. Snapshot still only clears `turnRewrites`, live `TurnRewrite` is not restated, `set_turn_rewrite` does not refresh the turn snapshot, and a retained session does not re-hydrate. If `rewritten` is dropped, the spinner stays even after the rewrite is stored. Only stamp `rewritten` text onto the session store; keep `rewriting`/`failed` as overlay-only, or restamp from the turn snapshot after lag.

naingthet:
Addressed the NO GO at f4fd29001 in 5e872cf24.

Only a rewritten recap with text is stamped onto the session store. rewriting and failed stay overlay-only, so a lagged snapshot cannot leave Writing a recap on a retained transcript.

naingthet:
**G2G** at 5e872cf24
Depth: medium

naingthet:
**G2G** at c3f027047
Depth: medium

naingthet:
**G2G** at 475b2828d
Depth: medium

Issue #2889, opened 2026-08-28
Remote sessions: expose the create paths and lift the typed refusals


naingthet:
Follow-up to #2882 (epic #2875). The runtime carries remote sessions end to end, but three edges were deliberately deferred and refuse with typed conflicts:

- No HTTP route or desktop surface calls `create_remote_workspace` / `create_remote_session` yet; the Slack adapter (stage 2 of `docs/slack-sessions.md`) is the intended first caller. If the desktop should offer remote workspaces before the adapter lands, that needs its own wire types and UI.
- Trigger turns refuse on remote sessions (`remote_triggers_unsupported`).
- Attachments refuse on remote sessions (`remote_attachments_unsupported`).

Also operator tunability: `default_settings` hard-codes the cap (3), the per-spawn ceiling ($5), and the per-session ceiling ($20); nothing exposes them to configuration, and the spend-exhausted refusal tells the owner the session takes no more turns because raising the ceiling is not offered anywhere.

Done when each edge either has a caller and works, or its refusal is a product decision recorded in the design page.

naingthet:
Starting the first bounded slice: expose authenticated HTTP create paths for remote repository workspaces and remote sessions, backed by the existing `CodeRuntime::create_remote_workspace` and `create_remote_session` methods. This slice will not add desktop UI, trigger support, attachment support, or settings controls.

naingthet:
The first three hosted follow-up slices are ready:

- #2898 exposes authenticated remote workspace and session create routes.
- #2900 makes the remote sandbox cap and spend ceilings configurable.
- #2901 records the trigger and attachment refusals as product decisions and gives owners actionable fallback messages.

All required checks are green, and Shipright is G2G on each head. This issue stays open for the desktop approval and Channels work.

naingthet:
Prerequisite for epic #3178 (Tidebreak in Slack). Attachments feed #3200, trigger turns feed #3204, and the tunable caps and ceilings are taken up by #3194; that issue closes the configuration half of this one.

PR #2898, opened 2026-08-28
feat(code): expose remote session create routes

naingthet:
This adds the first bounded slice of #2889.

It exposes authenticated HTTP routes to create remote workspaces and remote sessions through the existing owner-scoped runtime paths. Remote sessions always start in `Allow` mode with fast mode off, and managed permission ceilings can refuse that posture.

The route tests cover authentication, remote-disabled deployments, local-workspace refusal, remote row creation, fixed session posture, unsupported request fields, and managed-policy refusal.

Checks:

- `cargo test -p tidebreak-server remote_create_routes --lib`
- `cargo test -p tidebreak-server a_managed_ceiling_can_refuse_remote_allow_sessions --lib`
- `cargo test -p tidebreak-server code::remote::service --lib`

### DIFF
--- a/crates/tidebreak-server/src/code/remote/service.rs
+++ b/crates/tidebreak-server/src/code/remote/service.rs
@@ -273,8 +273,8 @@ mod tests {
 
     use tidebreak_core::db::code::{insert_repo, latest_turn};
     use tidebreak_core::{
-        CodeRepo, CodeSessionLifecycle, CodeTurnStatus, FenceReason, HarnessKind, OwnerId,
-        PermissionMode, RepoId,
+        CodeRepo, CodeSessionLifecycle, CodeTurnStatus, CodeWorkspaceStatus, FenceReason,
+        HarnessKind, OwnerId, PermissionMode, RepoId,
     };
 
     use super::super::super::runtime::{NewSessionSettings, SubmitTurnOutcome};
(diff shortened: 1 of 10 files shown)


naingthet:
**NO GO** at cea0c0063
Depth: deep

1. P2: `create_remote_session` still skips the workspace lifecycle lock that `create_session_of_kind` holds around the Active check and `insert_session`. Archive CAS's Active to Archiving under that lock, then drops it before `end_workspace_sessions` and `complete_workspace_archive`. `POST /code/remote/workspaces/{id}/sessions` can therefore insert an Idle session after the workspace has already left Active. Turns then 409 `workspace_not_ready`, but the row is stranded on an archived remote workspace. Take the same lock around get / status / insert now that this path is an HTTP route.

naingthet:
**NO GO** at 69d0c4300
Depth: deep

1. P2: `create_remote_session` still skips the workspace lifecycle lock that `create_session_of_kind` holds around the Active check and `insert_session`. Archive CAS's Active to Archiving under that lock, then drops it before `end_workspace_sessions` and `complete_workspace_archive`. `POST /code/remote/workspaces/{id}/sessions` can therefore insert an Idle session after the workspace has already left Active. Turns then 409 `workspace_not_ready`, but the row is stranded on an archived remote workspace. Take the same lock around get / status / insert now that this path is an HTTP route.

naingthet:
**G2G** at 38028fa75
Depth: deep

Issue #2894, opened 2026-08-28
External sessions: grant-authenticated HTTP routes


naingthet:
Slice of the stage 2 epic. The route surface the adapter speaks, authenticated by adapter tokens: external get-or-create, messages, session events over WebSocket, interrupt, and reap — nothing else (no settings, no repository administration).

Done when: each route refuses a token whose grant does not tag the session; the event stream carries the attention snapshot and per-turn assistant record a renderer needs; interrupt and reap behave as their desktop equivalents.

PR #2899, opened 2026-08-28
feat(code): make external message intake idempotent and ordered

### LINKED ISSUE #2892: External sessions: messages endpoint semantics with event idempotency
Slice of the stage 2 epic. The machine-side submit path a channel adapter calls: text plus the channel's `event_id` idempotency key and message `ts`.

Deliver:
- Outcomes `new_turn` (idle, reincarnating first if needed) and `queued` (turn running) on the shipped queue-default contract; `ended`/`fenced`/unknown as typed conflicts the adapter treats as close-the-route.
- `event_id` commits in the same transaction as the queue row or turn row it caused; a replay derives its response from that row's current state — no separate outcome snapshot.
- Messages inside a short window apply in `ts` order so A-then-B cannot invert.

Done when: a replayed `event_id` returns the row-derived outcome without a second turn; out-of-order `ts` within the window applies in order; the busy path queues durably.

naingthet:
Stage 2 of docs/slack-sessions.md, slice 2 (#2892). Builds on the binding table from #2896.

## What

- New `code_external_event` table: one row per channel event, unique on `(owner, session_id, event_id)`. The event row and the queue row it causes commit in one transaction, and `turn_id` names that row — the queue row's id becomes the promoted turn's id (decision 69), so one id follows the message through its whole life.
- `record_external_message` (core op): first delivery parks the message as a durable queue row and records the event; a replay answers `Replay { turn_id }` and writes nothing. A concurrent duplicate delivery loses on the unique key, rolls back whole, and answers with the winner's id.
- Ordering: each event carries the channel's ordering token (`channel_ts`, compared lexicographically — Slack's fixed-width epoch format sorts correctly this way). The insert reorders the session's still-queued external rows by that token, permuting them within their own position slots so desktop follow-ups keep their place. A row already promoted into a turn is outside the window and never moves, so "A then B" cannot become "B steered by A".
- `CodeRuntime::external_submit_message`: scopes by grant (`session_bound_to_grant`), refuses ended and fenced sessions, records the event, promotes the head immediately when the session is idle, and derives the outcome — `NewTurn`, `Queued`, or `Dropped` (the row was retracted before it ran) — from the row's current state. The remote sweep and the immediate promotion now share one per-session helper.

No route speaks this yet; the grant-authenticated routes are #2894.

## Done-when (from #2892)

- A replayed `event_id` returns the row-derived outcome without a second turn: pinned by `a_replayed_external_message_records_once` (core) and `an_external_message_is_idempotent_and_queues_busy` (runtime — replay of a promoted event answers the same turn with no second spawn).
- Out-of-order `ts` within the window applies in order: `out_of_order_external_messages_apply_in_channel_order`.
- The busy path queues durably, and a replay of a queued event answers the same row: the runtime test covers both.

## Checks run

`cargo fmt --all`, `cargo clippy --workspace --all-targets --locked -- -D warnings`, `cargo test -p tidebreak-core --lib` (796 passed), `cargo test -p tidebreak-server --lib remote` (68 passed).

Closes #2892.

### DIFF
--- a/crates/tidebreak-core/src/code/mod.rs
+++ b/crates/tidebreak-core/src/code/mod.rs
@@ -1368,6 +1368,25 @@ pub enum ExternalSessionResolution {
     GrantMismatch,
 }
 
+/// What recording one external message delivery committed
+/// (`docs/slack-sessions.md`, stage 2).
+///
+/// The event id and the queue row it caused commit in one transaction, so a
+/// replayed delivery never writes a second row: it answers `Replay` with the
+/// id the first delivery minted, and the caller derives the outcome from
+/// that row's current state.
+#[derive(Debug, Clone, PartialEq)]
+pub enum ExternalMessageRecord {
+    /// First delivery: the queue row and event row committed together.
+    Recorded(Box<CodeQueuedTurn>),
+    /// The event was already recorded. `turn_id` names the row the first
+    /// delivery caused — still queued, promoted into a turn, or retracted.
+    Replay {
+        /// The id shared by the queue row and the turn it promotes into.
+        turn_id: CodeTurnId,
+    },
+}
+
 /// The outcome of asking for a new incarnation under the owner's cap.
 #[derive(Debug, Clone, PartialEq)]
 pub enum IncarnationAdmission {
--- a/crates/tidebreak-core/src/db/entities.rs
+++ b/crates/tidebreak-core/src/db/entities.rs
@@ -1960,6 +1960,28 @@ pub mod code_external_binding {
     impl ActiveModelBehavior for ActiveModel {}
 }
 
+pub mod code_external_event {
+    use sea_orm::entity::prelude::*;
+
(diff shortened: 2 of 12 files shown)


naingthet:
**NO GO** at ff9719e70
Depth: medium

P2: Out-of-order insert reorders queue positions, then `external_submit_message` immediately promotes the head. Promotion still claims with `promote_queued_turn` matching `position` (and `updated_at`), and a failed claim in `start_turn_row` inserts a **new** turn id while leaving the original queue row in place. Slack delivers unordered, so the case this change exists to handle — B already snapshotted/sent, A arrives with an earlier `channel_ts` and permutes B — makes that claim fail. B is then both a running turn (fresh id, text already in the sandbox) and still queued under the event’s `turn_id`, so the same `event_id` runs twice. Reorder or claim without treating a position-only move as a stale delivery; do not keep the row if that text already spawned a turn.

naingthet:
**G2G** at b39ca8a29
Depth: medium

PR #2900, opened 2026-08-28
feat(code): make remote sandbox limits configurable

naingthet:
## What changed

- Add boot settings for each owner's remote sandbox cap, the per-spawn spend ceiling, and the cumulative per-session spend ceiling.
- Accept positive integer values in micro-USD and `none` for either spend ceiling while preserving the existing 3 / $5 / $20 defaults.
- Pass those settings into remote session provisioning and name the relevant setting in cap and spend refusals.
- Document the hosted deployment variables and restart requirement.

This is the operator-tunability slice of #2889. It does not close the remaining trigger and attachment work.

## Validation

- `cargo test -p tidebreak-core config::tests` (16 passed)
- `cargo test -p tidebreak-server code::remote::` (49 passed)
- Focused cap, spend, and configuration wiring tests (3 passed)
- `cargo fmt --all --check`
- `git diff --check`


naingthet:
**G2G** at 435a114bf
Depth: deep

PR #2901, opened 2026-08-29
fix(code): explain remote trigger and attachment refusals

naingthet:
## What changed

- Record why remote code-trigger turns remain refused: trigger delivery is at-most-once, while sandbox spawn and inbox calls have no idempotency key or replay result.
- Record why remote image attachments remain refused: the sandbox message contract carries text and has no bounded owner-scoped file transfer.
- Make both typed conflict messages explain the boundary and the immediate fallback.
- Add a regression test proving that neither refusal spawns a sandbox, sends a message, or creates a turn.
- Update the Slack sessions design page so these refusals are explicit product decisions, as #2889 allows.

This completes the refusal-decision slice of #2889. The refusals can lift when the runtime contract gains the missing delivery primitives.

## Validation

- `cargo test -p tidebreak-server remote_inputs_without_transport_contracts_are_refused_before_delivery`
- `cargo test -p tidebreak-server code::remote::` (49 passed)
- `cargo fmt --all --check`
- `git diff --check`

### DIFF
--- a/crates/tidebreak-server/src/code/remote/service.rs
+++ b/crates/tidebreak-server/src/code/remote/service.rs
@@ -1017,6 +1017,82 @@ mod tests {
         assert!(fake.spawns.lock().unwrap().is_empty());
     }
 
+    /// Remote trigger and attachment delivery stay refused until the runtime
+    /// can preserve their delivery contracts. Neither refusal may reach the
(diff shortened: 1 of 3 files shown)


naingthet:
**G2G** at c231694a2
Depth: medium

Issue #2890, opened 2026-08-28
Epic: external sessions and grants (Slack sessions stage 2)


naingthet:
Stage 2 of `docs/slack-sessions.md` (stage 1 was #2875): everything the machine must offer so a channel adapter can drive sessions without owning any session state.

Scope, from the design page: the channel-agnostic binding table (`owner, channel kind, external key → session`), external get-or-create, the messages endpoint (`new_turn`/`queued`, `ts` ordering, transactional `event_id` idempotency), grant-tagged sessions, adapter tokens with mint, rotate, reuse detection, list, and revoke, the connect approval surface on hosted web, and the desktop grants list.

Stage-level verify: two racing creates yield one session; a replayed `event_id` returns the row-derived outcome; out-of-order `ts` within the window applies in order; a revoked grant's WebSocket drops immediately and the next call fails; a grant cannot touch a session tagged to another grant; the approval page shows the Slack identity and a forwarded link binds nothing without the closing DM confirm.

Slices:

- [x] #2891 Channel-agnostic binding table and get-or-create
- [x] #2892 Messages endpoint semantics with event idempotency
- [x] #2893 Adapter grants with mint, rotate, reuse detection, and revoke
- [x] #2894 Grant-authenticated HTTP routes
- [x] #2895 Connect approval surface and desktop grants list

PR #2904, opened 2026-08-29
feat(code): mint, rotate, and revoke adapter grants

### LINKED ISSUE #2893: External sessions: adapter grants with mint, rotate, reuse detection, and revoke
Slice of the stage 2 epic. The credential a channel adapter holds per linked user.

Deliver:
- A grant row (owner, channel kind, external identity, workspace identity, created/revoked) and adapter tokens the machine mints with an adapter audience; the machine stores only a hash.
- Scopes limited to external get-or-create, submit messages, watch events, interrupt, and reap, each restricted to sessions tagged with the grant.
- Rotation with refresh tokens and reuse detection: a replayed rotated token revokes the grant and notifies the owner. Revocation severs live event WebSockets immediately.

Done when: a grant cannot touch a session tagged to another grant; a replayed rotated token kills the grant; a revoked grant's WebSocket drops and its next call fails.

naingthet:
Adapter credentials for the external surface, machine side (stage 2 of
`docs/slack-sessions.md`, slice 3 of #2890).

- A grant is minted per linked identity (`workspace_identity` +
  `channel_workspace`), one live grant per identity; minting over a live
  grant refuses, so replacement is an explicit revoke-and-replace.
- The machine stores only SHA-256 hashes of the access and refresh
  secrets. Secrets are two v4 UUIDs behind `tbg_`/`tbr_` prefixes and
  exist in full only in the mint or rotate response.
- Rotation retains the previous refresh hash. A replay of a rotated
  refresh token is treated as theft: the grant is revoked in the same
  transaction and `GrantRevocations` broadcasts so live event streams
  sever immediately.
- Revocation is idempotent and keeps the first recorded reason.
- Cross-owner lookups are named `_all_owners` and justified for
  credential-to-owner resolution; the owner-scope conformance test
  covers them.

Routes and the connect handshake build on this in the next two slices.

Closes #2893


naingthet:
**NO GO** at 790681e42
Depth: deep

1. P2: `mint_external_grant` enforces one live grant per identity with a read-then-insert and no uniqueness constraint on `(owner, channel_kind, external_identity, workspace_identity)` for live rows. Two concurrent mints for the same identity can both see no live row and both commit, each returning a distinct token pair. Either token authenticates, and revoking one leaves the other live. The binding slice already uses a unique key as the race gate; this needs the same treatment (a live-only unique index, and map the conflict to the existing refuse). Sequential mint-while-live is tested; the race is not.

2. P2: `rotate_external_grant_all_owners` loads by current `refresh_hash` and then `update`s by id, so two concurrent presents of the same refresh token can both read it as current and both return `Rotated` with different pairs. The second write wins; the first pair is unusable; `prev_refresh_hash` is the same old hash in both cases, so neither call is reuse. A unique index on `refresh_hash` does not serialize this, because each call writes a new hash. Rotate with `UPDATE … WHERE refresh_hash = presented` (and fall through to reuse detection when no row matches) so a refresh token cannot be redeemed twice.

naingthet:
**G2G** at 21918ecea
Depth: deep

naingthet:
**G2G** at 9df62051d
Depth: deep

PR #2907, opened 2026-08-29
feat(server): store self-host credentials in Vault KV

### LINKED ISSUE #2906: feat(server): store self-host credentials in Vault KV
Parent: #2322

Give the self-host profile remote custody for deployment-scoped credentials without routing them through the desktop OS keychain.

Acceptance criteria:

- Select a HashiCorp Vault KV v2 backend for stored self-host credentials through boot configuration.
- Read Vault authentication from a mounted token file on every request so token rotation does not require a restart.
- Require HTTPS except for literal loopback development, refuse redirects, bound response bodies, and keep tokens and secret values out of errors.
- Keep the existing provider environment variables as fallbacks when Vault is not configured.
- Skip desktop keychain migration and re-home work in the self-host profile.
- Cover remote read, overwrite, delete, missing-secret, authentication failure, malformed response, and rotating-token behavior with a local HTTP fixture.
- Document the Vault policy and operator configuration, then remove remote secret custody from the remaining self-host integration gaps.

Object storage and multi-process ownership remain separate slices.

naingthet:
Closes #2906

Stores self-host credentials in HashiCorp Vault KV v2 without routing them through the desktop OS keychain.

- Selects Keychain for Desktop, Vault for configured SelfHost, and an unavailable stored-secret provider when Vault is absent.
- Reads the mounted Vault token file for every request, refuses redirects, requires HTTPS outside literal loopback development, bounds bodies, and redacts sensitive failures.
- Keeps provider environment variables as fallbacks when Vault is absent while returning setup guidance for credential writes and deletes.
- Skips boot-time and manual keychain re-home work for SelfHost.
- Documents the Vault policy, mount path, namespace, token rotation, and Compose integration.

Validation:
- `cargo test -p tidebreak-core config::tests --lib`
- `cargo test -p tidebreak-server vault_secrets::tests --lib`
- `cargo test -p tidebreak-server profile_secret_tests --lib`
- `cargo test -p tidebreak-server error::tests --lib`
- `cargo check -p tidebreak-server --features postgres`
- `cargo check -p tidebreak-cli`
- `cargo clippy -p tidebreak-server --all-targets --features postgres -- -D warnings -A dead-code`
- `cargo clippy -p tidebreak-cli --all-targets -- -D warnings`
- `pnpm --dir docs-site build`
- `pnpm --dir docs-site lint`

### DIFF
--- a/crates/tidebreak-cli/src/main.rs
+++ b/crates/tidebreak-cli/src/main.rs
@@ -15,9 +15,10 @@
 //! `tidebreak mcp <workspace>` serves the built-in read-only filesystem tools over
 //! MCP stdio, confined to the explicit workspace directory.
 //!
-//! `tidebreak rehome-secrets` rewrites the profile's stored credentials so their
-//! keychain items belong to the running binary's code signature, which is what
-//! stops macOS asking for credentials an earlier build created.
+//! `tidebreak rehome-secrets` rewrites the desktop profile's stored credentials
+//! so their keychain items belong to the running binary's code signature, which
+//! stops macOS asking for credentials an earlier build created. Self-host
+//! profiles reject this command because they never use the OS keychain.
 //!
 //! `tidebreak output list|show|revisions|export <chat> …` reads a conversation's
 //! outputs and writes one to a path, and `tidebreak attach <chat> <file>` puts a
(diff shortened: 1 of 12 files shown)


naingthet:
**G2G** at 7bb93e6c4
Depth: deep

Self-host now keeps the credential bundle in Vault KV v2 (HTTPS outside literal loopback, no redirects, token file reread per request, redacted failures) and refuses the desktop keychain, while unset Vault still lets provider env fallbacks serve reads and turns Settings writes into operator setup guidance. Desktop re-home stays on Desktop, and this does not recast Helm self-host as the product path.

naingthet:
**G2G** at c1220d3de
Depth: deep

Vault still fails at bind-time planning before the instance lock or store, then the provider is built from that already-checked plan. Operator-facing custody, desktop refusal, and unset-Vault fallback reads are unchanged, and this still does not recast Helm self-host as the product path.

PR #2909, opened 2026-08-29
feat(server): open the grant-authenticated adapter routes

### LINKED ISSUE #2894: External sessions: grant-authenticated HTTP routes

naingthet:
The route surface a channel adapter speaks (stage 2 of
`docs/slack-sessions.md`, slice 4 of #2890).

- A separate unauthenticated router, outside `require_token` on the
  inference relay's pattern: external get-or-create, messages, session
  events over WebSocket, interrupt, reap, and grant rotation — nothing
  else.
- Every route authenticates per request by adapter token (fail-closed
  401), and every session route then checks that the token's grant tags
  the session through a binding. A session another grant holds answers
  the not-found shape, so the surface says nothing about other grants'
  work.
- The events WebSocket sends the session snapshot first, then the
  desktop event stream, and is severed the moment the grant is revoked
  via the `GrantRevocations` broadcast.

The connect handshake and the desktop grants panel follow in the next
slice.

Closes #2894

### DIFF
--- a/crates/tidebreak-server/src/code/grants.rs
+++ b/crates/tidebreak-server/src/code/grants.rs
@@ -64,8 +64,6 @@ impl Default for GrantRevocations {
 }
 
 impl GrantRevocations {
-    /// The events WebSocket route (#2894) is the production subscriber.
-    #[cfg_attr(not(test), allow(dead_code))]
     pub(crate) fn subscribe(&self) -> tokio::sync::broadcast::Receiver<CodeGrantId> {
         self.sender.subscribe()
     }
@@ -105,7 +103,6 @@ impl super::runtime::CodeRuntime {
 
     /// The live grant a presented access token authenticates. A revoked
     /// grant's next call dies here.
-    #[cfg_attr(not(test), allow(dead_code))]
     pub(crate) async fn authenticate_adapter_token(
         &self,
         token: &str,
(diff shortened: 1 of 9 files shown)


naingthet:
**NO GO** at 05103e82a
Depth: deep

1. P2: `external_events` reuses `session_events::stream_events`, which calls `mark_session_viewed` on connect. That call is the desktop meaning of an events socket: the owner is looking at the session, so `DoneUnreviewed` becomes `Idle`. The adapter is a renderer, not a viewer. A connect or reconnect — the resync path `docs/slack-sessions.md` already requires — clears the state Slack uses for the success/artifact render and the state the desktop inbox uses for the same session. Capture the snapshot and attach the journal without marking viewed, or the adapter watching a turn will look like the owner already reviewed it.

naingthet:
**G2G** at b7946c113
Depth: deep

The adapter events socket attaches as `Viewer::Adapter` and does not call `mark_session_viewed`, so a connect or resync leaves `DoneUnreviewed` in place for the desktop inbox and the Slack success render. Grant-scoped routes, the snapshot-then-journal stream, and revocation severing match the stage-2 attach contract.

naingthet:
**G2G** at eba2bfb9d
Depth: deep

The new commit only documents the adapter family as a grant-scoped bearer surface and pins that path in the existing route-scoping gate. Tests still prove get-or-create writes the binding they then speak, so the stage-2 attach contract does not move.

PR #2916, opened 2026-08-29
feat(server): enforce one PostgreSQL store owner

### LINKED ISSUE #2914: feat(server): enforce one self-host owner per PostgreSQL store
Parent: #2322

A self-host process locks only its local data directory today. Two processes with different data directories can open the same PostgreSQL database, start duplicate workers, and race process-local coordination.

Acceptance criteria:

- Acquire one database-scoped PostgreSQL advisory lease before migrations or workers start.
- Hold the lease on a dedicated connection for the server lifetime.
- Refuse a second self-host process that targets the same database, even when it uses another data directory.
- Stop serving if the lease connection fails so the process never continues without ownership.
- Leave desktop SQLite behavior unchanged and allow separate PostgreSQL databases to run independently.
- Cover acquisition, contention, release, and the boot refusal with focused PostgreSQL tests.
- Update self-host documentation to state that Tidebreak enforces one active process per database.

This slice establishes exclusive deployment ownership. Horizontal multi-process serving remains out of scope until every process-local worker and live-delivery path has a distributed owner.

naingthet:
Closes #2914

## What changed

- Acquire a database-scoped PostgreSQL advisory lease before migrations and workers start.
- Hold the lease on a dedicated connection for the server lifetime.
- Refuse a second self-host process against the same database, even with another data directory.
- Stop the server when the ownership connection fails.
- Add a PostgreSQL integration test for contention, lease loss, replacement startup, and credential-safe URL errors.
- Run the integration test in the PostgreSQL CI lane and document the single-process boundary.

## Validation

- `cargo test -p tidebreak-server --features postgres --test postgres_store_ownership -- --nocapture` — 1 passed.
- `cargo check -p tidebreak-server --all-targets` — passed.
- `cargo check -p tidebreak-server --features postgres --all-targets` — passed.
- `cargo clippy -p tidebreak-server --all-targets --all-features -- -D warnings` — passed.
- `cargo clippy -p tidebreak-server --features postgres --test postgres_store_ownership -- -D warnings` — passed.
- `cargo fmt --check` — passed.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -849,6 +849,10 @@ jobs:
         run: >-
           cargo test -p tidebreak-core --no-default-features --features postgres
           --test postgres_release_upgrade --locked
+      - name: Exercise self-host store ownership on PostgreSQL
+        run: >-
+          cargo test -p tidebreak-server --features postgres
+          --test postgres_store_ownership --locked
 
   self-host-build:
     name: self-host server build
--- a/crates/tidebreak-server/Cargo.toml
+++ b/crates/tidebreak-server/Cargo.toml
@@ -15,7 +15,7 @@ workspace = true
 # Compile the PostgreSQL driver in, so a self-host boot can open the store
 # `TIDEBREAK_DATABASE_URL` names. Off by default: the desktop profile is
 # SQLite-only and the driver is a heavy build-time dependency.
-postgres = ["tidebreak-core/postgres"]
+postgres = ["tidebreak-core/postgres", "sea-orm/sqlx-postgres"]
 # Compile the scripted model provider in, so a test in another crate can drive
 # a real turn through the binary. Off by default and never enabled by a release
 # build: see `src/scripted_provider.rs`.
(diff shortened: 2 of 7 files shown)


naingthet:
**NO GO** at fc7647caa
Depth: deep

1. P1: Losing the ownership backend releases the advisory lock immediately, but `serve` only notices on a 5s `SELECT 1` poll. A second process can `bind`, pass `try_acquire`, and start workers while the first process is still running its worker set and serving `/healthz`. That is the duplicate-owner failure this change exists to prevent. The integration test waits for the old `serve` to fail before starting a replacement, so it does not catch the overlap. Watch the ownership connection in the same `select!` as `axum::serve` so the process stops when the lock is released, not up to an interval later.
2. P2: `serve` moves `_store_ownership` out of `Server`. On lease loss the guard drops (and the session lock is released) before `AbortTask` fields drop. Even with faster detection, workers can still run after the lock is free. Keep the guard on `Server` and abort workers before the ownership connection closes.

naingthet:
**G2G** at 86393f6ce
Depth: deep

The lease is taken before migrations, held on a dedicated connection, and watched in the same select as the accept loop, so a second process is refused at boot and a lost backend stops serving without waiting out a poll. Workers abort and drain before the guard drops, and the boot refusal names the other owner without leaking the database URL.

PR #2917, opened 2026-08-29
fix(cli): require TLS for remote server URLs

### LINKED ISSUE #2915: Reject cleartext remote --server URLs
## Problem

`crates/tidebreak-cli/src/connect.rs::base_url` accepts `http://` for any host. `Server::resolve` then builds a bearer-authenticated client whose default `Authorization` header is sent with each request. A value such as `--server http://example.test:8080` can therefore send the server token over cleartext HTTP.

Decision 47 requires TLS unless the host is loopback, and its validation section requires a non-TLS, non-loopback URL to fail before connection. CodeQL alert 212 surfaced this path while reviewing #2913.

## Done when

- `http://` remains available for loopback development endpoints only.
- A non-loopback `http://` URL returns a stable configuration error before the client reads or sends its bearer token.
- Remote `https://` URLs remain accepted.
- Tests cover IPv4 loopback, IPv6 loopback, localhost, remote HTTP, and remote HTTPS.

naingthet:
Closes #2915

## What changed

- Reject bearer-authenticated `http://` server URLs outside loopback before the CLI reads the token environment variable.
- Apply the same URL validation inside `Client::attach` so direct callers cannot bypass the transport boundary.
- Accept `localhost`, IPv4 loopback, and IPv6 loopback development URLs.
- Disable redirects on the authenticated HTTP client so an HTTPS endpoint cannot downgrade the bearer to cleartext HTTP.
- Reject embedded URL credentials, queries, and fragments, and document the TLS rule.

## Validation

- `cargo test -p tidebreak-cli --bin tidebreak` — 113 passed.
- `cargo test -p tidebreak-cli --test serve a_setup_command_attaches_` — 2 passed.
- `cargo test -p tidebreak-cli authenticated_clients_do_not_follow_cleartext_redirects` — 1 passed.
- `cargo clippy -p tidebreak-cli --all-targets -- -D warnings` — passed.
- `cargo fmt --check` — passed.

### DIFF
--- a/crates/tidebreak-cli/src/api/client.rs
+++ b/crates/tidebreak-cli/src/api/client.rs
@@ -131,13 +131,18 @@ impl Client {
         local_import_token: Option<&str>,
         listen_data_dir: Option<PathBuf>,
     ) -> Result<Self> {
+        let base = validated_server_base_url(&base)?;
         let mut headers = reqwest::header::HeaderMap::new();
         let mut value = reqwest::header::HeaderValue::from_str(&format!("Bearer {token}"))
             .map_err(|error| AgentError::msg(format!("invalid server token: {error}")))?;
         value.set_sensitive(true);
         headers.insert(reqwest::header::AUTHORIZATION, value);
-        let mut http = reqwest::Client::builder().default_headers(headers);
-        if loopback_http_base(&base) {
+        let mut http = reqwest::Client::builder()
+            .default_headers(headers)
+            // A same-host HTTPS-to-HTTP redirect would otherwise resend the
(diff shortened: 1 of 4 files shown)


naingthet:
**G2G** at c6f6c92db
Depth: deep

`--server` and `Client::attach` both refuse non-loopback `http://` before the bearer is read or sent, keep loopback HTTP for local serve/attach, and disable authenticated redirects so HTTPS cannot downgrade the token. That matches decision 47 and the managed-hosting attach path; CLI help and self-hosting copy state the TLS rule without treating Helm HTTP as a product surface.

Issue #2344, opened 2026-08-20
feat(browser): constrain uploads and downloads to Tidebreak capabilities


naingthet:
Parent: #2335
Depends on: semantic macOS driver and browser consent/control.

## Outcome

Support useful browser file flows without granting a page or agent ambient filesystem access.

## Deliverable

- Browser upload commands accept logical connected-folder attachment ids or output ids, never host paths.
- The native executor resolves and reauthorizes the exact resource immediately before attaching it to a file input.
- Downloads land in an app-owned staging directory with bounded metadata and become Tidebreak outputs.
- Export or write-back remains an explicit user/capability action through the existing output path.
- Cross-origin redirects, duplicate filenames, oversized files, cancellation, engine crash, and malicious filenames fail safely.
- File bytes and paths never enter renderer browser state or page-visible Tidebreak IPC.

## Acceptance criteria

- The fixture uploads an approved logical resource and downloads a result that appears as a staged output.
- Guessed ids, stale grants, symlinks, unsupported file types, and paths outside approved roots are refused.
- A download cannot choose an arbitrary destination or overwrite an existing user file.
- Transfer confirmation, audit, cancellation, and restart recovery are test-covered.

naingthet:
Status update — August 21, 2026

This slice has not started and remains intentionally outside the initial browser release path that just landed. Downloads are still blocked rather than staged, file-input values are omitted, and no host path is exposed to the page or agent.

Keep this issue open for logical connected-resource uploads, app-owned staged downloads, explicit export/write-back, transfer confirmation/audit, and restart/cancellation hardening after the trusted action path is available.

naingthet:
Tracking audit — 2026-08-21

This slice remains intentionally unimplemented: browser downloads are blocked, file-input values are omitted, and no host path is exposed. Logical-resource uploads, app-owned staged downloads, explicit export/write-back, transfer confirmation and audit, and cancellation/restart recovery remain here.

The implementation depends on the trusted action path in #2339, so marking this issue blocked until that capability is available.

naingthet:
Bounded downloads that land as outputs (#2918) and bounded foreground uploads from logical resources (#2919) are on main. Closing.

PR #2918, opened 2026-08-29
feat(browser): publish bounded downloads as outputs

naingthet:
This adds the download half of the bounded browser-transfer work.

The desktop stages accepted browser downloads in private app storage, validates their names, size, and content, and publishes completed files into the conversation Outputs catalog. Durable receipts recover completed transfers after restart and discard interrupted transfers. Blocked downloads keep the existing Open externally fallback.

The renderer receives status events only. File paths and bytes stay outside renderer state and page-visible IPC.

Part of #2344

Tests:
- cargo clippy -p tidebreak-core -p tidebreak-server -p tidebreak-desktop --lib -- -D warnings
- cargo fmt --all --check
- focused Rust download, deliverable, and output-storage tests
- pnpm --dir crates/tidebreak-desktop/ui lint
- pnpm --dir crates/tidebreak-desktop/ui test
- pnpm --dir crates/tidebreak-desktop/ui build
- pnpm --dir crates/tidebreak-desktop/ui storybook:build
- Storybook screenshots at 1280x800 and 720x480 in light, dark, high-contrast light, and high-contrast dark


naingthet:
**NO GO** at da1b3d6af
Depth: deep

1. P2: After you accept a download, rejection deletes the staged file and emits `download_failed` with no URL. The failure notice is kind `download`, but Open externally only renders when `notice.url` is set, so the user cannot recover the bytes. `.svg` makes this worse: `begin` treats it as `image/svg+xml`, then `validate_download_content` hard-codes that type as never matching, so every SVG shows "Saving … to Outputs", fails, and drops the file. Pass the request URL on failure the way `download_blocked` already does, and refuse SVG at `begin` unless you actually validate it.

2. P2: CSV, markdown, JSON, and other curated-text downloads are stored as `application/octet-stream` binaries. `download_media_type` already classifies those names via `deliverable_media_type`, then throws the real type away so `accept_workspace_artifact` will take them. Outputs already has viewers for `text/csv`, markdown, and code; they never run, and the catalog shows export-only. Publish those names as text outputs with the media type you already computed.

naingthet:
**G2G** at ce4a44c56
Depth: deep

Accepted downloads stay in private staging, then publish into the conversation Outputs catalog with the canonical text or binary type so CSV, markdown, and code viewers actually run. Failures and unsupported types (including SVG) keep Open externally, and file paths never enter renderer state. That matches the bounded in-app transfer: the page cannot pick a disk destination, and the user still has a way out when Tidebreak will not keep the bytes.

PR #2919, opened 2026-08-29
feat(browser): add bounded foreground uploads

naingthet:
## Summary

- add a foreground-only `browser_upload` tool for current conversation outputs and connected-folder files
- resolve and reauthorize the exact resource after native confirmation and inside the serialized browser dispatch
- bind confirmation to the transfer capability, filename, byte length, and SHA-256 digest
- attach the confirmed file through Tidebreak's isolated WebKit content world without exposing host paths or file bytes through renderer state or renderer-facing IPC
- keep the code-session browser tool surface unchanged

## Safety

Uploads accept only opaque output or root identifiers and bounded root-relative paths. The native executor enforces the broker's 8 MiB read ceiling, portable filenames, symlink-safe reads, live root attachment, stable staged-folder selection, current output revision, and a fresh confirmation for every upload.

Part of #2344. Complements the bounded download work in #2918.

## Tests

- `cargo test -p tidebreak-core browser_upload`
- `cargo test -p tidebreak-server foreground_browser_ --lib`
- `cargo test -p tidebreak-server exact_output_reauthorization_rejects_other_chats_and_stale_heads --lib`
- `cargo test -p tidebreak-desktop browser`
- `node --test crates/tidebreak-desktop/tests/browser-fixture/*.test.mjs`
- `cargo clippy -p tidebreak-desktop --all-targets -- -D warnings`
- `cargo clippy -p tidebreak-core -p tidebreak-server --lib -- -D warnings`

### DIFF
--- a/crates/tidebreak-core/src/browser.rs
+++ b/crates/tidebreak-core/src/browser.rs
@@ -37,9 +37,9 @@ pub const BROWSER_UPLOAD_TOOL: &str = "browser_upload";
 /// The complete set of browser tools this contract supports.
 ///
 /// Wait and screenshot are available on any engine that can inspect a page.
-/// Semantic act requires native input synthesis and is registered only when
-/// the engine adapter reports [`BrowserEngineCapabilities::semantic_actions`]
-/// as true.
+/// Semantic act and upload require trusted native interaction and register
+/// only when the engine adapter reports
+/// [`BrowserEngineCapabilities::semantic_actions`] as true.
 pub const BROWSER_TOOLS: [&str; 7] = [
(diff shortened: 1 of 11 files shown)


naingthet:
**G2G** at ea62894a1
Depth: deep

Foreground `browser_upload` stays on opaque conversation resources, binds native confirmation to transfer capability plus filename/length/digest, and re-resolves those bytes inside serialized dispatch before isolated-world attach. Host paths never enter the tool contract, code-session tools stay unregistered, and the share/confirm copy matches a per-upload gate rather than ambient filesystem access.

PR #2922, opened 2026-08-29
feat(ui): show repository trigger targets

### LINKED ISSUE #2503: feat(ui): show the trigger surface on the notifications page
Finishes #2480. Part of #2477.

`CodeTriggerRules` landed in #2496 as a presentational component with Storybook
coverage, but nothing renders it. This wires it into `CodeNotificationsPage`
against a real repository.

## What is needed

- Pick the repository to scope to. Triggers bind per `RepoId`; the page works
  in `CodeGitHubRepositoryRef`, which carries `tidebreak_repo_id`. Only
  repositories with that id can hold a trigger.
- Load with `listCodeTriggers`, and wire `onArm`, `onSetEnabled` and
  `onChangeAction` to `createCodeTrigger`, `setCodeTriggerEnabled` and the
  action path. All four client methods exist.
- Build the `CodeTriggerTarget` the component renders its delivery line from:
  the workspace's most recently active interactive session, and whether its
  harness declares `mid_turn_steering`. That is what makes the surface say
  "interrupts" or "waits for quiet" instead of guessing.
- Hold the switches while a write is in flight — the component already takes
  `busy`.

## Done when

A user opens notifications, sees which triggers are armed on a repository,
arms and disarms one, and the section names the session a fire would reach and
how it would arrive.

naingthet:
## Summary

- Render repository trigger rules from the tracked-repositories dialog.
- Name the session a trigger would reach and whether delivery interrupts it or waits for a turn.
- Share the server timestamp used to rank trigger targets so the preview matches delivery.
- Preserve that timestamp through live digest updates.

## Validation

- `pnpm exec vitest run src/code/CodeUpdatesStore.test.ts src/code/CodeTriggerTarget.test.ts src/code/CodeDeliveryPage.dom.test.tsx src/code/parsers.test.ts` (141 tests passed)
- `pnpm exec tsc --noEmit`
- `pnpm exec biome check` on the changed UI files
- Full UI suite (2,261 tests passed)
- `pnpm storybook:build`
- `cargo test -p tidebreak-server trigger_target_time_prefers_the_latest_turn_and_falls_back_to_creation`
- Storybook screenshots across four themes at desktop and narrow widths

Closes #2503

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeDeliveryPage.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/CodeDeliveryPage.dom.test.tsx
@@ -28,6 +28,9 @@ import {
   deliveryRepositoriesSnapshot,
   deliveryRunDetails,
   deliveryRuns,
+  codeDigest,
+  codeWorkspace,
+  harnessDoctor,
 } from "../stories/fixtures";
 import {
   CodeDeliveryPage,
@@ -40,11 +43,13 @@ import {
   useCodeDeliveryStore,
 } from "./CodeDeliveryStore";
 import { useCodeUpdatesStore } from "./CodeUpdatesStore";
+import { useCodeCatalogStore } from "./CodeCatalogStore";
 
 afterEach(() => {
   cleanup();
   useCodeDeliveryStore.getState().reset();
   useCodeUpdatesStore.getState().reset();
+  useCodeCatalogStore.getState().reset();
 });
 
 vi.mock("@/openInBrowser", () => ({ openInBrowser: vi.fn() }));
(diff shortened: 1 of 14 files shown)


naingthet:
**NO GO** at e8ebca654
Depth: medium

1. P2: Live digest updates drop `trigger_target_at`, so the new copy goes blank after the first WS digest. `codeTriggerTargetForRepository` skips any session without that timestamp. Snapshot sessions keep it, but `noticeToAction` in `crates/tidebreak-desktop/ui/src/code/CodeUpdatesStore.ts` rebuilds a digest notice field-by-field and never copies `trigger_target_at`. `upsertDigest` then replaces the snapshot row. After any turn, attention, or lifecycle digest, the Triggers dialog shows "No session is open in a workspace with a pull request, so a fire waits until one is." even when a session would receive the fire. The delivery-page test only applies a snapshot, so it does not catch this.

naingthet:
**G2G** at f4b35b907
Depth: medium

PR #2923, opened 2026-08-29
feat(ui): migrate notification rules to triggers

### LINKED ISSUE #2504: feat(ui): run the notification-rule migration and retire the client rules
Finishes #2481. Part of #2477. Depends on the surface issue above.

`triggersForNotificationRules` landed in #2502 with tests, but nothing calls
it. This runs it and removes the second rule engine, which is the point record
0060 was making.

## What is needed

- Run the mapping **once**, on first read of `tidebreak.code-delivery`, and
  record that it ran so it cannot run twice. Arming is idempotent per
  `(repository, condition)`, but re-running would resurrect rules the user
  disabled on the server afterwards.
- Arm each result with `createCodeTrigger`.
- Then drop the client-side evaluation: `notificationRules`,
  `updateNotificationRule`, `DEFAULT_DELIVERY_NOTIFICATION_RULES`, the
  `NotificationRules` and `RuleRow` components, and the rule fields in the
  persisted state.

## Traps

- The `localStorage` write is one-way. Removing the key without migrating loses
  a user's rules silently, so the read path has to be the thing that migrates.
- A migration that fails partway must not mark itself done, or the rest is lost.
- Keep `notifications`, `seenFingerprints` and the saved views. Only the
  **rules** move; the feed stays client-side.

## Done when

Someone carrying existing notification rules keeps them, sees them as triggers,
and the client no longer evaluates rules of its own.

naingthet:
## Summary

- Migrate persisted client notification rules into server trigger rows once after repository discovery.
- Keep legacy rules recoverable until every missing trigger is armed, and skip existing disabled rows during a retry.
- Keep the one-way migration pending while repository discovery reports partial failures.
- Retire client-side rule evaluation while preserving notifications, seen fingerprints, and saved views.
- Remove the legacy-rule message and add Storybook coverage for the repository trigger states.

## Validation

- `pnpm test` (256 files and 2,265 tests passed before the review follow-up)
- `pnpm exec vitest run src/code/CodeDeliveryMonitor.test.ts src/code/CodeDeliveryStoreMigration.test.ts` (11 tests passed after the review follow-up)
- `pnpm exec tsc --noEmit`
- `pnpm exec biome check` on the changed UI files
- `pnpm storybook:build`
- Storybook screenshot capture remains unavailable because no browser connection is available.

Closes #2504

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/CodeDeliveryMonitor.test.ts
+++ b/crates/tidebreak-desktop/ui/src/code/CodeDeliveryMonitor.test.ts
@@ -33,17 +33,37 @@ describe("migrateLegacyNotificationRules", () => {
     const listCodeTriggers = vi.fn(async () => []);
     const createCodeTrigger = vi.fn(async () => ({}) as never);
 
-    await migrateLegacyNotificationRules(
-      { listCodeTriggers, createCodeTrigger },
-      [attentionRule],
-      [repository],
-    );
+    await expect(
+      migrateLegacyNotificationRules(
+        { listCodeTriggers, createCodeTrigger },
+        [attentionRule],
+        { repositories: [repository], errors: [] },
+      ),
+    ).resolves.toBe(true);
 
     expect(listCodeTriggers).toHaveBeenCalledWith("repo-1");
     expect(createCodeTrigger.mock.calls).toEqual([
       ["repo-1", "changes_requested", "notify"],
       ["repo-1", "conflicts", "notify"],
     ]);
+  });
+
+  it("keeps the migration pending when repository discovery is partial", async () => {
+    const listCodeTriggers = vi.fn(async () => []);
+    const createCodeTrigger = vi.fn(async () => ({}) as never);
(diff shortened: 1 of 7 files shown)


naingthet:
**NO GO** at 6a5dad947
Depth: medium

P2: `CodeDeliveryMonitor` marks the one-way rule migration complete after any non-throwing `migrateLegacyNotificationRules` call, including when `discovered.errors` is non-empty or the tracked catalog is a subset of the user's repos. `CodeDeliveryRepositoriesSnapshot` is a partial GitHub list plus failures. A first poll that resolves only some repositories arms those rows, then `completeNotificationRuleMigration` persists `notificationRulesMigrated` and drops the saved rules, so later discovery cannot arm the rest. Hold the complete until the catalog used for mapping is complete, or keep the legacy rules until every bindable repo from that snapshot has been considered without `errors`.

naingthet:
**G2G** at ca58f5ff3
Depth: medium

PR #2929, opened 2026-09-01
feat(settings): let users disable turn recaps

naingthet:
## What changed

- Add a **Write turn recaps** switch to Agent settings.
- Persist the choice in runtime settings and skip future utility-model recap calls when it is off.
- Keep existing stored recaps visible.
- Add enabled and disabled Storybook states plus server and UI coverage.

## Validation

- `cargo test -p tidebreak-server --lib tests::code_recap::`
- `cargo test -p tidebreak-server --lib tests::configuration::settings_default_then_update_roundtrips`
- `cargo test -p tidebreak-server --lib wire_types::tests::the_generated_bindings_are_current`
- `pnpm --dir crates/tidebreak-desktop/ui exec vitest run src/settings/AgentsPanel.dom.test.tsx`
- `pnpm --dir crates/tidebreak-desktop/ui exec tsc --noEmit`
- `cargo fmt --all -- --check`
- `pnpm --dir crates/tidebreak-desktop/ui exec biome check src/api/client.ts src/settings/AgentsPanel.tsx src/settings/AgentsPanel.dom.test.tsx src/stories/SettingsStoryHarness.tsx src/stories/SettingsPanels.stories.tsx`
- Production UI build and Storybook build passed before the rebase; the rebase applied without conflicts.

PR #2930, opened 2026-09-01
test(code): avoid cleartext alert in recap test

naingthet:
## What changed

- Keep the session identifier as its wire string when building the loopback-only test URL.
- Parse a separate typed identifier for direct database assertions.

This resolves the CodeQL cleartext-transmission alert that landed after #2929 merged. The URL targets a local test server and carries no credential or private data.

## Validation

- `cargo test -p tidebreak-server --lib tests::code_recap::a_disabled_recap_setting_skips_the_model_call`
- `git diff --check`


naingthet:
**G2G** at d44a9c1cf
Depth: medium

PR #2931, opened 2026-09-01
fix: address security and dependency alerts

naingthet:
## Summary

- restrict test ChatGPT authentication endpoints to credential-free loopback HTTP URLs
- override vulnerable Vercel CLI transitive dependencies while retaining the trusted CLI pin
- override the vulnerable mobile `uuid` dependency and pin the mobile pnpm version
- regenerate desktop and mobile wire types without trailing whitespace

## Validation

- `cargo fmt --check`
- focused ChatGPT authentication and wire-type tests
- mobile typecheck, lint, and 90 tests
- workflow security tests
- frozen pnpm installs, Vercel CLI smoke test, and dependency audits

## Remaining alerts

Two `image-size` advisories have no patched release. The patched `decode-uri-component` release is ESM-only and cannot safely replace the CommonJS dependency used by `query-string@7.1.3`. Eighteen Pillow advisories require Pillow 12.3.0, which does not support the Python 3.9 runtime that the local sandbox still guarantees.

### DIFF
--- a/.github/vercel-cli/package.json
+++ b/.github/vercel-cli/package.json
@@ -8,8 +8,17 @@
   },
   "pnpm": {
     "overrides": {
+      "@tootallnate/once@2.0.0": "2.0.1",
+      "ajv@8.6.3": "8.20.0",
+      "js-yaml@4.1.1": "4.3.1",
+      "minimatch@10.1.1": "10.2.6",
+      "path-to-regexp@6.1.0": "6.3.0",
       "path-to-regexp@8.2.0": "8.4.2",
-      "path-to-regexp@8.3.0": "8.4.2"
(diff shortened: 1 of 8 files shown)


naingthet:
**NO GO** at 30c903373
Depth: medium

1. P2: `mobile/package.json` overrides `decode-uri-component` to `0.5.0`, which is ESM-only (`"type": "module"`, no CJS export). The only lockfile consumer is CJS `query-string@7.1.3` (`require('decode-uri-component')`), used by `expo-router` and `@react-navigation/core`. That can throw `ERR_REQUIRE_ESM` on Node/Metro paths that still load query-string as CommonJS. Pin a CJS-compatible patched build, or scope the override so it does not replace `query-string`'s dependency.

PR #2933, opened 2026-09-01
test(code): keep recap turn request in process

naingthet:
The disabled-recap test now sends the session turn through the in-process router. This removes the loopback HTTP sink that CodeQL treats as cleartext transmission while keeping the handler path covered.

Test: `cargo test -p tidebreak-server tests::code_recap::a_disabled_recap_setting_skips_the_model_call -- --exact`

Also ran `cargo fmt --check` and `git diff --check`.


naingthet:
**G2G** at f3a6e0dc0
Depth: medium

Issue #2311, opened 2026-08-20
epic: one interaction model for chat and code


naingthet:
Tracking issue for [decision 48](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0048-one-interaction-model.md).

Chat and code still have parallel journals, permission enums, approvals, turn machines, attention surfaces, and id spaces. The plan merges them in independently useful steps. Code-mode structures survive. Chat contributes the content model (attachments, documents, grant ladders).

**Already landed**

- Owner scoping for `code_*` tables and `/code/*` (decision 48 step 1, also required by [decision 47](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0047-gateway-linked-hosting.md)).
- #2315 — one permission vocabulary
- #2314 — one attention surface
- #2312 — one turn-submission contract

**Remaining steps** (each is independently useful; land in order):

- [x] #2315 — one permission vocabulary
- [x] #2314 — one attention surface
- [x] #2312 — one turn-submission contract
- [x] #2313 — one conversation, engines behind the adapter

#2313 starts only after the remote hosting wire has been exercised. Open further issues only when a slice has an owner and a bounded outcome. Do not file per-file or per-session tasks.

Parked on purpose, still in [`docs/deferred.md`](https://github.com/brightwave-inc/tidebreak/blob/main/docs/deferred.md): checkpoint restore, concurrent sessions, changing permission mode after start, an in-app editor, and remote session execution. The supervision-first mobile client is #2644.

naingthet:
Two follow-ups filed against this plan:

- #2430 — finish dropping the `code` prefix. Group A (`code_session`, `code_turn`, `code_event`, `code_approval`, the `/code/*` routes) belongs **inside** #2313 rather than before it: that is the change that retires `chat` / `message` / `event` and frees the names. Decision 30 already named the unprefixed nouns.
- #2428 — switch the schema posture from epoch resets to migrations. Every step here has to land first; the record already says #2313 after 1.0 would cost migrations it never budgets for.

[...]

PR #2934, opened 2026-09-01
feat(harness): extend the adapter contract with parks and structured decisions

naingthet:
First slice of #2313 (decision 48 step 5, epic #2311): before the internal agent loop can join the decision-31 adapter contract, the contract has to be able to express what chat's turn machinery does. This slice is purely additive — no behavior change for any external engine.

## What changes

- **Durable parks.** `TurnOutcome::Parked { park_ref, waiting_on }` with `ParkWait` (approval, client tool call, or agent-run set), and a default-refusing `HarnessSession::resume_turn(park_ref, ResumeInput)`. Chat's waiting-for-client continuations become expressible as an engine checkpointing durably and resuming after the wait resolves — with no side door past the adapter.
- **Structured approval decisions.** `ApprovalDecision` gains `ApproveWithGrant`, `Answers`, and `PlanDecision`; `CodeApprovalKind` gains `ToolUse` (exact `ToolActionPreview` + offered grant rungs), `Questions`, and `Plan`; `ApprovalDecisionKind` mirrors the new resolutions in the journal. One approval surface, two honesty postures: an engine that classifies precisely ships `kind` on `ApprovalRequested` and no raw; external adapters keep verbatim raw and a server guess.
- **Three new capability flags** — `durable_parks`, `user_questions`, `standing_grants`. Exhaustive construction forces every adapter to answer; all six existing implementors answer `Unsupported`. Decision 33's no-standing-grants posture for external harnesses is now a declared capability rather than an implied subsystem boundary.
- **`SessionSpec.binary` becomes `Option<PathBuf>`** for engines that run in-process and spawn no child. CLI-driving adapters refuse a spec without one.
- External adapters reject the decision variants their channels cannot carry (`HarnessError::DecisionUnsupported`); the session worker fails a turn loudly if an engine parks, since no registered engine may yet declare the capability.
- The scripted test harness scripts parks, resumes, and capability-gated decisions, so the contract is testable before the internal engine exists.

## Tests

- `a_parked_turn_resumes_through_the_adapter`, `resume_is_refused_without_the_capability`, `rich_decisions_are_refused_unless_declared` (scripted harness)
- Serde round-trips pinning the new decision and park-wait tags (harness + core)
- Journal fixture unchanged: the new variants are additive and existing rows keep their shape
- Ran: core lib suite (812 passed), harness lib suite (321), server `code` suite (660), UI `tsc` + biome + vitest (2284), `wire.ts` regenerated for desktop and mobile via the generator tests

Part of #2313; three more slices follow (worker park handling, the internal engine behind the adapter, then the entity merge).


naingthet:
**NO GO** at 136f06796
Depth: deep

1. P2: `mobile/src/lib/api.ts` `parseApprovalKind` still returns null for `tool_use`, `questions`, and `plan`. `approvalTitle` / `approvalSummary` already switch on those kinds, and `listCodeApprovals` runs every row through `parseCodeApproval`. One such row throws away the whole pending list instead of rendering the card.

2. P2: Desktop `parseCodeEvent` still accepts only `approve` / `deny` / `abandoned` on `approval_resolved`, while `CodeSessionReducer` now maps `approved_with_grant`, `answered`, and `plan_decided`. A journaled structured resolution therefore parses as null and never reaches the reducer, so the live card stays pending.

naingthet:
**NO GO** at 514694eff
Depth: deep

1. P2: `mobile/src/lib/api.ts` `parseApprovalKind` still returns null for `tool_use`, `questions`, and `plan`. `approvalTitle` / `approvalSummary` already switch on those kinds, and `listCodeApprovals` runs every row through `parseCodeApproval`. One such row throws away the whole pending list instead of rendering the card.

2. P2: Desktop `parseCodeEvent` still accepts only `approve` / `deny` / `abandoned` on `approval_resolved`, while `CodeSessionReducer` now maps `approved_with_grant`, `answered`, and `plan_decided`. A journaled structured resolution therefore parses as null and never reaches the reducer, so the live card stays pending.

naingthet:
**NO GO** at f5d52f308
Depth: deep

1. P2: `mobile/src/lib/api.ts` `parseApprovalKind` still returns null for `tool_use`, `questions`, and `plan`. `approvalTitle` / `approvalSummary` already switch on those kinds, and `listCodeApprovals` runs every row through `parseCodeApproval`. One such row throws away the whole pending list instead of rendering the card.

2. P2: Desktop `parseCodeEvent` still accepts only `approve` / `deny` / `abandoned` on `approval_resolved`, while `CodeSessionReducer` now maps `approved_with_grant`, `answered`, and `plan_decided`. A journaled structured resolution therefore parses as null and never reaches the reducer, so the live card stays pending.

naingthet:
**G2G** at d64122146
Depth: deep

Prior findings are fixed: mobile `parseApprovalKind` now keeps `tool_use` / `questions` / `plan` rows in the pending list, and desktop `parseCodeEvent` now applies `approved_with_grant` / `answered` / `plan_decided` so those cards leave pending. The rest of the slice stays additive — parks and structured decisions are on the contract, every live adapter still refuses them, and no engine can raise those cards yet, so the existing approve/deny surfaces stay honest.

naingthet:
**G2G** at e717aea75
Depth: deep

Parks and structured decisions land on the adapter contract only. Every live engine still refuses them, the worker still fails a parked turn, and the approve/deny route is unchanged, so existing cards stay honest. The new kinds render the literal tool action, the questions, and the proposed plan mode, and they hide an empty harness payload; nothing in this slice can raise those cards yet, so Approve/Deny remaining the only buttons is the right surface until an engine actually ships them.

PR #2961, opened 2026-09-01
perf(server): stop querying turn runs per streamed event

### LINKED ISSUE #2941: Events WebSocket queries turn runs for every streamed delta
`crates/tidebreak-server/src/routes/events.rs:187-193` runs `list_turn_runs(chat)` per forwarded event, per client. Streaming deltas turn this into hundreds of turn-table queries per second. The model only changes at turn boundaries: cache the active turn's model in the WS loop and refresh on turn-boundary events. The `replay_after` catch-up path (`events.rs:224-231`) should reuse the same cache.

naingthet:
Closes #2941

The chat events WebSocket was calling `list_turn_runs` for every forwarded live event, including each `TextDelta`/`ReasoningDelta`. The model only changes at turn boundaries, so this session now keeps a turn→model cache, refreshes it on turn started/finished (or a miss), and reuses that cache on `replay_after` catch-up instead of rebuilding from a fresh query when the mapping is already populated.

Checked: `cargo check -p tidebreak-server` and `cargo test -p tidebreak-server events` could not be run here (offline Cargo could not check out the locked `symphonia` git dep). Unit tests in `routes/events.rs` pin that deltas skip refresh once warm and that a turn-boundary refresh still retains prior turn models.

### DIFF
--- a/crates/tidebreak-server/src/routes/events.rs
+++ b/crates/tidebreak-server/src/routes/events.rs
@@ -1,5 +1,6 @@
 //! Route handlers extracted from the parent `routes` module.
 
+use std::collections::HashMap;
 use std::future::pending;
 use std::time::Duration;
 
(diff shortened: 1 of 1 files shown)


naingthet:
**G2G** at 271d7904d
Depth: medium

PR #2967, opened 2026-09-01
fix(harness): allowlist the coding-engine child environment

### LINKED ISSUE #2943: Coding-engine children inherit the user's full shell environment
`crates/tidebreak-harness/src/probe.rs:454` (`filter_child_env`) strips only `TIDEBREAK_*`, so secrets exported by the user's shell rc (GITHUB_TOKEN, AWS_*, provider keys) reach the engine child and are readable by a prompt-injected turn via `env`. Switch to an allowlist of what the engines need (PATH, HOME, TERM, LANG, engine-specific vars), the way the host exec path already uses `.env_clear()` (`local.rs:899`).

naingthet:
Closes #2943

`filter_child_env` previously stripped only the reserved `TIDEBREAK_` namespace, so anything the user's shell rc exports — `GITHUB_TOKEN`, `AWS_SECRET_ACCESS_KEY`, `OPENAI_API_KEY`, `ANTHROPIC_API_KEY` — reached every spawned coding-engine child, where a prompt-injected turn can read it with one `env` call. The child environment is now composed from an explicit allowlist, mirroring the `env_clear()` exec path in `tidebreak-code-execution/src/local.rs`.

Every spawn site is covered by the one shared filter: the four engine session children (via `apply_child_env_tokio`), the probe/version/auth/model-listing commands in `tidebreak-harness`, and the `gh`/usage spawns in `tidebreak-server`.

### What passes through now, and why

- `PATH` — command resolution; the probe prepends the managed Node runtime and the login-shell PATH here, and pinned engine entrypoints need their interpreter found through it.
- `HOME` — engine state lives under it (`~/.claude`, `~/.codex`, `~/.config` fallbacks).
- `USER`, `LOGNAME`, `SHELL`, `TERM`, `LANG`, `TMPDIR` — POSIX session identity, terminal, locale, temp basics.
- `LC_*` — locale configuration.
- `XDG_*` — base directories engines use for config and cache trees (e.g. opencode).
- `GH_CONFIG_DIR` — `tidebreak-server` spawns `gh` under this filter and gh resolves a relocated config directory through it; a path, not a secret (contract covered by an existing test).
- `SSL_CERT_FILE`, `REQUESTS_CA_BUNDLE`, `CURL_CA_BUNDLE`, `GIT_SSL_CAINFO`, `NODE_EXTRA_CA_CERTS` — the supervised runtime merges `Trust::environment()` into the session snapshot (`HarnessEngine::launch`), not `extra_env`; behind the sidecar's TLS-intercepting egress an engine child without these loses every outbound HTTPS call. CA-bundle paths, not secrets.
- Windows session basics (`SYSTEMROOT`, `COMSPEC`, `PATHEXT`, `USERPROFILE`, `APPDATA`, `TEMP`, …) — children cannot start, resolve commands, or write temp files without them; absent on Unix.

Everything Tidebreak wires into a child through the launch plan — settings `extra_env`, the session relay key, the browser capability file — is applied after the filter, so it needs no allowlist entry. Anything else is dropped: a missing variable fails visibly in the child, a leaked secret fails silently.

### Deliberate consequences

- Ambient env credentials (`ANTHROPIC_API_KEY`, `OPENAI_API_KEY`, gateway base URLs from the shell rc) no longer reach engine children implicitly. A machine that authenticates that way should carry the credential through Tidebreak settings `extra_env` or the session relay instead. Auth-override *detection* is unaffected: the stored probe snapshot keeps its denylist-only strip (`strip_tidebreak_env`) and still sees the full captured environment.
- Proxy variables (`HTTP_PROXY` etc.) are not forwarded; nothing in the harness forwarded them deliberately before, and they can carry credentials in the URL. They can be added back per-engine via settings `extra_env` if needed.

### Tests

- New: `child_env_is_an_allowlist_that_drops_planted_secrets` plants `AWS_SECRET_ACCESS_KEY`, `GITHUB_TOKEN`, `ANTHROPIC_API_KEY`, `OPENAI_API_KEY` and asserts none reach the computed child env while `PATH`/`HOME`/`LC_*`/`XDG_*` and the supervised trust variables (`SSL_CERT_FILE`, `NODE_EXTRA_CA_CERTS`) survive. It fails against the old denylist.
- Updated: the interactive-profile probe test now asserts a profile-only export stays out of the child while the profile-built `PATH` still arrives; the lib-level filter test asserts an arbitrary ambient variable is dropped.

`cargo check -p tidebreak-harness` and `cargo test -p tidebreak-harness` run clean locally except `cancelling_output_collection_kills_a_pipe_owning_descendant`, which fails in this sandbox's gVisor kernel independent of this change (untouched `child.rs`, process-group kill semantics).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7476,6 +7476,7 @@ dependencies = [
  "tidebreak-core",
  "tidebreak-managed-node",
  "tokio",
+ "toml 1.1.4+spec-1.1.0",
  "tracing",
  "uuid",
  "windows-sys 0.61.2",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -35,6 +35,12 @@ brush-parser = "=0.4.0"
 fancy-regex = "=0.19.0"
 serde = { version = "=1.0.229", features = ["derive"] }
 serde_json = "=1.0.151"
+# Parses the same `$CODEX_HOME/config.toml` the Codex engine parses, so
+# auth-override detection and the engine child-env filter agree with the
+# engine (and each other) about what the config declares. A substring scan
+# cannot tell a declared provider from a comment, and the filter must
+# forward exactly the env keys the declared providers name.
+toml = "=1.1.4"
 schemars = { version = "=1.2.2", features = ["uuid1"] }
 # Validates tool-call arguments against the advertised input_schema at
 # dispatch. No resolvers: schemas come from the registry or a mounted MCP
--- a/crates/tidebreak-harness/Cargo.toml
+++ b/crates/tidebreak-harness/Cargo.toml
@@ -27,6 +27,7 @@ thiserror = { workspace = true }
 tidebreak-core = { path = "../tidebreak-core", default-features = false }
 tidebreak-managed-node = { path = "../tidebreak-managed-node" }
 tokio = { workspace = true, features = ["io-util", "macros", "process", "rt", "sync", "time"] }
+toml = { workspace = true }
 tracing = { workspace = true }
 uuid = { workspace = true }
 
--- a/crates/tidebreak-harness/src/browser_channel.rs
+++ b/crates/tidebreak-harness/src/browser_channel.rs
@@ -22,8 +22,9 @@ pub const BROWSER_CAPFILE_ENV_KEY: &str = BrowserChannelSpec::ENV_KEY;
 /// single ordering every adapter must use:
 ///
 /// 1. Clear the inherited process environment.
-/// 2. Restore the probe snapshot through [`filter_child_env`], which strips
-///    the reserved `TIDEBREAK_` namespace (case-insensitively).
+/// 2. Restore the probe snapshot through [`filter_child_env`], which narrows
+///    it to the child allowlist — ambient shell-rc secrets and the reserved
+///    `TIDEBREAK_` namespace do not survive.
 /// 3. Apply the sanitized launch-plan environment. Settings already rejected
 ///    reserved keys via [`BrowserChannelSpec::is_reserved_env_key`], so this
 ///    step can carry no adapter-owned override.
--- a/crates/tidebreak-harness/src/browser_channel.rs
+++ b/crates/tidebreak-harness/src/browser_channel.rs
@@ -10,7 +10,7 @@
 
 use std::ffi::OsString;
 
-use crate::{filter_child_env, BrowserChannelSpec};
+use crate::{filter_engine_child_env, BrowserChannelSpec};
 
 /// Adapter must inject this exact key; engines must consume it.
 ///
@@ -22,9 +22,10 @@ pub const BROWSER_CAPFILE_ENV_KEY: &str = BrowserChannelSpec::ENV_KEY;
 /// single ordering every adapter must use:
 ///
 /// 1. Clear the inherited process environment.
-/// 2. Restore the probe snapshot through [`filter_child_env`], which narrows
-///    it to the child allowlist — ambient shell-rc secrets and the reserved
-///    `TIDEBREAK_` namespace do not survive.
+/// 2. Restore the probe snapshot through [`filter_engine_child_env`], which
+///    narrows it to the child allowlist plus `kind`'s own auth-signal
+///    variables — ambient shell-rc secrets, other engines' credentials, and
+///    the reserved `TIDEBREAK_` namespace do not survive.
 /// 3. Apply the sanitized launch-plan environment. Settings already rejected
 ///    reserved keys via [`BrowserChannelSpec::is_reserved_env_key`], so this
 ///    step can carry no adapter-owned override.
@@ -35,6 +36,7 @@ pub const BROWSER_CAPFILE_ENV_KEY: &str = BrowserChannelSpec::ENV_KEY;
 /// Some/None/final-override path. `None` adds no environment entry.
 pub fn apply_child_env_tokio<I>(
     cmd: &mut tokio::process::Command,
+    kind: tidebreak_core::HarnessKind,
     snapshot: I,
     plan_env: &[(String, String)],
     browser: Option<&BrowserChannelSpec>,
@@ -42,7 +44,7 @@ pub fn apply_child_env_tokio<I>(
     I: IntoIterator<Item = (OsString, OsString)>,
 {
     cmd.env_clear();
-    for (key, value) in filter_child_env(snapshot) {
+    for (key, value) in filter_engine_child_env(kind, snapshot) {
         cmd.env(key, value);
     }
     for (key, value) in plan_env {
@@ -67,8 +69,22 @@ mod tests {
         plan_env: &[(String, String)],
         browser: Option<&BrowserChannelSpec>,
     ) -> std::collections::BTreeMap<String, String> {
+        final_env_for(
+            tidebreak_core::HarnessKind::ClaudeCode,
+            snapshot,
+            plan_env,
+            browser,
+        )
+    }
+
+    fn final_env_for(
+        kind: tidebreak_core::HarnessKind,
+        snapshot: impl IntoIterator<Item = (OsString, OsString)>,
+        plan_env: &[(String, String)],
+        browser: Option<&BrowserChannelSpec>,
+    ) -> std::collections::BTreeMap<String, String> {
         let mut cmd = tokio::process::Command::new("/bin/true");
-        apply_child_env_tokio(&mut cmd, snapshot, plan_env, browser);
+        apply_child_env_tokio(&mut cmd, kind, snapshot, plan_env, browser);
         cmd.as_std()
             .get_envs()
             .filter_map(|(name, value)| {
@@ -140,7 +156,13 @@ mod tests {
         let mut cmd = tokio::process::Command::new("/bin/true");
         cmd.env(AMBIENT_SENTINEL, "must-be-cleared");
 
-        apply_child_env_tokio(&mut cmd, Vec::new(), &[], None);
+        apply_child_env_tokio(
+            &mut cmd,
+            tidebreak_core::HarnessKind::ClaudeCode,
+            Vec::new(),
+            &[],
+            None,
+        );
 
         assert!(
             cmd.as_std()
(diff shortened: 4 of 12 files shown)


naingthet:
Addressed the review finding in fb05efd8: the five Trust::environment() names (SSL_CERT_FILE, REQUESTS_CA_BUNDLE, CURL_CA_BUNDLE, GIT_SSL_CAINFO, NODE_EXTRA_CA_CERTS) are now in the child allowlist with a comment pointing at the HarnessEngine::launch snapshot merge, and the allowlist test pins SSL_CERT_FILE and NODE_EXTRA_CA_CERTS surviving. The gateway wiring in that same launch path travels through extra_env/relay_key_env, which is applied after the filter, so the trust pairs were the only snapshot-merged injections.

naingthet:
**NO GO** at fb05efd8f
Depth: deep

1. P2: Session create and the doctor still treat captured shell credentials as a working auth mode, but the child no longer receives them. `refuse_signed_out_harness` and `resolve_auth_mode` read `auth_override_present` / `observe_auth_mode` against the unfiltered probe snapshot (`ANTHROPIC_*`, `OPENAI_*`, Bedrock/Vertex flags). `filter_child_env` drops those names, and desktop launch only fills `SessionSpec::extra_env` for the hosted inference relay. A signed-out machine that used to run via a shell-rc key still looks GatewayManaged with empty remediation, then 401s on the first turn — the failure issue 2653 asked create to refuse. The PR's stated escape hatch (settings `extra_env`) is not a desktop launch input; engine config under HOME still works, shell-rc auth does not. Point detection at the same allowlist the child inherits, or stop treating snapshot-only env keys as live auth.

naingthet:
Addressed the P2 from the review (head is now 2ce2eca99188260b6d12b097e80249b8b58e3dc0).

The gap was that create/doctor auth detection and the child filter read different sets: refuse_signed_out_harness and resolve_auth_mode counted ANTHROPIC_API_KEY / ANTHROPIC_AUTH_TOKEN / ANTHROPIC_BASE_URL / CLAUDE_CODE_USE_BEDROCK / CLAUDE_CODE_USE_VERTEX (Claude Code) and OPENAI_API_KEY / OPENAI_BASE_URL (Codex) in the unfiltered snapshot as working auth, while filter_child_env dropped all of them, so a shell-rc-authenticated machine was admitted as GatewayManaged and 401d on its first turn.

The auth-signal sets are now defined once in probe.rs (CLAUDE_AUTH_ENV_VARS, CODEX_AUTH_ENV_VARS) next to the child allowlist, and both the detection code (claude/codex observe_auth_override) and child_env_allowed read those same constants, so the two cannot drift: whatever detection counts as the engine's live auth, the child actually receives. CODEX_HOME also passes now, since detection reads a custom model provider out of $CODEX_HOME/config.toml and the child must resolve the same file.

For the Bedrock/Vertex switches, whose supporting credentials the flags alone do not carry: AWS_* passes only when CLAUDE_CODE_USE_BEDROCK is set non-empty, and GOOGLE_* / ANTHROPIC_VERTEX_* / CLOUD_ML_REGION only when CLAUDE_CODE_USE_VERTEX is, using the same non-empty rule as detection's env_sets_any. With a switch set, that provider's credential chain is the engine's configured auth; without it, the same variables are unrelated secrets and stay out, as does everything else (GITHUB_TOKEN, NPM_TOKEN, random rc exports).

Tests: the allowlist test now pins that ANTHROPIC_API_KEY and OPENAI_API_KEY survive the filter while GITHUB_TOKEN, NPM_TOKEN, and un-flagged AWS_SECRET_ACCESS_KEY do not, and keeps the CA-trust pins; a new test pins the Bedrock/Vertex gating including the empty-flag and does-not-widen-to-GITHUB_TOKEN cases. cargo check and cargo test -p tidebreak-harness pass locally (the one unrelated failure, child::unix_process_tree_tests::cancelling_output_collection_kills_a_pipe_owning_descendant, fails identically at the previous head in this sandbox), cargo fmt --all is clean.

naingthet:
**NO GO** at 2ce2eca99
Depth: deep

1. P2: The shared child filter now admits both engines’ auth variables unconditionally, so a credential used to authenticate one engine is disclosed to unrelated children. `OPENAI_API_KEY` and `OPENAI_BASE_URL` reach Claude, Grok, OpenCode, probe, `gh`, and `modelctl` spawns; `ANTHROPIC_API_KEY`, `ANTHROPIC_AUTH_TOKEN`, and `ANTHROPIC_BASE_URL` reach Codex and those same non-Anthropic children. With `CLAUDE_CODE_USE_BEDROCK` set, `AWS_*` follows every spawn; Vertex does the same for `GOOGLE_*`. Create and the doctor now share `CLAUDE_AUTH_ENV_VARS` / `CODEX_AUTH_ENV_VARS` with the filter, so the first-turn 401 at fb05efd8f is closed for Claude/Codex — but a prompt-injected turn in the other engine can still `env` the first engine’s secret, which is the disclosure this allowlist was meant to stop. Scope the exception to the engine that uses it (and keep probe/`gh`/usage on the tight list) rather than widening the one shared filter.

naingthet:
Addressed the P2 on cross-engine credential disclosure (head is now ac501028fc5808d06ce79121671157545da92da1).

The auth-signal passthrough is no longer in the shared filter. filter_child_env is back to the base allowlist with no auth variables at all, and that is what probe spawns (--version, model listings, the auth-status checks in each engine module), gh, and the gateway usage CLI use. Engine session launches now go through a new filter_engine_child_env(kind, snapshot), which adds only the auth variables that engine's own detection reads: ANTHROPIC_API_KEY / ANTHROPIC_AUTH_TOKEN / ANTHROPIC_BASE_URL plus the Bedrock/Vertex switches for Claude Code, OPENAI_API_KEY / OPENAI_BASE_URL for Codex, and nothing for opencode and Grok — their auth_override_present grants the benefit of the doubt without reading any environment names, so per the detection-mirrors-filter invariant their children receive no auth variables. Bedrock/Vertex supporting credentials (AWS_*, GOOGLE_* / ANTHROPIC_VERTEX_* / CLOUD_ML_REGION) pass only when the launching engine is Claude Code AND the matching switch is set non-empty in the snapshot; a Codex launch drops them even with the switch present.

The per-engine sets remain the single constants shared with the detection code (CLAUDE_AUTH_ENV_VARS / CODEX_AUTH_ENV_VARS in probe.rs), so the previous finding's invariant still holds per engine: whatever resolve_auth_mode counts as working auth for engine X reaches engine X's session child, and nothing else's. apply_child_env_tokio takes the engine kind, so all four adapters stay on the one shared launch path.

Tests: probe tests pin ANTHROPIC_API_KEY surviving a Claude Code launch and dropped for Codex/opencode/Grok, OPENAI_API_KEY vice versa, GITHUB_TOKEN dropped everywhere, the CA-trust pins unchanged, and the Bedrock/Vertex gating including the wrong-engine and empty-flag cases; a browser_channel test pins the same scoping on the real adapter env path. cargo check -p tidebreak-harness and -p tidebreak-server pass, cargo test -p tidebreak-harness passes (same single unrelated pre-existing sandbox failure in child::unix_process_tree_tests as before), cargo fmt --all is clean.

naingthet:
**G2G** at ac501028f
Depth: deep

The prior cross-engine leak is gone: `filter_child_env` is the tight base list (probe, `gh`, modelctl), and session launches go through `filter_engine_child_env(kind, snapshot)` so Claude and Codex children receive only the auth names their own detection reads. Create/doctor still admit shell-rc auth for those engines, and the child actually gets it, without handing the other engine, or a prompt-injected turn in Grok/OpenCode, a key it does not use.

naingthet:
Addressed the P2 on config-declared Codex provider credentials (head is now bed12c22).

The fix closes the invariant on both sides by parsing $CODEX_HOME/config.toml as real TOML — the same file, resolved from the same snapshot — in detection and in the child filter:

- filter_engine_child_env now forwards, for Codex launches only, exactly the environment keys the declared providers name: each [model_providers.*] entry's env_key plus every variable its env_http_headers map reads. Bounded to what the config names; nothing else widens. On the reviewer's example, a config naming GATEWAY_OPENAI_KEY now hands that variable to the Codex child and to no other engine's.
- config_declares_model_provider no longer admits on a substring match. It admits on a genuine declaration: a top-level or per-profile model_provider selection, or a non-empty [model_providers.*] table. A comment that merely mentions the word declared nothing and previously admitted a session whose first turn 401s with no credential for the filter to forward — that path stops counting as live auth. An unparseable config keeps the old textual over-read, because Codex then refuses to start with its own config error, which is more legible than a sign-in refusal.

Admission paths audited (refuse_signed_out_harness / resolve_auth_mode both route through observe_auth_override per engine):

- Claude env vars (ANTHROPIC_API_KEY / ANTHROPIC_AUTH_TOKEN / ANTHROPIC_BASE_URL, Bedrock/Vertex switches): forwarded to Claude children only, with the switch-gated AWS_*/Google credential chains. Unchanged.
- Claude settings.json (apiKeyHelper, env block): both are file-carried — the engine reads settings.json itself and the child reaches it via HOME, so the helper command and the env-block values travel with the file. Neither names a parent-environment key, so nothing to forward. Verified, no change.
- Codex env vars (OPENAI_API_KEY / OPENAI_BASE_URL): forwarded to Codex children only. Unchanged.
- Codex config-declared providers: env_key and env_http_headers names now forwarded (this fix). A provider naming no key — ChatGPT OAuth (requires_openai_auth), an auth helper, an unauthenticated local endpoint — authenticates through files under CODEX_HOME, which the child reaches via CODEX_HOME/HOME, and contributes no names; pinned by a test that an ambient variable stays dropped when such a provider is declared.
- opencode and Grok: detection reads no environment or config surface (blanket benefit of the doubt), so no admission ever rests on an environment credential and their children receive no auth variables; their file-based auth travels via HOME/XDG_*. Symmetric env-key-naming configs: neither engine's config is read by detection, so no equivalent of the Codex case exists to forward.
- Relay-managed launches (harness_llm) inject their provider config and key via plan_env after the filter, unaffected.

Tests pin the new case: a config declaring env_key GATEWAY_TEST_KEY (and an env_http_headers variable) survives for a Codex launch and is dropped for a Claude launch; a declared provider without env keys forwards nothing extra; comment-only and unparseable configs pin the detection behavior; config_provider_env_keys is pinned to exactly the named, deduped keys. Existing pins (ANTHROPIC only to Claude, OPENAI only to Codex, CA-trust vars, GITHUB_TOKEN never) are untouched and green.

cargo check, clippy, and the tidebreak-harness test suite pass locally (326 passed; the one failure, child::unix_process_tree_tests::cancelling_output_collection_kills_a_pipe_owning_descendant, also fails on the unmodified base in this sandboxed kernel and is untouched by this change). New direct dependency: toml, pinned to =1.1.4 matching the lockfile.

naingthet:
**G2G** at 910d25832
Depth: deep

The per-engine filter still holds at this head: Claude and Codex session children receive only the auth names their own detection reads, including Codex `env_key` / `env_http_headers` from the same `config.toml` create parses, and those names do not reach Claude, Grok, OpenCode, probe, or `gh`. Ambient secrets stay off the shared allowlist.

naingthet:
@shipright review

naingthet:
@shipright review

PR #2968, opened 2026-09-01
perf(server): wake the queued-turn promoter instead of fast-polling

### LINKED ISSUE #2949: Queued-turn promoter polls every 750ms with no wake wired
`tidebreak-server/src/lib.rs:2390` ticks a DISTINCT `chats_with_queued_turns` query every 750ms forever; the comment claims wake-driven but no wake exists. `code/session_worker.rs:770` shows the Notify-driven pattern: notify on enqueue and turn completion, keep a multi-second fallback tick.

naingthet:
Closes #2949

The queued-turn promoter ran a 750ms `interval` over a DISTINCT `chats_with_queued_turns` query forever, with no wake wired despite its comment. This wires the repo's Notify pattern (as in `code/session_worker.rs`):

- New `AppState::queued_turn_wake`, fired from every commit that can make a queued message promotable:
  - both enqueue paths in `post_message` (`BeginTurnAdmissionOutcome::Queued` and `ReservedQueuedTurnOutcome::Queued`)
  - a turn reaching a terminal state — end of a worker segment (`Completed`/`Cancelled`/`Failed`) and lease-expiry terminalization in `claim_once`
  - `post_cancel`, which can terminalize not-yet-running work atomically without a worker segment
  - both unpause paths: `PUT …/queue-paused` with `paused: false` and `POST …/queued/send-now`
- The promoter loop (`run_queued_turn_promoter`) sweeps, then `select!`s on `notified()` with the fallback tick stretched from 750ms to a 5s floor. Sweep-then-wait means a notification landing mid-sweep is held as the `Notify` permit and the next `notified()` completes immediately, so no wakeup is lost between checks. An enqueue while idle still promotes near-immediately via the notify.

New test `queued_turn_promotes_on_completion_wake_without_the_fallback_tick` runs the real promoter loop with a 300s floor, queues a follow-up behind a gated turn, releases the gate, and asserts the row promotes within seconds — only the wake path can do that in time.

Ran locally:
- `cargo check -p tidebreak-server` — clean
- `cargo test -p tidebreak-server queue` — 26 passed; 8 failures (`tests::code::*`, `chat_titling`, `sandbox_container_run` loopback tests) are TCP-connect timeouts in my sandbox and reproduce identically on unmodified `main`
- `cargo test -p tidebreak-server -- queued_turn_promotes_on_completion_wake send_now_releases_a_paused_queue` — both pass; the new wake test finishes in ~3s against its 300s floor

### DIFF
--- a/crates/tidebreak-server/src/lib.rs
+++ b/crates/tidebreak-server/src/lib.rs
@@ -2329,6 +2329,7 @@ async fn bind_inner(
         state.active_turns.clone(),
         state.turn_job_wake.clone(),
         state.agent_run_wake.clone(),
+        state.queued_turn_wake.clone(),
         state.agent_config.clone(),
         Some(state.config.data_dir.join("scratch")),
         turn_worker::TurnWorkerConfig {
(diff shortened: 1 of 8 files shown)


naingthet:
**G2G** at ae4a11c4b
Depth: medium

naingthet:
**G2G** at b9aaa2da9
Depth: medium

PR #2969, opened 2026-09-01
perf(ui): stop app-wide re-renders from context and drag

### LINKED ISSUE #2942: AppShell context value is rebuilt every render; sidebar drag re-renders the whole app
`ui/src/AppShell.tsx:975` builds the `AppContextProvider` value inline with ~15 fresh closures, so all `useApp()` consumers re-render whenever AppShell does. AppShell subscribes to `sidebarWidth`, which the drag handle (`sidebar/primitives.tsx:87`) writes per pointermove, so a drag re-renders the app per frame. Also `MessageList.tsx:336` calls `groupMessageItems` un-memoized, and `ComputerUseIndicator.tsx:28` ticks a 5s re-render unconditionally. Memoize the context value, drive drag width through a CSS variable committed on drag end, memoize the grouping, gate the indicator tick.

naingthet:
Closes #2942

## What changed

- **AppShell context** (`AppShell.tsx`): the `AppContextProvider` value was an inline object with ~15 fresh closures, so all ~34 `useApp()` consumers re-rendered on every shell render. The value is now built with `useMemo` keyed on the data it carries (client, catalog, status, update state). The mutations go through one stable set of forwarders that dispatch via a per-render ref, so their identities never change and their closures never go stale. Verified against consumers: the only context member appearing in a dependency array anywhere is `setStatus` (`ChatRoute.tsx`), which is a `useState` setter and was already stable — no consumer relied on the context identity changing per render.
- **Sidebar drag** (`sidebar/primitives.tsx`, `UiStore.ts`): pointermove wrote `sidebarWidth` to the store every frame, and AppShell subscribes to it for its `--sidebar-expanded-width` CSS variable — so a drag re-rendered the whole app, transcript included, per frame. The handle now writes the clamped width imperatively during the drag (the shell's CSS variable, which also moves the titlebar, plus the rail's own inline styles) and commits to the store once on release. Persistence is identical: one localStorage write per drag, on release; keyboard and double-click resizes are unchanged. The now-caller-less `persist: false` option is removed from `setSidebarWidth`.
- **MessageList** (`MessageList.tsx`): `groupMessageItems` — a full transcript walk that builds every row — ran un-memoized in the component body. It and the background-agent options object are now under `useMemo`; the background-agent prop defaults are hoisted to module constants so omitting those props no longer produces fresh identities that would bust the memo.
- **ComputerUseIndicator** (`ComputerUseIndicator.tsx`): an unconditional 5s `setNow` interval ticked from shell mount forever, even when no computer-use session ever ran. It now uses the existing `useNowWhile` pattern from `BackgroundAgentPanel`, gated on the active banner actually being on screen — no session, or a session already idled past its visibility window, means no timer and no re-renders. The crossing tick re-renders the banner away and stops the interval with it.

## What I ran

- `biome format` (no fixes) and `pnpm lint` — clean
- `tsc` and `pnpm build` — clean
- Targeted suites: `AppShell.dom.test.tsx`, `MessageList.test.tsx`, `MessageList.dom.test.tsx`, `MessageList.isolation.dom.test.tsx`, `sidebar/SettingsSidebar.dom.test.tsx` — 93/93 pass
- Full `pnpm test`: 2268 pass; 23 failures in 4 files (`CodeWorkspacePage`, `CodeDeliveryPage`, `PullRequestDetail`, `OutputsView`) were 5s timeouts from sandbox CPU starvation — each of those files passes in isolation on this branch (139/139), and one of them also flaked on a clean `main` checkout under the same load
- `pnpm storybook:build` — succeeds (Navigation/Sidebar and ConversationTranscript stories cover the touched surfaces)

### DIFF
--- a/crates/tidebreak-desktop/ui/src/AppShell.tsx
+++ b/crates/tidebreak-desktop/ui/src/AppShell.tsx
@@ -1,4 +1,10 @@
-import { useEffect, useRef, useState, type CSSProperties } from "react";
+import {
+  useEffect,
+  useMemo,
+  useRef,
+  useState,
+  type CSSProperties,
+} from "react";
 import { Outlet, useNavigate, useRouter } from "@tanstack/react-router";
 import { getVersion } from "@tauri-apps/api/app";
 import { isTauri } from "@tauri-apps/api/core";
@@ -13,7 +19,7 @@ import {
   type ProviderInfo,
   type ServerInfo,
 } from "./api";
-import { AppContextProvider } from "./AppContext";
+import { AppContextProvider, type AppContextValue } from "./AppContext";
 
 import { resolveServerInfo } from "./boot";
 import {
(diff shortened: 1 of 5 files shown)


naingthet:
**G2G** at 29844916d
Depth: medium

PR #2970, opened 2026-09-01
fix(code)!: make archive the storage cleanup flow

naingthet:
## What changed

- Archive now removes a local worktree, saves the branch's own commits as a git bundle, drops the branch, and records the workspace as `released`.
- Restore still rebuilds the exact worktree and branch from that bundle. Remote workspaces stay `archived` because they have no host worktree or local branch.
- Removed the Storage page, route, sidebar entry, palette row, API endpoint, parser, and wire types.
- Removed the separate release endpoint. Archive is the single cleanup action.

## Why

The storage page exposed internal reclaim tiers as a user workflow. Users only need one cleanup action: put the work away, free its space, and keep it recoverable from Archive.

## Validation

- `cargo test -p tidebreak-server archive_frees_the_branch_and_restore_rebuilds_it_from_the_bundle`
- `cargo test -p tidebreak-server released_restore_keeps_the_bundle_when_the_final_row_fails`
- `pnpm --dir crates/tidebreak-desktop/ui lint`
- `pnpm --dir crates/tidebreak-desktop/ui test`
- Captured and reviewed Archive and sidebar stories in light and dark.


naingthet:
**NO GO** at 073214a09
Depth: medium

1. P2: Archive can delete the branch and still report success when the Released row never lands. `finalize_removed_workspace` bundles the commits and drops the ref, then warns on a failed `save_workspace` and returns `Ok(released)`. The durable row stays `archived` with no `released_tip`. Restore therefore takes the kept-branch path, returns `branch_missing`, and the UI says the work cannot come back even though the bundle is on disk. Retrying archive is a no-op once the row is `Archived`, so nothing finishes the release. Fail the request if that write fails, or persist Released before you return success.

naingthet:
**NO GO** at 82712adcc
Depth: medium

1. P2: Archive can still drop the branch without a durable Released row, then restore hits `branch_missing`. `finalize_removed_workspace` now errors on `save_workspace_final` `Err`, but it already persisted `Archived` via `complete_workspace_archive` before bundling and deleting the ref. A failed write therefore fails the first request while the durable row stays `archived` with no `released_tip`. Retrying archive returns success because `Archived` is a no-op, and restore takes the kept-branch path. `save_workspace_final` `Ok(false)` is still treated as success (`await?` only propagates `Err`). Fail that write too, and do not return success until the Released row lands — or persist Released before you drop the branch.

naingthet:
**NO GO** at 559e946f2
Depth: medium

1. P2: Archiving a local workspace whose branch is at its base records Released with no `released_tip` or bundle, then deletes the branch. Restore then fails with `released_tip_missing` and cannot recover, even though `base_ref` is still in the repo. That is the freshly created workspace path — `archive_succeeds_for_a_branch_with_no_commits_past_its_base` asserts the unrestorable restore. Persist the tip and restore at that commit without a bundle, or keep a base-only restore path, before dropping the ref.

naingthet:
@shipright review

naingthet:
**G2G** at b6c66747f
Depth: medium

PR #2136, opened 2026-08-15
feat(code): commit, push, and pull-request flow for workspaces

naingthet:
## Summary

Workspaces can now commit, push, and open a pull request from the code-mode surface.

- `POST /code/workspaces/{id}/git/commit` stages the worktree and commits. A missing message is generated from the workspace title plus the staged shortstat. A clean tree is a typed `nothing_to_commit` 409.
- `POST /code/workspaces/{id}/git/push` pushes the workspace branch to `origin` with upstream set. Auth failures return remediation text; Tidebreak never prompts and never stores git credentials.
- `POST /code/workspaces/{id}/git/pr` creates a pull request through the user's `gh` CLI. Title and body are generated from the workspace title, commit list, and diffstat when omitted. The route never merges and never enables auto-merge.
- `GET /code/workspaces/{id}/pr` returns the local git facts plus a briefly cached `gh pr view` / `gh pr checks` digest, persisted on `code_workspace.pr` so list views can carry it.
- Absent or signed-out `gh` is a typed 409 with copyable terminal instructions. Auth is observed with `gh auth status`; tokens are never stored.
- `POST /code/workspaces/{id}/actions/{name}` runs a repo quick action in the worktree and returns bounded output. It does not write the session journal. `auto_run_on_create` actions run once after setup during workspace creation.
- The workspace page now has a `PrCard` for commit / push / create-PR, status and checks chips, and the gh-absent remediation state. Archive copy points at this card when leftover work would be discarded.

## Tests

- Temp repos with a bare `origin` remote for commit and push (no network).
- `gh` PATH shims for create / view / checks / absent / signed-out, including an assertion that merge and GraphQL are never invoked.
- Deterministic generated commit messages and nothing-to-commit refusal.
- Quick-action output bound and timeout.

Local: `cargo fmt`, `cargo clippy -p tidebreak-server --all-targets -- -D warnings`, `cargo test -p tidebreak-server --locked`, `cargo check --workspace --locked`, `pnpm test`, `pnpm build`.

PR #2971, opened 2026-09-01
fix(code): ship structured check counts on the digest

### LINKED ISSUE #2946: UI parses PR check counts back out of the human-readable summary string
`ui/src/code/prState.ts:161-165` regex-parses "(N) passing" etc. from `checks_summary` (produced at `tidebreak-core/src/code/mod.rs:474`) when the structured check list is absent. Any wording change silently zeroes the UI counts. Ship structured counts on the digest wire type and drop the prose round-trip.

naingthet:
Closes #2946

The digest wire type now carries `check_counts` (passing/pending/failing/skipped) beside `checks_summary`, tallied by the same pass that renders the summary string; `checks_summary` stays display-only. The UI reads the structured counts instead of regex-parsing the prose, and a core test pins that the counts and the summary line's numbers come from one tally.

The regex fallback in `prState.ts` is kept deliberately: digests are persisted in the `code_pull_request` live tier, and rows written before #2563 can carry a checks summary with no structured check list (the original #2136 producer built the summary straight from `gh pr checks` output), so an old row — or an older self-hosted server talking to a newer UI — can still surface a summary with neither list nor counts.


naingthet:
**G2G** at 2f56d0484
Depth: medium

PR #2974, opened 2026-09-01
fix(code): detect pull requests from changed checkouts

naingthet:
## Summary

- inspect a changed workspace checkout after each turn when command detection misses the create or push
- require a clean checkout whose branch and HEAD exactly match an open pull request
- cover successful recovery, different-head rejection, and dirty-checkout rejection

## Validation

- cargo fmt --all -- --check
- git diff --check origin/main...HEAD
- cargo test -p tidebreak-server code_pr_facts -- --nocapture

### DIFF
--- a/crates/tidebreak-core/src/code/mod.rs
+++ b/crates/tidebreak-core/src/code/mod.rs
@@ -807,10 +807,13 @@ impl CodePullRequestState {
 
 /// How strongly a workspace is tied to a pull request (decision 77).
 ///
-/// Only two acts mint attribution: `gh pr create` (authored) and a push whose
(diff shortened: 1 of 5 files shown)


naingthet:
**G2G** at 8a1a86cd3
Depth: medium

Checkout recovery is gated on a persisted turn checkpoint with files > 0, a clean worktree still on the workspace branch, and an exact open-head SHA match after a confirming `gh` read. It runs after checkpointing, does not attribute remote workspaces or review checkouts on another branch, and the new tests cover recovery plus the dirty, other-branch, and different-head misses.

PR #2982, opened 2026-09-01
feat(code): persist durable turn parks and route rich decisions

naingthet:
## What

Second slice of #2313 (decision 48 step 5): the session worker learns durable
parks and the routes learn rich decisions.

- A turn whose engine returns `TurnOutcome::Parked` persists as a new
  `waiting` status with the engine's `park_ref` and a typed `park_wait`
  (approval by engine call id, client tool call, or agent runs). The worker
  releases the leg, then resumes through `resume_turn` when the awaited
  decision arrives; a resumed turn may park again. Interrupt closes a parked
  turn cleanly.
- `POST /code/approvals/{id}/decision` accepts `approve_with_grant`
  (resolves `grant_index` against the approval's offered scopes — clients
  never name scopes), `answers` (validated against the question shape), and
  `plan_decision`. Each is gated on the engine's declared capability and the
  approval's structured kind; bare `approve` is refused on questions and
  plans.
- Migration `m20260901_000001_code_turn_park` widens the `code_turn` status
  check to admit `waiting` and adds the park columns. SQLite takes a table
  rebuild (the check is anonymous); PostgreSQL swaps the constraint in
  place. The v0.60 upgrade test now asserts a seeded turn row survives the
  rebuild.
- Recovery closes a dead worker's `waiting` turn as interrupted; engines
  resume from their own durable state until slice C gives the internal
  engine real recovery.
- Desktop and mobile treat `waiting` as an open turn: the session stays
  busy, no boundary renders, and the pending approval card is the call to
  action.

## Deferred

The plan-approval → `set_permission_mode` follow-up lands with the internal
engine in slice C, where the first engine that can propose a plan arrives.

Part of #2313 / epic #2311.

### DIFF
--- a/crates/tidebreak-cli/src/agent_mcp/code_follow.rs
+++ b/crates/tidebreak-cli/src/agent_mcp/code_follow.rs
@@ -137,7 +137,8 @@ fn from_turn_status(
         CodeTurnStatus::Completed => TurnStatus::Completed,
         CodeTurnStatus::Failed => TurnStatus::Failed,
         CodeTurnStatus::Interrupted => TurnStatus::Cancelled,
-        CodeTurnStatus::Running => TurnStatus::Running,
+        // A parked turn is still in flight from a follower's seat.
+        CodeTurnStatus::Running | CodeTurnStatus::Waiting => TurnStatus::Running,
     };
     CodeTurnResult {
         status,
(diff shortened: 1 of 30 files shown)


naingthet:
**G2G** at 82a272c35
Depth: deep

Waiting is an open turn on desktop, mobile, and MCP follow: the session stays busy, no boundary renders, and the pending approval card is the call to action. Parks persist with park_ref/park_wait and resume on the awaited approval, including a Decide that landed on the running leg; interrupt closes a parked turn, and recovery interrupts a dead worker's waiting row as this slice documents. Rich decisions are capability-gated, grant_index only names offered rungs, and bare approve is refused on questions and plans — kinds no shipped engine raises yet.

Issue #2430, opened 2026-08-21
refactor: finish dropping the code prefix from shared nouns


naingthet:
[Decision 30](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0030-code-mode-separate-surface.md) already named code mode's nouns: **repo**, **workspace**, **session**, **turn**. Unprefixed. The `Code` / `code_` prefix is a namespacing artifact of "its own tables and journal," not the product vocabulary.

[Decision 48](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0048-one-interaction-model.md) says code-mode structures survive the merge and chat merges into them. After #2313, a chat is a session with the internal engine and no workspace — at which point `code_session` holding an ordinary chat is a lie.

Done already: `CodeExecution*` → `Exec*` (#2423). That was an unrelated subsystem sharing the prefix.

## Group A — becomes universal, rename with #2313

`code_session`, `code_turn`, `code_event`, `code_approval`, `code_turn_attachment`, the matching `CodeSession` / `CodeTurn` / `CodeEvent` / `CodeApproval` types, the `/code/*` routes, and the generated TS names.

Do this **inside** #2313, not before. That is the change that retires `chat` / `message` / `event` and frees the names. Renaming earlier leaves `session` and `chat` as two live tables with nearly the same purpose, which reads worse than the prefix does. The route rename is client-visible, so spend it once, there.

## Group B — stays code-specific, and is not the easy half

`code_repo`, `code_workspace`, `code_watch`.

These survive the merge as code concepts, and the prefix is not what makes them specific. Half of this is already done and inconsistent:

- `RepoId` and `WorkspaceId` are **code mode's** id types and are already unprefixed, declared in `crates/tidebreak-core/src/code/mod.rs`.
- The entity structs beside them are still `CodeRepo` and `CodeWorkspace`.

So ids and entities disagree today. But finishing it hits a real collision, which is why this is not a sed:

**`Workspace` is a contested noun.** `WorkspaceFileEntry`, `WorkspaceListing`, and `WorkspaceLifecycle` are defined in `tidebreak-code-execution` and describe the **exec sandbox's** workspace, unprefixed. `WorkspaceId` in `tidebreak-core` describes a **code-mode** worktree, also unprefixed. Same word, two subsystems, no prefix separating them — the same problem #2423 just fixed one level up, and `ExecutionWorkspaceId` in that crate shows someone already hit it.

[...]

PR #3001, opened 2026-09-01
feat(code): run the internal engine behind the adapter contract

naingthet:
Third slice of #2313 (decision 48 step 5, epic #2311): the internal agent loop implements the decision-31 adapter contract as one engine among several. A conversation bound to a workspace selects a harness engine; one without selects the internal engine.

## What changes

- **`HarnessKind::Internal`.** In-process, reference tier. It probes as found with no binary, needs no pin, takes no relay wiring, and resolves inference through the server's own provider resolution. Every `HarnessKind` match site answers for it; the doctor skips it, since there is nothing to install or sign in to.
- **Sessions without a workspace.** `CodeSession.workspace_id` is now `Option<WorkspaceId>`. Migration `m20260901_000002_internal_engine_sessions` drops the NOT NULL (SQLite rebuilds `code_session` from its stored definition; PostgreSQL alters in place) and adds `chat.engine_private`. Every checkout-bound path (baseline, checkpoints, PR facts, browser channel, titling, worktree lock) skips a session with no workspace; the engine's working directory is a private root under `code/private/sessions/{id}`.
- **`POST /code/sessions`** creates such a session. The body names no engine: the absence of a workspace is what selects it. A workspace session that asks for `internal` is refused with `harness_unavailable`.
- **The engine** (`server/src/engine/internal`) drives the chat turn lane — admission, the durable `turn_run` queue, the `TurnWorker`, the chat journal — and translates that journal into `HarnessEvent`s for the session worker's sink. Its durable state is one engine-private `chat` row per session, keyed by the session id, which every owner-scoped chat read excludes; the conversation is reachable only through its session. Chat's continuations ride the contract as durable parks: a user-questions card or a plan proposal ends the leg as `Parked`, and the answer or plan decision resumes it through `resume_turn`. It declares `durable_parks`, `user_questions`, and `standing_grants`.
- **Engine-observed decisions.** An approval the engine decides on its own channel (a standing grant, the Auto judge) reaches the worker as `ApprovalResolved`; the worker now settles the still-pending, unclaimed row from it (`settle_engine_observed_approval`) instead of dropping the event. A row the decision route claimed never matches.
- **Plan accept → mode.** The deferred follow-up from #2982: `WorkerCommand::Decide` carries the accepted plan's proposed mode, and the worker applies it on delivery (`set_permission_mode` on the engine, the row, and a journal notice), because the turn is still open and the settings route refuses a running session.
- **`SessionSpec.owner` and `SessionSpec.session_id`** for engines that key their own durable state; external adapters ignore them.
- **Clients.** `wire.ts` regenerated and synced to mobile. The desktop rail is keyed by workspace, so a digest with `workspace: null` gets no row until the surfaces merge (slice E); labels, parsers, pickers, and analytics know the new kind.

## Tests

- `a_session_without_a_workspace_runs_on_the_internal_engine`: creates through `POST /code/sessions`, runs a turn end to end against the fake provider through the chat turn worker, asserts the answer reached the code journal, and asserts the engine's conversation is absent from `GET /chats` and 404 on `GET /chats/{id}`.
- `the_internal_engine_is_not_a_workspace_engine`: refusal on a workspace session and absence from the doctor.
- The v0.60 upgrade test now asserts the seeded session row survives the `code_session` rebuild with its workspace, that the column is nullable, and that `chat.engine_private` defaults to false.
- Translate unit tests pin the chat→adapter mapping.
- Ran: core migration + code store suites (93), harness lib (330 pass; `interactive_profile_env_is_captured_and_stripped` fails in this agent shell on main too — it reads the login profile), server `tests::code*` + `engine::` + `code::` (648), clippy on the touched crates, desktop `tsc` + biome + vitest (2320), mobile `tsc`.

## Deferred

Filed as issues at close-out: image input on the internal engine, recovery resume of a parked internal turn after a restart, and sandbox agents. Slice D (the entity merge with the #2430 Group A renames) follows.

Part of #2313 / epic #2311.

### DIFF
--- a/crates/tidebreak-core/src/code/mod.rs
+++ b/crates/tidebreak-core/src/code/mod.rs
@@ -150,11 +150,21 @@ pub enum HarnessKind {
     Opencode,
     /// Grok CLI.
     Grok,
+    /// Tidebreak's own agent loop, run in-process behind the same adapter
+    /// contract as the external engines (decision 0048 step 5). A session
+    /// with no workspace selects it.
+    Internal,
 }
 
 impl HarnessKind {
     /// Every known engine, in adapter-tier order.
-    pub const ALL: &'static [Self] = &[Self::ClaudeCode, Self::Codex, Self::Opencode, Self::Grok];
+    pub const ALL: &'static [Self] = &[
+        Self::ClaudeCode,
+        Self::Codex,
+        Self::Opencode,
+        Self::Grok,
+        Self::Internal,
+    ];
 
     /// Stable database and wire token.
     #[must_use]
@@ -164,6 +174,7 @@ impl HarnessKind {
             Self::Codex => "codex",
             Self::Opencode => "opencode",
             Self::Grok => "grok",
+            Self::Internal => "internal",
         }
     }
 
@@ -185,7 +196,16 @@ impl HarnessKind {
             Self::Codex => HarnessTier::Secondary,
             Self::Opencode => HarnessTier::Tertiary,
             Self::Grok => HarnessTier::BestEffort,
+            Self::Internal => HarnessTier::Reference,
         }
+    }
+
+    /// Whether this engine runs inside the server process and spawns no
+    /// child. Such an engine probes as found with no binary, needs no pin,
+    /// and takes its inference from the server's own provider resolution.
+    #[must_use]
+    pub const fn is_in_process(self) -> bool {
+        matches!(self, Self::Internal)
     }
 }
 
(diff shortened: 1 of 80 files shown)


naingthet:
Don't merge I'm reconsidering this whole approach DO NOT MERGE

naingthet:
Superseded by #3010, which merged the same slice from the other worktree. Not merged, per the note above. The plan-accept mode switch and the SQLite migration marker fix move to a follow-up.

PR #3024, opened 2026-09-02
docs(decisions): amend 0048 to merge internals and keep two surfaces

naingthet:
Amends [decision 48](docs/decisions/0048-one-interaction-model.md) after the team discussion on 2026-09-01: the chat and code internals still merge onto the code-shaped structures, but the user-facing collapse into one surface is deferred, the `/chats/*` routes become aliases over the session routes rather than a flag day, and the internal engine as merged in #3010 is recorded as the transitional shape the entity merge retires.

Also updates the `docs/deferred.md` entry to match.

Part of #2313 / epic #2311. The entity merge (slice D) follows as its own stack.

### DIFF
--- a/docs/decisions/0048-one-interaction-model.md
+++ b/docs/decisions/0048-one-interaction-model.md
@@ -1,6 +1,6 @@
 # 48. One interaction model for chat and code
 
-- Status: Proposed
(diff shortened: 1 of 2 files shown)


naingthet:
**G2G** at 5184cc0ff
Depth: fast

PR #3045, opened 2026-09-02
fix(legal): resolve the UI notices closure for every platform

naingthet:
## Why

The v0.83.0 publish (run 33642928252) failed on four of five platform lanes with `legal/THIRD-PARTY-NOTICES.md is stale`. Each lane regenerated a different TypeScript 7 platform package: the checked-in file listed `@typescript/typescript-linux-x64`, macOS wanted `darwin-arm64`, Windows `win32-x64`, Linux arm64 `linux-arm64`.

The generator read the UI's production closure from whatever pnpm had installed on the host, and pnpm installs only the optional dependencies matching the host's `os`/`cpu`. TypeScript 7 is the first production-closure package (it reaches it as vue's peer) that ships one native build per platform, so the output started to depend on where it was generated. CI passed because CI and the PR's regeneration both ran on x86-64 Linux.

## What changes

- `scripts/generate-third-party-notices.mjs` installs the UI's production closure into a scratch copy of the project with `supportedArchitectures` set to every platform, then reads `pnpm licenses list` from there. Same lockfile, same bytes, on any host. This matches how the Cargo graph is already read with `--all-features`. Production only, so development dependencies' native variants add nothing.
- `legal/THIRD-PARTY-NOTICES.md` regenerated on macOS: the existing Linux x86-64 entry stays and the other 19 TypeScript variants join it. Nothing else moves.
- CI's `third-party notices` lane drops its now-redundant install step, and the path filter also watches `pnpm-workspace.yaml`, `.npmrc`, and `.pnpmfile.cjs`.
- `docs/releases.md` describes the platform-complete generation.

## Verification

- `node scripts/generate-third-party-notices.mjs` on macOS arm64 produced the checked-in file; `--check` reports up to date.
- `node --test scripts/*.test.mjs`: 206 pass.
- `actionlint .github/workflows/ci.yml` clean.
- The release lanes keep their check unchanged, so this PR's own notices lane plus the next publish run are the end-to-end proof.

After this lands, re-run the desktop publish for v0.83.0.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -235,8 +235,9 @@ jobs:
               crates/tidebreak-desktop/ui/src/generated/*)
                 rust=true; ui=true; workspace=true ;;
               # The UI production graph is an input to the notices; the rest of
-              # the UI tree is not.
-              crates/tidebreak-desktop/ui/package.json|crates/tidebreak-desktop/ui/pnpm-lock.yaml)
+              # the UI tree is not. pnpm-workspace.yaml carries the overrides
+              # and .npmrc the settings the generator resolves that graph with.
+              crates/tidebreak-desktop/ui/package.json|crates/tidebreak-desktop/ui/pnpm-lock.yaml|crates/tidebreak-desktop/ui/pnpm-workspace.yaml|crates/tidebreak-desktop/ui/.npmrc|crates/tidebreak-desktop/ui/.pnpmfile.cjs)
                 notices=true; ui=true ;;
               crates/tidebreak-desktop/ui/*) ui=true ;;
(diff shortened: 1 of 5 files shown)


naingthet:
**G2G** at d49d1b8e3
Depth: medium

Scratch `--prod` install with explicit `supportedArchitectures` (no `current`) plus reading `node_modules/.pnpm` instead of `pnpm licenses list` is the right fix for host-filtered optional natives. CI dropping the pre-install is consistent with the generator owning that tree. Path filter covering workspace yaml / `.npmrc` / `.pnpmfile.cjs` matches what the scratch copy actually consults.

PR #3046, opened 2026-09-02
fix(desktop): keep the composer draft across a Settings round trip

### LINKED ISSUE #3037: bug(desktop): preserve composer draft when navigating to Settings
## What happened?\n\nText entered in the conversation composer disappears after navigating to Settings.\n\nExpected: unsent text remains available when returning from Settings, so opening configuration does not destroy a draft.\n\nActual: the entered text does not persist across navigation.\n\n## Steps to reproduce\n\n1. Open a conversation.\n2. Enter text in the message composer without sending it.\n3. Navigate to Settings.\n4. Return to the conversation.\n5. Observe that the composer text is gone.\n\n## Tidebreak version\n\nCurrent desktop build used for Connected apps testing; exact version was not captured.\n\n## Platform\n\nmacOS (architecture not captured).\n\n## Acceptance criteria\n\n- An unsent composer draft survives navigation to and from Settings.\n- Drafts remain scoped to the correct conversation.\n- Sending or explicitly clearing a draft removes the persisted draft.\n- A regression test covers the Settings navigation round trip.\n\n## Diagnostics\n\nNo error message is shown.\n\n## Before submitting\n\n- [x] I searched existing issues for the same problem.\n- [x] I removed API keys, tokens, private document contents, and other sensitive data.

naingthet:
## What you change

You keep unsent composer text across a Settings round trip. Settings replaces the main pane and unmounts the conversation. Chat already stored drafts in `ComposerDrafts`. Code sessions kept the text in `useState`, so that unmount dropped it.

You write those code-session drafts through the same store, keyed by session id (or the start composer's workspace). One session's draft cannot appear in another. Sending or clearing still removes the persisted draft.

You add a Settings round-trip test for the conversation composer. It types, opens Settings, comes back, and checks the text is still there. It also checks conversation scoping and that send or `clearDraft` forgets the stored text.

## Why

Opening Settings must not destroy an unsent message.

## Cause

Settings unmounts the composer. Chat already persisted unsent text in the zustand draft store backed by sessionStorage, so a remount restored it. CodeComposer held the draft in component state instead, so the remount started empty. You do not add a second draft mechanism.

## What you ran

`pnpm --dir crates/tidebreak-desktop/ui exec biome format src`

```
Checked 866 files in 920ms. No fixes applied.
```

`pnpm --dir crates/tidebreak-desktop/ui lint`

```
Checked 866 files in 1164ms. No fixes applied.
```

`pnpm --dir crates/tidebreak-desktop/ui test`

```
Test Files  269 passed (269)
      Tests  2376 passed (2376)
   Duration  237.34s
```

`pnpm --dir crates/tidebreak-desktop/ui build`

```
tsc succeeded
✓ 13113 modules transformed.
✓ built in 9.67s
```

Closes #3037

### DIFF
--- a/crates/tidebreak-desktop/ui/src/ChatView.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/ChatView.dom.test.tsx
@@ -1,6 +1,23 @@
 // @vitest-environment jsdom
-import { act, cleanup, screen, waitFor, within } from "@testing-library/react";
+import {
+  act,
+  cleanup,
+  render,
+  screen,
+  waitFor,
+  within,
+} from "@testing-library/react";
 import userEvent from "@testing-library/user-event";
+import {
+  createMemoryHistory,
+  createRootRoute,
+  createRoute,
+  createRouter,
+  RouterProvider,
+  useNavigate,
+  useParams,
+  useRouter,
+} from "@tanstack/react-router";
 import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
 import type { ApiClient, Chat } from "./api";
 import { ChatView, type ChatViewProps } from "./ChatView";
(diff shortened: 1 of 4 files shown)


naingthet:
**NO GO** at 029d52105
Depth: medium
P2: First-turn recovery re-appends on remount. Settings unmounts CodeComposer, so `appliedRecoveryRef` resets while `recovery` is still set. The mount effect runs `appendRecoveredDraft` against the persisted draft. That helper only skips when the restored text equals the recovered prompt or already ends with it. An edit after a refused first turn comes back as the edit plus another copy of the recovered prompt.

naingthet:
**G2G** at 8064646ed
Depth: medium

PR #3048, opened 2026-09-02
fix(code): resume a parked turn after a worker restart

### LINKED ISSUE #3013: fix(code): resume a parked turn after a worker restart
Recovery closes a `Waiting` turn as `Interrupted` and leaves `park_ref`/`park_wait` on the row (`crates/tidebreak-core/src/db/ops/code/recovery.rs`). The internal engine (PR #3010) keeps its checkpoint durable, so a restarted worker could call `resume_turn` against it instead. Change recovery to re-attach and resume a parked turn when the engine declares `durable_parks`, and keep closing as interrupted for engines that do not.

naingthet:
## What changed

`resumes_parked_turn_after_restart` in `crates/tidebreak-core/src/db/ops/code/recovery.rs` decides between resume and interrupt. You declare the capability as `HarnessCaps::durable_parks` in `crates/tidebreak-core/src/code/caps.rs`: `Supported` on the internal engine, `Unsupported` on the external harnesses.

Recovery used to close every `Waiting` turn as `Interrupted` and leave `park_ref` / `park_wait` on the row. You now keep a waiting turn with a park open when the engine declares durable parks, settle the dead worker to idle, and re-attach a worker that calls the same `resume_turn` path a live park resolution uses. Engines without durable parks still close as interrupted.

The journal records `TurnResumed` so the transcript shows the turn resumed rather than restarted. Pending approvals on that waiting turn stay pending and rebind onto the new worker epoch so you can still decide them.

## Commands

`cargo fmt --all --check` printed nothing and exited 0.

`cargo clippy -p tidebreak-core -p tidebreak-server --all-targets -- -D warnings` and the recovery tests did not run here: cargo cannot fetch the pinned Symphonia git rev (`HTTP 403` / offline git db empty). CI runs those lanes.

## Tests

- A durable-park engine: a waiting turn stays waiting through recovery, the restarted worker resumes it, the park resolves, and the journal records `TurnResumed`.
- A non-durable engine: a waiting turn still closes as `Interrupted`.

Closes #3013


naingthet:
**NO GO** at 23e007c8b
Depth: medium

1. P1: `continue_parked_turn` calls `engine.resume_turn` after a restart, but `InternalSession::resume_turn` still requires in-memory `active_turn()`. `launch` always starts `active` at `None`, so the only engine that declares durable parks cannot resume the parked chat turn. The worker already journals `TurnResumed`, then the resume errors with `no parked turn to resume for …` and `close_open_turn` fails the waiting turn. The recovery test never hits this: it relaunches a `ScriptedAdapter` that already has `with_parked_turn`, not a fresh `InternalSession`. Restore the parked chat turn onto `active` at launch (or look it up in `resume_turn`) before claiming a worker restart can continue it.

naingthet:
Finding 1: you were right. `InternalSession::launch` now restores a waiting turn with a park onto `active` from the turn row, and `resume_turn` does the same lookup if `active` is still empty. The journal tail is the `last_seq` so you do not replay the park that already landed.

naingthet:
**G2G** at efba2041f
Depth: medium

naingthet:
Review comments on efba2041f:

1. `restore_parked_turn` now looks up the live `turn_run` for the chat and puts that id on `active`, not the code turn row id.

2. Recovery (and rebind) now drop `decision_claim` / `claimed_at` on the waiting turn's pending approvals so you can decide the park again after the worker that held the claim is gone.

naingthet:
**G2G** at f2f7d0a1d
Depth: medium

naingthet:
**G2G** at 30acebe08
Depth: medium

PR #3050, opened 2026-09-02
feat(code): accept image attachments on the internal engine

### LINKED ISSUE #3012: feat(code): image input on the internal engine
The internal engine (PR #3010, decision 48 step 5) declares `image_input: Unsupported`. Chat carries images through its attachment model (blob publication plus `resolve_message_attachments`); the adapter contract hands hydrated bytes on `TurnInput.images`. Bridge the two so a workspace-less session accepts image attachments: publish the bytes to the blob store, attach them to the admitted user message, and declare the capability. Trigger: before slice E collapses the chat UI onto the session controller, since chat users have images today.

naingthet:
## What changed

Declare `image_input` as supported on the internal engine so a workspace-less session offers image attachment. When a code-session turn carries images, you publish the bytes through chat's attachment model, attach them to the admitted user message, and hand them to the model request through the existing resolution path. External harnesses stay as they were.

## Why

Chat already carries images. The internal engine (decision 48 step 5) dropped them on submit and advertised `Unsupported`. You need this wired before chat's UI collapses onto the session controller.

## Path from the turn request to the model call

1. `publish_session_image` sniffs and stores the bytes (`inspect_image_bytes`, then `BlobStore::put`) and reserves session authority.
2. The turn route accepts the attachment ids and the session worker hydrates them with `hydrate_turn_images`.
3. `InternalSession::run_turn` calls `submit`, which calls `publish_turn_images`: `inspect_image_bytes` (chat's size and type limits), `BlobStore::put`, then `publish_chat_image_scoped`.
4. `accept_reserved_turn_with_message_context` records the images on the admitted user message.
5. The chat turn lane rebuilds the transcript with `list_message_attachments` and `user_message_for_model`, then `hydrate_images` puts the bytes on `ChatRequest.images`.

## Commands

`cargo fmt --all --check` — clean.

`cargo clippy -p tidebreak-core -p tidebreak-server --all-targets -- -D warnings` — this environment cannot fetch the pinned Symphonia git rev. With that patch dropped locally, clippy compiles the crates and reports two existing findings outside this change (`result_large_err` in `harness_llm.rs`, `chunks_exact_to_as_chunks` in `exec_write_snapshot.rs`).

`cargo test -p tidebreak-server --lib an_internal_turn_with_an_image_reaches` — `ok`. 1 passed.

`cargo test -p tidebreak-core attachment` — `ok`. 42 passed.

`cargo test -p tidebreak-server code_` — this sandbox blocks loopback TCP, so HTTP `serve()` tests cannot connect. The new coverage uses in-process `oneshot` instead.

No wire types changed, so TypeScript was not regenerated.

Closes #3012

### DIFF
--- a/crates/tidebreak-server/src/engine/internal/adapter.rs
+++ b/crates/tidebreak-server/src/engine/internal/adapter.rs
@@ -69,9 +69,9 @@ impl HarnessAdapter for InternalAdapter {
             reasoning_levels: CapLevel::Supported,
             native_file_change_events: CapLevel::Unsupported,
             native_interrupt: CapLevel::Supported,
-            // Chat's attachment model is the image path for the internal
-            // engine; hydrated bytes on the turn input are not wired yet.
-            image_input: CapLevel::Unsupported,
+            // Hydrated bytes on the turn input publish through chat's
+            // attachment model and reach the model request as image blocks.
+            image_input: CapLevel::Supported,
             slash_commands: CapLevel::Unsupported,
             durable_parks: CapLevel::Supported,
             user_questions: CapLevel::Supported,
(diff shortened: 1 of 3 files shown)


naingthet:
**G2G** at 5dc755dc0
Depth: medium

Capability flip lines up with session-worker delivery of hydrated bytes, and publish-then-admit follows chat ingest (inspect, blob write lock, put, scoped publication). Duplicate puts and publications are idempotent.

naingthet:
**G2G** at 2ab4c5dbd
Depth: medium

Capability flip still matches session-worker delivery of hydrated bytes. Publish-then-admit follows chat ingest (inspect, blob write lock, put, scoped publication), and duplicate puts stay idempotent.

PR #3052, opened 2026-09-02
fix(code): start sessions from the repository's real default branch

### LINKED ISSUE #3040: bug(code): first turn cannot start when the repository has no main ref
## What happened?\n\nA large first prompt was preserved, but it cannot succeed because Code session startup fails with an HTTP 400 while trying to resolve a Git ref named main.\n\nExpected: Tidebreak starts the Code session from the repository's configured default branch or selected base ref. If no valid base can be determined, it asks the user to choose one without preventing reuse of the prompt.\n\nActual: startup tries to resolve main^{commit}. When that ref does not exist, the first turn fails and Try again repeats the same failure, so the prompt cannot run.\n\n## Error shown\n\nHttpError: 400: could not resolve main: fatal: ambiguous argument 'main^{commit}': unknown revision or path not in the working tree. Use '--' to separate paths from revisions, like this: 'git <command> [<revision>...] -- [<file>...]'\n\nThe screenshot showed this error while files including Dockerfile and package.json were selected. The prompt was not discarded. The screenshot contained no credentials, but it is not uploaded here because it was supplied through a private local attachment path.\n\n## Steps to reproduce\n\n1. Start a new Code session for a repository where a local main ref cannot be resolved.\n2. Enter and submit a prompt.\n3. Allow workspace/session startup to resolve its base Git revision.\n4. Observe the HTTP 400 for main^{commit}.\n5. Select Try again.\n6. Observe that the preserved prompt still cannot start successfully.\n\n## Tidebreak version\n\nCurrent desktop build; exact version was not captured.\n\n## Platform\n\nmacOS (architecture not captured).\n\n## Acceptance criteria\n\n- Code session startup uses the repository's actual default branch or explicitly selected base ref rather than assuming main exists.\n- Remote-only default refs and repositories using another branch name are handled.\n- If no valid base ref can be resolved, the UI identifies the missing ref and lets the user select or repair the base before retrying.\n- Try again can succeed after the base ref is corrected without requiring the prompt to be re-entered.\n- A regression test covers a repository with no main branch/ref.\n- The error message distinguishes an invalid base revision from a prompt or model failure.\n\n## Before submitting\n\n- [x] I searched existing issues for the same problem.\n- [x] I removed API keys, tokens, private document contents, and other sensitive data.

naingthet:
## Why

Code session startup assumed a `main` ref. When a repository uses another default branch, the first turn fails with HTTP 400 while git tries `main^{commit}`, and Try again repeats the same failure with the prompt still sitting there.

## What changed

Startup now resolves the base ref from the checkout.

`crates/tidebreak-server/src/code/worktree.rs` owns the resolver. `crates/tidebreak-server/src/code/runtime/repos.rs` and `crates/tidebreak-server/src/code/runtime/workspaces.rs` persist the ref you actually used. `crates/tidebreak-server/src/code/runtime/mod.rs` maps a miss to the typed `missing_base_ref` error. `crates/tidebreak-core/src/code/mod.rs` documents that the stored default comes from the repository when you do not set one. `crates/tidebreak-server/src/tests/code_create.rs` pins the HTTP path.

Candidates, in this order:

1. The explicitly configured base ref
2. The target of `refs/remotes/origin/HEAD` (`git symbolic-ref`)
3. `main`
4. `master`
5. The current branch

Each candidate is checked with `git rev-parse --verify --quiet <ref>^{commit}`. If the local branch is missing, `origin/<name>` is accepted. The workspace records that resolved ref, and the repo default is updated when you asked for the stored default.

When nothing resolves, you get `missing_base_ref` naming the ref you looked for and the candidates that were tried. That is distinct from a prompt or model failure. Workspace create still fails before a session starts, so the composer keeps the prompt and Try again works once the base is reachable.

## Tests

- `git init -b trunk` with one commit: register, create, and first turn start from `trunk` even when the stored default is `main`
- Empty `git init -b trunk` with no commit: workspace create returns `missing_base_ref` and names `main`
- Unit tests for origin-only refs, origin/HEAD over `main`, and the missing-ref error

## Commands

`cargo fmt --all --check` is clean.

`cargo test -p tidebreak-server --lib default_base_ref`: 4 passed.

`cargo test -p tidebreak-server --lib code::worktree::`: 29 passed.

`cargo clippy -p tidebreak-server -p tidebreak-core --all-targets -- -D warnings` compiles this change. This sandbox rustc is 1.98.0; CI pins 1.97.1. On 1.98 two existing lints fire in files this PR does not touch (`clippy::result_large_err` in `harness_llm.rs`, `clippy::chunks_exact_to_as_chunks` in `exec_write_snapshot.rs`). Allowing those two sandbox-only lints, clippy finishes clean.

`cargo test -p tidebreak-server code_create` cannot connect to `127.0.0.1` in this sandbox (`tcp connect error`, timed out). Existing `code_create` tests fail the same way. The new HTTP tests share that fixture; the worktree unit tests cover the resolver and the missing-ref error.

Closes #3040

### DIFF
--- a/crates/tidebreak-core/src/code/mod.rs
+++ b/crates/tidebreak-core/src/code/mod.rs
@@ -757,6 +757,8 @@ pub struct CodeRepo {
     /// Display name.
     pub display_name: String,
     /// Default base ref for new workspaces.
+    ///
+    /// Resolved from the repository when you do not set one.
     pub default_base_ref: String,
     /// Prefix applied to workspace branch names.
     pub branch_prefix: String,
--- a/crates/tidebreak-server/src/code/runtime/mod.rs
+++ b/crates/tidebreak-server/src/code/runtime/mod.rs
@@ -813,6 +813,9 @@ fn map_worktree(err: WorktreeError) -> ServerError {
                 ServerError::bad_request_kind("worktree", message)
             }
         }
+        missing @ WorktreeError::MissingBaseRef { .. } => {
+            ServerError::bad_request_kind("missing_base_ref", missing.to_string())
+        }
         WorktreeError::Internal(message) => ServerError::internal(message),
         WorktreeError::ArchiveUncertain(message) => {
             ServerError::conflict_kind("archive_inspection_uncertain", message)
--- a/crates/tidebreak-server/src/code/runtime/repos.rs
+++ b/crates/tidebreak-server/src/code/runtime/repos.rs
@@ -53,14 +53,27 @@ impl CodeRuntime {
                     .unwrap_or_else(|| "repo".into())
             });
         let default_branch_prefix = self.default_branch_prefix(owner).await;
+        let configured_base = default_base_ref
+            .map(|value| value.trim().to_owned())
+            .filter(|value| !value.is_empty());
+        let default_base_ref = match worktree::resolve_default_base_ref(
+            &validated.toplevel,
+            configured_base.as_deref(),
+        )
+        .await
+        {
+            Ok(resolved) => resolved,
+            Err(WorktreeError::MissingBaseRef { .. }) => {
+                configured_base.unwrap_or_else(|| "main".into())
+            }
+            Err(other) => return Err(map_worktree(other)),
+        };
         let repo = CodeRepo {
             id: RepoId::new(),
             owner: owner.clone(),
             root_path: toplevel,
             display_name: name,
-            default_base_ref: default_base_ref
-                .filter(|value| !value.trim().is_empty())
-                .unwrap_or_else(|| "main".into()),
+            default_base_ref,
             branch_prefix: branch_prefix
                 .filter(|value| !value.trim().is_empty())
                 .unwrap_or(default_branch_prefix),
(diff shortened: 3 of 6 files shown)


naingthet:
**NO GO** at 901cb1ad9
Depth: medium

1. P2: `create_workspace` runs every requested `base_ref` through `resolve_default_base_ref`, so a missing non-default branch still falls through origin HEAD, `main`, `master`, and the current branch. Persistence already treats only “no explicit ref, or the stored default” as the case that should rewrite `default_base_ref`, but the workspace is still created from the fallback. An explicit `feature-x` that does not exist therefore starts on `trunk` instead of returning `missing_base_ref`. Resolve the default chain only for that default case; pass any other requested ref through `resolve_ref` (or an equivalent that does not walk sibling default names).

naingthet:
Finding 1 is fixed. Workspace create walks the default-branch candidates only when you omit a base ref or ask for the stored default. An explicit missing ref now goes through resolve_named_base_ref, returns missing_base_ref, and does not start on trunk. Covered by named_base_ref_does_not_fall_through_to_the_current_branch and workspace_create_does_not_fall_through_an_explicit_missing_ref.

naingthet:
**G2G** at e84a9f44a
Depth: medium

PR #3054, opened 2026-09-02
refactor(server): move the chat turn worker under the internal engine

naingthet:
The chat turn worker lives under the internal engine. Slice D4a0 of #2313 (decision 48 step 5); follows D3 (#3042). No behavior change.

## What changed

`crates/tidebreak-server/src/turn_worker.rs` is now `crates/tidebreak-server/src/engine/internal/leg.rs`. The type is `LegDriver` (config and outcome follow). Bind, tests, and `docs/how-tidebreak-works.md` point at the new path. The process-wide claim loop, leases, and quiesce path are the same.

This exists so D4a's retarget is a reviewable diff instead of a 3,900-line move hiding inside the turn-row merge.

## Checks

- `cargo fmt --all --check`
- `cargo check -p tidebreak-server --all-targets`
- `cargo clippy -p tidebreak-server --all-targets -- -D warnings` (the only hit is the pre-existing `too_many_arguments` in `tidebreak-code-execution`, which CI allows)

Part of #2313 / epic #2311.

### DIFF
--- a/crates/tidebreak-server/src/engine/internal/leg.rs
+++ b/crates/tidebreak-server/src/engine/internal/leg.rs
@@ -47,7 +47,7 @@ use crate::retry::{LaneBackoff, RetryAttempt, RetrySchedule};
 use crate::state::{BlobWriteGuard, TurnGuard};
 
 #[derive(Debug, Clone, Copy)]
-pub(crate) struct TurnWorkerConfig {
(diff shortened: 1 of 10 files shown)


naingthet:
**G2G** at 9b501dac6
Depth: medium

PR #3055, opened 2026-09-02
feat(desktop): show live activity on the workspace rail

naingthet:
## Summary

The workspace rail row now says what a code session is doing instead of how many turns it has run.

- **Live activity text.** The session digest carries a new `activity_detail`: the subject of the tool the live turn is waiting on (the command, path, query, or task description), bounded to one line. A running row shows "cargo test" rather than "Shell running". When the turn is parked, the row shows the recap the server already derives.
- **Subagents unfold on the row.** A session with subagents gets a chevron. Expanding it lists each subagent as a child line with its own running, completed, or failed state.
- **One glyph set.** Every row state draws from the attention mark vocabulary the rail already uses: spinner while working, bot while waiting on subagents, radar while monitoring background work, check for done-unreviewed, alert for needs-you, clock for stalled.
- **Turn count dropped** from the row. It is a tally, not a state.
- **Stall sweep skips parked sessions.** A session whose activity is `monitor` or `subagents` is silent by design. The sweep and `compute_attention` no longer flip it to Stalled, which used to contradict the "Monitoring" label beside the badge.

## Testing

- `cargo test -p tidebreak-server` (attention unit tests, code doctor integration test for the stall skip, wire type and fixture currency).
- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets --locked -- -D warnings` (with `too_many_arguments` allowed locally: the one hit is a pre-existing macOS-only function CI never compiles).
- Desktop UI: biome, `tsc --noEmit`, `vitest run src/code src/stylesContract.test.ts` (85 files, 1140 tests), `storybook:build`.
- Storybook stories for `WorkspaceCard` cover running with detail, monitoring, subagents folded and unfolded, and done-unreviewed. Screenshots reviewed in all four themes.


naingthet:
**NO GO** at e365124d6
Depth: medium

1. P2: Live digest notices drop `activity_detail`, so the rail never keeps the tool subject this PR adds. `CodeSessionDigest` and `CodeUpdateNotice` carry the field, and `parseCodeUpdateNotice` accepts it, but `noticeToAction` in `crates/tidebreak-desktop/ui/src/code/CodeUpdatesStore.ts` still builds the store digest without it. A snapshot can show `cargo test -p …`; the next `/code/updates` digest replaces that row with `activity` only, and `sessionActivityLineLabel` falls back to "Shell running". Copy `activity_detail` the same way as `activity`.

naingthet:
**G2G** at d34e4b3f3
Depth: medium

PR #3057, opened 2026-09-02
feat(models): curate Claude Fable 5.1 and price every first-party row

naingthet:
## What changes

- `model_registry.rs`: Claude Fable 5.1 is the recommended Anthropic row, ahead of Fable 5. The default stays Opus 5. Gemini 3.8 Flash and 3.7 Flash are curated, 3.8 recommended, both starting the effort ladder at `low`.
- `anthropic.rs`: Fable 5.1 and Mythos 5.1 get their new contract. Structured output goes through `output_config.format` instead of a forced tool. A forced `tool_choice` fails with an error that names the model. Replayed thinking blocks opt into `drop_block` under the `thinking-binding-controls-2026-08-01` beta header. Older rows are untouched.
- `analytics.rs`: every curated Anthropic, OpenAI, xAI, and Gemini row now has a price. GPT-5.6 Sol moves to its published promotional rate, and the GPT-5.6 rows carry cache-write rates. Fireworks and Together stay unpriced on purpose, since they price per deployment.
- The StartSession story lists Fable 5.1.

## How it is guarded

- `every_curated_first_party_row_has_a_price` walks the four first-party providers, so a curated row without a price fails CI.
- `provider_prefixed_ids_reach_their_own_row` pins that a vendor-prefixed Flash-Lite id lands on its own row rather than the Flash row.
- Adapter tests cover the native output constraint, the refused forced tool choice, the block-binding field, and the family-plus-generation gate that keeps Opus 5 on the old contract.

## Checks run

rustfmt, workspace clippy, 53 router tests, 25 registry and analytics tests, biome on the story.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/stories/StartSession.stories.tsx
+++ b/crates/tidebreak-desktop/ui/src/stories/StartSession.stories.tsx
@@ -32,8 +32,8 @@ import { harnessDoctor, harnessDoctorDegraded } from "./fixtures";
 const models: Partial<Record<HarnessKind, ParsedHarnessModel[]>> = {
   claude_code: [
     {
-      id: "claude-fable-5",
-      label: "Claude Fable 5",
+      id: "claude-fable-5-1",
+      label: "Claude Fable 5.1",
       default: true,
       reasoning_efforts: [],
       fast_mode: true,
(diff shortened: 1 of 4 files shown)


naingthet:
**G2G** at 66dc6bf6e
Depth: medium

PR #3060, opened 2026-09-02
ci(release-draft): cancel superseded push runs

naingthet:
## What

Set the release-draft workflow's `cancel-in-progress` to true for push runs, not only `pull_request_target`. Manual dispatches still queue.

## Why

Every run of the draft job regenerates the draft from scratch, so when several merges land within a minute only the newest run matters. Today the earlier runs still complete, and each one spends a slice of the repository's shared `GITHUB_TOKEN` budget (1,000 requests per hour on the Team plan, shared by every workflow run in the repo).

On 2026-09-02 three merges landed between 17:15 and 17:16 UTC during a busy hour with about 180 workflow runs, and the drafter's final "update release" call failed with `API rate limit exceeded for installation` on all three ([run 33660072987](https://github.com/brightwave-inc/tidebreak/actions/runs/33660072987/job/100348166324)).

A cancelled run can leave an extra draft behind briefly. The next run's "Keep exactly one native release draft" step already reconciles that.

## Follow-ups

Moving the draft job onto a dedicated GitHub App token, so it stops sharing the Actions budget, is tracked separately.


naingthet:
**NO GO** at 623df72b7
Depth: medium

1. P2: `cancel-in-progress` is evaluated on the incoming run, not the one already in flight. `push` and `workflow_dispatch` share `release-draft-${{ github.ref }}` on main, and a push evaluates the new expression to true, so a merge cancels an in-progress backfill. The comment claims the opposite. An `apply: true` dispatch can die mid-loop; the push job never resumes that backfill. Either put dispatches in a separate concurrency group, or drop the unenforceable claim and accept that merges preempt them.

PR #3069, opened 2026-09-02
feat(memory): inject the digest, add the memory tool, and capture proposals in work mode

### LINKED ISSUE #2554: Work-mode memory: injection, explicit tool, turn capture
Part of #2551.

Memory section in the foreground prompt, digest pinned per conversation and refreshed only at prefix-rebuilding boundaries; an explicit memory tool in the internal registry; post-completion capture on the utility model producing proposals and tracked hypotheses; transcript proposal chips with inline approve/edit/dismiss; per-chat incognito. Includes the prompt-fingerprint test proving a mid-conversation memory change leaves the composed prompt byte-identical.

naingthet:
Implements the work-mode memory slice: digest injection with a per-conversation pin, the explicit memory tool, post-turn capture, transcript proposal chips, and per-chat incognito.

## What changed

- **Injection.** `compose_for_surface` gains a `## Memory` section, last among the stable sections, rendered from `MemoryBackend::assemble_context`. The turn worker pins one render per conversation and re-renders only at prefix-rebuilding boundaries (new conversation, compaction checkpoint, model switch), so a record changing mid-conversation never invalidates the prompt cache. An empty digest pins too, for the same reason.
- **The fingerprint test.** `a_mid_conversation_memory_change_leaves_the_composed_prompt_byte_identical` runs real turns against a prompt-recording provider, mutates the store mid-conversation (create, status change), and asserts the next turn's composed prompt is byte-identical; after a model switch it asserts the new digest appears. An implementation that re-renders per turn passes every functional test and fails only this one.
- **Explicit tool.** A `memory` tool (verbs `propose`, `search`, `read`) with its spec and parser in `tidebreak-core` and its executor in the server. `propose` writes a `proposed`, model-authored, evidence-backed record — never active — so the review lifecycle is the consent gate; the tool is read-only by consent class, excluded from the plan-mode surface by name like `update_task_plan`, and withheld (and double-checked at execution) for incognito chats.
- **Turn capture.** `LegDriver::process` gains work mode's first post-completion hook: `MemoryCapture` spawns fire-and-forget on `Completed`, resolves the utility model, and derives one structured candidate through `derive_text_with_retries` with a new `Proposal` implementation. Confident candidates land as `proposed`; weak signals land as `tracking` hypotheses with observation counts, graduate to review only after repeating in a distinct conversation, and never enter the digest. Titles already stored or recently rejected are never re-proposed. Capture defaults off behind the existing `memory.capture.enabled` setting.
- **Proposal chips.** Terminal turn snapshots carry the turn's model-authored records, a `memory_proposals_recorded` metadata notice triggers re-hydration, and the transcript renders a `ToolCardShell`-based proposal card with inline approve, edit (full-envelope CAS update), and dismiss, plus Storybook stories over shared typed fixtures and a jsdom test.
- **Per-chat incognito.** A `memory_incognito` column on the conversation row via appended migration `m20260902_000005_chat_memory_incognito`, settable at `POST /chats` and `PATCH /chats/{id}`, surfaced as a switch in the composer tools menu. Incognito suppresses injection, the tool, and capture for that chat.
- Two historical migrations read `code_session` through the live entity, which breaks the chain for any appended column; they now name the columns they need, keeping the appended-only rule workable.

## How I verified it

- `cargo fmt --all -- --check`; `cargo clippy -p tidebreak-server -p tidebreak-core --all-targets` (only two pre-existing warnings in untouched files).
- `cargo test -p tidebreak-core --lib` for the memory, migration (full chain, 21 tests), db, registry, and agent suites; `cargo test -p tidebreak-server --lib` for the memory suite (fingerprint, incognito, capture thresholds, tool verbs, transcript proposals) plus conversations, workers, compaction, and documents.
- UI: `tsc --noEmit`, biome lint, `vitest run` (271 files / 2430 tests), and `storybook:build`, all clean.
- Wire types regenerated through the pinned generator test and synced to mobile.

## Notes for review

- The digest pin is in-memory per worker with LRU shedding at 512 conversations; a shed or restart re-renders at the same boundary key, which only changes bytes if records changed meanwhile. Making the pin durable was not worth a schema row for this slice.
- The memory verb is advertised only when a digest composes (backend present, memory on, owner resolvable, not incognito), so the prompt never describes a verb the turn would refuse.

Closes #2554.


naingthet:
@ShipRight review

naingthet:
**NO GO** at abfa3afac
Depth: medium

P2: Hypothesis graduation writes a reviewable proposal, then announces it on the wrong conversation. `observe_hypothesis` keeps the original `origin.chat_id` / `turn_id` and only flips status to `proposed`. `derive` then publishes `MemoryProposalsRecorded` on the *current* chat. `list_chat_messages` only attaches records whose origin chat matches the transcript being loaded, so the confirming turn never gets the proposal chip and its live metadata refresh finds nothing. The card lands on the first tracking conversation instead, if that chat is reloaded at all. Either retarget origin (and evidence) to the graduating turn, or hydrate/announce against the origin chat — the two paths have to agree.

naingthet:
Addressed in ea89be6fb: graduation retargets the record origin to the graduating chat and turn and appends that turn's evidence, so the announcement and the transcript attach on the same conversation. The graduation test asserts the new origin and both evidence entries.

naingthet:
**NO GO** at ea89be6fb
Depth: medium

P2: `memory` is still advertised (and, when memory is off, still executable) on turns that compose no digest. `freeze_foreground_turn_surface_with_folders` drops the verb from the specs it feeds `compose_for_surface`, but `Agent::foreground_tool_specs` rebuilds the wire list from the full registry and never strips `MEMORY_TOOL` — unlike `web_search`, which sets `agent_config.web_search` so advertisement and the prompt stay aligned. Incognito and memory-off turns therefore still send the schema. Execute double-checks only incognito, so a globally disabled memory switch still lets `search`/`read`/`propose` run. Withhold the tool from `foreground_tool_specs` (or unregister it on the frozen registry) and refuse it at execute when the digest would not compose.

naingthet:
**G2G** at ea89be6fb
Depth: medium

naingthet:
Addressed in 53fd06804: an agent-config memory_tool flag set by the frozen surface withholds the verb from the wire list whenever no digest composes, and the executor now refuses when the global memory switch is off, not only for incognito chats. A leg test pins the wire and prompt agreeing.

naingthet:
**G2G** at 0994e95d9
Depth: medium

PR #3074, opened 2026-09-02
feat(core): one turn lane for chat and code

naingthet:
Closes nothing on its own. Slice D4a of #2313 (decision 48 step 5).

Chat and code now share one turn row, one durable lease, and one worker family.

`turn_claim` is renamed to `code_turn_claim`, not retired. Seven composite foreign keys make append idempotence and heartbeat fencing a property of the schema. Folding the claim into `code_turn` would drop that.

The admission ledger is gone. Identity lives on the turn row as `fingerprint`. `postgres_turn_admission_serializes_processes_and_recovers_expiry` is deleted with that ledger; the lease/expiry half has no subject once a session has one consumer. Concurrent same-id submits still converge on one row (`Existing` / `IdentityConflict`).

`turn_run.agent_run_id` is dropped. The coordinator run is per-session (`ensure_foreground_agent_run`), so the column was the same value on every turn of a chat.

`InternalSession::run_turn` drives `LegDriver::run_turn` under the session worker's lease. `drive_turn_inner` claims before the leg and heartbeats alongside it. Client waits still release the lease until D4b: a wait ends the leg, drops the lease, and returns the turn to the claim scan. Do not pin a worker on a client tool call.

`HarnessCaps::mid_turn_resume` is a new flag every adapter had to answer. Internal is `Supported`; every harness is `Unsupported`. Boot recovery skips a pid-less internal session so it does not close a live chat turn as interrupted. Both 0080 quiesce halves stay wired until D4b.

The queued-turn promoter is an attach sweep: sessions with a queued turn and no worker get `attach_and_spawn_worker`. Promotion itself stays in `drain_queued`.


naingthet:
@ShipRight review

naingthet:
@ShipRight review

naingthet:
**NO GO** at e8460fe12
Depth: medium

1. P1: `InternalSession::run_turn` loads the row through `Store::get_turn`, which returns `None` when `input_message_id` is missing. The session worker inserts a `code_turn` with no message, takes a lease, then passes that id in `TurnInput`. The internal engine therefore errors "was not claimed before the leg" on the path this PR says now drives the chat loop. `LegDriver::run_turn` cannot run a code-shaped row.

2. P1: The migration stores chat `cancelled` as `interrupted`, and the code worker writes `interrupted` for the same terminal. `delete_chat`, `list_terminal_turns`, and `list_cancelled_output_message_ids` still match only `TurnRunStatus::Cancelled` (`cancelled`). After upgrade, a chat whose last turn was cancelled looks busy and cannot be deleted. Those turns also drop out of terminal history and cancelled-output annotation.

3. P2: `take_lease_on_turn` refuses a `running` row that still has any `lease_token`, including an expired one. Boot recovery skips pid-less internal sessions so a claim scan can resume them, but `claim_turn` ignores rows without `input_message_id` and session-worker attach only continues `waiting` parks. A crashed internal turn stays `running`.

naingthet:
**NO GO** at 0c4514718
Depth: medium

1. P1: `get_turn` now returns a `code_turn` with no `input_message_id`, but that is not enough to run the leg. `drive_turn_inner` still inserts that row and ignores `TurnInput.text`; `LegDriver::run_turn` still calls `Agent::run_claimed_turn`, which rebuilds the prompt from the message table. A queued chat follow-up that the attach sweep hands to the session worker, or any Internal send, therefore runs without the new user text. The engine answers the previous transcript, or an empty one.

2. P2: Boot recovery expires the Internal lease, then re-attaches a session worker that resumes a `Running` turn without taking a new lease. `run_turn` reuses the expired token, or, if the claim scan already reclaimed, the live claim-loop token. The first case is `LeaseLost`; the second is two `LegDriver`s on one lease.

naingthet:
**NO GO** at 39238cfc1
Depth: medium

1. P1: After an Internal client or agent-run wait, `InternalSession` maps the leg to `TurnOutcome::Clean` and `drive_turn_inner` leaves the row `waiting_for_client` / `waiting_for_agent_run`, then sets the session `Idle`. A follow-up send or `drain_queued` inserts another `running` row with `input_message_id` NULL, which bypasses `idx_code_turn_one_active`. `ensure_turn_input_message` then unique-conflicts, so that row stays `running` with no lease and no message. Claim scan will not pick it up (`expired_turn_candidate` requires a lease expiry).

2. P2: `ensure_turn_input_message` writes only the text. `InternalSession::run_turn` still ignores `TurnInput.images`, and `LegDriver` rebuilds the prompt from `message`. An Internal turn with images runs without them.

naingthet:
**NO GO** at 62231dd5d
Depth: medium

1. P1: `drive_turn_inner` inserts a `running` Internal row with `input_message_id` NULL, then `ensure_turn_input_message` writes the transcript before `take_lease_on_turn`. `code_turn_lease_check` requires a live lease whenever `input_message_id` is set and status is `running`/`cancelling`. The update fails, the row stays `running` with no lease and no message, and `claim_turn` will not pick it up (`expired_turn_candidate` needs a lease expiry).

2. P1: After an Internal client or agent-run wait, `InternalSession` maps the leg to `TurnOutcome::Clean` and `drive_turn_inner` leaves the row `waiting_for_client` / `waiting_for_agent_run`, then sets the session `Idle`. A follow-up send inserts another `running` row with `input_message_id` NULL, which bypasses `idx_code_turn_one_active`. `ensure_turn_input_message` then unique-conflicts, so that row also stays `running` with no lease.

3. P2: `ensure_turn_input_message` writes only the text. `insert_turn` stores images with `message_id` NULL, `InternalSession::run_turn` ignores `TurnInput.images`, and the adapter `LegDriver` is constructed without blobs. An Internal turn with images runs without them.

PR #3075, opened 2026-09-02
feat(connected-apps): allow explicitly approved loopback HTTP APIs

### LINKED ISSUE #3063: feat(connected-apps): allow explicitly approved loopback HTTP APIs
## What are you trying to do?\n\nConnect Tidebreak to an API or MCP-compatible service running locally on the same computer, such as a service bound to 127.0.0.1.\n\n## What happens today?\n\nThe Connected app REST form rejects a loopback base URL such as:\n\nhttp://127.0.0.1:23373/v0/mcp\n\nwith:\n\n400: connected app base URL is not admissible: scheme must be https\n\nRequiring HTTPS is appropriate for remote hosts, but many local development and desktop companion services expose plain HTTP only on the loopback interface. There is currently no clear supported path for that case.\n\n## What would a good outcome look like?\n\nTidebreak permits plain HTTP only for tightly validated loopback destinations, with explicit user consent and the existing connected-app capability boundaries. Remote and non-loopback HTTP destinations should remain rejected.\n\nThe UI should also clarify whether a supplied endpoint is a REST/OpenAPI service or an MCP endpoint and route the user to the appropriate setup flow.\n\n## Acceptance criteria\n\n- Explicit loopback hosts such as 127.0.0.1 and IPv6 loopback can be configured without TLS after clear consent.\n- Hostname resolution cannot be used to bypass the loopback restriction or create an SSRF path.\n- Redirects and subsequent requests cannot escape the approved loopback destination.\n- Non-loopback HTTP URLs remain rejected.\n- The UI distinguishes REST/OpenAPI base URLs from MCP endpoints.\n- The security posture and supported local-server configurations are documented and tested.\n\n## Additional context\n\nThe screenshot shows a Beeper configuration rejected solely because the loopback URL uses HTTP. The private screenshot is not uploaded to this public issue.

naingthet:
Closes #3063

## Security model

Plain HTTP is admitted only when all of the following hold:

- The host is an **IP literal** in `127.0.0.0/8` or exactly `::1`. `localhost` and every other DNS name are refused for `http` with a message that says to use `127.0.0.1` or `[::1]`, so hostname resolution cannot widen the exemption.
- The `rest_api` record carries `allow_loopback_http: true`. Without the flag, admission names `allow_loopback_http` in the refusal. The flag is persisted on the definition and is part of the consent fingerprint (`v:3`), so toggling it re-prompts.
- The exemption applies only to that record's own admitted origin (scheme, host, port). Every executed request is pinned to it.

https keeps today's behavior, including the denied-network list for IP literals (loopback https is still refused). Remote and RFC1918 `http` (for example `http://10.0.0.1`) still fail with `scheme must be https`. Managed profiles still refuse REST connected-app upserts.

Redirects: the reqwest client already uses `Policy::none()` for operation execution (https and loopback-http alike), so a 302 is reported and never followed. Spec fetches still follow https hops with per-hop re-admission; loopback-http spec fetches do not follow redirects, so a 302 cannot walk the opt-in off that origin.

Connected-app traffic does not pass a second policy layer in `tidebreak-egress`; admission stays in `rest_executor`.

## What changed

- `admit_base_url` / spec fetch take `allow_loopback_http` and admit loopback HTTP IP literals only under that flag.
- `RestApiDefinition` and the upsert body persist the flag; listing projects it so the form can re-show consent.
- Executor and spec-fetch clients set `https_only` from the URL scheme; loopback-http still uses `no_proxy` and `Policy::none()`.
- Settings form: consent checkbox (warning surface, Save disabled until checked) when the typed base URL is loopback HTTP; MCP hint when the path ends in `/mcp` or a preview ingest looks like a non-OpenAPI document. Story: `ConnectedAppsLoopbackHttpConsent`.
- Docs: "Local services on this computer" in `docs/connected-apps.md`.

## Commands run

```
cargo fmt --all
cargo check -p tidebreak-server --tests
cargo test -p tidebreak-server rest_executor
cargo test -p tidebreak-server connected_apps
cargo clippy -p tidebreak-server --tests -- -D warnings
UPDATE_WIRE_TYPES=1 cargo test -p tidebreak-server --lib wire_types
```

`cargo fmt --all` and `cargo check -p tidebreak-server --tests` succeeded (the sandbox comments the root `symphonia` git patch locally and restores `Cargo.toml` / `Cargo.lock` before every commit).

`cargo test -p tidebreak-server rest_executor` (lib): 13 passed, including admission cases for `127.0.0.1`, `127.9.9.9`, `[::1]`, `localhost`, `10.0.0.1`, and http without the flag.

`cargo test -p tidebreak-server connected_apps`: `loopback_http_rest_app_requires_explicit_consent` passed; `the_listing_carries_mounted_tool_names_only` returned 400 on `PUT /mcp/servers` for a loopback MCP URL in this environment (loopback TCP is often dropped here). Fingerprint and upsert tests passed.

`cargo clippy -p tidebreak-server --tests -- -D warnings` failed on existing lints in `harness_llm.rs` (`result_large_err`) and `exec_write_snapshot.rs` (`chunks_exact_to_as_chunks`), not on this change.

UI: `CI=true pnpm install --frozen-lockfile` failed with `ERR_PNPM_FETCH_403` from `https://cdn.sheetjs.com/xlsx-0.20.3/xlsx-0.20.3.tgz`, so biome, lint, test, build, and `pnpm storybook:build` could not run here.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/api/client/apps.ts
+++ b/crates/tidebreak-desktop/ui/src/api/client/apps.ts
@@ -55,6 +55,7 @@ export function withAppsApi<TBase extends Constructor<HttpCore>>(Base: TBase) {
          * the document is not judged. */
         operation_ids?: string[];
         credential: RestCredentialUpdate;
+        allow_loopback_http?: boolean;
       },
     ): Promise<ConnectedAppsInfo> {
       return this.json(`/connected-apps/rest/${encodeURIComponent(id)}`, {
@@ -67,11 +68,15 @@ export function withAppsApi<TBase extends Constructor<HttpCore>>(Base: TBase) {
     /** List what an OpenAPI document declares, for the operation picker. */
     previewRestSpec(
       source: { url: string } | { document: string },
+      options?: { allow_loopback_http?: boolean },
     ): Promise<SpecPreviewInfo> {
       return this.json("/connected-apps/rest/spec-preview", {
         method: "POST",
         headers: this.headers(true),
-        body: JSON.stringify({ source }),
+        body: JSON.stringify({
+          source,
+          allow_loopback_http: options?.allow_loopback_http ?? false,
+        }),
       });
     }
 
--- a/crates/tidebreak-desktop/ui/src/generated/wire.ts
+++ b/crates/tidebreak-desktop/ui/src/generated/wire.ts
@@ -2272,7 +2272,12 @@ placement: CredentialPlacement | null, updated_at: string,
  * a count only, never app names or ids, the renderer-safety posture
  * of this surface. Grants of library-deleted apps do not count.
  */
-used_by_app_count: number, };
+used_by_app_count: number,
+/**
+ * Whether this record opted into loopback HTTP. Surfaced so the
+ * form can re-show consent on edit.
+ */
+allow_loopback_http: boolean, };
 
 /**
  * The renderer's listing of every configured connected app, across kinds.
--- a/crates/tidebreak-desktop/ui/src/settings/ConnectedAppsPanel.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/settings/ConnectedAppsPanel.dom.test.tsx
@@ -76,6 +76,7 @@ const listing: ConnectedAppsInfo = {
       placement: "bearer",
       updated_at: "2026-08-03T00:00:00Z",
       used_by_app_count: 1,
+      allow_loopback_http: false,
     },
   ],
 };
@@ -298,6 +299,7 @@ describe("ConnectedAppsPanel", () => {
           base_url: "https://api.example.com/v2",
           openapi_document: '{"openapi": "3.0.3"}',
           credential: { set: { value: SECRET, placement: "bearer" } },
+          allow_loopback_http: false,
         },
       ),
     );
@@ -379,6 +381,7 @@ describe("ConnectedAppsPanel", () => {
           document_sha256: "cd".repeat(32),
           operation_ids: ["listOrganizations"],
           credential: "none",
+          allow_loopback_http: false,
         },
       ),
     );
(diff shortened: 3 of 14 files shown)


naingthet:
@ShipRight review

naingthet:
**NO GO** at e388cfaa1
Depth: deep

1. P2: Spec-fetch HTTPS redirects still re-admit with `allow_loopback_http`, so a remote document URL can 302 onto `http://127.0.0.1:<other-port>` and the fetch follows it. Loopback-http originals already refuse redirects so the opt-in cannot leave that origin; HTTPS hops need the same pin (or must re-admit without the flag). Execution stays origin-pinned; this is config-time SSRF onto a different loopback origin than the operator typed.
2. P2: Fetch operations is not gated on the consent checkbox. URL-source save requires a successful preview first, so the first action on a loopback document URL 400s with `allow_loopback_http` unless the box is already checked. Disable Fetch until consent when the document URL is loopback HTTP, and surface that refusal in the checkbox copy instead of the flag name.

naingthet:
**NO GO** at 90ecb4771
Depth: deep

1. P2: Spec-fetch HTTPS redirects still re-admit with `allow_loopback_http`, so a remote document URL can 302 onto `http://127.0.0.1:<other-port>` and the fetch follows it. Loopback-http originals already refuse redirects so the opt-in cannot leave that origin; HTTPS hops need the same pin (or must re-admit without the flag). Execution stays origin-pinned; this is config-time SSRF onto a different loopback origin than the operator typed.
2. P2: Loopback HTTP still fails through the form's fetch paths. "Find the OpenAPI document" admits https-only and probes with the flag false, so it cannot discover a document on the origin this PR admits. Fetch operations stays clickable without consent, and choosing a discovery hit calls preview without the flag, so a URL-source save's required preview 400s naming `allow_loopback_http`. Disable Find/Fetch until the box is checked when the URL is loopback HTTP, pass the flag through discovery and choose-discovered, and put that refusal in the checkbox copy instead of the flag name.

PR #3078, opened 2026-09-02
feat(connected-apps): make REST API and OpenAPI discovery practical

### LINKED ISSUE #3035: feat(connected-apps): make REST API and OpenAPI discovery practical
## What are you trying to do?\n\nConnect a company's REST API to Tidebreak without first spending substantial time searching for the company's API domain and OpenAPI document.\n\n## What happens today?\n\nThe REST connected-app flow assumes the user already knows both the correct API domain and where the company publishes an OpenAPI document. Those values are often difficult to locate or are documented under unexpected URLs, so setup becomes a lengthy external research task before the Tidebreak flow can even begin.\n\n## What would a good outcome look like?\n\nThe flow helps users locate or derive the API domain and OpenAPI document, or clearly guides them through the discovery process. A user starting with a company or product name should be able to reach a valid import target without manually hunting across documentation sites.\n\nThe UI should also explain what to do when a provider does not publish an OpenAPI document.\n\n## Acceptance criteria\n\n- The REST setup flow no longer assumes users already know the API base domain and OpenAPI URL.\n- The flow provides actionable discovery guidance or product-assisted discovery.\n- Invalid or unsupported inputs explain the next corrective step.\n- The no-public-OpenAPI case has an explicit path forward.\n\n## Additional context\n\nReported from hands-on testing of Connected apps. Searching existing issues found related completed implementation work, but no open issue covering the current discovery friction.

naingthet:
Closes #3035

## What changed

The REST connected-app form no longer assumes you already have the OpenAPI document URL.

**Server.** `POST /connected-apps/rest/spec-discovery` takes one https origin or base URL (same managed-profile refusal as spec preview). It probes a fixed list of well-known paths on the origin and, when present, the path prefix (`/openapi.json`, `/openapi/v3.json`, `/openapi.yaml`, `/swagger.json`, `/swagger/v1/swagger.json`, `/v3/api-docs`, `/api-docs`, `/api/openapi.json`, `/.well-known/openapi.json`, `/docs/openapi.json`). Each fetch uses the existing spec-document admission (https, SSRF, size, time). Probes run with concurrency 4 and a 5s overall deadline. Cross-origin redirects are not followed. The response lists candidates (operation count, or why they cannot be used: YAML, Swagger 2.0, HTML) and every location tried.

**UI.** Next to the base URL, **Find the OpenAPI document** calls that route, shows candidates, and choosing one fills the document URL and runs the existing preview. On failure it lists the probed locations. **No public OpenAPI document?** is always on the form: check the vendor portal; paste a minimal OpenAPI 3 JSON example (one GET with a path parameter and a bearer header); or add an MCP server on this page. Ingest errors for http, HTML, YAML, and Swagger 2.0 keep the server text and add a next step.

**Docs.** `docs/connected-apps.md` has a short "Finding the OpenAPI document" section.

## Follow-up (deliberately not in this slice)

- Model-assisted or web-search discovery of a company's API domain / OpenAPI URL from a product name.

## What I ran

- `cargo fmt --all`
- `cargo check -p tidebreak-server --tests`
- `cargo test -p tidebreak-server connected_apps` — discovery tests pass. `the_listing_carries_mounted_tool_names_only` fails here because this sandbox cannot accept loopback TCP (pre-existing MCP mock); it is unrelated to discovery.
- `cargo test -p tidebreak-server rest_executor` — pass
- `cargo clippy -p tidebreak-server --tests -- -D warnings` — fails on two findings in `harness_llm.rs` and `exec_write_snapshot.rs` that this change does not touch
- Desktop UI: `CI=true pnpm install --frozen-lockfile` could not finish (cdn.sheetjs.com `xlsx` fetch failed in this sandbox), so biome/lint/test/build and `pnpm storybook:build` were not run here. CI should run those lanes.

Wire types were regenerated with `UPDATE_WIRE_TYPES=1 cargo test -p tidebreak-server --lib the_generated_bindings_are_current` and copied to `mobile/src/generated/wire.ts`.

### DIFF
--- a/crates/tidebreak-desktop/ui/src/api/client/apps.ts
+++ b/crates/tidebreak-desktop/ui/src/api/client/apps.ts
@@ -22,6 +22,7 @@ import {
   type PromptBody,
   type RestCredentialUpdate,
   type SkillInstructions,
+  type SpecDiscoveryInfo,
   type SpecPreviewInfo,
 } from "../types";
 import { attachedRemotely } from "../../host";
@@ -71,6 +72,15 @@ export function withAppsApi<TBase extends Constructor<HttpCore>>(Base: TBase) {
         method: "POST",
         headers: this.headers(true),
         body: JSON.stringify({ source }),
+      });
+    }
+
+    /** Probe well-known OpenAPI locations for one https origin or base URL. */
+    discoverRestSpec(origin: string): Promise<SpecDiscoveryInfo> {
+      return this.json("/connected-apps/rest/spec-discovery", {
+        method: "POST",
+        headers: this.headers(true),
+        body: JSON.stringify({ origin }),
       });
     }
 
--- a/crates/tidebreak-desktop/ui/src/api/types.ts
+++ b/crates/tidebreak-desktop/ui/src/api/types.ts
@@ -28,6 +28,8 @@ import {
   type RestCredentialStatus as WireRestCredentialStatus,
   type SpecPreviewInfo as WireSpecPreviewInfo,
   type SpecPreviewOperation as WireSpecPreviewOperation,
+  type SpecDiscoveryInfo as WireSpecDiscoveryInfo,
+  type SpecDiscoveryCandidate as WireSpecDiscoveryCandidate,
   type ExecCredentialReadiness as WireExecCredentialReadiness,
   type ExecProviderKind as WireExecProviderKind,
   type CodeDeliveryActionResult as WireCodeDeliveryActionResult,
@@ -552,6 +554,8 @@ export type ConnectedAppInfo = WireConnectedAppInfo;
 export type RestCredentialStatus = WireRestCredentialStatus;
 export type SpecPreviewInfo = WireSpecPreviewInfo;
 export type SpecPreviewOperation = WireSpecPreviewOperation;
+export type SpecDiscoveryInfo = WireSpecDiscoveryInfo;
+export type SpecDiscoveryCandidate = WireSpecDiscoveryCandidate;
 /** Where a stored REST credential is injected: `"bearer"` or a named header. */
 export type CredentialPlacement = WireCredentialPlacement;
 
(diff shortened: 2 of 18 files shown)


naingthet:
@ShipRight review

naingthet:
**NO GO** at 3296bd519
Depth: medium

P2: `discover_openapi_documents` awaits spawned probes in list order and, when the 5s deadline fires, replaces whatever it already collected with an empty `candidates` list (`timeout(...).unwrap_or_default()`). A slow first path (`/openapi.json`) blocks later hits, so a document that already answered is reported as nothing found. The form then shows "No OpenAPI document turned up at the usual locations." Collect finished probes as they complete, keep what arrived before the deadline, and treat timeout as timeout rather than as an empty origin.

P2: `fetch_spec_document_detailed` now sends `Accept: application/json, application/yaml, text/yaml;q=0.8` on the shared preview and upsert fetch, not only on discovery. JSON and YAML sit at q=1.0, so a content-negotiating origin can return YAML for a URL that used to serve JSON. Ingest still refuses YAML. Keep preview/upsert preferring JSON; do not advertise YAML at equal quality on the path that cannot use it.

naingthet:
**NO GO** at 3296bd519
Depth: medium

1. P2: Discovery's 5s deadline awaits probe tasks in spawn order and, on timeout, returns an empty candidate list, discarding any documents that already answered. A hanging first well-known path (`/openapi.json`) hides a finished later hit. Each in-flight probe still uses the 30s spec-fetch timeout, so the first four slow 404s consume the budget before later paths start. The UI then lists every URL as tried and says nothing turned up, which is the same shape as a complete miss.

2. P2: `fetch_spec_document_detailed` now sends `Accept: application/json, application/yaml, text/yaml;q=0.8` for every spec fetch, including preview and upsert. `application/yaml` is q=1.0, same as JSON, so a content-negotiating origin can return YAML. `/v3/api-docs` is in the probe list; Springdoc will often pick YAML when both types are equally acceptable. Discovery then reports the document unusable, and preview/upsert ingest fails, for a URL that previously received JSON.

PR #3081, opened 2026-09-02
fix(code): open the workspace conversation as soon as the session exists

naingthet:
## What

Creating a workspace with a first message left the starting steps on screen until you refreshed, even after the agent had begun answering.

`POST /turns` waits until that first turn finishes. The create handoff waited for the same response, so the workspace page never opened the session socket while the turn ran.

The overlay now clears as soon as the session exists. The page also reads a remembered session during render, so dropping the overlay does not flash the empty start prompt.

## Why

A refresh should not be required to watch the first reply.

## Test plan

- [ ] Create a workspace with a first message. The conversation appears while the agent is still working.
- [ ] Create a workspace with no first message. The start composer still appears after the session exists.
- [ ] If the first turn fails to send, the prompt returns to the workspace composer.

Focused checks: `biome check` on the four touched files; `vitest run` for `NewWorkspaceDialog.dom.test.tsx`, `useWorkspaceSessions.test.ts`, and `CodeWorkspacePage.dom.test.tsx` (113 passed).

### DIFF
--- a/crates/tidebreak-desktop/ui/src/code/NewWorkspaceDialog.dom.test.tsx
+++ b/crates/tidebreak-desktop/ui/src/code/NewWorkspaceDialog.dom.test.tsx
@@ -942,6 +942,57 @@ describe("NewWorkspaceDialog", () => {
     );
   });
 
+  it("drops the handoff once the session exists, before the first turn finishes", async () => {
+    const repos = [repo("repo-new", "tidebreak")];
+    useCodeCatalogStore.setState({
+      repos,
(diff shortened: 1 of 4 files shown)


naingthet:
**G2G** at 6532d57e0
Depth: medium

Handoff clears after `rememberSession`, and `listedSessions` folds that snapshot in during render so dropping the overlay does not flash the empty start prompt. Failed first turns still `holdPrompt` into the workspace composer. No P1/P2.

PR #3082, opened 2026-09-02
fix(desktop): use the shared Spinner on busy refresh controls

naingthet:
Busy refresh buttons spun Lucide `RefreshCw` or `RotateCw`. Those glyphs orbit under `animate-spin` because the ink box is not the viewBox center.

The buttons now swap to the shared `Spinner` while work is in flight. The styles contract fails if a refresh icon is spun again.

## Validation
- `pnpm --dir crates/tidebreak-desktop/ui exec biome check` on the changed files
- `pnpm --dir crates/tidebreak-desktop/ui test src/stylesContract.test.ts`


naingthet:
**G2G** at c25c87e72
Depth: medium

Busy refresh controls swap to the shared Spinner. Adjacent copy or aria-label stays, and the control is disabled while work is in flight. The contract test covers the Lucide orbit case this PR is fixing.

PR #3085, opened 2026-09-02
ci: compile Windows native tests once and link with rust-lld

naingthet:
## What changed

Windows-scoped CI rebuilt four test crates in sequence under `link.exe`, which is what made the required `Windows cargo check` job take ~20 minutes. After `cargo check`, that job now compiles the native test binaries in two `cargo test --no-run` graphs and links them with `rust-lld`. The later run steps reuse those binaries. `cargo check` still uses the MSVC linker so a linker mismatch cannot fail the rust-scoped compile gate.

A rust-only change still pays only `cargo check`. This does not switch the job onto a larger runner.

## Validation

Could not time the Windows job locally. The first windows-scoped CI run on this PR is the measurement.

## Compatibility and risk

Windows CI linker for native test binaries only. `cargo check` and non-Windows lanes are unchanged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -560,6 +560,11 @@ jobs:
       SCCACHE_GHA_ENABLED: "true"
       SCCACHE_GHA_RW_MODE: READ_WRITE
       RUSTC_WRAPPER: sccache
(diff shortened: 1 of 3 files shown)


naingthet:
**G2G** at 0ce55121f
Depth: medium

PR #3087, opened 2026-09-02
ci: skip the Windows native test precompile

naingthet:
## What changed

#3085 compiled native Windows test binaries in `cargo test --no-run` graphs before the suites. That step took 9.4 minutes, and the later test steps still compiled, so the job went from ~20 minutes to 28.

This skips those graphs. The cargo lines stay in the workflow so the base-branch release-policy test still matches; a full revert of #3085 fails that test.

## Validation

## Compatibility and risk

Windows native tests still run. They compile in the existing suite steps, as they did before #3085.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -659,9 +659,9 @@ jobs:
           fi
       # `cargo check` does not build test binaries, and `--lib` is workspace-wide,
       # so host-broker's sidecar integration test cannot share an invocation with
-      # server and core. Two `--no-run` graphs compile everything the later run
-      # steps need; rustc can then keep those binaries instead of rebuilding
-      # them four times in sequence.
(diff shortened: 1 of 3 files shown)


naingthet:
**G2G** at c0f2122e5
Depth: medium

Dropping the Windows `cargo test --no-run` graphs and `rust-lld` restores the job from before #3085. Native suites still run behind the same `NATIVE_RUN` gate, `cargo check` no longer needs a `RUSTFLAGS` wipe, and the workflow test plus releases note match the workflow.

naingthet:
**G2G** at 20e13a6ce
Depth: medium

The precompile step now exits before the `--no-run` graphs, so the nine-minute extra compile is gone. Native suites still run behind `NATIVE_RUN`, `cargo check` still gates the job, and the cargo lines stay only so the policy test matches. Docs and the workflow test follow that skip.

PR #3088, opened 2026-09-02
ci: attest the published server image

naingthet:
Every pushed server manifest, each per-architecture image and the multi-arch index, now carries a GitHub build-provenance attestation (`actions/attest-build-provenance`, pinned by commit). The bundle is stored under this repository and pushed to GHCR as an OCI referrer.

Why: Model Gateway hosts Tidebreak as a managed add-on and admits images against a trust policy that enumerates digests today, so every release and every weekly rebuild needs a deployment change to admit the new digest. With provenance, the gateway can admit any digest that this repository's publish workflow provably built, verified offline against the Sigstore trust root.

Decision record 84 states the contract: the workflow file path is now the image's public identity, so moving it is a breaking change for verifying consumers.

Verify on the next publish:

```
gh attestation verify oci://ghcr.io/brightwave-inc/tidebreak-server:<version> -R brightwave-inc/tidebreak
```

`actionlint` passes on the workflow. No image build changes.

### DIFF
--- a/.github/workflows/publish-server-image.yml
+++ b/.github/workflows/publish-server-image.yml
@@ -57,6 +57,16 @@ run-name: >-
 # whole point of that lane. Consumers that need an immutable reference pin the
 # digest, which every run prints in its step summary.
 #
+# Every pushed manifest — each per-architecture image and the multi-arch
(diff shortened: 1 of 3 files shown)


naingthet:
**G2G** at c0c8d243c
Depth: medium

PR #3218, opened 2026-09-04
feat(self-host): a standalone machine signs a browser in

### LINKED ISSUE #3182: Standalone browser sign-in: token paste now, an OIDC authenticator next
Part of #3178 (track B).

## Context

Decision 82 serves the renderer from the machine, but a machine on `TIDEBREAK_AUTH_TOKENS_FILE` has no browser sign-in: `HostedSignIn` says so and the desktop app stays the only client. `docs/self-hosting.md` documents that gap. A Slack link to a session on a standalone machine therefore lands nowhere. Decision 6 sequences the end state: roster-provisioned tokens first, then an authenticator behind the credential-to-principal seam.

## Outcome

A standalone machine signs a browser in.

## Deliver

1. Token paste. When discovery reports `static_token`, the sign-in page takes a token, probes the principal read with it, and holds it in memory for the tab's life the way the handoff bearer is held (no cookie, no storage). The page honors `return_to` so a session link lands on the session.
2. An OIDC authenticator: `TIDEBREAK_AUTH_OIDC_ISSUER`, client id and secret, and a claim-to-user mapping, selected at boot as a third exclusive mode beside the gateway and the token file (`crates/tidebreak-server/src/auth.rs`). Authorization-code with PKCE in the browser, a machine-issued short-lived bearer bound to the session, and the same fail-closed rules as decision 49: a provider refusal admits nobody. The token file remains the bootstrap for the first admin and for CLI access.
3. `docs/self-hosting.md`: replace "has no browser sign-in" with both paths.

## Done when

A standalone machine's session link, opened in a fresh browser, asks for sign-in and lands on the session; an OIDC-mode machine refuses a static token and a token-file machine refuses an OIDC bearer. Record: "standalone browser sign-in".

naingthet:
Closes #3182. Implements [decision 0087](https://github.com/brightwave-inc/tidebreak/blob/main/docs/decisions/0087-standalone-browser-sign-in.md).

A machine on `TIDEBREAK_AUTH_TOKENS_FILE` had no browser sign-in at all: the page said so and sent the reader to the desktop app. A Slack link to a session on such a machine therefore landed on a screen nobody could pass, which quietly made every feature in this epic a gateway-only feature. This branch gives a standalone machine two ways to sign a browser in, and the page shows whichever one its machine can run.

## Token paste

When discovery reports `static_token`, the sign-in screen takes a token. On submit the page probes it against `/models` — the member-plane capability read every principal may make, and the kind of request the shell makes with a bearer the moment boot finishes — and only then holds it, in the same tab memory the hand-off bearer lives in: no cookie, no storage, gone on reload. A refusal stays on the same screen with the field still filled. A pasted value that could not be a bearer at all is refused without a request. `return_to` is honored the way `consoleSignInUrl` does it, so a session link lands on its route.

## OIDC authenticator

`TIDEBREAK_AUTH_OIDC_ISSUER`, `_CLIENT_ID`, and `_CLIENT_SECRET` select a third exclusive boot mode. Combining them with `TIDEBREAK_AUTH_GATEWAY_URL` is a boot error whose message names both variables; the token file may sit beside them, because OIDC only ever names a member and a standalone deployment still needs a first administrator and a CLI credential.

`/auth/oidc/start` redirects to the issuer's authorization endpoint with a fresh `state`, `nonce`, and PKCE challenge. `/auth/oidc/callback` takes the flow that `state` names — removing it, so a `state` answers once — exchanges the code with that flow's verifier, and validates the ID token against the issuer's discovery document and JWKS: signature by a published key, a signing algorithm from a fixed set, the configured issuer, a live expiry, this client id as the audience, and the flow's nonce. Only then does the machine mint its own bearer, good for an hour, and land the page through the same `#handoff=` envelope the console hand-off uses — so `captureHandoffToken` needed no change. Every failure lands the page with a reason and mints nothing. The code, the ID token, and the bearer are never logged.

`TIDEBREAK_AUTH_OIDC_CLAIM` overrides the login claim (`sub` by default), and the authorization request asks for the scope that claim needs (`email` or `profile`) alongside `openid` — otherwise pointing the override at `email` would fail closed on a claim the provider was never told to send. Discovery reports `mode: oidc` with the start URL, and the page shows one button named for the issuer.

### Which crate, and why

**`oauth2` 5.0.0 plus `jsonwebtoken` 11.0.0**, not `openidconnect`. `openidconnect` is not in the workspace lockfile, so it was not the cheaper option, and it brings a large generic client this code would use for two round trips. `oauth2` supplies the PKCE challenge and the CSPRNG behind `state`, `nonce`, and the minted bearer; `jsonwebtoken` supplies JWKS parsing and JWT validation, which is the part worth borrowing. Four packages join the lockfile (`oauth2`, `jsonwebtoken`, `signature`, `untrusted` 0.7.1), all under licenses `deny.toml` already allows.

## Fail-closed and hardening

- A provider refusal, an unreachable issuer, an oversized body, or a key set that does not publish the signing key all admit nobody.
- The issuer and every endpoint its discovery document names must be HTTPS, loopback excepted, so an issuer that answers with an `http://` token endpoint does not get the client secret. Discovery must also name the configured issuer.
- Both in-memory maps are capped. `/auth/oidc/start` is public, so a machine that cannot remember a flow refuses to start one rather than accepting a callback it could not check.
- `/auth/discovery` stays public and unauthenticated, and so do the two new routes, because a page reads all three before it holds any credential. They 404 on a machine in another mode.
- Cross-mode refusals are tests in both directions: an OIDC machine accepts only its bootstrap file's tokens and its own `tb_oidc_` bearers, and a token-file machine has nothing that could answer for one.

## Docs

`docs/self-hosting.md` loses the "has no browser sign-in" paragraph. In its place: what every path does with the bearer, the gateway console path kept, the two new paths, the four OIDC variables, the redirect URI to register, the claim mapping and what it costs when it drifts, and the corrected boot-fails-closed bullet, prerequisites, and variable table.

## Verification

Run locally on this branch; each block ends with the command's final line.

```
$ cargo fmt --all --check
(no output)

$ cargo clippy -p tidebreak-core -p tidebreak-server-core -p tidebreak-server --all-targets -- -D warnings
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 55.21s

$ cargo test -p tidebreak-server-core auth::tests
test result: FAILED. 22 passed; 1 failed; 0 ignored; 0 measured; 956 filtered out; finished in 5.01s

$ cargo test -p tidebreak-server-core -- an_id_token_this_machine_did_not_ask_for_names_nobody \
    the_login_claim_override_names_the_user_or_nobody \
    a_minted_bearer_lasts_its_hour_and_then_names_nobody \
    a_started_flow_is_single_use_and_its_request_carries_the_whole_check \
    the_number_of_waiting_sign_ins_is_bounded \
    an_oidc_machine_refuses_a_static_token_and_a_token_file_machine_refuses_a_bearer \
    oidc_is_a_third_exclusive_mode_that_keeps_its_bootstrap_token_file \
    an_oidc_provider_this_machine_would_not_trust_fails_boot
test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 971 filtered out; finished in 0.03s

$ cargo test -p tidebreak-server auth
test result: FAILED. 11 passed; 2 failed; 0 ignored; 0 measured; 615 filtered out; finished in 127.48s

$ cargo test -p tidebreak-server a_standalone_machine_publishes_the_one_sign_in_it_can_run
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 627 filtered out; finished in 0.13s

$ cargo test -p tidebreak-core config
test result: ok. 26 passed; 0 failed; 0 ignored; 0 measured; 848 filtered out; finished in 0.01s
```

In `crates/tidebreak-desktop/ui`:

```
$ pnpm exec biome format --write src
Formatted 876 files in 964ms. No fixes applied.

$ pnpm exec vitest run HostedSignIn hostedSession boot
 Test Files  3 passed (3)
      Tests  37 passed (37)

$ pnpm exec tsc --noEmit -p tsconfig.json
(no output)
```

Also run, beyond what was asked: the whole UI suite (`267 files, 2306 tests passed`), `pnpm lint` (clean), and `pnpm storybook:build` (`Storybook build completed successfully`).

### The three Rust failures above are the sandbox, not this branch

This sandbox blocks loopback TCP, so every test that stands up a local HTTP server and connects to it fails here. The three are `auth::tests::gateway_auth_maps_live_identity_and_fails_closed`, `tests::websocket::ws_subprotocol_auth_succeeds`, and `tests::code_turns::reap_replaces_browser_authority_without_tombstoning_the_session` — none of them touched by this branch. Confirmed by running the same command in a clean worktree of unmodified `main`:

```
$ git worktree add /tmp/tb-main origin/main && cd /tmp/tb-main
$ cargo test -p tidebreak-server auth
test tests::websocket::ws_subprotocol_auth_succeeds ... FAILED
test tests::code_turns::reap_replaces_browser_authority_without_tombstoning_the_session ... FAILED
test result: FAILED. 11 passed; 2 failed; 0 ignored; 0 measured; 614 filtered out; finished in 127.96s
```

Same two, same counts. The same limitation is why the OIDC route tests here cover everything except the two HTTP round trips to the issuer: the wiring around them is covered by unit tests over `begin`, `claim_flow`, `authorization_url`, `mint_bearer`, and `validate_oidc_id_token`, and by route tests that reach the callback's refusal paths without a network. A reviewer with loopback available may want to exercise a real issuer once.

## Storybook

`Shell/Hosted browser session` gains `StaticToken`, `StaticTokenRefused` (with a play function that types a wrong token and asserts the refusal), and `Oidc`, and the existing gateway states are unchanged. `.boot-token-form` is a new block in `styles.css` because the boot screen is centered copy and a labelled field is not.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -491,6 +491,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ce2b2dcc879c3bae0d371e77c99f2238400ef24ec001394befa67b6e543add9e"
 dependencies = [
  "aws-lc-sys",
+ "untrusted 0.7.1",
  "zeroize",
 ]
 
@@ -3381,6 +3382,22 @@ dependencies = [
 ]
 
 [[package]]
+name = "jsonwebtoken"
+version = "11.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "881733cbc631fc9e472e24447ce32a64bedf2da498d6d8570b08edc87de71f65"
+dependencies = [
+ "aws-lc-rs",
+ "base64 0.22.1",
+ "getrandom 0.2.17",
+ "js-sys",
+ "serde",
+ "serde_json",
+ "signature",
+ "zeroize",
+]
+
+[[package]]
 name = "keyboard-types"
 version = "0.7.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -3945,6 +3962,25 @@ dependencies = [
  "proc-macro2",
  "quote",
  "syn 2.0.119",
+]
+
+[[package]]
+name = "oauth2"
+version = "5.0.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "51e219e79014df21a225b1860a479e2dcd7cbd9130f4defd4bd0e191ea31d67d"
+dependencies = [
+ "base64 0.22.1",
+ "chrono",
+ "getrandom 0.2.17",
+ "http",
+ "rand 0.8.8",
+ "serde",
+ "serde_json",
+ "serde_path_to_error",
+ "sha2 0.10.9",
+ "thiserror 1.0.69",
+ "url",
 ]
 
 [[package]]
@@ -5137,7 +5173,7 @@ dependencies = [
  "cfg-if",
  "getrandom 0.2.17",
  "libc",
- "untrusted",
+ "untrusted 0.9.0",
  "windows-sys 0.52.0",
 ]
 
@@ -5308,7 +5344,7 @@ dependencies = [
  "aws-lc-rs",
  "ring",
  "rustls-pki-types",
- "untrusted",
+ "untrusted 0.9.0",
 ]
 
 [[package]]
@@ -6018,6 +6054,15 @@ checksum = "c4db69cba1110affc0e9f7bcd48bbf87b3f4fc7c61fc9155afd4c469eb3d6c1b"
 dependencies = [
  "errno",
  "libc",
+]
+
+[[package]]
+name = "signature"
+version = "2.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "77549399552de45a898a580c1b41d445bf730df867cc44e6c0233bbc4b8329de"
+dependencies = [
+ "rand_core 0.6.4",
 ]
 
 [[package]]
@@ -7521,7 +7566,9 @@ dependencies = [
  "image",
  "infer 0.22.0",
  "jsonschema",
+ "jsonwebtoken",
  "libc",
+ "oauth2",
  "plist",
  "portable-pty",
  "reqwest 0.12.28",
@@ -8227,6 +8274,12 @@ name = "unicode-xid"
 version = "0.2.6"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ebc1c04c71510c7f702b52b7c350734c9ff1295c464a03335b00bb84fc54f853"
+
+[[package]]
+name = "untrusted"
+version = "0.7.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "a156c684c91ea7d62626509bce3cb4e1d9ed5c4d978f7b4352658f96a4c26b4a"
 
 [[package]]
 name = "untrusted"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -111,6 +111,8 @@ tokio-util = { version = "=0.7.19", features = ["io-util"] }
 # loopback chat socket itself is plain ws://.
 tokio-tungstenite = { version = "=0.30.0", default-features = false, features = ["connect", "handshake", "rustls-tls-webpki-roots"] }
 reqwest = { version = "=0.12.28", default-features = false, features = ["json", "multipart", "stream", "rustls-tls"] }
+oauth2 = { version = "=5.0.0", default-features = false }
+jsonwebtoken = { version = "=11.0.0", default-features = false, features = ["aws_lc_rs"] }
 sha2 = "=0.11.0"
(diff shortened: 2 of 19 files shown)


naingthet:
Review notes from the maintainer: the authenticator (flow claim, PKCE, algorithm allowlist, audience/issuer/nonce checks, bounded maps, HTTPS-or-loopback on every issuer endpoint), the token-paste probe, and the discovery routing all read as the contract in decision 0087 asks. The two failing groups the body reports do not reproduce here: `cargo test -p tidebreak-server-core auth::tests` passes 23 and `cargo test -p tidebreak-server auth` passes 13 on this head, so those were the sandbox's environment. I pushed the regenerated `legal/THIRD-PARTY-NOTICES.md` for the four new crates, which the notices lane wanted.

PR #3429, opened 2026-09-15
fix(code): silence routine sandbox retirement notices

naingthet:
Normal sandbox retirement appended “The remote supervisor stopped (stop_requested)” to the conversation after completed turns. Suppress routine stop and idle-ceiling notices while preserving the durable event cursor and terminal-flush marker. Turn interruptions, engine failures, checkpoint failures, and unexpected shutdown reasons retain their visible outcomes.

Validation: nine event-ingestion tests passed, including retirement replay after completed and interrupted turns and preservation of unexpected stop notices. `cargo check -p tidebreak-code-remote`, formatting, and `git diff --check` passed.


naingthet:
Released and verified in v0.105.3.

Desktop release 34976849537 and server-image release 34982301847 passed. Both container images publish for amd64 and arm64 and pass attestation verification. Hosted Tidebreak adopted the release after its configured 30-minute soak; the matching supervised runtime is deployed.

A bounded follow-up in an existing Slack DM returned its exact result once in Slack and the same hosted workspace. The workspace button opened the matching conversation. Gateway retained the same user-owned subscription with notional billing and no fallback. The sandbox then retired normally. The workspace kept its two historical shutdown notices and added no new notice.

Validation: nine focused ingestion tests, the crate check, all 22 full-validation jobs, and main CI passed.

Issue #3191, opened 2026-09-04
Child sessions: parent link, lifecycle, and a generalized wait park


naingthet:
Part of #3178 (track E).

## Context

Decision 52 renders a harness's `Task` spans as subagents: not sessions, not steerable, never in another workspace. Chat's background agent runs are in-process, depth one, four per turn. Neither is "spin up five workspaces and work each". The internal engine already parks on `ParkWait::AgentRuns` and never resumes (#3014).

## Outcome

A session may create child sessions, in new or existing workspaces, on the machine or in a sandbox, wait on them, and read their results. Children are real sessions.

## Deliver

- `session.parent_session_id` (appended migration), a children read on the session, and the parent shown on a child's snapshot.
- Lifecycle: a parent's wait park resumes when the children it named settle; a parent's end does not end its children; reap and delete are per session; a child's fence surfaces on the parent as attention.
- Generalize `ParkWait::AgentRuns` into a wait on a set of child-session or sandbox-run ids with one resume input, so #3014 closes on the same shape.
- Rail and notifications: a child shows under its parent; a parent's digest carries a bounded children summary the way subagents ride the digest today.
- Wire types and stories for the tree (track E5 covers the rendering).

## Done when

A parent creates three children in three workspaces, waits, and resumes with their results in the order it asked; killing one child fences the parent's wait with the reason; ending the parent leaves the children running and reachable.

## Depends on

#3014.

naingthet:
Child-tree rendering and wait freshness are merged in #3475 (`fda73e3a0efa21b64c93dfb19194459557da1a3b`). Keep this issue open for durable child waits and the remaining lifecycle contract.

Covered by implementation and tests:
- `db/migration/session_context.rs` persists parent links; `code/session_tree.rs` projects authorized children and fences. Tests include `live_child_status_reaches_a_quiet_parent_journal` and `ending_the_parent_leaves_children_running_and_on_the_tree`.
- `session_tools.rs::a_conversation_chooses_two_repositories_and_waits_without_a_parent_workspace` verifies independent workspaces and requested result order.
- #3475 adds recursive rail rendering, `CodeSessionTree.stories.tsx`, and reconnect/expiry regressions. Its focused validation passed 11 WebSocket, 12 tree, 13 session-tool, 16 owner-scope, and one external-tree test, plus scoped Clippy.

Outstanding acceptance:
[...]

PR #3459, opened 2026-09-16
feat(code): publish authorized child session trees

naingthet:
Parent Slack threads cannot show their delegated sessions because Tidebreak does not publish a child tree. Add direct-child snapshots and journal updates using the existing session relationships and grants. Quiet parents receive child lifecycle and attention changes, including fault clearing. Ending a parent leaves its children running.

Snapshots explicitly include empty children and a null wait, so reconnects can clear stale cards. Adapter grants and shared-session readers see only authorized children. Parent updates serialize computation and journal writes to prevent stale concurrent updates. The desktop parser accepts the wire fields and rejects malformed states.

Validation: 9 focused tree tests, 1 gate-lifetime test, and 1 external WebSocket test pass. The WebSocket test covers snapshots, replay, quiet-parent updates, fault clearing, grant isolation, ended parents, and empty reconnect trees. Generated wire/fixture tests pass (1 each), UI parser tests pass (118), TypeScript passes, scoped Clippy with warnings denied passes, and formatting/diff checks pass.

Related to #3191 and brightwave-inc/model-gateway#1856. Pairs with brightwave-inc/model-gateway#2070. This implements the direct-child event contract; code_wait remains polling, so wait stays null. No live producer/consumer canary is claimed, and this does not close the full generalized-wait issue.


naingthet (commented):
Reviewed 819bf42eb4619f0b24295c4acfec0853f3230fcf (direct-child session_tree production). This does not replace the Shipright gate.

No actionable correctness/security findings.

Checked owner/grant isolation on snapshot, replay, and live send; shared-reader snapshot and debug filtering; empty reconnect clears; concurrent parent-tree journaling; fault clearing via persist/reap; parent end leaving children running.

Limits: wait stays null (code_wait still polling). Attach failures warn and leave the default empty children/null wait, which reconnect treats as authoritative; WS replay can still restock. Binding/access changes are filtered on the next send or reconnect and do not themselves journal a tree. Did not re-run the suite.

Issue #3178, opened 2026-09-04
epic: Tidebreak in Slack: channel sessions, the web view, and orchestration


naingthet:
Tidebreak in Slack, whole: a person mentions Tidebreak in a channel or DM and gets an agent that answers, triages, works in one or several repositories, spins up workspaces and child sessions, and ships pull requests, with anyone in the channel able to contribute and anyone on the deployment able to open the session on the web. It works without Model Gateway, and with Model Gateway but without a sandbox runtime.

`docs/slack-sessions.md` stays the adapter's design page; this epic names where it changes. The adapter and gateway slices live in the model-gateway epic (brightwave-inc/model-gateway#1847).

## Outcomes

- Mention Tidebreak in a channel or DM and get an agent that answers, triages, and works. No repository is required to start. A task that names a repository runs in a workspace; a task that spans repositories runs across several.
- Anyone in the channel can contribute in the thread. Every turn shows who submitted it. Approvals, questions, and plan decisions are answerable in Slack when the session runs where approvals can be answered.
- Anyone on the deployment opens the session on the web in one click, signed in, on a gateway-backed machine or a standalone self-host. A private channel's session is visible only to its members.
- A session that opens a pull request keeps working it: CI results, review findings, and merge state land in the thread, and `fix`, `merge`, and friends from the thread are turns on the same session.
- The agent can spin up workspaces and child sessions, on the machine or in gateway sandboxes, wait for them, and report back. The thread and the web show the tree.
- The agent can read the thread it was mentioned in when the channel allows it. A reply during a turn steers where the engine can take a steer.
- One session can be reached from more than one thread or channel, and a Slack Code channel can be a session.
- Later: the person's own laptop can be the machine Slack reaches.

## Baseline when this epic opened — September 4, 2026

- External get-or-create refuses with `remote_disabled` when no sandbox runtime is configured and always builds a remote workspace (`crates/tidebreak-server/src/code/runtime/remote.rs`).
- A Slack session is owned end to end by the mentioning user; other replies never reach the engine. `ScopedCode` binds every `/code/*` query to one owner key; turns record no actor.
- The hosted page has no way in from Slack: the only bearer comes from the gateway console (decision 82), a standalone machine has no browser sign-in, and the link the adapter posts is a path against a hash router.
- Workspace-less internal-engine sessions exist (`POST /code/sessions`); their agent-run park never resumes (#3014) and they have no notifications or rail rows (#3015).
[...]

PR #3485, opened 2026-09-18
fix: keep hosted browser sign-in across reload

naingthet:
## What changed

Reloading a hosted Tidebreak tab that had already opened a Slack parent or child transcript showed Sign in required, even though execution kept running and Open the console silently signed in through the existing Gateway session and returned to the same workspace and task query.

The bearer still lives in memory only (decision 82). After a successful attach, the tab now records a presence-only sessionStorage marker and, on reload without a bearer, renews through the existing console or OIDC hand-off with `return_to` set to the current hash route. First visit, hand-off failure, sign-out, static-token machines, missing Gateway login, and a second refusal inside the loop window stay on the sign-in screen. The credential is never written to storage or a cookie.

## Validation

Focused Vitest (jsdom, no live credentials):

- `hostedSession.test.ts`: 46 tests pass, including valid reload with workspace/task query, OIDC renewal, first visit, static-token contract, expired/invalid/unavailable hand-off, sign-out, missing Gateway login loop, loop-window expiry, concurrent tabs, and no bearer in storage.
- `HostedSignIn.dom.test.tsx`: 3 tests pass, including the console link preserving `/code/w/workspace-1?task=session-2`.
- `ManagedGate.dom.test.tsx`: 27 tests pass (existing expiry/loop screens).

`tsc --noEmit` passes. Biome format and lint pass on the changed files. No Rust changes; no cargo loop.

Story copy for the first-visit sign-in screen was updated. No visual redesign, so no new Storybook screenshots.

## Compatibility and risk

Does not store the bearer, extend its lifetime, or add cookies. Renewal still goes through Gateway console / OIDC, so expiry, revocation, and an absent Gateway login remain fail-closed. `return_to` stays the existing same-origin hash-route check. sessionStorage is origin- and tab-scoped. Closing the tab drops the marker.

No Gateway change is required: the live console already silently mints a bearer from the existing Gateway session.

## Tracking

Part of #3178 (hosted reload sign-in continuity).


naingthet:
Review against 354ec11920c131b982d1800afa87a6e445ecfafa: P2 — bound unsuccessful automatic renewal across slow round trips.

The stored attempt stops another redirect for only 15 seconds. I executed the exact `hostedSession.ts` from this head with an established continuity marker, no returned bearer, and calls at timestamps 1,000,000; 1,015,001; and 1,045,002. All three return `redirect` to the console. A failed or abandoned round trip taking longer than 15 seconds therefore starts another automatic renewal instead of reaching the sign-in screen. The test named `allows another renewal after the loop window` currently pins this behavior.

Consume one automatic renewal attempt until a session is successfully established or the user explicitly retries. Add a regression for an unsuccessful round trip longer than 15 seconds and repeated reloads afterward. Preserve the existing in-memory bearer policy and deep-link path/query. The sign-out fix from the earlier review is present; this is a separate merge blocker.

naingthet:
P2 addressed on `5bd6ce897ea9be173fad805266424d58e58dbf6b` (existing branch, still draft).

The attempted-reentry marker is now presence-only. One automatic renewal is consumed until `noteHostedSessionEstablished` or `allowHostedReentryRetry`. Continuity with no returned bearer at 1,000,000 / 1,015,001 / 1,045,002 is `redirect`, then `sign_in`, then `sign_in`. The former `allows another renewal after the loop window` test is replaced.

Focused Vitest (jsdom): `hostedSession.test.ts` 51, `HostedSignIn.dom.test.tsx` 3, `ManagedGate.dom.test.tsx` 27 — 81 passed. Coverage added for a >15s unsuccessful round trip, repeated reloads with no bearer, successful sign-in reset, explicit retry, and sign-out after the loop window. Deep-link path/query is kept on the first redirect and on retry; bearers stay out of storage; sign-out still clears continuity.

Biome format/lint on the two files, and `tsc --noEmit`, passed. No workflow, merge, or auto-merge changes.

naingthet:
Root verification against 5bd6ce897ea9be173fad805266424d58e58dbf6b: the slow-renewal P2 from comment 5734423365 is resolved.

I executed the exact `hostedSession.ts` at this head. With continuity present and no successful session, timestamps 1,000,000 / 1,015,001 / 1,045,002 now produce `redirect`, `sign_in`, `sign_in`. I also verified that session establishment resets the attempt, explicit retry works, sign-out clears storage, and the first redirect preserves the workspace route and query.

The worker reports 81 focused tests, type checking, and formatting/lint passing. Full CI is now dispatched for this head in run 35383024145 so skipped Rust checks on the UI-only PR do not stand in for validation. Keep the PR draft until required checks and a fresh automatic review pass.

